PR #19, opened 2019-03-08
frontend

marcinbodnar:


Gemfile
```diff
@@ -1,3 +1,2 @@
 source "https://rubygems.org"
 gem "minima", "~> 2.0"
-gem "github-pages", group: :jekyll_plugins
```
azban: why is this removed? seems like this is suggested dependency if you are using jekyll + github pages.

from https://github.com/github/pages-gem#project-goals

> The GitHub Pages gem relies on explicit requirements shared between both users' computers and the build servers to ensure that the result of a user's local build is consistently also the result of the server's build.

PR #29, opened 2019-04-27
Allow to set recovery phone when creating account

vgrichina:


src/actions/account.js
```diff
@@ -41,6 +43,17 @@ export function handleRefreshAccount(wallet, history) {
          })
          .catch(e => {
             console.log(e)
+
+            if (e.message && e.message.indexOf('is not valid') !== -1) {
```
janedegtiareva: pull 'is not valid' into a constant somewhere?

janedegtiareva: or better yet this function might be good as a util function

src/components/accounts/SetRecoveryInfo.js
```diff
@@ -0,0 +1,88 @@
...
+   }
+
+   handleChange = (e, { name, value }) => {
+      this.setState(() => ({
+         [name]: value,
+         isLegit: this.isLegitField(name, value)
+      }))
+   }
+
+   isLegitField(name, value) {
```
janedegtiareva: let's rename legit to something else, maybe valid?

src/components/accounts/SetRecoveryInfoContainer.js
```diff
@@ -0,0 +1,88 @@
...
+
+         <Grid.Row className='creating-info'>
+            <Grid.Column computer={8} tablet={8} mobile={16} className=''>
+               <Header as='h1'>Protect your Account 🔒</Header>
+               <Header as='h2'>
+                  Connect account to phone number – save recovery key on our servers.
+               </Header>
+               <Header as='h3' className='color-blue'>
+                  You can skip this if you plan to backup account keys manually.
+                  We won't be able to help you with account recovery otherwise.
```
janedegtiareva: Is this text from the official UX? I think the sentence should read as "You can skip this blah blah... If you do skip this, we won't be able to help with your account recovery"

src/components/accounts/SetRecoveryInfoContainer.js
```diff
@@ -0,0 +1,88 @@
...
+         </Grid.Row>
+      </Grid>
+
+      {children}
+
+      <Grid className=''>
+         <Grid.Row className='disclaimer border-top-bold'>
+            <Grid.Column computer={16} tablet={16} mobile={16} className=''>
+               <span className='disclaimer-info'>DISCLAIMER: </span>
+               This is a developers&apos; preview Wallet. It should be used for
```
janedegtiareva: Is this still true about the preview?

src/components/accounts/SetRecoveryInfoForm.js
```diff
@@ -0,0 +1,172 @@
...
+import { Form, Input, Button } from 'semantic-ui-react'
+import PhoneInput from 'react-phone-number-input'
+
+import ProblemsImage from '../../images/icon-problems.svg'
+import CheckBlueImage from '../../images/icon-check-blue.svg'
+
+import styled from 'styled-components'
+
+const RecoveryInfoForm = styled(Form)`
+   &&& input {
```
janedegtiareva: is it normal in react to have css inline? I definitely prefer separate css files

src/components/accounts/SetRecoveryInfo.js
```diff
@@ -0,0 +1,88 @@
...
+   }
+
+   handleSubmit = e => {
+      e.preventDefault()
+
+      if (!this.state.isLegit) {
+         return false
+      }
+
+      const { dispatch } = this.props;
```
janedegtiareva: Let's start adding tests for the wallet code. This looks like it doesn't have a lot of UI logic, so should be possible to write unit tests for this.

src/components/accounts/SetRecoveryInfoForm.js
potatodepaulo: > is it normal in react to have css inline? I definitely prefer separate css files

This is somewhat common in react for component-level styles that aren't going to be reused elsewhere to avoid css anti-pattern of overriding, then overriding again with `!important`. I'm not personally a fan, but this is pretty widely accepted.

potatodepaulo: That being said, this is a big enough block of style that I would argue it should be taken out of here.

src/components/accounts/SetRecoveryInfoContainer.js
vgrichina: yes

src/components/accounts/SetRecoveryInfoForm.js
vgrichina: that's normal pattern, the way to refactor it is to cut components into smaller pieces

the good part about this is that it's scoped by component, so very easy to manage vs global styles which can get applied to whatever

src/components/accounts/SetRecoveryInfo.js
vgrichina: I plan to add [integration tests for contract helper](https://github.com/nearprotocol/near-contract-helper/issues/29) and [end-to-end tests for wallet](https://github.com/nearprotocol/near-wallet/issues/33).

Unit tests are going to be easy to write for Redux-based app, but I don't think they are going to be very useful. Most of our real world problems are with broken integrations.

src/components/accounts/SetRecoveryInfoContainer.js
vgrichina: There is no official UX for now, going to update text after comments from Jake and Erik

src/actions/account.js
vgrichina: proper way to fix this would be to haver error codes returned by `nearcore`. This works as a temporary kludge as is.

src/components/accounts/SetRecoveryInfo.js
vgrichina: this seems subjective, `Legit` is used in all other code in wallet.

PR #34, opened 2019-05-13
Add access key when user clicks "authorize"

janedegtiareva:


janedegtiareva:
Tested on localhost: key is added and page is redirected to success url

src/components/login/Login.js
```diff
@@ -27,12 +27,17 @@ class Login extends Component {
...
-      e.preventDefault()
+      e.preventDefault();
       if (this.props.account.url.failure_url) {
          window.location.href = this.props.account.url.failure_url
       }
    }
 
+   handleAllow = e => {
+      e.preventDefault()
+      this.wallet.addAccessKey(this.props.account.accountId, this.props.account.url.contract_id, this.props.account.url.public_key, this.props.account.url.success_url);
```
vgrichina: Ideally this should flow through Redux. `dispatch` an action which adds a key and then handle state changes through reducers.

PR #35, opened 2019-05-14
Recover account

vgrichina:
Follow-up to https://github.com/nearprotocol/near-wallet/pull/29


src/components/accounts/RecoverAccount.js
```diff
@@ -0,0 +1,89 @@
...
+
+   handleChange = (e, { name, value }) => {
+      this.setState(() => ({
+         [name]: value,
+         isLegit: this.isLegitField(name, value)
+      }))
+   }
+
+   isLegitField(name, value) {
+      // TODO: Use some validation framework?
```
potatodepaulo: https://www.npmjs.com/package/react-validation

vgrichina: looks pretty cool, thanks!

janedegtiareva: let's not use "isLegitField". It's not common language for this. https://www.merriam-webster.com/thesaurus/leg%20(it)
the word to use here is "valid"

vgrichina: you are linking verb definition, which has no relation to "legit" adjective:

Screen Shot 2019-05-16 at 12 43 27 PM

PR #38, opened 2019-05-15
Error handling refactoring

vgrichina:


src/components/common/RequestStatusBox.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react'
+import { Grid, Responsive } from 'semantic-ui-react'
+import styled from 'styled-components'
+
+const RequestStatusBoxGrid = styled(Grid)`
```
janedegtiareva: Do we want to start adding docs for reusable components?

vgrichina: What would you put in docs for this component?

janedegtiareva: What it does and maybe a usage example (if we are relatively sure it's not going to change radically).

vgrichina: It is hard to keep non-running usage examples up to date TBH. Tests prob work better for that purpose.

> if we are relatively sure it's not going to change radically

not sure it's going to survive introduction of design system

PR #42, opened 2019-05-17
Send signature when validating code

vgrichina:


src/utils/wallet.js
```diff
@@ -270,11 +270,14 @@ export class Wallet {
    }
 
    requestCode(phoneNumber, accountId) {
-      return sendJson('POST', `${ACCOUNT_HELPER_URL}/account/${phoneNumber}/${accountId}/requestCode`);
+      return sendJson('POST', `${ACCOUNT_HELPER_URL}/account/${phoneNumber}/${accountId}/requestCode`)
```
janedegtiareva: do we have lint fix for this to fix things like missing ;?

vgrichina: @janedegtiareva this project has to use different JS conventions (enforced by `create-react-app`), but yes there is `npm run fix`.

PR #45, opened 2019-05-21
Wallet access keys: wire up adding key through dispatch for error han…

janedegtiareva:
…dling

Note that actual error handling does not work yet


src/components/login/Login.js
```diff
@@ -49,6 +51,11 @@ class Login extends Component {
 
    render() {
       const { account } = this.props
+      const combinedState = {
+         ...this.props,
+         ...this.state,
+         isLegit: this.state.isLegit && !this.props.formLoader
```
vgrichina: I don't think there is either `isLegit` in this component state or used by `LoginForm`. Why pass it downstream?

PR #68, opened 2019-06-05
Make account recovery work

vgrichina:


package.json
```diff
@@ -7,7 +7,7 @@
       "connected-react-router": "^6.4.0",
       "fetch-send-json": "0.0.2",
       "gh-pages": "^2.0.1",
-      "nearlib": "^0.6.2",
+      "nearlib": "^0.8",
```
janedegtiareva: does this default to 0.8.0?

src/utils/wallet.js
```diff
@@ -277,11 +281,28 @@ export class Wallet {
...
-      const signer = new nearlib.SimpleKeyStoreSigner(this.key_store);
-      const { signature } = key ? signer.signBuffer(Buffer.from(securityCode), accountId) : undefined;
-      return sendJson('POST', `${ACCOUNT_HELPER_URL}/account/${phoneNumber}/${accountId}/validateCode`, { securityCode, signature })
+   validateCode(phoneNumber, accountId, postData) {
+      return sendJson('POST', `${ACCOUNT_HELPER_URL}/account/${phoneNumber}/${accountId}/validateCode`, postData)
+   }
+
+   async setupAccountRecovery(phoneNumber, accountId, securityCode) {
+      // TODO: Don't hardcode key
+      const HELPER_KEY = '22skMptHjFWNyuEWY22ftn2AbLPSYpmYwGJRGwpNHbTV';
```
janedegtiareva: is this supposed to be a secret? are we going to change it for production, since this was already out in the public?

src/actions/account.js
```diff
@@ -76,14 +76,28 @@ export function handleRefreshUrl(location) {
...
 const wallet = new Wallet()
 
-export const { requestCode, validateCode } = createActions({
+export const redirectToApp = () => (dispatch, getState) => {
+   const state = getState()
+   const nextUrl = `/login/${(state.account.url && state.props.url.next_url) || '/'}`
+   setTimeout(() => {
+      // TODO: Navigate to page
+      // this.props.history.push(nextUrl)
+      console.log("redirect to ", nextUrl)
```
janedegtiareva: Let's remove the log, since it looks like it's not doing what it's saying it's doing.

src/middleware/index.js
```diff
@@ -13,25 +13,28 @@ const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose
...
+    }
+  )
+}
+
 export default (history) => composeEnhancers(
     applyMiddleware(
         routerMiddleware(history),
-        readyStatePromise,
-        thunk))
+        thunk,
```
janedegtiareva: did applyMiddleware change recently, or was this just not working before?

package.json
vgrichina: according to this https://semver.npmjs.com it should match versions 0.8.0 - 0.8.2 (if you take current published)

I am still kind of confused whether this will update to latest minor version by itself, or only when any of other dependencies depends on it.

src/utils/wallet.js
vgrichina: This isn't supposed to be a secret, it's public key, but the intent is to pass it via environment variable (same as for contract helper)

src/actions/account.js
vgrichina: intent is to fix this to actually redirect before merging

src/middleware/index.js
vgrichina: not sure this change is actually needed, likely just leftover from debugging

package.json
vgrichina: Resolving, cause now it already says 0.8.2 anyway

src/utils/wallet.js
vgrichina: Done

PR #84, opened 2019-06-18
Nginx config to handle React routes properly

vgrichina:


nearmax:
What is the usage of `Dockerfile.fast`? My understanding of the use cases was that we only build docker when we cut the new release. In the future it is going to be all automatic on CI. So build speed does not seem to be a factor?

vgrichina:
> What is the usage of Dockerfile.fast? My understanding of the use cases was that we only build docker when we cut the new release. In the future it is going to be all automatic on CI. So build speed does not seem to be a factor?

I used it to iterate on Nginx config faster. Not sure if there are other uses besides it, might be useful for other niche debugging. Maybe makes sense to use in some CI jobs, so that we can test end to end with Nginx config on CI relatively fast.

nearmax:
>Maybe makes sense to use in some CI jobs, so that we can test end to end with Nginx config on CI relatively fast.

CI jobs should be using prod images, the ones that we actually deploy to prod, and in the future will be building automatically on gitlab. So the only use case I see for it is:

>I used it to iterate on Nginx config faster. Not sure if there are other uses besides it, might be useful for other niche debugging.

vgrichina:
> CI jobs should be using prod images,

Depends for which purpose. If we want to run e2e tests as fast as possible (so that build is disqualified earlier than later), it helps to build using CI cache vs from scratch in Docker. 

Still would need to run slower process on CI as well to verify actual prod, but faster smoke check would help iteration speed.

PR #92, opened 2019-07-10
V0 for shell integration (adding key from a shell-generated url works)

janedegtiareva:
TODOs left:
* show status message in wallet as opposed to redirecting to success url
* icon for near shell

### DIFF
--- a/src/components/login/LoginContainer.js
+++ b/src/components/login/LoginContainer.js
@@ -138,7 +138,7 @@ const CustomContainer = styled(Container)`
    }
 `
 
-const LoginContainer = ({ loader, children, appTitle }) => (
+const LoginContainer = ({ loader, children, appTitle, contractId }) => (
    <CustomContainer>
       <Dimmer.Dimmable as={Segment} basic>
          <Grid>
@@ -185,7 +185,7 @@ const LoginContainer = ({ loader, children, appTitle }) => (
                   tablet={16}
                   mobile={16}
                >
-                  Contract: @contractname.near
+                  Contract: {contractId}
                </Grid.Column>
             </Grid.Row>
             <Grid.Row>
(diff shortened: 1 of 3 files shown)


src/components/login/LoginContainer.js
```diff
@@ -138,7 +138,7 @@ const CustomContainer = styled(Container)`
    }
 `
 
-const LoginContainer = ({ loader, children, appTitle }) => (
+const LoginContainer = ({ loader, children, appTitle, contractId }) => (
```
vgrichina: what does `contractId` mean in this context? The idea is that you create master account through wallet and then use this master account to deploy whatever contracts from CLI. Developer shouldn't need to login through wallet for every created contract.

src/utils/wallet.js
```diff
@@ -151,11 +151,13 @@ export class Wallet {
          '', // fundingOwner
          0 // fundingAmount
       )
-      const parsedUrl = new URL(successUrl)
-      parsedUrl.searchParams.set('account_id', accountId)
-      parsedUrl.searchParams.set('public_key', publicKey)
-      const redirectUrl = parsedUrl.href
-      window.location.href = redirectUrl
+      if (successUrl) {
```
vgrichina: to provide good UX there still needs to be success URL, but pointing to localhost, so flow is like:

1. Run `near login`
2. `near-shell` starts webserver on `localhost` listening for callback and prints login URL in console. `successUrl` should point to this webserver.
3. Click login URL in console.
4. Chose account in wallet and get redirected to `localhost`.
5. `near-shell` handles callback and exits.

src/components/login/LoginContainer.js
janedegtiareva: contract id is needed for the regular access key flow. It will be null for login.

src/utils/wallet.js
janedegtiareva: yeah that's why PR description says it's V0 and it's a todo

janedegtiareva: I really don't think we want to be starting webservers for this. People may have weird firewall setups/shells we are not expecting. Having a complex system like this will just make it not work period for those use cases. 
We can ask them to rerun login command with a diff account id if the orginal account id is not available. In any case this is all not V0.

janedegtiareva:
Rebased on master an merged in a diff pr

PR #99, opened 2019-07-17
Balance Component setup

icerove:
Create a new Component named Balance and it shows the amount more flex and has near and millinear token image


src/components/Balance/Balance.js
```diff
@@ -0,0 +1,20 @@
...
+    let amount = props.amount / BALANCE
+    let style = {
+        width: "1.2em",
+        height: "1.2em",
+        marginLeft: "4px", 
+        bottom: "-2px",
+        position: "relative"
+    }
+    return (<div>
+        {(1234 < 0.01) ?
```
ilblackdragon: This should be amount, right?

src/components/Balance/Balance.js
```diff
@@ -0,0 +1,20 @@
...
+            <div>{amount.toFixed(4)}<img style={style} src={props.milli} alt="" /></div> : <div>{amount} Ⓝ</div>}
```
ilblackdragon: And if it's in millis - it should be * 1000

src/components/Balance/Balance.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+
+const BALANCE = 1000000000000000000
```
ilblackdragon: It's not really a balance, but a NEAR nomination
Also better to write 10**18 and a comment that this is denomination of one near in minimal non divisible units (attoNears)

src/components/dashboard/DashboardContainer.js
```diff
@@ -2,7 +2,9 @@ import React from 'react'
 import { Link } from 'react-router-dom'
 
 import DashboardOtherAssets from './DashboardOtherAssets'
+import Balance from '../Balance/Balance'
 
+import milli from '../../images/n-1000.svg'
```
ilblackdragon: I don know but is there a reasob not to import image once in Balance component?

src/components/profile/Profile.js
```diff
@@ -32,9 +32,11 @@ class Profile extends Component {
             loader: false
          }))
       }, 1000)
+
```
ilblackdragon: These new lines are required?

src/components/Balance/Balance.js
icerove: yes, I just test with 1234

src/components/dashboard/DashboardContainer.js
icerove: it has black one and write one, this is easier to understand as giving as props

src/components/profile/Profile.js
icerove: I used to console log the props, now it is no more used

src/components/Balance/Balance.js
icerove: right, added

src/components/Balance/Balance.js
ilblackdragon: just a tip: it's easier to test components, by passing props vs changing code in the component

icerove: thank you

PR #113, opened 2019-07-22
finish displaying and continue change enter amount

icerove:
create new component balance for amount and change the enter into different format


vgrichina:
Your PR Server URL is https://near-wallet-pr-113.onrender.com.

Follow its progress at https://dashboard.render.com/web/srv-bkr1kvb60hnf2kkmja3g.

Cheers,

vgrichina:
Required by https://github.com/nearprotocol/near-shell/pull/72

src/components/common/Balance.js
```diff
@@ -0,0 +1,34 @@
...
+       {(amount < 0.01) ?
+           <div>{(amount * 1000).toFixed(5)}<img style={style} src={props.milli} alt="" /></div> : 
+           (amount < 1 ? <div>{amount.toFixed(5)} Ⓝ</div> : 
+           (amount < 1000 ? <div>{amount.toFixed(5)} Ⓝ</div> : <div>{toThousands(amount)} Ⓝ</div>)
+           )}
+   </div>)
+}
+
+const toThousands = (num) => {
+   let num_i = parseInt(num)
```
vgrichina: our JS style is using `camelCase` for symbol names, so `num_i` should be something like `numInt`

src/components/common/Balance.js
```diff
@@ -0,0 +1,34 @@
+
+import React from 'react'
+
+// denomination of one near in minimal non divisible units (attoNears)
+const NEAR_NOMINATION = 10**18
+
+const REG = /(?=(\B)(\d{3})+$)/g;
+
+const Balance = (props) => {
+   let amount = props.amount / NEAR_NOMINATION
```
vgrichina: float numbers generally shouldn't be used with money amounts. We have 128-bits of precision in amounts, JS floats at most have 53 bits of precision.

So all code working with amounts need to be rewritten operating either on strings or long numbers (https://github.com/indutny/bn.js/ is preferred cause already used elsewhere).

Note that for this specific case it looks like operating directly on decimal string is the easiest solution.

src/components/common/Balance.js
```diff
@@ -0,0 +1,38 @@
...
+        height: "1em",
+        marginLeft: "3px",
+        verticalAlign: "middle"
+    }
+
+    let index = props.amount ? props.amount.indexOf(".") : null
+    let amount = index > 0 ?  props.amount.slice(0,index): props.amount
+    let zerosSmall = "0".repeat(NOMINATION - 3)
+    let zeros = "0".repeat(NOMINATION)
+    let amountShow = amount ? ((amount.length <= NOMINATION - 3) ?
```
vgrichina: this is somewhat hard to read as a single expression. Maybe extract 2 separate functions: 
1) format as full NEAR tokens
2) format as milliNEAR tokens

It's perfectly ok to have functions which return JSX.

Also this is something that would be good idea to unit test – shouldn't require much setup. See e.g. https://stackoverflow.com/questions/49468126/how-to-test-a-stateless-component

src/components/common/Balance.test.js
```diff
@@ -0,0 +1,23 @@
...
+
+configure({adapter: new Adapter()});
+
+describe('<Balance.js>', ()=>{
+    const contextSmall = "1"+"0".repeat(10) 
+    const contextBig = "1"+"0".repeat(21)
+
+    it('balance should return properly for small number',()=>{
+        let wrapper = shallow(<Balance amount={contextSmall} />)
+        expect(wrapper.contains("0.00001")).toEqual(true);
```
vgrichina: you might also want to match DOM against snapshot https://jestjs.io/docs/en/snapshot-testing

src/components/send-money/SendMoneyAmountInput.js
```diff
@@ -62,9 +67,9 @@ class SendMoneyAmountInput extends Component {
    }
 
    handleChangeAmount = (e, { name, value }) => {
-      const amountStatus = !Number.isInteger(Number(value))
-         ? 'Please enter a whole number.'
-         : value > Number(this.props.amount)
+      const amountStatus = value && !/^[0-9]*(|[.][0-9]{1,5})$/.test(value)
```
vgrichina: I think it makes sense to extract this regexp into separate function, something like `isDecimalString`

src/components/send-money/SendMoneyAmountInput.js
```diff
@@ -62,9 +67,9 @@ class SendMoneyAmountInput extends Component {
...
+            ? 'Invalid Input'
+            :  Number(value)  > Number(this.props.amount)
```
vgrichina: you probably want to use `BN` for this comparison

src/components/common/Balance.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react'
+
+// denomination of one near in minimal non divisible units (attoNears)
+// const NEAR_NOMINATION = 10 ** 18
+const NOMINATION = 18
+
+const REG = /(?=(\B)(\d{3})+$)/g;
+
+const Balance = (props) => {
+
```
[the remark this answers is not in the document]
icerove: changed

src/components/common/Balance.js
```diff
@@ -0,0 +1,50 @@
...
+
+    let index = props.amount ? props.amount.indexOf(".") : null
+    let amount = index > 0 ? props.amount.slice(0, index) : props.amount
+
+    
+    let amountShow = amount ? ((amount.length <= NOMINATION - 3) ? 
+                    convertToShowMilli(amount,props.milli) : convertToShow(amount))
+                    : null
```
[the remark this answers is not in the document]
icerove: change to throw error

src/components/common/Balance.test.js
icerove: thank you for telling me this. I think it might be more useful for bigger component. Next time, when the component needs to check many places, I will use this. :)

src/components/send-money/SendMoneyAmountInput.js
icerove: changed

src/components/send-money/SendMoneyAmountInput.js
icerove: changed

src/components/common/Balance.js
```diff
@@ -0,0 +1,46 @@
...
+// denomination of one near in minimal non divisible units (attoNears)
+// NEAR_NOMINATION is 10 ** 18 one unit
+const NOMINATION = 18
+const REG = /(?=(\B)(\d{3})+$)/g;
+
+const Balance = ({amount, milli}) => {
+    if (!amount) {
+        throw new Error("the account balance is wrong")
+    }
+    let index = amount.indexOf(".")
```
vgrichina: in which case will amount string have dot?

icerove: in the future, will user use near token to transfer or use string like "1111111111111111"? I just dont want to miss that situation

vgrichina: not sure I understand.  Isn't this working on amount from backend (i.e. always integer)?

src/components/common/Balance.js
```diff
@@ -0,0 +1,46 @@
...
+    amount = index > 0 ? amount.slice(0, index) : amount
+    let amountShow = (amount.length <= NOMINATION - 3) ?
+        convertToShowMilli(amount, milli) : convertToShow(amount)
+        
+    return (<div>
+        {amountShow}
+    </div>)
+}
```
vgrichina: nit: missing empty line

src/components/common/Balance.js
```diff
@@ -0,0 +1,46 @@
...
+        let zeros = "0".repeat(NOMINATION)
+        return (<div>{"0." + (zeros.substring(amount.length) + amount).slice(0, 5)} Ⓝ</div>)
+    } else {
+        let len = amount.length - NOMINATION
+        let numInt = len > 3 ? amount.slice(0, len).replace(REG, ",") : amount.slice(0, len)
+        let numDec = amount.slice(len, amount.length)
+        let num = numInt + "." + numDec.slice(0, 5)
+        return (<div>{num} Ⓝ</div>)
+    }
+
```
vgrichina: nit: unnecessary empty line

src/components/dashboard/DashboardDetail.js
```diff
@@ -116,15 +118,16 @@ class DashboardDetail extends Component {
...
+      const { authorizedApps, amount } = this.props
 
       return (
          <PageContainer
             title={(
                <Fragment>
                   <span className='balance'>Balance: </span>
-                  <span className='color-black'>{amountStr}</span>
-                  <span className='near'>Ⓝ</span>
+                  {amount ?  <Balance
```
vgrichina: nit: better to format in a way when it's clear what is which branch of ternary, i.e. something like:

```
{amount
    ? <Balance. amount={amount} milli={milli} />
    : "NaN"}
```

vgrichina: (this applies to all uses of ternary operator, it becomes hard to read when also mixed with JSX)

package.json
```diff
@@ -15,11 +15,15 @@
     "babel-plugin-named-asset-import": "^0.3.1",
     "babel-preset-react-app": "^7.0.2",
     "bfj": "6.1.1",
+    "big.js": "^5.2.2",
+    "bn.js": "^5.0.0",
```
vgrichina: are both libraries needed now?

src/components/common/Balance.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
+
+// denomination of one near in minimal non divisible units (attoNears)
+// NEAR_NOMINATION is 10 ** 18 one unit
+const NOMINATION = 18
+const REG = /(?=(\B)(\d{3})+$)/g;
+
+const Balance = ({amount, milli}) => {
+    if (!amount) {
```
vgrichina: shouldn't also check for `milli`? also I'd be explicit and just spell in error that given param is null. So something like "<Balance> `amount` property shouldn't be null`

src/components/send-money/SendMoneyAmountInput.js
```diff
@@ -55,58 +60,70 @@ const CustomDiv = styled(`div`)`
...
+
    handleChangeAmount = (e, { name, value }) => {
-      const amountStatus = !Number.isInteger(Number(value))
-         ? 'Please enter a whole number.'
-         : value > Number(this.props.amount)
-            ? 'Not enough tokens.' 
-            : ''
+      let amountStatus = ''
+      if (value && !this.isDecimalString(value)) {
+         amountStatus = 'NO MORE THAN 5 DECIMAL DIGITS'
```
vgrichina: won't this show  'NO MORE THAN 5 DECIMAL DIGITS' when I enter letters instead of digits? Probably should disallow typing (or copy-pasting) anything besides digits and decimal point / commas.

src/components/send-money/SendMoneyAmountInput.js
```diff
@@ -55,58 +60,70 @@ const CustomDiv = styled(`div`)`
...
+            {amountInput ? <Balance milli={milli} amount={amountInput} /> : "How much would you want to send? Please send at least 10,000,000,000 "}
          </CustomDiv>
       )
    }
 }
 
 SendMoneyAmountInput.propTypes = {
    handleChange: PropTypes.func.isRequired,
-   amount: PropTypes.string
+   amountInput: PropTypes.string
```
vgrichina: why rename?

src/components/common/Balance.js
icerove: changed

src/components/common/Balance.js
icerove: changed

src/components/dashboard/DashboardDetail.js
icerove: changed

package.json
icerove: no, I just use big.js

icerove: uninstall bn.js

src/components/common/Balance.js
icerove: changed

src/components/send-money/SendMoneyAmountInput.js
icerove: You should try letters and other things, it is already forbidden that

src/components/send-money/SendMoneyAmountInput.js
icerove: `amount` here means walletAccount.amount, and there is also local state used to name amount, I change that to amountInput

PR #120, opened 2019-07-31
move initial account checking to Route component

marcinbodnar:


vgrichina:
Your PR Server URL is https://near-wallet-pr-120.onrender.com.

Follow its progress at https://dashboard.render.com/web/srv-bl0tr88rvnps3kfbl230.

Cheers,

src/components/common/PrivateRoute.js
```diff
@@ -0,0 +1,31 @@
...
+import { Route, withRouter, Redirect } from "react-router-dom"
+import { connect } from "react-redux"
+
+import { isEmpty } from '../../actions/account'
+
+const PrivateRoute = ({component: Component, ...rest}) => (
+   <Route 
+      {...rest} 
+      render={(props) => (
+         isEmpty().payload
```
vgrichina: @marcinbodnar I'm not sure I understand the way `isEmpty` is used. Looks like here it's the one from `import { isEmpty } from '../../actions/account'` and so doesn't get dispatched. What is the intended flow of using `isEmpty` as action?

marcinbodnar: @vgrichina I'm not sure if it is the best variant, but the idea was not to duplicate wallet class object and at the same time check directly if there is an account available. In the future, we can also easily add request status or global alert to it.

vgrichina: @marcinbodnar If this isn't dispatching Redux action, it shouldn't be a Redux action. If the goal is just to have wallet singleton, I assume we can solve this by providing it using context? https://reactjs.org/docs/context.html

Also I assume this can be solved if `isEmpty` is part of Redux state which gets updated as end result of dispatching `handleRefreshAccount`?

PR #122, opened 2019-08-02
Fix bug with displaying amount bigger than 21 digits

icerove:
change bug for displaying amount more than 21 digits

detail: make Big object to Big.toFixed()

### DIFF
--- a/src/components/common/Balance.js
+++ b/src/components/common/Balance.js
@@ -12,7 +12,6 @@ const Balance = ({ amount, milli }) => {
     if (!milli) {
         throw new Error("token image should not be null")
     }
-    amount = amount.toString()
     let amountShow = (amount.length <= NOMINATION - 3) ?
         convertToShowMilli(amount, milli) : convertToShow(amount)
     return (<div>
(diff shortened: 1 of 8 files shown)


vgrichina:
Your PR Server URL is https://near-wallet-pr-122.onrender.com.

Follow its progress at https://dashboard.render.com/web/srv-bl278dnn59ak2ue37850.

Cheers,

src/components/common/Balance.js
```diff
@@ -12,9 +12,9 @@ const Balance = ({ amount, milli }) => {
     if (!milli) {
         throw new Error("token image should not be null")
     }
+    amount = amount.toString()
```
vgrichina: this shouldn't be needed, If there is any more cases when `amount` is number, we need to find and fix the root cause, e.g. like here:
https://github.com/nearprotocol/near-wallet/commit/ec6c1c687e73b587b34c6264f3c9e3170a8be16c#diff-f60c724c7de0318f2d59895aadc93156L112

vgrichina:
@icerove CI is broken (also I am not sure why test would pass locally for you – it fails cause `milli` not set)

PR #134, opened 2019-09-04
Fix full access keys by updating to nearlib with hotfix

janedegtiareva:


vgrichina:
@janedegtiareva needs yarn.lock

janedegtiareva:
> @janedegtiareva needs yarn.lock

done. 
I am confused, I though we were removing .lock files from everywhere?

janedegtiareva:
Verified fix works on https://near-wallet-pr-134.onrender.com/

vgrichina:
@janedegtiareva we totally shouldn't remove lock files from apps to keep deployments deterministic. Libraries is more debatable as lock file would only affect dev experience and CI.

vgrichina:
There are some arguments to keep them for all projects though (e.g. https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/).

Issue #138, opened 2019-09-16
Decimal error when sending fraction of NEAR Token


kcole16:
Issue:
- When attempting to send money, if I enter 0.0001 (3 or more leading zeros), the UI renders 0.10000 ([see video](https://drive.google.com/file/d/1YwKfkq2iQbP1YM9y1it8pMc2B3IjCJ-b/view?usp=sharing))

Steps to reproduce:
- Navigate to https://wallet.nearprotocol.com/send-money/test
- Enter 0.0001 (3 or more leading zeros)

marcinbodnar:
This is related to Balance.js component that Yifang prepared, that is responsible for showing the amount of tokens. There are a few other issues with this component:
 - console: Warning: Invalid prop `style` supplied to `React.Fragment`
 - it shows balance always in a separate line (for example Dashboard)
 - there are inline styles related to this component
 - we always send `milli` prop, which is not good approach
 - it should use `styled-components` like other components.

Want me to fix it as well?

kcole16:
Ah, thanks for pointing those out. Yes, I think it makes sense to refactor the component then (especially the decimal issue and separate line balance)

icerove:
when you really enter 0.0001, the token will change the image to n/1000, which is a millitoken for near token, and if this is confused, just cancel that token image and use near token only

icerove:
when user really enters 0.0001 amount, the token will change the image to n/1000, which is a millitoken for near token, and if this is confused, just cancel that token image and use near token only.

PR #149, opened 2019-09-24
Node & Staking and visual changes

marcinbodnar:


kcole16:
Couple of comments:

- On /staking, placeholder text should be “example: 15” and 
- “0% of Total” is shown in attoNear, should be in Near (divide by 10^18)

[Screenshot](https://drive.google.com/file/d/1Coqu0VdtduK0mKKMe9AqMURwBPLjXUHQ/view?usp=sharing)

marcinbodnar:
@kcole16 placeholder fixed, when it comes to “0% of Total” it is related to issue #138, so I will fix it when Balance component will work fine.

icerove:
when you really enter 0.0001 amount, the token will change the image to n/1000, which is a millitoken for near token, and if this is confused, just cancel that token image and use near token only

PR #151, opened 2019-09-26
MVP: Display full access keys in dashboard

vgrichina:
Fixes https://github.com/nearprotocol/near-wallet/issues/146 as well


kcole16:
I think we should move the Full Access Key list to /profile. It's not something user's will need to deal with very often, and we'll need to make it clear that creating and deleting full access keys has important consequences. Thoughts @jakestutzman?

vgrichina:
> I think we should move the Full Access Key list to /profile. It's not something user's will need to deal with very often, and we'll need to make it clear that creating and deleting full access keys has important consequences. Thoughts @jakestutzman?

the whole thing needs a lot of work before it's decent UX. It's not even really following mockups from Jake, as most info needed for this is still not available.

for now I want to have some quick MVP to allow me plug in Ledger into the system, I'll work on UX more deeply separately from that

kcole16:
> > I think we should move the Full Access Key list to /profile. It's not something user's will need to deal with very often, and we'll need to make it clear that creating and deleting full access keys has important consequences. Thoughts @jakestutzman?
> 
> the whole thing needs a lot of work before it's decent UX. It's not even really following mockups from Jake, as most info needed for this is still not available.
> 
> for now I want to have some quick MVP to allow me plug in Ledger into the system, I'll work on UX more deeply separately from that

Fair enough. Looks good to me then

jakestutzman:
Let’s sync up on this when I’m in the office next week

Sent via Mobile

________________________________

PR #159, opened 2019-10-02
improve refresh account data

marcinbodnar:


src/actions/account.js
```diff
@@ -3,26 +3,17 @@ import { createActions } from 'redux-actions'
...
 
 export function handleRefreshAccount(history, loader = true) {
    return (dispatch, getState) => {
-      if (loader) {
-         dispatch({
-            type: LOADER_ACCOUNT,
-            loader: true
-         })   
-      }
+      dispatch(loginPending(loader))
```
vgrichina: Any reason to not go with async function (similar to wallet actions)? Seems like it'll allow to make this method simpler as request lifecycle management will be handled by middleware and you'll only need to dispatch another action for clearing accounts.

src/actions/account.js
```diff
@@ -42,6 +33,7 @@ export function handleRefreshAccount(history, loader = true) {
                   accounts: wallet.accounts
                }
             })
+            dispatch(loginSuccess(true))
          })
          .catch(e => {
             console.error('Error loading account:', e)
```
vgrichina: I think stuff in ` if (e.message && e.message.indexOf('does not exist while viewing') !== -1) {` block should be moved into separate RESET_ACCOUNTS action.

marcinbodnar:
@vgrichina Prepared new version, now we are using readyStatePromise middleware to handle pending status. Also, refreshUrl was moved to redux-actions and I removed some code from wallet.js that will not be used anymore.

src/actions/account.js
```diff
@@ -139,4 +61,29 @@ export const { switchAccount } = createActions({
    SWITCH_ACCOUNT: wallet.selectAccount.bind(wallet)
 })
 
-export const { loginPending, loginSuccess, loginError } = createActions('LOGIN_PENDING', 'LOGIN_SUCCESS', 'LOGIN_ERROR')
+export const { refreshAccount, refreshUrl, resetAccounts } = createActions({
+   REFRESH_ACCOUNT: [
+      wallet.loadAccount.bind(wallet),
+      (loader) => ({ 
+         loader,
```
vgrichina: how does this work? Is it generating some `meta` for action? I think we need to standardize the way we pass `meta` to allow generic loader/error handling.

marcinbodnar:
@vgrichina this PR can be closed due to #390 changes

PR #168, opened 2019-10-18
Ledger support

vgrichina:


kcole16:
@vgrichina Getting the following error when trying to sign any TXs (sending NEAR and adding keys).

```
Error in background action: TypeError: Cannot read property 'signature' of undefined
    at Object.t.signTransaction (transaction.js:178)
    at async d.signAndSendTransaction (account.js:71)
index.js:30 
```

vgrichina:
> @vgrichina Getting the following error when trying to sign any TXs (sending NEAR and adding keys).
> 
> ```
> Error in background action: TypeError: Cannot read property 'signature' of undefined
>     at Object.t.signTransaction (transaction.js:178)
>     at async d.signAndSendTransaction (account.js:71)
> index.js:30 
> ```

ah, latest commit didn't get deployed apparently (needed settings change on Render).

Try now

PR #169, opened 2019-10-18
Localization POC for RequestStatusBox

vgrichina:


src/actions/account.js
```diff
@@ -102,32 +102,34 @@ export const redirectToApp = () => (dispatch, getState) => {
...
    ],
    CREATE_NEW_ACCOUNT: [
       wallet.createNewAccount.bind(wallet),
-      () => ({ successCode: 'Congrats! this name is available.', errorCode: 'Username is taken. Try something else.' })
+      defaultCodesFor('account.create')
    ],
    CHECK_ACCOUNT_AVAILABLE: [
       wallet.checkAccountAvailable.bind(wallet),
-      () => ({ successCode: 'User found.', errorCode: 'User not found.' })
+      defaultCodesFor('account.available')
```
icerove: check_account_available one is kind of weird, when success, it will be "account.available.sucess", and error would be "account.available.error"

src/components/Routing.js
```diff
@@ -24,13 +28,27 @@ import { SignWithRouter } from './sign/Sign'
...
 import { handleRefreshAccount, handleRefreshUrl } from '../actions/account'
 
-import { ThemeProvider } from 'styled-components'
 import GlobalStyle from './GlobalStyle'
 const theme = {}
 
 const PATH_PREFIX = process.env.PUBLIC_URL
 
 class Routing extends Component {
+   constructor(props) {
```
icerove: are we going to have language option button for wallet? We should test the translation is good enough or not. Since in our team, many language is known from different people. If need Chinese one, I will definitely help to see whether the translation is weird on the page.

src/translations/en.global.json
```diff
@@ -0,0 +1,12 @@
...
+    "account.requestCode.error": "Failed to send SMS with code.",
+    "account.setupAccountRecovery.success": "Success",
+    "account.setupAccountRecovery.error": "Failed to setup account recovery.",
+    "account.recoverAccount.success": "Success",
+    "account.recoverAccount.error": "Failed to recover account.",
+    "account.create.success": "Congrats! this name is available.",
+    "account.create.error": "Username is taken. Try something else.",
+    "account.available.success": "User found.",
+    "account.available.error": "User not found."
+}
```
icerove: just small suggestion: only "success" to setup and recover is not a good log, give every message with specific info to help logging

src/components/Routing.js
vgrichina: we probably going to start with just supporting `Accept-Language` https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.4

src/translations/en.global.json
vgrichina: that's not for logging, this is for end user

also these are default error messages, if error itself provides `messageCode` that should override this

src/translations/en.global.json
```diff
@@ -0,0 +1,12 @@
+{
+    "account.requestCode.success": "Sent SMS with code.",
+    "account.requestCode.error": "Failed to send SMS with code.",
+    "account.setupAccountRecovery.success": "Success",
+    "account.setupAccountRecovery.error": "Failed to setup account recovery.",
+    "account.recoverAccount.success": "Success",
+    "account.recoverAccount.error": "Failed to recover account.",
+    "account.create.success": "Congrats! this name is available.",
```
kcole16: "Congrats! This name is available."

vgrichina: Feel free to edit copy directly next time

PR #197, opened 2019-10-24
Use Parcel instead of Webpack

vgrichina:


icerove:
question: what impact for parcel is better than webpack? I also want to update corgi to remove webpack. 
problem: when I open the render, it show the font is not what wallet used to use. Does wallet font is affected by this? maybe have a look at if font file is not effective with parcel?

vgrichina:
> question: what impact for parcel is better than webpack? I also want to update corgi to remove webpack.

Parcel is faster and needs less config / manually specified dependencies.

> problem: when I open the render, it show the font is not what wallet used to use. Does wallet font is affected by this? maybe have a look at if font file is not effective with parcel?

That's known issue, I'm looking into it.

PR #203, opened 2019-10-24
Save target URL and redirects user after account creation

kcole16:


src/actions/account.js
```diff
@@ -95,8 +95,8 @@ const wallet = new Wallet()
 
 export const redirectToApp = () => (dispatch, getState) => {
    const state = getState()
-   const nextUrl = (state.account.url && (state.account.url.success_url || state.account.url.public_key)) ? `/login/?${stringify(state.account.url)}` : '/'
```
vgrichina: why /login got removed from path? I'm finding hard to follow what's going on.

src/components/accounts/CreateAccount.js
```diff
@@ -4,15 +4,17 @@ import { connect } from 'react-redux'
...
 
 class CreateAccount extends Component {
    state = {
       loader: false,
       accountId: ''
    }
 
-   componentDidMount = () => {}
+   componentDidMount = () => {
+      this.props.handleRefreshUrl(this.props.history.location)
```
vgrichina: Isn't this already handle automatically in `Routing.js`?

src/components/common/PrivateRoute.js
```diff
@@ -11,7 +11,7 @@ const PrivateRoute = ({component: Component, account, ...rest}) => (
                <Redirect
                   to={{
                      pathname: '/create/',
-                     search: props.location.search,
+                     search: `${props.location.search}${(props.location.search !== '' ? '&' : '?')}redirect_url=${props.location.pathname}`,
```
vgrichina: feel like less confusing would be to use https://www.npmjs.com/package/query-string

something like:

`stringify({...parse(props.location.search), redirect_url: props.location.pathname)`

src/actions/account.js
kcole16: Generalized this function to redirect to any route, rather than just 'login'. 

The user's initial target url is stored in ```url.redirect_url``` when they're redirected to ```CreateAccount```. 

Should probably rename it from ```redirectToApp``` to something like ```redirectToRoute``` since it's not necessarily an app login.

src/components/accounts/CreateAccount.js
kcole16: Refactored to remove the ```handleRefreshUrl``` call from ```CreateAccount```.

PR #214, opened 2019-10-31
Add Amplitude analytics tracking

kcole16:
Trying out [Amplitude](https://amplitude.com). So far seems much more powerful and easy to use than GA, will add in addition to GA to start. 

Large free tier too.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -34,7 +34,7 @@
       ;if(!n._iq.hasOwnProperty(e)){n._iq[e]={_q:[]};v(n._iq[e])}return n._iq[e]}
       ;e.amplitude=n})(window,document);
 
-      amplitude.getInstance().init("1adc1891955a4a54b335f89c2dae9d3f");
+      amplitude.getInstance().init('{{ WALLET_AMPLITUDE_ID || "" }}');
     </script>
     <meta charset="utf-8" />
     <link rel="shortcut icon" href="./favicon.ico" />
--- a/src/index.html
+++ b/src/index.html
@@ -9,6 +9,33 @@
       gtag('js', new Date());
       gtag('config', '{{ WALLET_GA_MEASUREMENT_ID || "" }}');
     </script>
+    <script type="text/javascript">
+      (function(e,t){var n=e.amplitude||{_q:[],_iq:{}};var r=t.createElement("script")
+      ;r.type="text/javascript"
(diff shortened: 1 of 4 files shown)


vgrichina:
> Trying out [Amplitude](https://amplitude.com). So far seems much more powerful and easy to use than GA, will add in addition to GA to start.
> 
> Large free tier too.

yeah, GA UX is terrible

https://heap.io this one also looks interesting as you don't have to decide what data to collect beforehand

kcole16:
> https://heap.io this one also looks interesting as you don't have to decide what data to collect beforehand

Interesting, will take a closer look

kcole16:
@vgrichina Heap uses CSS selectors to track events. A bit worried we'll break those pretty frequently

src/middleware/index.js
```diff
@@ -46,6 +46,10 @@ const analyticsMiddleware = store => next => action => {
       event_label: JSON.stringify(action.type)
     })
   }
+  if (window.amplitude) {
+    console.log("Yes")
```
vgrichina: looks like leftover from debugging

src/index.html
```diff
@@ -9,6 +9,33 @@
       gtag('js', new Date());
       gtag('config', '{{ WALLET_GA_MEASUREMENT_ID || "" }}');
     </script>
+    <script type="text/javascript">
```
vgrichina: is there anything done to prevent tracking on development / staging environment?

vgrichina:
@kcole16 weird, I don't see any banner on https://near-wallet-pr-214.onrender.com/

kcole16:
@vgrichina, @jakestutzman pushed back on the banner, pre-mainnet at least. I think it'll be explicit enough without being overbearing if we have this text + URL change to something like testnet.wallet.nearprotocol.com

vgrichina:
> @vgrichina Heap uses CSS selectors to track events. A bit worried we'll break those pretty frequently

I think it happens not that frequently in practice. Also we can always add special selectors for analytics if needed.

src/index.html
kcole16: Should I use the same template tag approach as you did with GA? Wasn't quite sure how we're handling env variables in production

vgrichina: same approach can work, but make sure it does disable tracking (vs crashing) when used in prod

PR #220, opened 2019-11-05
Dashboard transaction history

marcinbodnar:


src/components/dashboard/ActionsList.js
```diff
@@ -0,0 +1,196 @@
...
+            />
+            {/* <span className='font-small'>{row[3]}</span> */}
+         </Grid.Column>
+      )}
+   </CustomGridRow>
+)
+
+const ActionMessage = ({ transaction, action, actionKind }) => (
+   <Fragment>
+      {actionKind === 'createAccount' && `New account created: ${transaction.receiverId}`}
```
vgrichina: Let's make sure new stuff is localizable https://github.com/nearprotocol/near-wallet/issues/217

src/components/dashboard/ActionsList.js
```diff
@@ -0,0 +1,196 @@
...
+         typeof action.accessKey.permission === 'object'
+            ? `Access key added for contract: ${action.accessKey.permission.functionCall.receiverId}`
+            : `New key added for ${transaction.receiverId}`
+      )}
+      {actionKind === 'deleteKey' && `Key deleted`}
+   </Fragment>
+)
+
+const ActionIcon = ({ actionKind }) => (
+   <div>
```
vgrichina: any reason to do this vs something more generic?

For example:

```jsx
<img src={`images/action-${actionKind}`} />
```

is there long-term plan to pass stuff like icon color as parameter (so SVG must be inlined)?

src/components/dashboard/DashboardActivity.js
```diff
@@ -62,17 +60,22 @@ const DashboardActivity = ({ image, title, to, activity }) => (
...
+            transaction={transaction} 
+            actions={transaction.actions}
+            wide={false}
+         />
       ))}
 
       <Grid.Row>
          <Grid.Column textAlign='left' width={16}>
-            <Link to={to}>
+            <a href={to} target='_blank' rel='noopener noreferrer'>
```
vgrichina: I don't think it's desired to open a new window on VIEW ALL click (i.e. I think should be no `target='_blank'`)

BTW, is there any specific difference between `react-router-dom`'s `<Link>` and normal `<a>` link?

src/components/dashboard/DashboardDetail.js
```diff
@@ -133,46 +71,30 @@ class DashboardDetail extends Component {
...
-            ) : null}
+            <DashboardSection
+               notice={notice}
+               handleNotice={this.handleNotice}
+            >
+               <DashboardActivity
+                  loader={loader}
+                  image={activityGreyImage}
+                  title='Activity'
+                  to={`https://explorer.nearprotocol.com/accounts/${accountId}`}
```
vgrichina: let's make this URL configurable through `process.env.EXPLORER_URL`

frol:
Here is a code snippet that queries Explorer caching backend via WAMP-proto (`autobahn`; add `"autobahn": "19.7.3"` to `package.json`):

```js
const wampNearExplorerUrl = "wss://near-explorer-wamp.onrender.com/ws";

const autobahn = require("autobahn");

const wamp = new autobahn.Connection({
  realm: "near-explorer",
  transports: [
    {
      url: wampNearExplorerUrl,
      type: "websocket"
    }
  ],
  retry_if_unreachable: true,
  max_retries: Number.MAX_SAFE_INTEGER,
  max_retry_delay: 10
});

async function connect() {
  try {
    return await new Promise((resolve, reject) => {
      wamp.onopen = session => resolve(session);
      wamp.onclose = reason => reject(reason);
      wamp.open();
    });
  } catch (error) {
    console.error("Connection failure due to:", error);
    return;
  }
}

async function getTransactionsByAccountId(accountId) {
  const wampSession = await connect();
  if (!wampSession) {
    return;
  }

  try {
    return await wampSession.call(
      "com.nearprotocol.testnet.explorer.select",
      [
        `SELECT * FROM transactions WHERE signer_id = :accountId OR receiver_id = :accountId`,
        { accountId }
      ]
    );
  } catch (error) {
    console.error("Failed to call the query function due to:", error);
    return;
  } finally {
    wamp.close();
  }
}

async function main() {
  const transactions = await getTransactionsByAccountId('test');
  console.log("Transactions: ", transactions);
}

main();
```

You may find some inspiration in:

* https://github.com/nearprotocol/near-explorer/blob/master/frontend/src/libraries/explorer-wamp/index.ts - setup wamp API
* https://github.com/nearprotocol/near-explorer/blob/master/frontend/src/libraries/explorer-wamp/transactions.ts - queries

frol:
Actually, it would be better to have a support for querying transactions by account ID in nearcore RPC: https://github.com/nearprotocol/nearcore/issues/1054

src/components/dashboard/ActionsList.js
marcinbodnar: Yes, Jake proposed to make all icons in one color and color change option ultimately.

src/components/dashboard/DashboardActivity.js
marcinbodnar: I think, that Link does not support the external URL. If we want to use Link we can add relative path like `to='/explorer'` and then in Route component add redirection to external URL. Would you like me to implement it?

kcole16:
@vgrichina @marcinbodnar How do you feel about @frol's data fetching solution? Not sure where we are timeline-wise with adding TXs by account to RPC @bowenwang1996

marcinbodnar:
@kcole16 Looks like it's not working with Wallet, throws `Uncaught Error: Cannot find module './when'`.

frol:
@marcinbodnar Have you changed the `require()` with ES6 import for your code-base?

Try replacing:

```js
const autobahn = require("autobahn");
```

with

```js
import autobahn from "autobahn";
```

marcinbodnar:
@frol yes, that's not the issue.

src/utils/wallet.js
```diff
@@ -101,6 +103,55 @@ export class Wallet {
       return await this.getAccount(this.accountId).getAccessKeys()
    }
 
+   async connectWapm(wamp) {
```
vgrichina: let's move explorer-related stuff into separate module, e.g. `utils/explorer-api.js`

src/utils/wallet.js
```diff
@@ -101,6 +103,55 @@ export class Wallet {
...
+      try {
+         return await new Promise((resolve, reject) => {
+            wamp.onopen = session => resolve(session)
+            wamp.onclose = reason => reject(reason)
+            wamp.open()
+         });
+      } catch (error) {
+         console.error('Connection failure due to:', error)
```
vgrichina: why is error swallowed instead of being returned to caller?

src/utils/wallet.js
```diff
@@ -101,6 +103,55 @@ export class Wallet {
...
+      try {
+         return await wampSession.call(
+            'com.nearprotocol.testnet.explorer.select',
+            [
+               `SELECT * FROM transactions WHERE signer_id = :accountId OR receiver_id = :accountId`,
+               { accountId }
+            ]
+         );
+      } catch (error) {
+         console.error('Failed to call the query function due to:', error)
```
vgrichina: same question as above: why not return error to caller?

src/components/dashboard/DashboardDetail.js
```diff
@@ -133,60 +76,62 @@ class DashboardDetail extends Component {
...
-const mapStateToProps = ({ account }) => ({
-   ...account,
-   authorizedApps: account.authorizedApps,
-   fullAccessKeys: account.fullAccessKeys
-})
+const mapStateToProps = ({ account }) => {
+   const transactions = account.transactions 
+      ? account.transactions.reduce((x, t) => {
+            try {
+               t.actions = JSON.parse(t.actions)
```
vgrichina: I feel like JSON should be parsed in `getTransacrions` call

src/components/dashboard/DashboardDetail.js
```diff
@@ -133,60 +76,62 @@ class DashboardDetail extends Component {
...
+            } catch {}
+            return x.concat(t.actions.map((a) => {
+               return {
+                  ...t,
+                  action: typeof a == 'object' ? [a] : [{[a]: {}}]
```
vgrichina: not sure why this is needed, maybe needs some explanation in comment

src/components/dashboard/DashboardDetail.js
vgrichina: feels like in pure functional style you want something like:
```
t = { ...t, actions: JSON.parse(t.actions) }
```

src/components/dashboard/DashboardActivity.js
```diff
@@ -62,17 +60,22 @@ const DashboardActivity = ({ image, title, to, activity }) => (
          </Grid.Column>
       </Grid.Row>
 
-      {activity.map((row, i) => (
-         <ListItem key={`d-${i}`} row={row} />
+      {transactions.slice(0,5).map((transaction, i) => (
```
vgrichina: `5` probably should be configurable through props, something like `maxItems`

src/components/dashboard/DashboardDetail.js
frol: @vgrichina I know JS devs don't care, but you shallow copy `t` that way, which is not the case in the original implementation.

kcole16:
@marcinbodnar Looks like this is showing the chronologically first 5 transactions, as opposed to the last (most recent) 5

src/components/dashboard/DashboardDetail.js
vgrichina: @frol 
> but you shallow copy t that way

That is the whole point, that what should happen inside functional-style code unless you have good performance-optimization reasons.

BTW, there are even some cases where creating copies results in faster code. E.g. if you use immutable data structures for everything (not the case here), it'll take less time to determine what has changed between re-renders.

src/components/dashboard/DashboardDetail.js
frol: @vgrichina Oh, I was also confused. The reason for this thing is that nearcore serializes a list of actions as a list of (object | string) (serde serializes empty struct as a string as opposed to an object):

```
[
  "CreateAccount",
  {"Transfer": {"deposit":"10000000000000000000"}},
  {"AddKey":{"access_key":{"nonce":0,"permission":"FullAccess"},"public_key":"ed25519:4x9s9hpk5ipYJufQXPFEPxrYjW9VStLJtfpRmETuwVHG"}}
]
```

This line translates `"CreateAccount"` into `{"CreateAccount": {}}`.

I hope to address this on the nearcore side one day...

src/components/dashboard/DashboardDetail.js
frol: To be honest, I avoid reduce/fold operators in non-functional languages as a simple for loop is usually much readable.

src/utils/wallet.js
```diff
@@ -101,6 +103,55 @@ export class Wallet {
...
+      })
+
+      const wampSession = await this.connectWapm(wamp)
+      if (!wampSession) return
+
+      try {
+         return await wampSession.call(
+            'com.nearprotocol.testnet.explorer.select',
+            [
+               `SELECT * FROM transactions WHERE signer_id = :accountId OR receiver_id = :accountId`,
```
frol: Oh, let's LIMIT this with some sane number, and you may also need ORDER BY (well, to do that, you would need block.height, so JOIN is also required).

```js
`SELECT transactions.hash, transactions.signer_id as signerId, transactions.receiver_id as receiverId, transactions.actions, transactions.block_hash as blockHash, blocks.timestamp as blockTimestamp
          FROM transactions
          LEFT JOIN blocks ON blocks.hash = transactions.block_hash
          WHERE signer_id = :accountId OR receiver_id = :accountId
          ORDER BY blocks.height DESC
          LIMIT :limit`,
{ accountId, limit: 10 }
```

src/components/dashboard/DashboardDetail.js
vgrichina: > To be honest, I avoid reduce/fold operators in non-functional languages as a simple for loop is usually much readable.

JS is a functional language though (just very unsure), with object system more or less built on top :)

@frol @marcinbodnar however a good point is that `.reduce` usually not the most clear way to express stuff, like my understanding is that here you actually just want to `.map` transactions list first and then join arrays e.g. using `[].concat.apply([], listOfArrays)`. `.reduce` is a bit an overkill to join arrays.

src/components/dashboard/DashboardDetail.js
marcinbodnar: yes, it's necessary for now.

src/utils/wallet.js
marcinbodnar: yes, I will change it

src/utils/wallet.js
marcinbodnar: cool

src/components/dashboard/DashboardDetail.js
marcinbodnar: @vgrichina ok, I can use `.map` and some additional variable to store new transactions array. 

btw I think `.reduce` is very handy for lops when you need one additional variable.

frol: flatMap seems to be more appropriate here

vgrichina: @frol didn't even realize `.flatMap` is already a thing in modern browsers :)

src/components/dashboard/DashboardDetail.js
vgrichina: @marcinbodnar just add a comment, as it's not immediately obvious what's going on. Or maybe even better – extract it to some function, e.g. `postprocessSerdeStruct` which would have this explanation.

src/translations/en.global.json
```diff
@@ -8,5 +8,19 @@
...
     "account.create.error": "Username is taken. Try something else.",
     "account.available.success": "User found.",
-    "account.available.error": "User not found."
+    "account.available.error": "User not found.",
+
+    "actions": {
+       "CreateAccount": "New account created: ${receiverId}",
+       "DeleteAccount": "Account deleted: ${receiverId}",
+       "DeployContract": "Contract deployed: ${receiverId}",
+       "FunctionCall": "Called method in contract: ${methodName}",
```
kcole16: Let's change this to ```Called method in contract: ${receiverId}``` to keep consistent with Explorer (and since we don't have access to method names yet)

src/translations/en.global.json
```diff
@@ -8,5 +8,19 @@
...
+       "Transfer": "Transferred ${deposit}Ⓝ to ${receiverId}",
```
kcole16: ```${deposit}``` is currently in attoNear, needs to be in NEAR (attoNear/(10^18)). @vgrichina Do we have a helper function in NEARlib for this?

src/translations/en.global.json
vgrichina: we have access to method names, so we should display both.

src/translations/en.global.json
vgrichina: we have https://github.com/nearprotocol/near-wallet/blob/2f7e707b3f5c96ff6de8b690d427882b8a97da75/src/components/common/Balance.js

src/utils/wallet.js
```diff
@@ -103,19 +104,7 @@ export class Wallet {
...
-            wamp.open()
-         });
-      } catch (error) {
-         console.error('Connection failure due to:', error)
-         return
-      }
-   }
-
+   
    async getTransactions(accountId = '') {
```
vgrichina: @marcinbodnar  I meant that `getTransaction` should also be part of `explorer-api`.

src/utils/explorer-api.js
```diff
@@ -0,0 +1,12 @@
+export async function connectWapm(wamp) {
```
vgrichina: it's wamp, not wapm (wapm is completely different project https://wapm.io)

src/translations/en.global.json
kcole16: We currently do? If that's the case, we can display ```Called method ${methodName} in contract @${receiverId}```

The current implementation doesn't seem to have method access: [Screenshot](https://drive.google.com/file/d/1EWc2DfPpiAGkg9GF84n43gBwmuIHE4r2/view?usp=sharing)

vgrichina: > The current implementation doesn't seem to have method access: Screenshot

maybe it's lost somewhere in the plumbing, but it's totally is available in transaction itself and so needs to be displayed (both in wallet and in explorer)

src/translations/en.global.json
```diff
@@ -11,15 +11,15 @@
...
     "actions": {
-       "CreateAccount": "New account created: ${receiverId}",
-       "DeleteAccount": "Account deleted: ${receiverId}",
-       "DeployContract": "Contract deployed: ${receiverId}",
-       "FunctionCall": "Called method ${methodName} in contract @${receiverId}",
-       "Transfer": "Transferred ${deposit} to ${receiverId}",
+       "CreateAccount": "New account created: @${receiverId}",
+       "DeleteAccount": "Account deleted: @${receiverId}",
+       "DeployContract": "Contract deployed: @${receiverId}",
+       "FunctionCall": "Called method: ${methodName} in contract: @${receiverId}",
```
vgrichina: maybe it makes sense to use dot notation?

`Called method ${receiverId}.${methodName}`

PR #240, opened 2019-11-19
Implement API for signing individual transaction

vgrichina:


vgrichina:
@kcole16 not really done yet, but I plan to merge to avoid diverging from master too much (existing apps won't call `/sign` route for now)

kcole16:
@vgrichina Makes sense. What should I review in this PR for now?

vgrichina:
> @vgrichina Makes sense. What should I review in this PR for now?

that's more like just FYI at this point

something to pay attention to though is that I had to hide some stuff we had in design, e.g. contract name in multiple places. It didn't match the model of multiple transactions (to different contracts) per single flow.

PR #251, opened 2019-11-25
Setup Seed Phrase changes

marcinbodnar:


marcinbodnar:
it would be good to have some kind of confirmation after `COPY PHRASE` click

kcole16:
> it would be good to have some kind of confirmation after `COPY PHRASE` click

Agreed. It's difficult to tell if the copy succeeded. Do we have a tooltip element we can reuse? Can say "Copied to clipboard". [Example](https://drive.google.com/file/d/18DFVHOuLJXnX1Uw054XVPPPB6fV5m-FL/view?usp=sharing)

kcole16:
@marcinbodnar It looks like the "Copy" button is copying the numbers of the words.

This is what gets copied to my clipboard:
```eye 2want 3rival 4tell 5regular 6fog 7affair 8sample 9vanish 10absurd 11weekend 12okay```

kcole16:
Getting an error after submitting the "Verify" step with the correct word.

```Error: [-32000] Server error: Wrong response for transaction status query```

marcinbodnar:
@kcole16 we have something like this (when you hover lock icon):
https://wallet.nearprotocol.com/profile
but it's more like additional info.

vgrichina:
@kcole16 why are we including Copy to Clipboard button? It increases security risk for our users https://github.com/nearprotocol/near-wallet/issues/218#issuecomment-555133097

vgrichina:
@marcinbodnar looks like something wrong with "Start over" text. Seems like it's getting cut off

Screen Shot 2019-11-26 at 12 56 47 PM

vgrichina:
@kcole16 this screen currently flashes for a second or so. I think if we want users to read the text it should be explicit "Continue" instead. Alternatively text can be merged into verify screen and user redirected immediately after verification.

I think generally we should never use "delay a little bit and then redirect" pattern (there are some other cases for now). It adds delays into flow without providing benefit to anyone but fastest readers (i.e. people most likely to be annoyed by these delays in the first place). 

Screen Shot 2019-11-26 at 1 01 40 PM

bowenwang1996:
@vgrichina also "undefined is now authorized to use your account" seems to be a bug

vgrichina:
> @vgrichina also "undefined is now authorized to use your account" seems to be a bug

yes, but it's not clear whether this message will even exist as is after other changes

kcole16:
@vgrichina The "Copy" button is pretty crucial on mobile, though I'm open to removing it on desktop.

vgrichina:
> The "Copy" button is pretty crucial on mobile, though I'm open to removing it on desktop.

Why is it crucial? We shouldn't encourage people to copy the phrase digitally, very few people know how to store stuff securely in a digital form.

kcole16:
> > @vgrichina also "undefined is now authorized to use your account" seems to be a bug
> 
> yes, but it's not clear whether this message will even exist as is after other changes

Is this part of a larger error message refactoring, or can we change this individual message to ```Seed phrase recovery setup is complete```?

vgrichina:
@kcole16 it totally should be possible to change the specific message. I'm just not sure the bubble should even be part of this flow.

kcole16:
> > The "Copy" button is pretty crucial on mobile, though I'm open to removing it on desktop.
> 
> Why is it crucial? We shouldn't encourage people to copy the phrase digitally, very few people know how to store stuff securely in a digital form.

@vgrichina The copy button is ubiquitous on mobile wallets at this point (less so on the security-conscious desktop wallets), and there are reasonable ways to store the seed by digitally copying it (password manager, for example).

kcole16:
>  it totally should be possible to change the specific message. I'm just not sure the bubble should even be part of this flow.

@vgrichina I completely agree re: removing the "Success" screen. I do think the bubble is useful given an error can occur after the user finishes verification (encountered this yesterday when testnet was down).

@marcinbodnar Let's remove the "Success" screen (below) after the verification step, and change the green banner text to ```Seed phrase recovery setup is complete```.

Screen Shot 2019-11-26 at 1 01 40 PM

vgrichina:
@kcole16 
> @vgrichina I completely agree re: removing the "Success" screen. I do think the bubble is useful given an error can occur after the user finishes verification (encountered this yesterday when testnet was down).

When error occurs we probably don't want to redirect user to dashboard and need to display error. I meant the success case. Anyway, removing extra screen makes sense, and message would make more sense when displayed on dashboard screen.

marcinbodnar:
>When error occurs we probably don't want to redirect user to dashboard and need to display error. I meant the success case. Anyway, removing extra screen makes sense, and message would make more sense when displayed on dashboard screen.
@kcole16 @vgrichina it's not redirecting if error occurs I think, it stays on Verify page and show error

marcinbodnar:
@kcole16 @vgrichina sure, so it will work like before: it will show Global Alert (on Verify page) and then wait for 1,5s for redirection.
I think Vlad is right here:
> I think generally we should never use "delay a little bit and then redirect" pattern

and a better idea is to stay with Success page, add button for redirection and hide Global Alert. We have similar solution on Send Money Success page.

vgrichina:
@marcinbodnar 
> and a better idea is to stay with Success page, add button for redirection and hide Global Alert. We have similar solution on Send Money Success page.

I prefer the solution with less steps (i.e. just redirect to dashboard without "Success" page before that).

vgrichina:
@marcinbodnar merged this PR already as there are no critical issues, but please fix remaining stuff in separate PR

PR #253, opened 2019-11-27
Update nomination so that wallet correctly displays account balance

janedegtiareva:
this is due to the fact that nearcore now keeps account balance in units of
yoctonear

### DIFF
--- a/src/components/common/Balance.test.js
+++ b/src/components/common/Balance.test.js
@@ -13,8 +13,7 @@ describe('<Balance.js>', ()=>{
 
     it('balance should return properly for non 0 for 0.0987',()=>{
         let wrapper = shallow(<Balance amount={contextNull} />)
-        expect(wrapper.contains("<0.00001")).toEqual(true)
-                        
+        expect(wrapper.contains("<0.00001")).toEqual(true);                  
     })
 
     it('balance should return properly for small number',()=>{
--- a/src/components/common/Balance.test.js
+++ b/src/components/common/Balance.test.js
@@ -14,6 +14,7 @@ describe('<Balance.js>', ()=>{
     it('balance should return properly for non 0 for 0.0987',()=>{
         let wrapper = shallow(<Balance amount={contextNull} />)
         expect(wrapper.contains("<0.00001")).toEqual(true)
+                        
     })
 
     it('balance should return properly for small number',()=>{
@@ -23,6 +24,6 @@ describe('<Balance.js>', ()=>{
 
     it('balance should return properly',()=>{
         let wrapper = shallow(<Balance amount={contextBig} />)
-        expect(wrapper.contains("1,234.56700")).toEqual(true);
+        expect(wrapper.contains("1,234.567")).toEqual(true);
     })
 })
--- a/src/components/send-money/SendMoneyAmountInput.js
+++ b/src/components/send-money/SendMoneyAmountInput.js
@@ -5,7 +5,7 @@ import { Segment, Form } from 'semantic-ui-react'
 
 import styled from 'styled-components'
 
-import Balance, { NOMINATION, formatNEAR } from '../common/Balance'
+import { utils } from 'nearlib';
 
 const CustomDiv = styled(`div`)`
    &&&&& {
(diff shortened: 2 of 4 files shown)


kcole16:
@janedegtiareva Should we update this to use the nearlib function? (Once that [PR](https://github.com/nearprotocol/nearlib/pull/126) is merged)

src/components/send-money/SendMoneyAmountInput.js
```diff
@@ -80,10 +81,9 @@ class SendMoneyAmountInput extends Component {
       }
       let amountAttoNear = ''
       if (value !== '') {
-         let input = new Big(value).times(new Big(10).pow(NOMINATION))
-         amountAttoNear = input.toFixed()
+         amountAttoNear = utils.format.parseNearAmount(value);
```
vgrichina: Let's replace all mentions of AttoNear to avoid confusion.

src/components/send-money/SendMoneyAmountInput.js
```diff
@@ -80,10 +81,9 @@ class SendMoneyAmountInput extends Component {
...
          let balance = new Big(this.props.amount)
-         if (balance.sub(input).s < 0) {
+         if (balance.sub(new Big(amountAttoNear)).s < 0) {
```
vgrichina: if this is the only remaining use of `Big`, I think we can avoid it by using `BN` to do the same math (will reduce JS bundle size).

src/components/common/Balance.test.js
```diff
@@ -14,6 +14,7 @@ describe('<Balance.js>', ()=>{
     it('balance should return properly for non 0 for 0.0987',()=>{
         let wrapper = shallow(<Balance amount={contextNull} />)
         expect(wrapper.contains("<0.00001")).toEqual(true)
+
```
vgrichina: nit: not clear why empty line added here

vgrichina:
Looks good overall, but I assume we wait with merge for now (before nearcore is updated)?

src/components/send-money/SendMoneyAmountInput.js
janedegtiareva: done

src/components/send-money/SendMoneyAmountInput.js
janedegtiareva: I agree, it's a good follow up item. Would prefer to postpone to after this change ships to not block core.

janedegtiareva:
> Looks good overall, but I assume we wait with merge for now (before nearcore is updated)?

It can be merged right now - the current version of nearlib has constants for attoNear. When we ship core change, we just need to update the constant in nearlib and bump the nearlib version.

src/components/common/Balance.test.js
janedegtiareva: done

vgrichina:
Screen Shot 2019-12-09 at 2 01 18 PM

@janedegtiareva @kcole16 found interesting bug: after I send `0.1` NEAR once, next time I try to send `10` NEAR – I have `0.1` stuck. Same goes for follow up screen and I think the amount actually sent as well. Also I feel like we should remove that second number altogether, not sure why it's displayed.

Note, that I haven't tested yet if it's a regression from production wallet.

src/components/send-money/SendMoneyAmountInput.js
vgrichina: Small stuff like that tends to not get fixed for long time if not addressed immediately on the spot.
I've created PR https://github.com/nearprotocol/near-wallet/pull/275

janedegtiareva:
> Screen Shot 2019-12-09 at 2 01 18 PM
> 
> @janedegtiareva @kcole16 found interesting bug: after I send `0.1` NEAR once, next time I try to send `10` NEAR – I have `0.1` stuck. Same goes for follow up screen and I think the amount actually sent as well. Also I feel like we should remove that second number altogether, not sure why it's displayed.
> 
> Note, that I haven't tested yet if it's a regression from production wallet.
> 
> UPD: Looks like it doesn't happen on production. So this shouldn't be merged for now.

I can't repro the bug! How do you navigate to the send page again? I go through "Go to dashboard -> send money", and I see 0 as the amount to send.

ilblackdragon:
Repros for me.
Also:
 - after sending money first time "Send Money" button doesn't work - you can only go to dashboard.
 - at registration I'm getting 10.1 NEAR ??? 
 - after sending 0.1, I get now balance `9.900000999999996562` shown which is super unclear. Especially given I had 10.1 before --- where is ~0.1 more?

PR #255, opened 2019-11-28
Setup Seed Phrase changes

marcinbodnar:


src/actions/account.js
```diff
@@ -147,6 +147,15 @@ export const { addAccessKey, clearAlert } = createActions({
          errorCodeDescription: '' 
       })
    ],
+   ADD_ACCESS_KEY_SEED_PHRASE: [
+      wallet.addAccessKey.bind(wallet),
+      () => ({
+         successCodeHeader: 'Success',
```
vgrichina: Let's make this localizable, same as `RequestStatusBox`. I think ideally it should also use same format of `meta`, if we ever need custom headers – one way is to lookup `${messageCode}.header` message as header.

src/components/accounts/SetupSeedPhraseVerify.js
```diff
@@ -53,7 +54,7 @@ const SetupSeedPhraseVerify = ({
         <FormButton
             type='submit'
             color='blue'
-            disabled={!enterWord}
+            disabled={enterWord ? (globalAlert && globalAlert.success) : true}
```
vgrichina: hmm, why is this change needed?

marcinbodnar: thanks to that VERIFY button will stay disabled when addAccessKeySeedPhrase action is complete and it's waiting for redirection.

vgrichina: >  it's waiting for redirection.

we shouldn't be waiting for redirection, we should just redirect immediately and then show alert on dashboard

marcinbodnar: ok, by dashboard you mean dashboard or login? In this case, we would want to rebuild redirectToApp so it will do `history.push` instead of `window.location` for no redirect_url param case, so it will be possible to show Global Alert info from redux after redirection.

Also, we should handle `CLEAR_ALERT` action so the alert will be cleared from redux. We can simply clear Global Alert on every page change (for example in `componentDidMount` in `Routing` component), or we can clear it on `Dashboard` and `Login` components unmount. Or would you like to handle it another way?

Btw I'm not sure if `redirectToApp` is working correctly in one case, for example, if you start wallet with `/send-money` URL and create an account, it will redirect you to `/send-money` again (instead of dashboard). If we would like it to redirect to dashboard we should change a bit this condition:
https://github.com/nearprotocol/near-wallet/blob/c34b395f313484d46c5177b4d26dd2c95d85a299/src/actions/account.js#L100
or is that intentional?

marcinbodnar:
@vgrichina @kendall what error messages would you like to set for Setup Seed Phrase and Login?

src/components/accounts/SetupSeedPhraseVerify.js
kcole16: > ok, by dashboard you mean dashboard or login? In this case, we would want to rebuild redirectToApp so it will do history.push instead of window.location for no redirect_url param case, so it will be possible to show Global Alert info from redux after redirection.

Let's do this.

> Btw I'm not sure if redirectToApp is working correctly in one case, for example, if you start wallet with /send-money URL and create an account, it will redirect you to /send-money again (instead of dashboard). If we would like it to redirect to dashboard we should change a bit this condition:

This is intentional. We want to redirect the user to their original destination (/send-money in this example) after they complete the login flow.

vgrichina: > ok, by dashboard you mean dashboard or login? 

I mean `redirectToApp`, i.e. either next URL or dashboard if there is no next URL. 

> so it will do history.push instead of window.location for no redirect_url param case, so it will be possible to show Global Alert info from redux after redirection.

sure, whatever works

> We can simply clear Global Alert on every page change (for example in componentDidMount in Routing component)

yes, it's desired to clear alert automatically on every page change

kcole16:
> @vgrichina @kcole16 what error messages would you like to set for Setup Seed Phrase and Login?

@vgrichina Which errors can we expect here? Assuming
- RPC error
- RPC timeout

Should we just give a generic error message or will we have more info to give the user?

kcole16:
@marcinbodnar Let's go with generic error messages for now. Really, we should be displaying the actual RPC error to the user, but as @vgrichina points out in [this issue](https://github.com/nearprotocol/nearcore/issues/1813), we need to clarify those for end users.

Setup Seed error: ```An error occurred while setting up seed phrase recovery. Please try again!```

Create Account: ```An error occurred while creating your account. Please try again!```

kcole16:
@vgrichina Is this ready to be merged?

src/components/responsive/GlobalAlert.js
```diff
@@ -127,11 +128,9 @@ const GlobalAlert = ({ globalAlert, clearAlert, closeIcom = true }) => (
                <Image className='left' src={globalAlert.success ? IconCheckImage : IconsProblemImage} />}
             <Message.Content>
                <Message.Header>
-                  {globalAlert.messageCodeHeader}
+                  {globalAlert.success ? 'Success' : 'Error'}
```
vgrichina: this also needs `<Translate>`

src/actions/account.js
```diff
@@ -94,66 +94,67 @@ export function handleRefreshUrl(location) {
...
 const wallet = new Wallet()
 
-export const redirectToApp = () => (dispatch, getState) => {
-   const state = getState()
-   const params = (state.account.url && (state.account.url.success_url || state.account.url.public_key)) ? `?${stringify(state.account.url)}` : ''
-   const nextUrl = state.account.url.redirect_url ? `${state.account.url.redirect_url}${params}` : `/${params}`
-   setTimeout(() => {
-      window.location = nextUrl
-   }, 1500)
+export const redirectToApp = (history) => (dispatch, getState) => {
```
vgrichina: my understanding is that this can be better handled if we integrate with https://github.com/reactjs/react-router-redux

@marcinbodnar what do you think?

vgrichina:
@kcole16 @marcinbodnar merging this, but there are few minor comments (can be done in separate PRs)

src/components/accounts/SetupSeedPhraseVerify.js
vgrichina: @kcole16 I think there is unintentional consequence:
https://github.com/nearprotocol/near-wallet/issues/276

/cc @Patrick1904

PR #257, opened 2019-11-28
Add time stamp in 'time ago' format to Activity items

Patrick1904:
https://github.com/nearprotocol/near-wallet/issues/242

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "connected-react-router": "^6.6.0",
     "fetch-send-json": "0.0.2",
     "js-sha256": "^0.9.0",
+    "moment": "^2.24.0",
     "near-hd-key": "^1.0.0",
     "nearlib": "^0.17.2",
     "prop-types": "^15.7.2",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -9,6 +9,7 @@
     "connected-react-router": "^6.6.0",
     "fetch-send-json": "0.0.2",
     "js-sha256": "^0.9.0",
+    "moment": "^2.24.0",
```
vgrichina: https://bundlephobia.com/result?p=moment@2.24.0 feels a bit heavy if we only going to use for time stamps.  Maybe worth checking alternatives?

vgrichina:
https://bundlephobia.com/result?p=javascript-time-ago@2.0.4 seems like good alternative.

package.json
Patrick1904: Changed to use: [](https://bundlephobia.com/result?p=timeago.js@4.0.1)

Patrick1904: Changed to use: https://bundlephobia.com/result?p=timeago.js@4.0.1

PR #268, opened 2019-12-05
Add UI for email-based recovery

Patrick1904:
Feature needs hookup with email provider before it's ready for master.


vgrichina:
@Patrick1904 https://near-wallet-pr-268.onrender.com/recover-account still just shows "Phone Number" field for me.

src/components/accounts/SetRecoveryInfo.js
```diff
@@ -1,17 +1,22 @@
...
 
 class SetRecoveryInfo extends Component {
    state = {
       loader: false,
       phoneNumber: '',
+      email: '',
+      validEmail: false,
+      sentEmail: false,
       isLegit: false,
+      BackupWithEmail: true
```
vgrichina: why is this capitalized differently than everything else?

src/components/accounts/SetRecoveryInfoFormEmail.js
```diff
@@ -0,0 +1,112 @@
+import React, { Fragment } from 'react'
```
vgrichina: looks like this uses tabs for formatting?

it's a bit in a mess now (some files have 3 (!) spaces, some have 4 spaces). Let's stick to 4 spaces to be consistent with nearlib and near-shell

src/components/common/AccountSkipThisStep.js
```diff
@@ -1,30 +1,47 @@
...
-      <FormButton
-         onClick={skipRecoverySetup}
-         color='link'
-      >
-         I choose to backup my account manually
-      </FormButton>
-   </CustomSegment>
+   <SetUpPhraseBlock onClick={skipRecoverySetup}>
+      <div className='headline'>Switch to Manual Backup</div>
+      <div className='desc'>With this option, NEAR will not be able to assist you with account recovery.</div>
```
vgrichina: NEAR won't be able to assist with any option to be fair

src/components/accounts/SetRecoveryInfo.js
Patrick1904: Mistake from my side - will fix

src/components/accounts/SetRecoveryInfoFormEmail.js
Patrick1904: Yes there's currently an inconsistency with spaces. Most files seem to have 3 spaces, but will stick to 4 spaces going forward.

src/utils/account.js
```diff
@@ -0,0 +1,5 @@
+
+export const validateEmail = (email) => {
+    let re = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
```
vgrichina: This is way too complicated check, it's very hard to determine by reading what would match it. I think it makes sense to just make much simpler check (e.g. check whether string contains two non-empty parts separated by @).

We'll send e-mail to do final validation anyway.

src/components/common/AccountSkipThisStep.js
Patrick1904: Yes, suggest we change copy to "With this option, only your seed phrase will help with account recovery." @kcole16 any thoughts on the copy?

src/components/accounts/SetRecoveryInfo.js
```diff
@@ -86,22 +123,42 @@ class SetRecoveryInfo extends Component {
...
-            </AccountFormSection>
-         </AccountFormContainer>
-      )
+      const { sentSms } = this.props;
+      const { sentEmail } = this.state;
+
+      if (this.state.BackupWithEmail) {
+         return (
+            <AccountFormContainer 
+               title={sentEmail ? 'Confirm Recovery Setup' : 'Protect your Account'}
```
vgrichina: I think it makes sense to rename `sentSms` to `sentMessage` and use it in both cases instead of 2 different fields to signal same status.

Patrick1904:
@vgrichina thought we'd only send the email once as part of account/recovery setup? Basically the user won't be able to enter their email post account creation to receive another email?

> @Patrick1904 https://near-wallet-pr-268.onrender.com/recover-account still just shows "Phone Number" field for me.

src/utils/account.js
Patrick1904: Will add simple validation

vgrichina:
> @vgrichina thought we'd only send the email once as part of account/recovery setup? Basically the user won't be able to enter their email post account creation to receive another email?
> 

Good point, but we'll also change phone to work the same basically. So the form needs to be updated e.g. to provide instructions on how to search your recovery info.

src/components/accounts/SetRecoveryInfo.js
Patrick1904: Good point - will change

src/components/common/AccountSkipThisStep.js
kcole16: How about "With this option, you must write down and save a 12 word phrase to recover your account."

PR #279, opened 2019-12-11
Hide nav links during signup flow

Patrick1904:
- Move some code up to ResponsiveContainer
- Clean up


kcole16:
Looks like the NEAR logo link is still active. We should disable that as well during signup (mainly because users might accidentally clear the redirect_url, which is problematic if they're coming from an app and need to sign a transaction).

vgrichina:
@Patrick1904 nit: please keep formatting changes and logic changes in different commits next time

PR #280, opened 2019-12-11
Remove leftover GA code and track pathname

kcole16:


vgrichina:
@kcole16 I'd prefer we still have GA tracking for time being. Unless the plan is to completely remove GA from all properties.

kcole16:
> @kcole16 I'd prefer we still have GA tracking for time being. Unless the plan is to completely remove GA from all properties.

Fair point. Added back

PR #284, opened 2019-12-12
Modify recovery page to include phone, email, phrase options

Patrick1904:
This feature depends on [PR #268](https://github.com/nearprotocol/near-wallet/pull/268) and should not be launched before it has been merged. A strategy for handling existing accounts with current phone recovery method is also needed.


vgrichina:
> This feature depends on [PR #268](https://github.com/nearprotocol/near-wallet/pull/268) and should not be launched before it has been merged. A strategy for handling existing accounts with current phone recovery method is also needed.

Should just merge it to that branch then? We'll likely need to iterate on #268 for a while, as backend changes also needed.

Issue #276, opened 2019-12-11
Wrong redirect if browser refreshed during recovery setup


vgrichina:
**Problem**

If browser is refreshed during recovery setup, user is redirected to the point of refresh after completing setup.

**Expected Behavior**

User is redirected to dashboard or the app initiating login flow.

**Steps to reproduce**

1. Create new account.
2. Refresh browser page on "Protect your Account"
3. Select "I choose to backup my account manually" and complete the flow.

**NOTE**

Issue might be more generic than that, I didn't test refreshing on other pages.

kcole16:
I don't think we're persisting the redirect URL, so anytime a user refreshes, the new redirect URL is set (which in this case is ```/set-recovery```). 

Seems like the best way to do this is to set the ```redirect_url``` during the ```PrivateRoute``` redirect, as this is the only time the user's flow will be interrupted by account creation.

marcinbodnar:
Yes, it's a more generic problem. One thing is that we set redirect_url on every page refresh, but the main problem is that redux state is cleared on every page reload so we are not able to reproduce redirect_url after a page refresh. 

We can use redux middleware like `redux-storage` or `redux-persist` to reproduce state from before refresh, and additionally make sure that in this case redirect_url will not be set again (in Routing.js).

Want me to implement this solution?

vgrichina:
@marcinbodnar yes, I think it's good idea to use it, as long as we can control what is persisted. For now let's limit to redirect-related stuff.

kcole16:
Closed by #291

PR #291, opened 2019-12-18
Persist redirect related state from clear after page reloads (#276)

marcinbodnar:


marcinbodnar:
@vgrichina @kcole16 I start working with `redux-persist` but it was too much functionality for our needs because we basically want to save state ones (and only partially), and then be sure not to overwrite it in one session. That's why I prepared changes based on sessionStorage without any additional package.

There is one situation that I'm wondering if we want to handle. With this functionality, if a user changes URL manually, without changing the browser window or tab, it will not overwrite the URL state. We can handle it by adding additional conditions and make sure that the URL will be updated when another URL with `/login?...`  appears. But I'm not sure if this is necessary, what do you think?

src/actions/account.js
```diff
@@ -9,169 +9,178 @@ export const LOADER_ACCOUNT = 'LOADER_ACCOUNT'
...
+            const sessionStorageState = JSON.parse(sessionStorage.getItem('wallet:url'))
+
+            if (!sessionStorageState) {
+                sessionStorage.setItem('wallet:url', JSON.stringify(accountUrl))
+            }
+            else {
+                accountUrl = sessionStorageState
+            }
+        } catch(err) {
+            console.log('sessionStorage unavailable')
```
vgrichina: let's log actual error. Also should be `console.warn` IMO

src/actions/account.js
```diff
@@ -9,169 +9,178 @@ export const LOADER_ACCOUNT = 'LOADER_ACCOUNT'
...
-   }
+        }
+
+        try {
+            const sessionStorageState = JSON.parse(sessionStorage.getItem('wallet:url'))
+
+            if (!sessionStorageState) {
+                sessionStorage.setItem('wallet:url', JSON.stringify(accountUrl))
+            }
+            else {
```
vgrichina: nit: `} else {`

vgrichina:
> @vgrichina @kcole16 I start working with `redux-persist` but it was too much functionality for our needs because we basically want to save state ones (and only partially), and then be sure not to overwrite it in one session. That's why I prepared changes based on sessionStorage without any additional package.
> 

cool, session storage probably is a better match for this specific problem.


> There is one situation that I'm wondering if we want to handle. With this functionality, if a user changes URL manually, without changing the browser window or tab, it will not overwrite the URL state. We can handle it by adding additional conditions and make sure that the URL will be updated when another URL with `/login?...` appears. But I'm not sure if this is necessary, what do you think?

> make sure that the URL will be updated when another URL with `/login?...` appears.

Isn't this necessary so that user won't get redirected to wrong app when logging in second time?

Also I think generally we want to wipe this saved state once user performs any action outside of login / create account flows.

kcole16:
> There is one situation that I'm wondering if we want to handle. With this functionality, if a user changes URL manually, without changing the browser window or tab, it will not overwrite the URL state. We can handle it by adding additional conditions and make sure that the URL will be updated when another URL with `/login?...` appears. But I'm not sure if this is necessary, what do you think?

@marcinbodnar In this scenario, assuming the user has no `accountId`, wouldn't the new `/login?` route would be set as the `redirect_url`? If that's the case, this is the desired behavior.

marcinbodnar:
> Isn't this necessary so that user won't get redirected to wrong app when logging in second time?
@vgrichina Only if it will be in the same browser tab (the same that previous login). I will add this condition just in case.

marcinbodnar:
@kcole16 I'm not sure if I understand the question correctly, but redirect_url is set when user enters Wallet first time in a new browser tab/window. If there is no accountId user will be redirected to Create Account but redirect_url will be set earlier (before redirection).

kcole16:
> @kcole16 I'm not sure if I understand the question correctly, but redirect_url is set when user enters Wallet first time in a new browser tab/window. If there is no accountId user will be redirected to Create Account but redirect_url will be set earlier (before redirection).

Got it. Ideally, we should be setting `redirect_url` whenever we force the user into the "Create Account" flow, as the goal is to restore them to their intended destination. 

As Vlad mentioned, if we only set `redirect_url` on the first session in a browser/window, we could send them to an outdated destination.

marcinbodnar:
@vgrichina @kcole16 I did rebuild previous functionality, now there is a lot of possible cases so it's a bit more complicated, but I think I handled all of them.

One additional thought. Now on Login page, user is able to jump to other pages and in this case lose the url params. I think that maybe instead of removing login related url state (on other pages than Create Account flow and Login), we can keep it and in this case, shows some additional button (visible on every page) so the user will be able to come back to '/login' page and finish authorization. It's a small thing but could be useful.

vgrichina:
> One additional thought. Now on Login page, user is able to jump to other pages and in this case lose the url params. I think that maybe instead of removing login related url state (on other pages than Create Account flow and Login), we can keep it and in this case, shows some additional button (visible on every page) so the user will be able to come back to '/login' page and finish authorization. It's a small thing but could be useful.

Wonder if we should for now just send analytics event when it happens. Then we can see if abandoned logins are a big enough problem to handle it better in UI.

@marcinbodnar make sure to merge `master` here, looks like for now there are some merge conflicts

marcinbodnar:
@vgrichina conflicts resolved

kcole16:
Is this ready to merge? @marcinbodnar @vgrichina

marcinbodnar:
@vgrichina 
> WALLET_CREATE_NEW_ACCOUNT_URL and WALLET_CREATE_NEW_ACCOUNT_FLOW_URLS seems like they belong to Routing component, as routes are defined there.

ok

> I think they should be converted to Redux actions instead.

you mean `handleLoginUrl` and `handleRedirectUrl`?

vgrichina:
> an `handleLoginUrl` and `handleRedirectUrl`?

no, I mean `redirectToCreateAccount` and `redirectIfEmpty` functions in `wallet.js`. Seems like having them as Redux actions makes more sense.

marcinbodnar:
@vgrichina ok np, `redirectIfEmpty` is never used - want to keep it or remove it?

vgrichina:
@marcinbodnar remove

kcole16:
@marcinbodnar We had some issues with `/login` functionality as a result of this PR. @vgrichina reverted it for now. Can you take a look and resubmit? Have been using this URL to test: `/login/?title=NEAR+Corgi&contract_id=new-corgis&success_url=https%3A%2F%2Fcorgis.nearprotocol.com%2F&failure_url=https%3A%2F%2Fcorgis.nearprotocol.com%2F&app_url=https%3A%2F%2Fcorgis.nearprotocol.com&public_key=ed25519%3ACAEr31kB5UQewPjVHCSpNqjExxPL5aQodLYHM4qphxoJ`

PR #317, opened 2020-01-08
Receive page: Re-design and enable copying address

Patrick1904:


Patrick1904:
> Could use the confirmation toast after copying the link, especially on mobile. Already a solid improvement on the existing version though.

We could merge now or wait until we've added the toaster? Toaster ETA is Friday

PR #318, opened 2020-01-08
Change error message when user enters own accountId in Username input

marcinbodnar:


src/components/accounts/AccountFormAccountId.js
```diff
@@ -40,10 +40,16 @@ class AccountFormAccountId extends Component {
...
-            autoFocus
-        } = this.props;
+            autoFocus,
+            type
+        } = this.props
 
-        const { accountId } = this.state;
+        const { accountId } = this.state
+
+        const requestStatusSameAccount = type === 'send-money' && this.props.accountId === accountId && {
```
marcinbodnar: @vgrichina If we removed `type` prop for `AccountFormAccountId` component (#339), I will also rebuild this one. This change is only for Send Money page, so maybe just check it by URL instead of `type` prop?

vgrichina: @marcinbodnar I think it makes sense to make check / pass `requestStatus` from outside component instead. Isn't it passed as prop already?

marcinbodnar: @vgrichina 
>Isn't it passed as prop already?
It is at #339 changes. I will add it also here and move this check to the top-level component.

vgrichina: @marcinbodnar any update on this?

marcinbodnar: @vgrichina planning to finish it today

marcinbodnar: @vgrichina changes ready

src/components/send-money/SendMoney.js
```diff
@@ -113,6 +113,17 @@ class SendMoney extends Component {
         this.setState(() => ({
             [name]: value
         }))
+
+        if (name === 'accountId') {
```
vgrichina: let's not use state to track computed properties, it makes it harder to follow data flow

e.g. here you can just add a getter for `requestStatusSameAccount`:

```
get requestStatusSameAccount() {
    if (this.props.accountId !== this.state.accountId) {
        return null;
    }
    return {
                         success: false,
                         messageCode: 'account.available.errorSameAccount',
                     }
}
```

then later just use it as `this.requestStatusSameAccount || requestStatus`

marcinbodnar: done

PR #321, opened 2020-01-09
Add messaging for non monetary transactions

Patrick1904:


kcole16:
Update "is requesting to transfer" to "is requesting the transfer of" in the monetary case.

Add `Current Balance: <balance>` above `More Information`

vgrichina:
@Patrick1904 `totalAmount` is `BN`, let's not do `totalAmount > 0` comparison. Should be `totalAmount.gt(new BN(0))`

vgrichina:
@Patrick1904 I think there is still code which doesn't use `BN` with `totalAmount`

Issue #323, opened 2020-01-10
Phone number recovery didn't work on .cn mirror


ilblackdragon:
**Problem**
Trying to recover my account on https://wallet.cn.nearprotocol.com but it just goes back to create new account after entering code

**Expected Behavior**
Account is recovered

**Steps to reproduce**

1. Go to https://wallet.cn.nearprotocol.com
2. Recover account
3. Enter account and phone number
4. Enter sms code 
5. Observe create account screen

kcole16:
@marcinbodnar Can you take a look at this? Looks like it's still an issue in #291 as well

marcinbodnar:
@kcole16 @vgrichina This change will fix main issue, but we still need to merge #291 so it will work as it supposed to. Because now after account recovery it will redirect you to URL from where you start wallet, so for example if you start with `/` it will redirect you to `/`, but if you start with `/create` it will redirect to `/create`. #291 fixes this problem.

kcole16:
Closed by #324

PR #324, opened 2020-01-10
Fix redirect after account recovery

marcinbodnar:


src/components/accounts/RecoverAccountSeedPhrase.js
```diff
@@ -2,96 +2,98 @@ import React, { Component } from 'react'
...
-               isLegit: false
-            }))
-         })
-   }
+        const accountId = this.state.accountId
+        this.props.recoverAccountSeedPhrase(this.state.seedPhrase, accountId)
+            .then(({ error }) => {
+                if (error) return
+                this.props.handleRefreshAccount()
+                this.props.redirectToApp()
```
vgrichina: not sure I fully understand the logic of the fix. What will happen on `redirectToApp()`, will it always just short-circuit to `/authorized-apps`, etc?

marcinbodnar: The problem with #323 was that during redirection (after account is recovered) the Redux state is not updated yet with accountId, so it's redirecting again to `/create`. Adding `handleRefreshAccount()` before redirection will fix the problem.

Changes might look weird because of changing indention but I only add `handleRefreshAccount` before `redirectToApp()`.

vgrichina: What will happen in such scenario:
1) App requests login using NEAR Wallet (i.e. redirects to wallet with all stuff specified in URL)
2) There is no account yet, so user is redirected to /create page where there is option to recover.
3) User choses to recover.

Basically, will original redirect intent get preserved?

marcinbodnar: Yes, after recovery, user will be redirected to `/login` and will be able to proceed.

vgrichina: Awesome, I’ll test and merge later!

Issue #311, opened 2020-01-06
Allow users to send to subdomains


kcole16:
Users should be able to send NEAR to any subdomain, not just TLDs and ```.near```. Currently, we disallow ```.``` in all input fields. 

Examples:
```test.eth```
```google.com```

### Acceptance Criteria
- [ ] Users can enter subdomains in the "Username to send to" input field on ```/send-money```
- [ ] Tokens are sent to the correct subdomain

kcole16:
Closed by #336

PR #336, opened 2020-01-14
Allow to add subdomain to accountId on Send Money page

marcinbodnar:
@vgrichina do we need any additional change for it (#311)? For example, will `getAccount` wallet function work with a subdomain in accountId?

### DIFF
--- a/src/components/accounts/AccountFormAccountId.js
+++ b/src/components/accounts/AccountFormAccountId.js
@@ -17,7 +17,7 @@ class AccountFormAccountId extends Component {
     }
 
     handleChangeAccountId = (e, { name, value }) => {
-        if (value.match(/[^a-zA-Z0-9_-]/)) {
+        if ((this.props.type !== 'send-money' && value.match(/[^a-zA-Z0-9_-]/)) || value.match(/[^a-zA-Z0-9._-]/)) {
             return false
         }
 
--- a/src/components/accounts/AccountFormAccountId.js
+++ b/src/components/accounts/AccountFormAccountId.js
@@ -17,7 +17,7 @@ class AccountFormAccountId extends Component {
     }
 
     handleChangeAccountId = (e, { name, value }) => {
-        if ((this.props.type !== 'send-money' && value.match(/[^a-zA-Z0-9_-]/)) || value.match(/[^a-zA-Z0-9._-]/)) {
+        if (value.match(this.props.pattern)) {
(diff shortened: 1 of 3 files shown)


src/components/accounts/AccountFormAccountId.js
```diff
@@ -17,7 +17,7 @@ class AccountFormAccountId extends Component {
     }
 
     handleChangeAccountId = (e, { name, value }) => {
-        if (value.match(/[^a-zA-Z0-9_-]/)) {
+        if ((this.props.type !== 'send-money' && value.match(/[^a-zA-Z0-9_-]/)) || value.match(/[^a-zA-Z0-9._-]/)) {
```
vgrichina: let's not hard code `send-money`. How about we pass a pattern in a prop and have a default value allowing subdomains?

I think the only place where we shouldn't allow them is account creation.

vgrichina:
> For example, will getAccount wallet function work with a subdomain in accountId?

@marcinbodnar I think everything should work fine downstream in wallet. You can try creating `account_name.test` (`.test` should be allowed on backend) and testing with that.

src/components/accounts/AccountFormAccountId.js
marcinbodnar: Changes made. Send Money working fine, you can test it by sending it to `marcin.test`.

vgrichina:
@marcinbodnar can be a follow-up PR (as not part of these changes), but I think we also need to refactor this:

```
        this.timeout = setTimeout(() => {
            this.props.type === 'create'
                ? this.props.checkNewAccount(value)
                : this.props.checkAccountAvailable(value)
        }, 500)
```

1. Pass the method to call directly in props instead of passing 2 different methods and type.
2. 500 should be store in named constant (smth like `ACCOUNT_CHECK_TIMEOUT`)

BTW, is there any other reason this component is connected to Redux directly? E.g. how is bound `account` from state used? I think generally we should try to only bind Redux to top-level components, that avoids confusion and makes stuff less tightly coupled.

marcinbodnar:
@vgrichina
> BTW, is there any other reason this component is connected to Redux directly?
I don't think so.
> I think generally we should try to only bind Redux to top-level components, that avoids confusion and makes stuff less tightly coupled.
100% agreed.

Good idea with changes. I will take care of it all first thing tomorrow.

PR #343, opened 2020-01-21
Add CSS 'classNames' util function

Patrick1904:


src/utils/common.js
```diff
@@ -0,0 +1,28 @@
...
+    let len;
+    let tmp = typeof names;
+    let out = "";
+
+    if (tmp === "string" || tmp === "number") {
+        return names || "";
+    }
+    if (isArray(names) && names.length > 0) {
+        for (i = 0, len = names.length; i < len; i++) {
+            if ((tmp = classNames(names[i])) !== "") out += (out && " ") + tmp;
```
vgrichina: I think it's a bit confusing  to build string directly vs using map/filter functions and `.join()`

```
if (isArray(names) && names.length > 0) {
      return names.map(name => classNames(name)).filter(name => !!name).join(" ");
}

return Object.keys().filter(key => names[key]).join(" ");
```

src/utils/common.js
```diff
@@ -0,0 +1,28 @@
+/*  
+    classNames example usage:
+    <div className={classNames(['base-class', {'second' : someFlag, 'third' : true}])}></div>
+*/
+const classNames = (names) => {
+    let isArray = Array.isArray;
+    let i;
```
vgrichina: it's best to declare variables close to where they are used and in most limited scope, i.e. in this case `i` and `len` should be declared in `for` loop.

src/utils/common.js
```diff
@@ -0,0 +1,28 @@
...
+    let len;
+    let tmp = typeof names;
```
vgrichina: Let's use more descriptive names for variables, e.g. here it can be `let namesType = typeof names;`

src/utils/common.js
```diff
@@ -0,0 +1,28 @@
...
+    let out = "";
+
+    if (tmp === "string" || tmp === "number") {
```
vgrichina: in which cases `number` would be used?

src/utils/common.js
```diff
@@ -0,0 +1,28 @@
+/*  
+    classNames example usage:
+    <div className={classNames(['base-class', {'second' : someFlag, 'third' : true}])}></div>
+*/
+const classNames = (names) => {
```
vgrichina: I think it might be convenient to do this:

```
const classNames = (baseClass, ...names) => {
```

then you'd be able to use it as 

```
<div className={classNames('base-class', 'another-class',{'second' : someFlag, 'third' : true})}></div>
```

(without extra array wrapper)

src/utils/common.js
Patrick1904: 💯

vgrichina:
@Patrick1904 not sure why `yarn.lock` is changed, probably best to revert it for clean merge

Issue #258, opened 2019-11-28
Account creation when the network is slow


bowenwang1996:
**Problem**
If the network is slow, the account creation might get stuck saying `username is taken` after the user clicks the `create account` button, which is caused by an timeout error from rpc. However, at this point the account might have been already created on the network, which means that the user cannot create the same account again if they retry.

**Expected Behavior**
There should be some way to inform the user about the timeout error and then allow them to do some manual refresh of the page (or some better way) so that they can proceed with the same account id.

**Steps to reproduce**

1. Try to create an account on https://wallet.tatooine.nearprotocol.com
2. Observe that it says `Username taken` after the `create account` button is clicked. However, from near-shell `near state <account_id>` shows that the account exists.

joesixpack:
Still occurring.  Can't login for stakewars staking because of the username taken already bug.  I'm wondering exactly where there recovery phrase is provided also.

kcole16:
**Steps to reproduce locally (using Chrome)**
1. In dev tools "Network" tab, set speed to "Slow 3G"
2. Try to create an account
3. View recent [Explorer](https://explorer.nearprotocol.com) TX to see transaction succeeded

PR #349, opened 2020-01-22
Fix account creation when the network is slow (#258)

marcinbodnar:
@vgrichina @kcole16 
With these changes, when an error occurs (on blockchain) while creating an account it will be still saved in Local Storage. There are two cases now:
1. If the account was created in blockchain but some error occurred, then the user will be able to continue with set recovery, like nothing happens.
2. If the account was not created in blockchain, it will be still saved in Local Storage, but immediately after this action (or anytime when user enters wallet) check is performed and the account will be removed from Local Storage.

With case 2. there might be a situation where user have other accounts, so I added additional functionality that will log him into another (first of list) account, so he will be able to continue with Wallet. Without that, user will not be able to login to different accounts until creating new one (because if user is not logged in, the list of accounts is not available).

I think we should also refine text message for case 2. for something like "The account is saved in Local Storage, but it doesn't exist on blockchain. The account was removed from Local Storage and you ware logged in to another account. You can try creating account with the same name again.". @kcole16 ?

### DIFF
--- a/src/utils/wallet.js
+++ b/src/utils/wallet.js
@@ -9,7 +9,6 @@ import { store } from '..'
 import { getAccessKeys } from '../actions/account'
 
 export const WALLET_CREATE_NEW_ACCOUNT_URL = `/create`
-export const WALLET_CREATE_NEW_ACCOUNT_FLOW_URLS = [`create`, 'set-recovery', 'setup-seed-phrase', 'recover-account', 'recover-seed-phrase']
 
 const NETWORK_ID = process.env.REACT_APP_NETWORK_ID || 'default'
 const ACCOUNT_HELPER_URL = process.env.REACT_APP_ACCOUNT_HELPER_URL || 'https://near-contract-helper.onrender.com'
@@ -117,26 +116,6 @@ export class Wallet {
 
     async sendMoney(receiverId, amount) {
         await this.getAccount(this.accountId).sendMoney(receiverId, amount)
-    }
-
-    redirectToCreateAccount(options = {}, history) {
-        const param = {
-            next_url: window.location.search
-        }
-        if (options.reset_accounts) {
-            param.reset_accounts = true
-        }
-        //  let url = WALLET_CREATE_NEW_ACCOUNT_URL + "?" + $.param(param)
-        let url =
-            WALLET_CREATE_NEW_ACCOUNT_URL +
-            '/?' +
-            Object.keys(param).map(
-                (p, i) =>
-                    `${i ? '&' : ''}${encodeURIComponent(p)}=${encodeURIComponent(
-                        param[p]
-                    )}`
-            )
-        history ? history.push(url) : window.location.replace(url)
     }
 
     isEmpty() {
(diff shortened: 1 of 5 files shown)


kcole16:
@marcinbodnar I think we should assume the majority of users are unfamiliar with local storage, which is why I kept the message pretty generic.

Though it is confusing that we show `Username is taken. Try something else.` when any error occurs during account creation.

marcinbodnar:
@kcole16 

> I think we should assume the majority of users are unfamiliar with local storage, which is why I kept the message pretty generic.

got it. I mostly worry that the user might not notice that was logged automatically into another account, and it might not be the same account that before.

> Though it is confusing that we show Username is taken. Try something else. when any error occurs during account creation.

I think I can overwrite it, what do you propose?

kcole16:
> I think I can overwrite it, what do you propose?

Ideally, we would show the actual error message (I think this is covered in https://github.com/nearprotocol/near-wallet/issues/289).

In the meantime, we can display "An error occurred. Please try again!". 

We should also remove the red outline from the username input box, and enable the "Create Account" button. 

Important note: when the error is actually "Username is taken" we do still want to display that, highlight the input box, and disable "Create account".

marcinbodnar:
@kcole16 ok

src/utils/wallet.js
```diff
@@ -208,11 +208,14 @@ export class Wallet {
...
-            newAccountPublicKey: keyPair.publicKey.toString()
-        })
-        await this.saveAndSelectAccount(accountId, keyPair);
+        try {
+            await sendJson('POST', CONTRACT_CREATE_ACCOUNT_URL, {
+                newAccountId: accountId,
+                newAccountPublicKey: keyPair.publicKey.toString()
+            })
+        } finally {
+            await this.saveAndSelectAccount(accountId, keyPair);
```
vgrichina: this should only be done only specifically for timeout errors (either network request timeout or neacore side timeout).

all other errors shouldn't try to use account as it's clearly not created

marcinbodnar:
@vgrichina I also implemented changes from PR#276 and added them here.

In `redirectToCreateAccount` we don't need `next_url` - redirection to the previous page is implemented with earlier changes and it's working fine.

`WALLET_CREATE_NEW_ACCOUNT_URL` stays in `wallet.js` because moving it to `Routing.js` cause dependencies issue - in this case, `Routing.js` will import from `actions/account.js` and `actions/account.js` will import from `Routing.js`.

@vgrichina @kcole16 Also, it seems that this functionality could be done a bit better, at this moment if account exists in local storage but does not exist in the blockchain it will always redirect user to Create Account page, even if there is another account available. I think it would be better to clear the account and if user has another account available we should leave him on the same page and show GlobalAlert with info, and if user does not have another account he will be redirected to Create Account anyway and we will show the same globalAlert. It's not very important change but I think it will be a bit better flow.

kcole16:
@marcinbodnar Picking up this thread again. I think your proposed changes re: the flow makes sense (especially re: clearing the non-existent account from localStorage).

marcinbodnar:
@kcole16 ok, I will implement this flow, but it would be best to merge #390 first (@vgrichina ) because there are a lot of changes made, including some changes for this functionality.

PR #350, opened 2020-01-22
TX signing process: disable ability to switch account

Patrick1904:


src/components/login/SelectAccountDropdown.js
```diff
@@ -87,8 +91,13 @@ const CustomSegment = styled(Segment)`
...
 `
 
-const SelectAccountDropdown = ({ handleOnClick, account, dropdown, handleSelectAccount, redirectCreateAccount }) => (
-   <CustomSegment onClick={handleOnClick} basic>
+const SelectAccountDropdown = ({ handleOnClick, account, dropdown, handleSelectAccount, redirectCreateAccount, disabled }) => (
+   <CustomSegment
+      basic
+      onClick={!disabled ? handleOnClick : () => {}}
+      className={disabled && 'disabled'}
+      title={!disabled ? "Switch account" : "This app doesn't allow account changes"}
```
vgrichina: Doesn't block, but let's start localizing strings

Patrick1904: Adding translate now :)

vgrichina:
We can merge this as quick fix, however I think it's weird to keep disabled dropdown:

Screen Shot 2020-01-22 at 4 39 17 PM

@kcole16 @Patrick1904 I think we should change UI to not use dropdown at all.

PR #351, opened 2020-01-23
Revert "TX signing process: disable ability to switch account"

vgrichina:
Reverts nearprotocol/near-wallet#350

Got problem logging in. Test by running app https://studio.nearprotocol.com/?f=ttyvb4ert and trying to login. UI is showing that full access is requested for some reason.


vgrichina:
OOPS reverted wrong PR

PR #352, opened 2020-01-23
Revert "Persist redirect related state from clear after page reloads (#276)"

vgrichina:
Reverts nearprotocol/near-wallet#291

Got problem logging in. Test by running app https://studio.nearprotocol.com/?f=ttyvb4ert  (change `walletUrl` in `src/main.js` as necessary) and trying to login. UI was showing that full access is requested for some reason.

Another problem I got – looks like persisted state gets in the way of new logins (same public key gets reused).


marcinbodnar:
@vgrichina @kcole16 Sorry, did a mistake while resolving conflicts. All changes are ready here #354.

PR #354, opened 2020-01-23
Persist redirect related state from clear after page reloads

marcinbodnar:


kcole16:
@marcinbodnar @vgrichina Looks good now

PR #361, opened 2020-01-28
Recover w/link: Add UI for error scenario

Patrick1904:


src/components/accounts/RecoverWithLink.js
```diff
@@ -92,17 +111,32 @@ class RecoverWithLink extends Component {
...
             seedPhrase: this.props.seedPhrase,
-            successSnackbar: false
+            successSnackbar: false,
+            successView: true
         };
     }
 
+    componentWillMount = () => {
+        if (!this.isLegit) {
+            this.setState({ successView: false });
```
vgrichina: if you want to create derived property, why not just create it like this?

```
get successView() {
      return this.isLegit;
}
```

trying to track it in state is far more error prone

Patrick1904: Very nice

vgrichina:
@Patrick1904 please move it to our repo, otherwise it doesn't get deployed to Render (and so hard to test for me and @kcole16).

Patrick1904:
This PR is obsolete now. Closing.

PR #364, opened 2020-01-29
Fix button feedback delayed on Protect your Account and Setup Recovery Phrase pages

marcinbodnar:
@vgrichina The thing is that `setupRecoveryMessage` action contains another action, in this case `getAccessKeys`, so after `getAccessKeys` action is completed (and before `setupRecoveryMessage` is completed) the `loaderReducer` did set `formLoader` state to `false`, so the button was available. 
I added a stack to `loaderReducer` that will store all actions that started, so we will wait with setting `formLoader` to `false` till all actions are completed.

@kcole16  The same problem was on Setup Recovery Phrase page, so it's fixed with this change as well.

### DIFF
--- a/src/reducers/account/index.js
+++ b/src/reducers/account/index.js
@@ -26,7 +26,7 @@ const loaderReducer = (state, { type, ready }) => {
       return state
    }
 
-   const actionsPending = !ready ? [...state.actionsPending, type] : [...state.actionsPending.slice(0, -1)]
+   const actionsPending = !ready ? [...state.actionsPending, type] : state.actionsPending.slice(0, -1)
    return { 
       ...state, 
       formLoader: !!actionsPending.length,
--- a/src/reducers/account/index.js
+++ b/src/reducers/account/index.js
@@ -17,25 +17,19 @@ import {
 
 const initialState = {
    formLoader: false,
-   sentMessage: false
+   sentMessage: false,
+   actionsPending: []
 }
-
-const actionsPending = []
 
 const loaderReducer = (state, { type, ready }) => {
    if (typeof ready === 'undefined') {
       return state
    }
 
-   if (!ready) {
-      actionsPending.push(type)
-   } else {
-      actionsPending.pop()
-   }
    return { 
       ...state, 
-      formLoader: !!actionsPending.length,
-      actionsPending
+      formLoader: !!state.actionsPending.length,
+      actionsPending: !ready ? [...state.actionsPending, type] : [...state.actionsPending.slice(0, -1)]
    }
 }
 
--- a/src/reducers/account/index.js
+++ b/src/reducers/account/index.js
@@ -20,7 +20,7 @@ const initialState = {
    sentMessage: false
 }
 
-const loaderReducerPendingStack = []
+const actionsPending = []
 
 const loaderReducer = (state, { type, ready }) => {
    if (typeof ready === 'undefined') {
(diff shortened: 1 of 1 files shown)


src/reducers/account/index.js
```diff
@@ -20,11 +20,19 @@ const initialState = {
...
+const loaderReducerPendingStack = []
+
+const loaderReducer = (state, { type, ready }) => {
    if (typeof ready === 'undefined') {
       return state
    }
-   return { ...state, formLoader: !ready }
+
+   if (!ready) {
+      loaderReducerPendingStack.push(type)
```
vgrichina: @marcinbodnar what is the plan for `type`?  Looks like for now just using a counter would be enough as `.pop()` return value is never used.

marcinbodnar: @vgrichina I thought it will be good to know what exactly is happening there with actions, it might also be useful in the future to know what actions are pending. But it can be easily replaced with a counter. Want me to change it?

vgrichina: @marcinbodnar more like I wanted to understand what is the plan. 

> I thought it will be good to know what exactly is happening there with actions, it might also be useful in the future to know what actions are pending.

if we want to have it inspectable, shouldn't it be part of Redux state then so that it's visible in Redux dev tools?

vgrichina:
> The thing is that setupRecoveryMessage action contains another action, in this case getAccessKeys

I'm not sure I'm following this. My understanding is that it just delegates to `wallet.setupAccountRecovery()` which calls into `account.getAccessKeys()` without dispatching any other action.

marcinbodnar:
@vgrichina 
> My understanding is that it just delegates to wallet.setupAccountRecovery() which calls into account.getAccessKeys() without dispatching any other action.

yes, and after `account.getAccessKeys()` is finished `loaderReducer` is setting `formLoader` state to false, but `setupRecoveryMessage` action is still running.

src/reducers/account/index.js
marcinbodnar: @vgrichina Sure, good idea. Done.

vgrichina:
> yes, and after account.getAccessKeys() is finished loaderReducer is setting formLoader state to false, but setupRecoveryMessage action is still running.

So I am not sure how this can be happening. Promise from `account.getAccessKeys()` is not returned to Redux. So something else must be in play. 

Note, that what you describe is still a real issue (for cases when there are multiple actions in progress) and needs to be fixed, but I just want to make sure we don't miss some other underlying problem.

marcinbodnar:
@vgrichina 
> So something else must be in play.

Yes, it's `readyStatePromise` middleware.

vgrichina:
> Yes, it's `readyStatePromise` middleware.

so that's what I mean, there is no way for `account.getAccessKeys()` promise inside of `wallet.setupRecoveryMessage()` to propagate to  `readyStatePromise` middleware. Middleware only processes stuff that has been dispatched as action.

Looks like something else is dispatching `getAccessKeys` action though. I also noticed weird thing when testing: it doesn't look (in Redux dev tools) like `getAccessKeys` action is changing `loaderReducerPendingStack`. Is it behaving as expected on your side?

src/reducers/account/index.js
```diff
@@ -20,11 +20,23 @@ const initialState = {
...
+const actionsPending = []
+
+const loaderReducer = (state, { type, ready }) => {
    if (typeof ready === 'undefined') {
       return state
    }
-   return { ...state, formLoader: !ready }
+
+   if (!ready) {
+      actionsPending.push(type)
```
vgrichina: I think there might be problems with mutating this collection directly:
https://redux.js.org/faq/immutable-data/#why-is-immutability-required-by-redux
That might also explain why I had weirdness in Redux dev tools (getAccessKeys action not changing stack).

Also you generally want to take current state of the stack from what redux passes as `state`, as otherwise stuff like time travel would be broken.

marcinbodnar:
@vgrichina 
> Looks like something else is dispatching `getAccessKeys` action though.

It's `await account.addKey(publicKey)` - it's finally runs `getLedgerKey` which dispatching `getAccessKeys`.

src/reducers/account/index.js
vgrichina: @marcinbodnar do you plan to fix this?

marcinbodnar: @vgrichina Sorry for the delay. Done.

Issue #290, opened 2019-12-18
Correctly display small amounts of NEAR (<0.00001)


kcole16:
Same as this [explorer issue](https://github.com/nearprotocol/near-explorer/issues/152)

### Acceptance Criteria

- [x] Amounts of NEAR smaller than 0.00001 are displayed as "<0.00001 N"
- [ ] On hover, when a NEAR balance is <= 0.0001, the amount in yoctoNEAR is displayed "1000000 yoctoN"

**Original Thread**
> We should fix `formatNEAR` to actually reflect that the amount is not 0, but rather very small.

> @janedegtiareva Can nearlib format small yoctoNEAR amounts in a human-friendly format, e.g. `1 yoctoN`? We used to have `<0.00001 N` here. Do you think it is reasonable?

> @icerove @kcole16 It seems that to make the information non-confusing and useful, we should present both `<0.00001 N (1 yoctoNEAR)` (we may want to show the exact amount, e.g. "1 yoctoNEAR", as an on-hover popup)

kcole16:
Closed by #368

PR #368, opened 2020-01-31
Display yoctoN in title tag if amount is <=0.0001 (#290)

marcinbodnar:
I added `FRAC_DIGITS` so it will be easy to change number of fractional digits in the future.

`formatWithCommas` is copied from nearlib utils, it can be removed if we can export this function from nearlib.

### DIFF
--- a/src/components/common/Balance.js
+++ b/src/components/common/Balance.js
@@ -1,7 +1,8 @@
 import React from 'react'
 import styled from 'styled-components'
-import { List } from 'semantic-ui-react';
-import { utils } from 'nearlib';
+import { List } from 'semantic-ui-react'
+import { utils } from 'nearlib'
+import { BN } from 'bn.js'
 
 const CustomDiv = styled(List)`
     position: relative;
@@ -9,26 +10,38 @@ const CustomDiv = styled(List)`
 `
 
 const FRAC_DIGITS = 5
+const YOCTO_NEAR_THRESHOLD = new BN('10', 10).pow(new BN(utils.format.NEAR_NOMINATION_EXP - FRAC_DIGITS + 1, 10))
 
 const Balance = ({ amount }) => {
     if (!amount) {
-        throw new Error("amount property should not be null")
+        throw new Error('amount property should not be null')
     }
     let amountShow = formatNEAR(amount)
 
     return (
-        <CustomDiv title={['0','0.0001'].includes(utils.format.formatNearAmount(amount, (FRAC_DIGITS || 1) - 1)) ? `${formatWithCommas(amount)} yoctoⓃ` : ``}>
+        <CustomDiv title={showInYocto(amount)}>
             {amountShow} Ⓝ
         </CustomDiv>
     )
 }
 
 export const formatNEAR = (amount) => {
-    let ret =  utils.format.formatNearAmount(amount, FRAC_DIGITS);
+    let ret =  utils.format.formatNearAmount(amount, FRAC_DIGITS)
     if (ret === '0') {
-        return `<${!FRAC_DIGITS ? `0` : `0.${'0'.repeat((FRAC_DIGITS || 1) - 1)}1`}`
+        return `<${
+            !FRAC_DIGITS 
+                ? `0` 
+                : `0.${'0'.repeat((FRAC_DIGITS || 1) - 1)}1`
+        }`
     }
-    return ret;
+    return ret
+}
+
+const showInYocto = (amountStr) => {
+    const amount = new BN(amountStr)
+    if (amount.lte(YOCTO_NEAR_THRESHOLD)) {
+        return formatWithCommas(amountStr) + ' yoctoⓃ'
+    }
 }
 
 const formatWithCommas = (value) => {
--- a/src/components/common/Balance.js
+++ b/src/components/common/Balance.js
@@ -8,24 +8,35 @@ const CustomDiv = styled(List)`
     display: inline;
 `
 
+const FRAC_DIGITS = 5
+
 const Balance = ({ amount }) => {
     if (!amount) {
         throw new Error("amount property should not be null")
     }
     let amountShow = formatNEAR(amount)
+
     return (
-        <CustomDiv>
+        <CustomDiv title={['0','0.0001'].includes(utils.format.formatNearAmount(amount, (FRAC_DIGITS || 1) - 1)) ? `${formatWithCommas(amount)} yoctoⓃ` : ``}>
             {amountShow} Ⓝ
         </CustomDiv>
     )
 }
 
 export const formatNEAR = (amount) => {
-    let ret =  utils.format.formatNearAmount(amount, 5);
+    let ret =  utils.format.formatNearAmount(amount, FRAC_DIGITS);
     if (ret === '0') {
-        return "<0.00001";
+        return `<${!FRAC_DIGITS ? `0` : `0.${'0'.repeat((FRAC_DIGITS || 1) - 1)}1`}`
     }
     return ret;
+}
+
+const formatWithCommas = (value) => {
+    const pattern = /(-?\d+)(\d{3})/
+    while (pattern.test(value)) {
+        value = value.replace(pattern, '$1,$2')
+    }
+    return value
 }
 
 export default Balance


src/components/common/Balance.js
```diff
@@ -8,24 +8,35 @@ const CustomDiv = styled(List)`
...
+
 const Balance = ({ amount }) => {
     if (!amount) {
         throw new Error("amount property should not be null")
     }
     let amountShow = formatNEAR(amount)
+
     return (
-        <CustomDiv>
+        <CustomDiv title={['0','0.0001'].includes(utils.format.formatNearAmount(amount, (FRAC_DIGITS || 1) - 1)) ? `${formatWithCommas(amount)} yoctoⓃ` : ``}>
```
chadoh: Can we break up this long line? Maybe something like this:

```jsx
<CustomDiv title={
  ['0', '0.0001'].includes(
    utils.format.formatNearAmount(amount, (FRAC_DIGITS || 1) - 1)
  )
    ? `${formatWithCommas(amount)} yoctoⓃ`
    : ``
}>
```

I find that much easier to look at!

And now that I formatted it like that, I have some other questions:

* `FRAC_DIGITS || 1` – we know that `FRAC_DIGITS` is always `5`, so the `|| 1` will never be evaluated
* is there value in saying `FRAC_DIGITS - 1` instead of `4`?
* `['0', '0.0001']` seems like a very precise list – is this generic enough to catch all edge cases we're interested in catching? 
* do we need our own `formatWithCommas`? It seems like maybe `formatNearAmount` or `parseNearAmount` can do this for us: https://github.com/nearprotocol/nearlib/blob/master/test/utils/format.test.js#L7
* using `formatNearAmount` and then comparing the result to an array of strings seems like a roundabout way to check the magnitude of a number – can we directly use `<` instead? `amount <= 10 ** 21` – would this be correct?

src/components/common/Balance.js
```diff
@@ -8,24 +8,35 @@ const CustomDiv = styled(List)`
...
         </CustomDiv>
     )
 }
 
 export const formatNEAR = (amount) => {
-    let ret =  utils.format.formatNearAmount(amount, 5);
+    let ret =  utils.format.formatNearAmount(amount, FRAC_DIGITS);
     if (ret === '0') {
-        return "<0.00001";
+        return `<${!FRAC_DIGITS ? `0` : `0.${'0'.repeat((FRAC_DIGITS || 1) - 1)}1`}`
```
chadoh: Here again, we know that `FRAC_DIGITS` always exists, right? (I don't think the returned string makes sense anyhow, in the case that `FRAC_DIGITS` doesn't exist here – I think it would return `'<01'`?)

src/components/common/Balance.js
```diff
@@ -8,24 +8,35 @@ const CustomDiv = styled(List)`
...
     }
     return ret;
 }
 
+const formatWithCommas = (value) => {
```
chadoh: JavaScript has [the `Intl.NumberFormat` utility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat) built in, which will allow us to display large numbers in the format users are most used to, rather than always using commas. It can be used like this:

```js
new Intl.NumberFormat().format(amount)
```

I'd prefer we use this! 

Note that passing no arguments to `NumberFormat()` results in it selecting the user's locale. See [the docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat) for info about how to try it with other locale settings.

src/components/common/Balance.js
vgrichina: I think it's best to extract complex logic into another helper function vs have it inline in JSX.

Also I am nor sure why set `FRAC_DIGITS` in a way where you always have to subtract `1`.

src/components/common/Balance.js
vgrichina: @chadoh @marcinbodnar  it doesn't work because we have more than 53-bit numbers. Generally we cannot use built-in JS number support for anything related to NEAR amounts (or any other number that can have more than 53 bits of precision required).

chadoh: Could we use a library like [bignumber.js](https://github.com/MikeMcl/bignumber.js/) rather than implementing this logic ourselves?

vgrichina: @chadoh we already bring https://github.com/indutny/bn.js/ for big number math, ideally we shouldn't be bringing 2 libraries (as it adds weight).  It might be reasonable to replace `BN` with `bignumber.js`,  but it would require a lot of changes around `nearlib`, `near-shell` and app examples. Given that we are trying to stabilize for release, I think it's best to avoid doing it pre-mainnet.

src/components/common/Balance.js
marcinbodnar: > * `FRAC_DIGITS || 1` – we know that `FRAC_DIGITS` is always `5`, so the `|| 1` will never be evaluated

It's added in case `FRAC_DIGITS` will be set to `0`

> * is there value in saying `FRAC_DIGITS - 1` instead of `4`?

Yes, so that it can be controlled by `FRAC_DIGITS` only. From issue description, I understand that we want to show the amount in `yoctoN` in a situation where fractional digits number is one less than we are showing. So if we round it to `0.00001` we want to show it for amounts `<=0.0001`. Or did I misunderstood @vgrichina?

> * `['0', '0.0001']` seems like a very precise list – is this generic enough to catch all edge cases we're interested in catching?

Yes, for `FRAC_DIGITS = 5`. But `0.0001` needs to be changed to handle other `FRAC_DIGITS` values. Thanks for pointing that out.

> * do we need our own `formatWithCommas`? It seems like maybe `formatNearAmount` or `parseNearAmount` can do this for us:

`formatNearAmount` divides given amount by 24
https://github.com/nearprotocol/nearlib/blob/master/src.ts/utils/format.ts#L27
`parseNearAmount` multiplies given amount by 24
https://github.com/nearprotocol/nearlib/blob/master/src.ts/utils/format.ts#L51

> * using `formatNearAmount` and then comparing the result to an array of strings seems like a roundabout way to check the magnitude of a number – can we directly use `<` instead? `amount <= 10 ** 21` – would this be correct?

We are operating on strings there, so I wanted it to be consistent.

> Also I am nor sure why set `FRAC_DIGITS` in a way where you always have to subtract `1`.

In `title` tag it's because we want to show title if the amount is `<=0.0001`, so it's one fractional digit less. In `formatNEAR` it's because It's adding `1` at the end.

src/components/common/Balance.js
marcinbodnar: In case that `FRAC_DIGITS` is `0` it will return correct `<0`.

src/components/common/Balance.js
vgrichina: I think we need to create `utils.format.formatYoctoNearAmount` and use it here.

Also I feel like less hacky way to compare for threshold would be to use `BN` for math and compare against specific threshold number. So something like:

```
const YOCTO_NEAR_THRESHOLD = new BN('10', 10).pow(new BN(NEAR_NOMINATION_EXP - FRAC_DIGITS +1, 10));

// ... later in method
const amount = new BN(amountStr);
if (amount.lt(YOCTO_NEAR_THRESHOLD)) {
   return formatYoctoNearAmount(amount) + ' yoctoⓃ';
}
return  formatNearAmount(amount) + ' Ⓝ';
```

also likely best to use https://en.wikipedia.org/wiki/Non-breaking_space instead of regular space here

src/components/common/Balance.js
chadoh: I've enjoyed using bignumber.js more than BN, but I agree that we should not add a second library or do refactoring now. Looks like BN does not support formatting with commas or other separators, so we will need to implement some sort of `formatWithCommas` ourselves for now.

vgrichina: @chadoh I agree with your sentiment on bignumber.js vs BN, just think that changing that globally is pretty big project by itself, so we shouldn't block this change.

src/components/common/Balance.js
marcinbodnar: Done. `formatWithCommas` is copied from nearlib utils, it can be removed if we can export `formatWithCommas` from nearlib.

frol:
This thing is useful for Explorer as well. May I ask you guys to move it to nearlib, eventually?

PR #371, opened 2020-01-31
Add support for showing generic profiles

chadoh:
Previously, this app would only show you your own profile at `/profile`

This changes the route to `/your-username`, and allows showing anyone else's profile at this route as well

If viewing your own profile, data is loaded immediately from the redux store, which is cached in localStorage. If viewing someone else's, data is fetched directly in the component, since no other components use this data currently and we do not need to pollute the global state with it.

## Screenshots

near wallet new profile route

### DIFF
--- a/src/components/Routing.js
+++ b/src/components/Routing.js
@@ -26,7 +26,7 @@ import { AuthorizedAppsWithRouter } from './access-keys/AccessKeys'
 import { FullAccessKeysWithRouter } from './access-keys/AccessKeys'
 import { SendMoneyWithRouter } from './send-money/SendMoney'
 import { ReceiveMoneyWithRouter } from './receive-money/ReceiveMoney'
-import { ProfileWithRouter } from './profile/Profile'
+import { Profile } from './profile/Profile'
 import { SignWithRouter } from './sign/Sign'
 import { NodeStakingWithRouter } from './node-staking/NodeStaking'
 import { AddNodeWithRouter } from './node-staking/AddNode'
@@ -58,10 +58,10 @@ class Routing extends Component {
         })
         this.props.addTranslationForLanguage(translations_en, "en")
     }
-    
+
     componentDidMount = () => {
         const { handleRefreshAccount, handleRefreshUrl, history, clearAlert, clear, handleRedirectUrl, handleLoginUrl, router } = this.props
-        
+
         handleRefreshUrl()
         handleRefreshAccount(history)
 
@@ -79,12 +79,12 @@ class Routing extends Component {
             }
 
             handleRefreshAccount(history, false)
-            
+
             const { state: { globalAlertPreventClear } = {} } = history.location
             if (!globalAlertPreventClear) {
                 clearAlert()
             }
-            
+
             clear()
         })
     }
@@ -164,8 +164,8 @@ class Routing extends Component {
                                     />
                                     <PrivateRoute
                                         exact
-                                        path='/profile'
-                                        component={ProfileWithRouter}
+                                        path='/profile/:accountId'
+                                        component={Profile}
                                     />
                                     <PrivateRoute
                                         exact
@@ -192,8 +192,8 @@ class Routing extends Component {
                                         path='/staking'
                                         component={StakingWithRouter}
                                     />
-                                    <PrivateRoute 
-                                        component={DashboardDetailWithRouter} 
+                                    <PrivateRoute
+                                        component={DashboardDetailWithRouter}
                                     />
                                 </Switch>
                             )}
--- a/src/components/profile/Profile.js
+++ b/src/components/profile/Profile.js
@@ -1,44 +1,20 @@
 import React from 'react'
-import { useSelector } from 'react-redux'
 
 import PageContainer from '../common/PageContainer';
 import ProfileDetails from './ProfileDetails'
 import ProfileSection from './ProfileSection'
 import ProfileQRCode from './ProfileQRCode';
-import { Wallet } from '../../utils/wallet'
-
-const wallet = new Wallet()
+import { LOADING, NOT_FOUND, useAccount } from '../../hooks/allAccounts'
 
 export function Profile({ match }) {
     const { accountId } = match.params
-    const currentUser = useSelector(state => state.account)
-    const [ account, setAccount ] = React.useState(
-        currentUser.accountId === accountId && currentUser
-    )
-    const [ loading, setLoading ] = React.useState(!account)
+    const account = useAccount(accountId)
 
-    React.useEffect(() => {
-        if (currentUser.accountId === accountId) {
-            setAccount(currentUser)
-        } else {
-            setLoading(true)
-            wallet.getAccount(accountId).state()
-                .then(({ amount }) => {
-                    setLoading(false)
-                    setAccount({ accountId, amount })
-                })
-                .catch(err => {
-                    setLoading(false)
-                    setAccount(null)
-                })
-        }
-    }, [accountId])
-
-    if (loading) {
+    if (account.__status === LOADING) {
         return <PageContainer title="Loading..." />
     }
 
-    if (!account) {
+    if (account.__status === NOT_FOUND) {
         return <PageContainer title={`Account @${accountId} not found`} />
     }
 
--- a/src/components/profile/Profile.js
+++ b/src/components/profile/Profile.js
@@ -1,51 +1,53 @@
-import React, { Component } from 'react'
-import { connect } from 'react-redux'
-
-import { withRouter } from 'react-router-dom'
+import React from 'react'
+import { useSelector } from 'react-redux'
 
 import PageContainer from '../common/PageContainer';
 import ProfileDetails from './ProfileDetails'
 import ProfileSection from './ProfileSection'
-import ProfileYourKeys from './ProfileYourKeys'
-import ProfileNotice from './ProfileNotice'
 import ProfileQRCode from './ProfileQRCode';
+import { Wallet } from '../../utils/wallet'
 
-class Profile extends Component {
-   state = {
-      loader: false
-   }
+const wallet = new Wallet()
 
-   componentDidMount() {}
+export function Profile({ match }) {
+    const { accountId } = match.params
+    const currentUser = useSelector(state => state.account)
+    const [ account, setAccount ] = React.useState(
+        currentUser.accountId === accountId && currentUser
+    )
+    const [ loading, setLoading ] = React.useState(!account)
 
-   render() {
-      const { account } = this.props
+    React.useEffect(() => {
+        if (currentUser.accountId === accountId) {
+            setAccount(currentUser)
+        } else {
+            setLoading(true)
+            wallet.getAccount(accountId).state()
+                .then(({ amount }) => {
+                    setLoading(false)
+                    setAccount({ accountId, amount })
+                })
+                .catch(err => {
+                    setLoading(false)
+                    setAccount(null)
+                })
+        }
+    }, [accountId])
 
-      return (
-         <PageContainer
-            title={`Account: @${account.accountId ? account.accountId : ``}`}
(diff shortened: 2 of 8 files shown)


Patrick1904:
@vlad what do you think?

src/components/profile/Profile.js
```diff
@@ -1,51 +1,44 @@
...
-         </PageContainer>
-      )
-   }
+import { Wallet } from '../../utils/wallet'
+
+const wallet = new Wallet()
+
+export function Profile({ match }) {
+  const { accountId } = match.params
+  const [ loading, setLoading ] = React.useState(true)
```
vgrichina: Let's not become more inconsistent with adding React Hooks as well.

For now we have a bit half-assed Redux integration, I'd prefer we clean it up as generally it's been quite useful when debugging complex flows (as Redux Dev tools allows to view state diffs, time-travel, etc).

vgrichina:
@chadoh any reason we need separate `OwnProfile` component vs using one component for all profiles?

src/components/profile/Profile.js
vgrichina: Looks like there are good ways to use Redux together with Hooks as well: https://thoughtbot.com/blog/using-redux-with-react-hooks

Probably worth deeper discussion, as functional style of components looks cleaner.

kcole16:
@chadoh Thanks for the PR!

We've been using the `/profile` as more of an account settings-type page (which is probably a bit of a misnomer) thus far. 

While we have this branded "Wallet", I don't think generic profile pages are a priority, at least not in addition to the explorer accounts. In the mid-long term, if we focus more on discovery and a social experience, it is worth revisiting.

Thoughts? @vgrichina @Patrick1904

Patrick1904:
Since profile activity is currently public through the explorer it might make sense to enable it for wallet too? Potentially with an option for the user to ‘check’ a box to enable/disable it depending on default. Especially if we add more personal data to a profile. It could make it easier for users to send/receive by sharing profile url. 

@chadoh do you see value in the wallet profile including more personal information and being used as a social feature? Curious about thoughts here

vgrichina:
I think it's useful to have profiles in wallet, explorer has way too technical focus.

vgrichina:
> Especially if we add more personal data to a profile. It could make it easier for users to send/receive by sharing profile url.

@Patrick1904 @chadoh @kcole16  yeah, we for sure want to add more data once @evgenykuzyakov is done with his open web stuff for data storage on user's account. Like you should be able to maintain profile like on Twitter, etc.

src/components/profile/Profile.js
chadoh: Having worked on several hefty projects with redux and then a few without, I've come down on "redux is probably not worth all the indirection it adds, especially now that React has hooks."

It looks like there's only one piece of global state that we need in this project: the current user. We could make this easily accessible to all components that need it using React's context api. I recently helped implement something similar here: https://github.com/AutarkLabs/open-enterprise/commit/96ea396349208b8e064832caf42a050f3d8340ed#diff-ae42d1e8553b2812cb903dd2add50dd4

As for "state diffs, time-travel, etc" – I have not missed these features at all. I'd wager that the number of complex flows that require debugging can be reduced by removing redux.

chadoh:
Regarding @vgrichina's question:

> any reason we need separate `OwnProfile` component vs using one component for all profiles?

this relates to @kcole16's comment:

> We've been using the `/profile` as more of an account settings-type page (which is probably a bit of a misnomer) thus far.

I wasn't sure if we wanted to keep `/profile` and make it something more specific to the current user. 

For now, I think a sensible next step would be to remove `/profile` altogether. Instead, we can use a single `Profile` component, served from `/:accountId`. When you click your name in the top right of wallet.nearprotocol.com and select "Profile", it would bring you to your own URL, the same way that clicking on your avatar in the sidebar of twitter.com brings you to your profile's URL. 

For now, viewing your own profile will look exactly like viewing someone else's. Once we add more social features, we can customize the `Profile` component based on whether you're viewing your own profile ore someone else's.

src/components/profile/Profile.js
vgrichina: > It looks like there's only one piece of global state that we need in this project: the current user. 

Ideally we need to keep preloaded data about account globally as well, so that UI is more snappy (i.e. we show cached data before new data is loaded). Stuff like "what message to show after redirect" or "where to redirect after login" is also global state.

vgrichina: @marcinbodnar @Patrick1904 what are your thoughts on this? Do you feel like Redux Dev Tools are useful for debugging state changes or not really?

chadoh:
@Patrick1904 [asked](https://github.com/nearprotocol/near-wallet/pull/371#issuecomment-581456640):

> @chadoh do you see value in the wallet profile including more personal information and being used as a social feature? Curious about thoughts here

Yes, I think that seems useful

src/components/profile/Profile.js
chadoh: Another angle I'd like to consider: I think near-wallet may end up being a great reference project for other NEAR developers building apps with React. To that end, I want to make its code as easy to understand as possible, and as close to current React "best practices" as we can get. My anecdotal impression is that redux is falling out of favor, now that React has better built-in ways to manage data than it did when redux first gained popularity. (A quick search for "redux popularity over time" yielded this as the top result: https://hackernoon.com/goodbye-redux-26e6a27b3a0b.)

vgrichina: @chadoh I agree that Redux would be an overkill if all the data is loaded every time.

However we need to make a reference blockchain app and that ideally should account for far slower responses from backend than in normal web app. That means we'll need stuff like caching data locally. Which is generally pretty hard to get working and debug when state can be modified from anywhere.

marcinbodnar: > Do you feel like Redux Dev Tools are useful for debugging state changes or not really?

useful for me, regarding some complex flows we have implemented

chadoh: I'm willing to use redux for now. If I want to re-raise the question of whether it's worth moving away, I'll do so by refactoring one of those hard-to-get-working parts, and see what we think. It's hard to evaluate such architectural alternatives without seeing a good side-by-side comparison.

vgrichina: @chadoh I think it's good idea to switch to functional style with hooks when using Redux though:
https://thoughtbot.com/blog/using-redux-with-react-hooks

chadoh: @vgrichina would you like to see this done all at once, or are you ok with it being done piecemeal as we add and edit existing components?

vgrichina: @chadoh seems like smth much easier to do piecemeal.

chadoh:
See updated version. All profiles are now served from `/:accountId`, and the old component has been removed.

src/components/profile/Profile.js
chadoh: @vgrichina I've updated this component to fetch the current user from the redux state, but I doubt we want to add the profile-being-viewed to the global state, since no other components use this data currently. Let me know what you think.

src/components/responsive/DesktopPopup.js
```diff
@@ -138,7 +138,7 @@ const DesktopPopup = ({
                 <List className='submenu'>
                     <List.Item>
                         <List.Icon as={Image} src={AccountImage} />
-                        <List.Content as={Link} to='/profile' onClick={handleClose}>
+                        <List.Content as={Link} to={`/${account.accountId}`} onClick={handleClose}>
```
vgrichina: URL should have some prefix before `${account.accountId}` or we'll end up having name clashes for accounts and routes in wallet.

chadoh: The route is listed last in `Routing` for just this reason

src/components/profile/Profile.js
```diff
@@ -1,51 +1,53 @@
...
+        currentUser.accountId === accountId && currentUser
+    )
+    const [ loading, setLoading ] = React.useState(!account)
+
+    React.useEffect(() => {
+        if (currentUser.accountId === accountId) {
+            setAccount(currentUser)
+        } else {
+            setLoading(true)
+            wallet.getAccount(accountId).state()
```
vgrichina: I'd prefer we load through Redux as discussed before. In this specific example these reasons come in play (in addition to consistency / debugging possibility):
1) it'll allow us to cache results globally, so that on next call we can already show data and update it vs load from scratch
2) we won't need boilerplate for triggering loading flag on/off

src/components/responsive/DesktopPopup.js
vgrichina: That still doesn't allow to have e.g. `send-money` as account name.

src/components/profile/Profile.js
chadoh: This means that we will add items to localStorage that potentially get very stale, with no plan to clear them. For the foreseeable future this will probably cause no problems, but it seems like transient data and I like that the current implementation treats it as such.

src/components/responsive/DesktopPopup.js
chadoh: Ah, great point – maybe just an `@` before the username would be a good approach? Should we get design insight on this?

src/components/profile/Profile.js
vgrichina: I mean just cache in memory, not in local storage.

src/components/responsive/DesktopPopup.js
vgrichina: @chadoh let's just make it `/account/${account.accountId}`

src/components/Routing.js
```diff
@@ -156,10 +156,10 @@ class Routing extends Component {
                                         path='/receive-money'
                                         component={ReceiveMoneyWithRouter}
                                     />
-                                    <PrivateRoute
+                                    <Route
```
chadoh: I noticed that the dropdown in the top right shows a link to your profile _even when you're not logged in_ – I really think we should allow showing profiles of other people, even if you're not logged in, but that feels like a whole design challenge that we might want to avoid for now. Should I keep this as a `PrivateRoute`?

chadoh:
@vgrichina I found the existing places where we use `createActions` throughout this codebase hard to reason about, so I specifically avoided using `redux-actions`  here. To me, using it feels like hiding complexity in a bad way. I would rather make the boilerplate explicit, rather than requiring anyone modifying this code to understand the boilerplate that the `redux-actions` helpers will generate for us.

chadoh:
@vgrichina I've updated to use `redux-actions` – let me know what you think

src/components/Routing.js
chadoh: @kcole16 I think this might be a question for you – do we want to allow signed-out people to view other people's profile pages? If so, does this require a fairly deep redesign of the top bar, potentially getting rid of the entire dropdown which currently contains a link to the signed-in user's profile? If so, I think we have three options:

1. do that redesign & implementation as part of this PR
2. keep this PR as-is, which results in linking people to `/profile/undefined` if they're signed out (note that someone has created an account named `undefined`, so this results in a confusing experience)
3. only allow viewing someone else's profile if you're signed in

I think `3` is probably the easiest for now, until we implement a new signed-out design for the top bar.

chadoh:
@vgrichina 

> a bit uneasy around mixing strings for status and object for account info...  I suggest we just add a field to account object (e.g. `_status`)

I was also uneasy about mixing strings & objects, but couldn't think of a design I liked more. I liked the `_status` suggestion a lot, though! I've changed it to `__status` and I've also swapped strings like `'LOADING'` for constants. Let me know what you think!

src/components/Routing.js
kcole16: I think we should go with `3` for now. In some ways it weakens the purpose a bit, but I'd like to save any redesigns for a bigger overhaul we're kicking off with @corwinharrell soon.

Issue #228, opened 2019-11-08
Recovery Options Management


kcole16:
## Story

As a user, I want to view all of my setup recovery options in one place, so I can add additional options and remove existing ones.

## Acceptance Criteria

- [x] User can view email recovery
  - [x] If enabled
    - [x] Email address
    - [x] Date enabled
    - [x] “Resend email” CTA
    - [x] “Disable” CTA
  - [x] If disabled
    - [x] “Enable”
- [x] User can view phone recovery
    - [x] If enabled
      - [x] Phone number
      - [x] Date enabled
      - [x] “Resend message” CTA
      - [x] “Disable” CTA
    - [ ] If not enabled
      - [ ] “Enable”
- [x] User can view seed phrase recovery
  - [ ] If enabled
    - [ ] Date enabled
    - [ ] “Disable” CTA
  - [ ] If not enabled
    - [ ] “Enable”
- [x] If a user tries to view a recovery setup page (`/setup-recovery`) while the option is enabled, redirect to `/profile`
- [x] Mobile ok
- [x] Tested

## Assets
[User Flows](https://docs.google.com/document/d/1AdQyVtxUTxRUhDMGrItp5AwiCa8BEnzVb2QaRLJdtxU/edit)
[Designs](https://projects.invisionapp.com/share/SKQKQ6B4PYU#/screens/359877550_Account)

vgrichina:
I think we should replace "Full Access Keys" with some more user-focused language. E.g. "Authorized Devices" or smith like that. 

>  User must confirm intent to deauthorize

Also shouldn't be possible to remove last key without completely removing account.

> Full Access Keys do not appear in Summary

I think that it's good idea to show device that have access in summary. That way user is more likely to notice that there are some old device still in the list and can clean them up. It's likely best to highlight which devices haven't been used for signing for a while.

kcole16:
> I think we should replace "Full Access Keys" with some more user-focused language. E.g. "Authorized Devices" or smith like that.

Agreed. I think we can show both "Recovery Options" (keys associated w/ email, seed phrase, etc) and "Authorized Devices", and avoid using the word "key" entirely.

vgrichina:
> Agreed. I think we can show both "Recovery Options" (keys associated w/ email, seed phrase, etc) and "Authorized Devices", and avoid using the word "key" entirely.

+1.

PR #381, opened 2020-02-06
Recovery management

Patrick1904:
This adds a 'Recovery Management' module to the profile page, allowing users to enable, disable and re-send recovery links.
recovery-methods

#228 

<a href="https://gitpod.io/#https://github.com/nearprotocol/near-wallet/pull/381"></a>


src/components/profile/Recovery/RecoveryContainer.js
```diff
@@ -0,0 +1,196 @@
...
+        } else if (kind === 'phone') {
+            phoneNumber = detail;
+        }
+
+        this.setState({ resendingLink: method.kind })
+        sendNewRecoveryLink({ accountId, phoneNumber, email, publicKey, seedPhrase, method })
+            .then(({ error }) => {
+                if (error) return
+
+                loadRecoveryMethods(accountId);
```
Patrick1904, line 126: @vgrichina I `loadRecoveryMethods` here after re-sending link to get the latest data. However, it causes a bug. If you setup email recovery and then click ‘resend’, wait until it’s done and click ‘resend’ again, two email methods now show up. Re-freshing the page fixes it. I tried to do `refreshAccount` before `loadRecoveryMethod` but it doesn’t help.
rec-methods

vgrichina:
Few high level notes:
- feels like we need to have more prominent confirmation when disabling recovery method. It is equivalent to removing access key, so same dance with enter your account name to confirm seems warranted
- feels a bit glitchy in terms of UI when a box is small and then expands when data gets loaded (not blocking this PR)
- when I "Resend Email" – I end up with 2 recovery methods with same e-mail, which IMO is confusing
- need to understand what is the end game for key management (e.g. we need better UI for app keys and non-recovery full access keys as well).  (not blocking this PR)
- pro users need some way to see the actual public key (e.g. "i" icon with details popup)  (not blocking this PR)
- this is not related to this PR specifically, but I think we need to be more clear about which data is public on chain and which data is private (e.g. your e-mail)

src/components/profile/Recovery/RecoveryContainer.js
vgrichina: https://github.com/nearprotocol/near-wallet/blob/78fb7f0366af88be074d3c331fb9fb1833ec5a9d/src/hooks/recoveryMethods.js#L16

looks like this check results in data being cached and stale

src/hooks/recoveryMethods.js
```diff
@@ -13,7 +13,7 @@ export function useRecoveryMethods(accountId) {
     const dispatch = useDispatch()
 
     useEffect(() => {
-        if (accountId === wallet.getAccountId() && !recoveryMethods) {
```
Patrick1904, line 16: @vgrichina, even though I changed it here?

src/utils/wallet.js
```diff
@@ -310,6 +310,26 @@ class Wallet {
         });
     }
 
+    async deleteRecoveryMethod(method) {
+        await sendJson('POST', `${ACCOUNT_HELPER_URL}/account/deleteRecoveryMethod`, {
+            accountId: wallet.accountId,
+            recoveryMethod: method.kind,
+            publicKey: method.publicKey,
+            ...(await wallet.signatureFor(wallet.accountId))
+        }).then(({ error }) => {
```
vgrichina, line 319: why use `await` and `.then` together?  Also `{ error }` would only be returned for promises returned from `dispatch`, stuff not wrapped into Redux actions would just throw error.

src/utils/wallet.js
```diff
@@ -310,6 +310,26 @@ class Wallet {
...
+            if (error) return
+            this.removeAccessKey(method.publicKey)
+        })
+    }
+
+    async sendNewRecoveryLink({ phoneNumber, email, accountId, seedPhrase, publicKey, method }) {
+        await this.setupRecoveryMessage({ accountId, phoneNumber, email, publicKey, seedPhrase })
+            .then(({ error }) => {
```
vgrichina, line 327: same as above

Patrick1904:
> Few high level notes:
> 
> * feels like we need to have more prominent confirmation when disabling recovery method. It is equivalent to removing access key, so same dance with enter your account name to confirm seems warranted
> * feels a bit glitchy in terms of UI when a box is small and then expands when data gets loaded (not blocking this PR)
> * when I "Resend Email" – I end up with 2 recovery methods with same e-mail, which IMO is confusing
> * need to understand what is the end game for key management (e.g. we need better UI for app keys and non-recovery full access keys as well).  (not blocking this PR)
> * pro users need some way to see the actual public key (e.g. "i" icon with details popup)  (not blocking this PR)
> * this is not related to this PR specifically, but I think we need to be more clear about which data is public on chain and which data is private (e.g. your e-mail)

I agree 100%.

> feels like we need to have more prominent confirmation...
One alternative is to show a username input on 'confirm disable' view instead

> feels a bit glitchy in terms of UI...
I think we should consider 'skeleton loading' or similar. We should add that to 'full access keys' page too because you get the same loading glitch there w/ the rows.

> pro users need some way to see the actual public key...
I agree. Right now you can delete a recovery method key from the full access key page

We've had some initial discussions around what a revamp of key/device/permission management could look like and it's probably something we want to do sooner than later cc @kcole16

Patrick1904:
@kcole16 should we update the copy here? A bit scary 'account not found' when/if recovery link has been been replaced.

Screen Shot 2020-04-15 at 4.59.16 PM

kcole16:
> One alternative is to show a username input on 'confirm disable' view instead

Something like this?
Screen Shot 2020-04-15 at 4 56 24 PM

> when I "Resend Email" – I end up with 2 recovery methods with same e-mail, which IMO is confusing

Good point. We have some copy for this when user clicks an invalid link, but as @Patrick1904 pointed out, not clear enough. Will update.

Can also add a timestamp or sequence to the email? Not sure that is actually very helpful though.

> need to understand what is the end game for key management

@vgrichina Planned to address this more in #422 (for full access). I like the idea of labeling full access keys with some metadata re: the device they are pertain too, perhaps as a first version of multi-device support? 

I think most users should only see the list of Recovery Options and Registered Devices, which will represent their full access keys. Might involve moving the `/full-access-keys` page to an Advanced Section, as it is confusing for many users.

kcole16:
@Patrick1904 

**Link Expired**
Please check your email or phone for the most recent message. Links in all earlier messages are invalid.

Patrick1904:
> @Patrick1904
> 
> **Link Expired**
> Please check your email or phone for the most recent message. Links in all earlier messages are invalid.

[Updated](https://github.com/nearprotocol/near-wallet/pull/381/commits/3310d689cf15c284b4acd1cff4b93f89d5ca1dd0)

vgrichina:
> Something like this?

LGTM

> Can also add a timestamp or sequence # (e.g. Recovery Email #2) to the email? Not sure how helpful this will be.

from other comment by @Patrick1904 seems like 2 e-mails was just known bug (in fact old one supposed to be removed).


> I think most users should only see the list of Recovery Options and Registered Devices, which will represent their full access keys. Might involve moving the /full-access-keys page to an Advanced Section, as it is confusing for many users.

I think we should remove /full-access-keys page altogether and instead have following type of times (maybe even on a same page like "Security"):
- Apps
- Devices
- Recovery Options

if a key isn't know to fall under any of these categories (i.e. we have no metadata) – it makes sense to just show it as "Unknown Device" in Devices and recommend removing it.

Patrick1904:
@kcole16 @vgrichina please take a look at the loading animation

vgrichina:
I still had 2 e-mails (even after refresh) and when I tried to remove one got error.

Screen Shot 2020-04-16 at 5 29 15 PM

Patrick1904:
> I still had 2 e-mails (even after refresh) and when I tried to remove one got error.
> 
> Screen Shot 2020-04-16 at 5 29 15 PM

@vgrichina interesting. Can you try with a different account? I'm not able to repro

kcole16:
Loader looks good.

@Patrick1904 @vgrichina I'm able to reproduce creating multiple emails (by selecting "Enable" next to Phone and then just adding another email). Both are correctly adding keys though, and recover and delete correctly. 

This is definitely unintended, though not necessarily problematic (outside of the error case in @vgrichina example), especially if the user adds multiple email addresses. Multiple keys for the same email could be confusing, though.

Patrick1904:
> @Patrick1904 @vgrichina I'm able to reproduce creating multiple emails (by selecting "Enable" next to Phone and then just adding another email). Both are correctly adding keys though, and recover and delete correctly.

Yes we haven't disabled ability to add an additional email recovery method.

The bug I was referring to (and believe that's what @vgrichina was experiencing) was that clicking 'resend' added another email instead of replaced it on profile page. But that should be fixed.

vgrichina:
@Patrick1904 I still cannot remove my second e-mail

Patrick1904:
I’ve tested and tried to break things but it's working fine after latest changes by @chadoh.

@kcole16 @vgrichina

vgrichina:
Screen Shot 2020-04-23 at 2 17 30 PM

somehow getting error when trying to setup a recovery method (using e-mail) on existing account

Patrick1904:
> somehow getting error when trying to setup a recovery method (using e-mail) on existing account

cc @chadoh 
I'm able to repro the same thing with an existing account that's in a 'bad state' from before the latest changes
Screen Shot 2020-04-23 at 3 12 17 PM

chadoh:
@vgrichina can you confirm whether this happens with a newly-created account, or an account otherwise known to be in a good state?

vgrichina:
@Patrick1904 @chadoh does setting up account recovery when there are no recovery methods work for you? I've tried it on testnet using `test-create.test` account and couldn't setup using any of the methods,

vgrichina:
> @Patrick1904 @chadoh does setting up account recovery when there are no recovery methods work for you? I've tried it on testnet using `test-create.test` account and couldn't setup using any of the methods,

ah, looks like that account somehow doesn't have any keys for me locally

PR #382, opened 2020-02-06
Fix navbar console warning

Patrick1904:


vgrichina:
@Patrick1904 not sure I understand the fix. Why pass `'true'` and `'false'` as strings?

Patrick1904:
> @Patrick1904 not sure I understand the fix. Why pass `'true'` and `'false'` as strings?

They're not actual booleans but styled components doesn't like boolean. I can do `account.accountId ? "true" : null` instead of `account.accountId ? "true" : "false"` but doesn't really matter

PR #399, opened 2020-02-20
feat: fetch recovery methods for current user

chadoh:
This adds some data-handling to make use of some of the data & endpoints added in nearprotocol/near-contract-helper#88

* refactor wallet to export instance, not class (see first commit for full explanation of & reasoning for this change)
* add action, reducer, and hook for fetching recovery methods
* fetch these on the profile page
* only fetch them if someone is viewing their own profile
* `addAccessKey` now POSTs to `near-contract-helper` to record the publicKey & time that this recovery method was added

## Testing

To make sure this works as expected, visit the [preview URL](https://near-wallet-staging-pr-399.onrender.com/) and create a new account. Keep your browser console open as you do this.

Visit your profile. In the Network tab of your inspector, you should see a successful POST to the `recoveryMethods` endpoint of `near-contract-helper`:

Screen Shot 2020-03-30 at 09 56 39

In the Console tab of your inspector, you should see the recovery methods logged out:

chrome console showing recoveryMethods array of length one

And if you have the [Redux Dev Tools extension for Chrome](https://chrome.google.com/webstore/detail/redux-devtools/lmhkpmbekcpmknklioeibfkpmmfibljd) installed, you can visit the Redux tab of your inspector and you should be able to see recovery methods loaded in:

Screen Shot 2020-04-03 at 14 58 37


vgrichina:
Is this hooked up? Seems to show fake data. Does this need to be combined with other PR from @Patrick1904?

Screen Shot 2020-03-10 at 12 17 35 AM

chadoh:
@vgrichina this lays some groundwork for #381 – it does not use the data, it just fetches it for now. #381 can branch off of this, or wait until this one gets merged to `master`. I thought a standalone PR would be useful so that this logic could be vetted separately.

src/utils/wallet.js
```diff
@@ -153,10 +154,31 @@ export class Wallet {
...
+
+        const recoveryMethods = {}
+        if (fromDb.email) {
+            recoveryMethods.email = { data: fromDb.email, timeStamp: fromDb.emailAddedAt }
+        }
+        if (fromDb.phoneNumber) {
+            recoveryMethods.phone = { data: fromDb.phoneNumber, timeStamp: fromDb.phoneAddedAt }
+        }
+        if (fromDb.phraseAddedAt) {
+            recoveryMethods.phrase = { timeStamp: fromDb.phraseAddedAt }
```
vgrichina: @chadoh @Patrick1904  I think we have pretty big problem that none of these have info on what are associated public keys, so we have no idea which key to remove if you want to replace/remove given recovery method.

vgrichina: My original plan to address it was to add `meta` to the access keys which is loaded from elsewhere:
https://github.com/nearprotocol/near-wallet/blob/b7bbfbeef97c64d668cfe07385b377a6d0bac6e3/src/utils/wallet.js#L170

vgrichina:
> this lays some groundwork for #381 – it does not use the data, it just fetches it for now. #381 can branch off of this, or wait until this one gets merged to master.

best to branch off this from #381 and get it connected. Stuff merged to `master` should be shippable (as it gets deployed as testnet wallet).

chadoh:
> Stuff merged to master should be shippable (as it gets deployed as testnet wallet).

This is shippable – it adds data to the redux store, and that's it. It has no adverse effect in the UI.

src/utils/wallet.js
chadoh: Oh, great point! Should we be storing that information in the `near-contract-helper` database?

Patrick1904:
> > Stuff merged to master should be shippable (as it gets deployed as testnet wallet).
> 
> This is shippable – it adds data to the redux store, and that's it. It has no adverse effect in the UI.

I agree. Once there are public keys associated with each recovery method, I think we should merge this to master. Once it's on master, I can hook it up with https://github.com/nearprotocol/near-wallet/pull/381 as there are some UI changes needed for it to work with this data structure.

src/utils/wallet.js
vgrichina: @chadoh I think long term we should be storing this information on chain using open web contract from @evgenykuzyakov. We can store it in near-contract-helper short term if it's easier to get working.

src/utils/wallet.js
```diff
@@ -153,10 +154,31 @@ export class Wallet {
...
+        const [ fromChain, fromDb ] = await Promise.all([
+            this.getAccount(this.accountId).state(),
+            sendJson('POST', `${ACCOUNT_HELPER_URL}/account/recoveryMethods`, {
+                accountId: this.accountId,
+                ...(await this.signatureFor(this.accountId))
+            })
+        ]);
+
+        const recoveryMethods = {}
+        if (fromDb.email) {
```
vgrichina: why not return data from contract helper in already desired format?

vgrichina:
> I agree. Once there are public keys associated with each recovery method, I think we should merge this to master. Once it's on master, I can hook it up with #381 as there are some UI changes needed for it to work with this data structure.

np, up to you folks to merge in whatever order works best, just make sure it doesn't bring anything down!

src/utils/wallet.js
chadoh: Done!

src/utils/wallet.js
chadoh: Stored in near-contract-helper now (pending merge of nearprotocol/near-contract-helper#88)

vgrichina:
@chadoh seems like it's working alright

> loadAccount now fetches recovery methods from near-contract-helper, and these recoveryMethods are added to the redux state in reducers/account/index

not sure why it's needed to load these every time we load account. Seems like it's only relevant for specific page (when recovery method setup is happening) unless I misunderstand smth.

vgrichina:
@Patrick1904 you might have better context here on when this info is needed (e.g. if we plan to use it to display banner on homepage to nudge user – then it indeed makes sense to have it in `loadAccount`) /cc @chadoh

Patrick1904:
> @Patrick1904 you might have better context here on when this info is needed (e.g. if we plan to use it to display banner on homepage to nudge user – then it indeed makes sense to have it in `loadAccount`) /cc @chadoh

Screen Shot 2020-02-06 at 11 14 40 AM

We'll need recovery information on the profile page for the recovery management module and I believe we're planning on using it to encourage setting up additional/specific methods as part of [communicating security expectations](https://github.com/nearprotocol/near-wallet/issues/442).

Patrick1904:
Screen Shot 2020-04-02 at 10 09 50 AM

@chadoh I'm not able to create account. Please see attached

Patrick1904:
@chadoh, btw, not sure if this has anything to do with this PR, but recovery links sent using onrender don't get a domain resulting in: `http://undefined/recover-with-link/someusername/phrase`.

chadoh:
@Patrick1904 looks like we were missing the `WALLET_URL` environment variable in the betanet-account-helper service which gets used here: https://github.com/nearprotocol/near-contract-helper/blob/45645748e9879ab3865fd893604ce1c5b8147700/app.js#L219

I've updated `WALLET_URL` to point at https://near-wallet-pr-399.onrender.com, so you should see correct links in these emails now.

@vgrichina great point about this data not being needed app-wide, and it therefore being wasteful to fetch it in the main `account` reducer. I'll update the logic here to add a new action & reducer.

chadoh:
@Patrick1904 @vgrichina I've reworked the logic here and the PR description. I've also updated the environment variables for https://near-wallet-pr-399.onrender.com to point at the main testnet endpoints now that nearprotocol/near-contract-helper#88 has been merged. Please re-review and let me know what you think!

src/reducers/recoveryMethods.js
```diff
@@ -0,0 +1,20 @@
...
+
+const initialState = {}
+
+const reducer = (state, event) => {
+    const { error, meta: { accountId }, payload, ready } = event
+
+    if (!ready) return state
+
+    if (error) {
+        console.error('error loading recoveryMethods!', payload)
```
vgrichina, line 12: nit: this is redundant, should be enough to have logging once in the middleware for executing promises

I would just do `if (!ready || error) return state` above (unless there is need to update state based on error).

src/utils/wallet.js
```diff
@@ -254,12 +250,20 @@ export class Wallet {
...
     async addAccessKey(accountId, contractId, publicKey) {
-        return await this.getAccount(accountId).addKey(
+        const res = await this.getAccount(accountId).addKey(
             publicKey,
             contractId,
             '', // methodName
             ACCESS_KEY_FUNDING_AMOUNT
         )
+
+        await sendJson('POST', `${ACCOUNT_HELPER_URL}/account/seedPhraseAdded`, {
```
vgrichina, line 260: this only needs to happen for `addAccessKeySeedPhrase` action, having it here would result in every app-specific access key to be registered as a seed phrase

chadoh: Great point! I've moved this logic to the action creator – I don't love the look of it, but maybe it's ok. Do you see a quick way to clean it up?

I have verified that it still works locally:

Screen Shot 2020-04-07 at 13 02 33

vgrichina: @chadoh I think it makes sense to move more logic to actions. `wallet.js` seems like just historically ended up being almost "god object" kind of thing.

We probably just need to modularize actions better (for now way too much stuff ended up in `account.js`)

PR #401, opened 2020-02-21
add code of conduct, contributing, license

mikedotexe:
Addresses: https://github.com/nearprotocol/near-wallet/issues/386
This is a request that came in a while ago, getting to it with the repos in closest reach.

I am not super familiar with this repo, but added some basics to the bottom of the README to run `./build_image.sh` but believe this could be fleshed out some more. If folks who touch this project more want to jump in on this PR that could be helpful. After running that command I got:

> Publish image to hub (y/n)? n

And said no. If we want to expand on this, that'd be great.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,76 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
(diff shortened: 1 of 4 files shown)


vgrichina:
@mikedotexe don't think anyone is actively using Docker setup at the moment

vgrichina:
@mikedotexe did you try running it locally? Generally should work our of box (connected to testnet) with `yarn start` + some tests should be runnable using `yarn test`. No Docker stuff needed to operate.

PR #405, opened 2020-02-22
Add Web Share Api w/ fallback

Patrick1904:
This replaces our current 'single tap to copy' ux with native share dialog for **supported mobile** users.

Currently used on:
/recover-with-link/username/phrase
/setup-seed-phrase/username
/receive-money


src/components/accounts/RecoverWithLink.js
```diff
@@ -117,12 +118,21 @@ class RecoverWithLink extends Component {
     }
 
     handleCopyUrl = () => {
-        const selection = window.getSelection();
-        selection.selectAllChildren(this.recoverUrl.current);
-        document.execCommand('copy');
+        if (window.matchMedia("(max-width: 767px)").matches) {
```
vgrichina: @Patrick1904 I think we should better use explicit check https://developer.mozilla.org/en-US/docs/Web/API/Navigator/canShare

note that desktop Safari also has this API, but copying the link isn't available further in the flow

src/utils/common.js
```diff
@@ -17,4 +17,28 @@ const classNames = (names) => {
 
 }
 
-export { classNames };
\ No newline at end of file
+const webShare = (data) => {
```
vgrichina: let's put this into separate file/module, doesn't seem very related to `classNames` in use

src/utils/common.js
```diff
@@ -17,4 +17,28 @@ const classNames = (names) => {
...
+    if (navigator.share) { 
+        navigator.share({
+           title: data.title,
+           text: data.text,
+           url: data.url
+        }).then(() => {
+            if (typeof data.successCallback === 'function') {
```
vgrichina: why not return promise (i.e. make `webShare` async function) instead of using callbacks, especially given that `navigator.share` is already returning a promise?

src/utils/common.js
```diff
@@ -17,4 +17,28 @@ const classNames = (names) => {
...
+                data.successCallback();
+            }
+        }).catch(console.error);
```
vgrichina: seems like it's more robust not just log error, but also do fallback method (i.e. copy)?

src/components/accounts/RecoverWithLink.js
Patrick1904: We decided to only show on **mobile** because copy ux is not great on desktop as mentioned [here](https://github.com/nearprotocol/near-wallet/issues/394). Therefore, I'm adding the check for browser width for this use case. Didn't want to put it in the util function since it limits future use.

I'm currently doing `if (navigator.share)` to check for support. I can replace with `navigator.canShare(data)` if needed?

src/utils/common.js
Patrick1904: Yes it's not related to classNames but they're both in common.js? Not sure if any of the current common functions deserve their own file?

src/components/accounts/RecoverWithLink.js
vgrichina: > I can replace with navigator.canShare(data) if needed?

Actually I just checked on desktop Safari and it's unavailable, while `.share` is available. Weird.

I think might be most robust to just hardcode iPhone/Android user agent checks, as current check would have false positive if you resize window and false negative on iPad.

src/utils/common.js
vgrichina: it's not like files are scarce, usually changes are easier to merge when unrelated stuff is in different file. Navigation is often also easier when you have `classNames.js`, `webShare.js` and can quickly jump by file name.

src/components/accounts/RecoverWithLink.js
Patrick1904: Sounds good should we use [this](https://stackoverflow.com/a/21742107/5574318) method of checking mobile device OS?

src/utils/common.js
Patrick1904: Makes sense I'll fix it

src/utils/common.js
Patrick1904: What do you think about this?

In `webShare.js`: 
```
class NoSupportError extends Error {}

const webShare = (data) => {
    if (navigator.share) { 
        return navigator.share({
           title: data.title,
           text: data.text,
           url: data.url
        })
    } else {
        return Promise.reject(new NoSupportError());
    }
}
```

Usage:

```
webShare({
            url: window.location.href,
}).catch(err => {
            if (err instanceof NoSupportError) {
                this.handleCopyDesktop();
            } else {
                console.error(err);
                this.handleCopyDesktop();
            }
});
```

Patrick1904:
I removed `navigator.share` as helper function because of the minimal boiler plate needed to use anyway.

Patrick1904:
> @Patrick1904 looks good to me. Do you want to merge this as is or add separate share button?

@kcole16 thoughts?
Screen Shot 2020-02-25 at 10 19 32 PM
Screen Shot 2020-02-25 at 9 53 53 PM

corwinharrell:
From @Patrick1904 

> I'm gonna add a generic-looking 'share' button and I pulled an icon from thenounproject but the style is not perfectly aligned with the 'copy phrase' icon that we're using. do you think we should align the design before implementing or this works?

@Patrick1904 I've added a new custom share icon `src/images/icon-share.svg` and pushed a commit. The icon should better match the style of our existing 18px solid set. Can change the fill color of the svg if needed.

Patrick1904:
@vgrichina - after discussion w/ @kcole16 and @corwinharrell - we decided to replace 'copy' w/ the 'share' icon on supported mobile devices on the `/receive-money` page only.

vgrichina:
@Patrick1904 makes sense to me. On iOS in `/recover-with-link/username/phrase` context it looks like if you opened in wrong browser – there is usually a built-in button to open in Safari (and it doesn't seem available as an action from share sheet).

Patrick1904:
> @Patrick1904 makes sense to me. On iOS in `/recover-with-link/username/phrase` context it looks like if you opened in wrong browser – there is usually a built-in button to open in Safari (and it doesn't seem available as an action from share sheet).

Yes it seems that iOS omits the 'Open in Safari' option in the dialog vs. showing the option if you click on a link in Slack, etc. I'm assuming that in most cases, the user will click on the recovery link from email or sms - and iOS defaults to Safari. So if that's the case the Chrome option is relatively prominent as an option in the dialog.

PR #434, opened 2020-03-05
Navigation re-factor

Patrick1904:
Effort to get rid of semantic-ui, clean up/reduce code, create re-usable components and make UI pixel perfect


vgrichina:
@Patrick1904 LGTM, but please resolve conflicts and make sure all UI strings are translatable before merging

corwinharrell:
@Patrick1904 I think these changes look great. Code is squeaky-clean too ✨ 

In response to your question in slack re the sticky header, I can see why having the header be sticky on the create account page doesn't carry huge benefit. That being said this isn't something I'd think users would really notice, so I don't have a strong opinion.

src/components/responsive/MobileView.js
```diff
@@ -1,321 +0,0 @@
...
-                                    {availableAccounts.length < 2 && <Translate id='desktopPopup.noAccount' />}
-                                </List>
-                                <Button onClick={this.redirectCreateAccount}>
-                                    <Translate id='button.createNewAccount' />
-                                </Button>
-                            </Segment>
-                        </Menu.Menu>
-                    </Segment>
-                </Segment>
-                <GlobalAlert />
```
marcinbodnar, line 315: @Patrick1904 I think you forgot about GlobalAlert, it's not appearing in the Wallet now.

Patrick1904: @marcinbodnar sorry I forgot to tell you that I left it out. It's being used properly in LoginDetails and a few other components. Seems suboptimal to default errors/successes to an inline 'GlobalAlert' at the top of the page, pushing down the rest of the content. It's quick/easy but UX suffers. A global 'status/flag' to reference makes sense. It makes sense to put a reusable fixed position components in Routing, like a modal, snackbar (I should have done that) since they can be added to any page regardless of the page content.

What is the current case that you're going to use the GlobalAlert for? Maybe there's a better way to display error/success in this case

Patrick1904: @marcinbodnar thanks for clarifying. I guess it’s the solution requiring the least effort right now. Reg login success, just note that we recently disabled the GlobalAlert since it only appeared for short sec before re-direct making it appear like a bug.   You can add the GlobalAlert component to Routing.js and it'll be available on every page. The nav is fixed so can’t be inside the nav component. You can put it below it and should work fine.

marcinbodnar: @Patrick1904 If that's not a problem, it would be better if you would add it since you were this part of the project and you know better where to put it now.

Patrick1904: Sure thing here's the commit: https://github.com/nearprotocol/near-wallet/pull/497

marcinbodnar: cool thank you

PR #440, opened 2020-03-09
Linkdrop account funding

Patrick1904:
https://github.com/nearprotocol/near-wallet/issues/409


src/components/accounts/CreateAccount.js
```diff
@@ -42,11 +48,11 @@ class CreateAccount extends Component {
...
-        const { accountId } = this.state
-
-        this.props.createNewAccount(accountId).then(({ error }) => {
-            if (error) return
+        const { accountId, fundingKey } = this.state
 
+        this.props.createNewAccount(accountId, fundingKey).then(({ error }) => {
+            if (error) return;
+                
             this.props.refreshAccount()
```
Patrick1904: @vgrichina , when creating and funding account using link drop, there’s an issue w/ `refreshAccount` and `history.push(nextUrl)`, throwing the following error: `Error loading account: [-32000] Server error: account testaccount33 does not exist while viewing.` Seems `createNewAccount` doesn’t finish properly before continuing. If I wrap them in a 3s timeout it works fine. Do you have any idea how to solve?

src/utils/wallet.js
```diff
@@ -209,15 +222,39 @@ export class Wallet {
...
+            viewMethods: [],
+            changeMethods: ['create_account_and_claim', 'claim'],
+            sender: LINKDROP_CONTRACT_ID
+        });
+        const publicKey = keyPair.publicKey.toString().split(':')[1]
+        await contract.create_account_and_claim({
+            new_account_id: accountId,
+            new_public_key: publicKey},
+            BOATLOAD_OF_GAS
+        ).catch(err => console.log(err));
```
vgrichina: best to avoid mixing `await` and manual usage of promises. So to catch error it's best to wrap into `try` / `catch`

src/utils/wallet.js
```diff
@@ -209,15 +222,39 @@ export class Wallet {
...
+                newAccountPublicKey: keyPair.publicKey.toString()
+            })
+        }
         await this.saveAndSelectAccount(accountId, keyPair);
+
+    }
+
+    async createNewAccountLinkdrop(accountId, fundingKey, keyPair) {
+        const near = await nearlib.connect(fundingConfig);
+        window.localStorage.setItem(`nearlib:keystore:${LINKDROP_CONTRACT_ID}:default`, `ed25519:${fundingKey}`)
```
vgrichina: why not use `keyStore` instead of manipulating storage directly?

src/utils/wallet.js
```diff
@@ -209,15 +222,39 @@ export class Wallet {
...
+        const contract = await near.loadContract(LINKDROP_CONTRACT_ID, {
+            viewMethods: [],
+            changeMethods: ['create_account_and_claim', 'claim'],
+            sender: LINKDROP_CONTRACT_ID
+        });
+        const publicKey = keyPair.publicKey.toString().split(':')[1]
+        await contract.create_account_and_claim({
+            new_account_id: accountId,
+            new_public_key: publicKey},
+            BOATLOAD_OF_GAS
```
vgrichina: this shouldn't be needed, nearlib is AFAIK already passing maximum prepaid gas by default (if it's not – we should update nearlib)

src/utils/wallet.js
```diff
@@ -209,15 +222,39 @@ export class Wallet {
...
 
-        const keyPair = nearlib.KeyPair.fromRandom('ed25519')
-        await sendJson('POST', CONTRACT_CREATE_ACCOUNT_URL, {
-            newAccountId: accountId,
-            newAccountPublicKey: keyPair.publicKey.toString()
-        })
+        if (fundingKey) {
+            await this.createNewAccountLinkdrop(accountId, fundingKey, keyPair);
+            window.localStorage.removeItem(`nearlib:keystore:${LINKDROP_CONTRACT_ID}:default`)
+            // TODO: Figure out better solution than adding and removing from local storage
```
vgrichina: use `keyStore`

src/utils/wallet.js
```diff
@@ -209,15 +222,39 @@ export class Wallet {
...
+                newAccountId: accountId,
+                newAccountPublicKey: keyPair.publicKey.toString()
+            })
+        }
         await this.saveAndSelectAccount(accountId, keyPair);
+
+    }
+
+    async createNewAccountLinkdrop(accountId, fundingKey, keyPair) {
+        const near = await nearlib.connect(fundingConfig);
```
vgrichina: why is separate instance with different config needed?

src/components/accounts/CreateAccount.js
vgrichina: @Patrick1904 I think it's bug in nearcore (at minimum there is mismatch of what finality is used for view calls and submitting transactions). Best to create minimum reproducible example in studio. I think it basically needs existing code from wallet, but also add code to view account state immediately after creation:

```
await contract.create_account_and_claim({
             new_account_id: accountId,
             new_public_key: publicKey},
             BOATLOAD_OF_GAS);
const account = await near.account(accountId);
await account.state();
```

src/utils/wallet.js
```diff
@@ -209,15 +211,49 @@ export class Wallet {
...
+                newAccountId: accountId,
+                newAccountPublicKey: keyPair.publicKey.toString()
+            })
+        }
         await this.saveAndSelectAccount(accountId, keyPair);
+
+    }
+
+    async createNewAccountLinkdrop(accountId, fundingKey, keyPair) {
+        const near = await nearlib.connect({
```
Patrick1904: @vgrichina can you please let me know exactly what I should do here?

src/components/accounts/CreateAccount.js
bowenwang1996: @vgrichina @Patrick1904 what vlad said is correct. By default view call will execute on final blocks, whereas transaction submission works with the latest head. I am not sure what is needed here exactly, but if you are fine with working with non-final result an easy fix would be just to pass the finality requirement to the rpc (probably not well supported on nearlib side right now. You might need to manually construct the view request). 

Actually what I said above only applies to latest code but not testnet. If you have encountered this on testnet it is probably a separate issue.

vgrichina: @bowenwang1996 
>  By default view call will execute on final blocks, whereas transaction submission works with the latest head. 

IMO this should be considered a bug. Default finality should be consistent everywhere.

> probably not well supported on nearlib side right now. 

also would be really nice to not add features to nearcore without adding the same feature on nearlib side

vgrichina: @bowenwang1996 also this is for sure happening on testnet right now. I guess we can temporary fix by spinning tx status check in a loop? Will this be too much load on RPC?

bowenwang1996: transactions cannot not work with latest block. That is how it is supposed to work. Is there some repro that I can take a look?

vgrichina: > transactions cannot not work with latest block. 

I don't understand what this means.

> Is there some repro that I can take a look?

Yes, this PR is repro. There is a link to running instance on Render (as for every PR). Try creating account with this link:

https://near-wallet-pr-440.onrender.com/create?funding_key 8A2DVKfvobNEyrKLJHcLWbPf5s6T1yf9dEX8kFfzAGgeSTbbRFeUiXNBsH4X23sTRgGo1mtTvQ9CoJZPLDwVE5b

it will redirect you to "sorry all accounts removed" or smth like that – because we couldn't find account after contract method which creates account has finished execution

bowenwang1996: I meant that transactions are naturally applied on top of the latest block of the chain.

vgrichina: @bowenwang1996 how does that prevent view calls from operating on previous block? Also maybe it's something else – e.g. send_tx_commit just returns without waiting for all operations to complete when there is create account call from contract?

bowenwang1996: It shouldn't prevent view calls from operating on previous blocks. What can happen is that if you do view on the state that is of the last final block and the block that includes the transaction that creates some account hasn't been finalized yet, this will lead to an error. But I don't think this is what happened given that we don't have this change on testnet yet. I'll take a look soon.

bowenwang1996: Actually the change I made on nearcore side is backward compatible so the problem I described shouldn't exist :)

bowenwang1996: I found the reason after some debugging. The receipts execution results are saved before the head is updated, which means that when querying the state immediately after the `send_tx_commit` returns, it will query the state root as of the previous chunk, not the new chunk which contains the receipts. I don't think this is a bug and you just need to wait maybe 100ms (for the block processing to finish) before querying the state.

Patrick1904: @bowenwang1996 @vgrichina, seems it needs to wait at least 1000ms before loading account. Barely noticeable from UX perspective but unsure what the threshold actually is and maybe not reliable?

vgrichina: >  I don't think this is a bug and you just need to wait maybe 100ms (for the block processing to finish) before querying the state.

@bowenwang1996 this is a bug. We cannot be relying on arbitrary client-side RPC timeouts for system to function properly. We can workaround this to fix current testnet, but it is still a state inconsistency bug on nearcore side.

vgrichina: 1000ms is pretty big from UX standpoint:
https://www.gigaspaces.com/blog/amazon-found-every-100ms-of-latency-cost-them-1-in-sales/

bowenwang1996: @vgrichina @Patrick1904 actually I was wrong. There is a bug on nearcore side that prevents the request from tracking all receipts result. `send_tx_commit` doesn't actually wait until all receipts are executed before returning. I will try to fix it but before it gets into testnet probably some hackery is required to make things work on wallet side.

bowenwang1996: I submitted a PR to fix this issue https://github.com/nearprotocol/nearcore/pull/2252

evgenykuzyakov: It's not how it suppose to work. The contract doesn't care about the account being created, that's why the transaction returns the result before the account is being created here:
https://github.com/nearprotocol/near-linkdrop/blob/master/src/lib.rs#L50

The way it should work contract should return the result of the promise creating a new account if it's important for the function. Here is how it should be implemented:
https://github.com/near-examples/pow-faucet/blob/master/contract-rs/src/lib.rs#L76

Notice function `create_account` returns `PromiseOrValue<()>` which means the function may continue execution if it returns a promise. To fix it in linkdrop contract you need to add `.into()` that will convert `Promise` into `PromiseOrValue`
See example here: https://github.com/near-examples/pow-faucet/blob/master/contract-rs/src/lib.rs#L106

evgenykuzyakov: Here is a linkdrop contract fix: https://github.com/nearprotocol/near-linkdrop/pull/3

Patrick1904: @bowenwang1996 @vgrichina from comments on #2252 it seems that it's not a nearcore bug? Any thoughts on how we can move forward with this?

src/utils/wallet.js
```diff
@@ -246,7 +247,7 @@ export class Wallet {
         await contract.create_account_and_claim({
             new_account_id: accountId,
             new_public_key: publicKey
-        });
+        }, BOATLOAD_OF_GAS);
```
Patrick1904: @vgrichina, the changes made to [linkdrop contract](https://github.com/nearprotocol/near-wallet/pull/440#discussion_r392423775) fixed it, but I can only make it work by manually adding gas. Do you know if there's a way around this right now?

Patrick1904: Funding keys:
`WvFanAHobBP3MsHY2Rckn8aNAiuV8rYxRpYZ4gybSz1dgjTWAAEytVYRza4GMsWApH8a5pydkuzfCSsfHBpGTqr`
`62zFv9jCMMvkjzEXMsp17aPJmGcf3Xh7JdZiAcKQs1qeSxLf1a8RiTCGaza9qZgeCgbnaj5Vmbz7om286jnMXFFs`

Patrick1904: @vgrichina do you have any advice on how to avoid adding gas manually? thanks

vgrichina: How much gas does tx cost?

Patrick1904: @vgrichina: https://explorer.nearprotocol.com/transactions/8VRCPvxWEUnn986DKJmeqmQLUZBR27J3DKz5mFtEdQ4Y

vgrichina: @Patrick1904 I think it should be fine with default attached gas value then (it's `10000000000000000` in latest nearlib).

src/utils/wallet.js
```diff
@@ -209,15 +212,42 @@ export class Wallet {
...
+    async createNewAccountLinkdrop(accountId, fundingKey, keyPair) {
+        const account = this.getAccount(LINKDROP_CONTRACT_ID);
+
+        await this.keyStore.setKey(
+            NETWORK_ID, LINKDROP_CONTRACT_ID, 
+            nearlib.KeyPair.fromString(fundingKey)
+        )
+
+        const contract = new nearlib.Contract(account, LINKDROP_CONTRACT_ID, {
+            viewMethods: [],
```
vgrichina: nit: don't really need `viewMethods` here

src/components/accounts/CreateAccount.js
```diff
@@ -25,6 +27,10 @@ class CreateAccount extends Component {
         }
     }
 
+    handleFundingKey = () => {
+        this.setState({fundingKey: new URLSearchParams(window.location.search).get('funding_key')});
```
vgrichina: I think it might be better not to rely on `window.location` to make sure all works well with router:
https://learnwithparam.com/blog/how-to-handle-query-params-in-react-router/

src/utils/wallet.js
Patrick1904: Interesting it works now without adding manually. It used to not work. Fixing now.

vgrichina: @Patrick1904 nearlib update may be what changed this.

vgrichina:
@Patrick1904 is this still supposed to be draft or ready for final review / merge?

Patrick1904:
> @Patrick1904 is this still supposed to be draft or ready for final review / merge?

Last fix will be: https://github.com/nearprotocol/near-wallet/pull/440#discussion_r396850474

src/utils/wallet.js
```diff
@@ -25,6 +25,8 @@ const ACCOUNT_ID_REGEX = /^(([a-z\d]+[-_])*[a-z\d]+[.@])*([a-z\d]+[-_])*[a-z\d]+
 export const ACCOUNT_ID_SUFFIX = process.env.REACT_APP_ACCOUNT_ID_SUFFIX || '.test'
 export const ACCOUNT_CHECK_TIMEOUT = 500
 
+const LINKDROP_CONTRACT_ID = 'linkdrop-test-1';
```
vgrichina: @Patrick1904 sorry that I missed this earlier, but I think we need to pass this as part of URL, e.g. `fundingContract`.

Patrick1904: @vgrichina I've added `fundingContract` to URL. E.g.

http://localhost:1234/create/linkdrop-test-1/xGY3TYSTmXEtqAT4C1L5vDdCwHBCmvvUM8XVpRuqHiu3u9ttoDMje3iZTbYFqrrwjGADWkRYNKztxjvctzoazEG

Issue #422, opened 2020-02-27
Multi-Device Support


kcole16:
### Overview
Users will likely access the web wallet from multiple devices. This transition should be both seamless and secure.

### Stories
As a user, I want to use the NEAR wallet on multiple devices, so I can easily access NEAR applications on desktop and mobile. 

As a user, I want to set limits on my transaction amounts, requiring two factor authentication above the limits, so I can use the web wallet to safely store $100-$1000 worth of assets.
 

### Flows

**Desktop to Mobile Flow**
1. User selects “Add a new mobile device” on desktop
2. Signs a transaction creating a new full access key
3. Scans QR code with mobile device (URL w/ full access `private_key`)
4. Redirected to “New Device Confirmation” screen on mobile device
5. Selects “Complete Device Setup”
- Send batch transaction adding new full access key and removing full access key contained in the URL
6. Redirected to “Profile”

**Mobile to Desktop Flow**
1. User selects “Already have an account? Add this device” on desktop
2. Redirected to “Add Device” page on desktop
3. Selects “Add this device”, QR code generated
4. Scans QR code on mobile device Scans QR code with mobile device (URL w/ `public_key`)
5. Confirms “Add device”, signs a transaction creating a full access key using the `public_key` from the QR code
6. Redirected to “New Device Confirmation” screen on mobile device
7. Selects “Complete Device Setup”

ilblackdragon:
I want to also add that right now it's unclear which devices which keys are from. e.g. in the list of Full Access Keys - which one is the key in the wallet UI vs keys from somewhere else.

Issue #289, opened 2019-12-18
Improve error handling


vgrichina:
Need to do fine-grained error handling once https://github.com/nearprotocol/nearcore/issues/1839 lands.

kcole16:
@vgrichina @marcinbodnar nearprotocol/nearcore#1839 is now merged

vgrichina:
@kcole16 it's not live yet.

kcole16:
@vgrichina Re: your comments on [Notion](https://www.notion.so/nearprotocol/Wallet-Sync-5171c3e1e4f9413ea5e1f1897b7418a3)

> I think it's counterproductive to list all possible errors. Instead it makes sense to start from what user can do in case of failure and distinguish recoverable / unrecoverable errors for given action.

How about determining the potential error states for these three categories, and what the user should do for each (as opposed to each action)?

- HTTP errors
- RPC errors
- Client-side errors

kcole16:
Adding more detail to the above:

- HTTP errors
    - 400
        - Redirect to home
    - 500
        - If error message: Display error
        - If not: "Something went wrong. Please try again!"
    - 400
        - If error message: Display error. e.g. #95
        - If not: Display "You may have entered something incorrectly. Please try again!"
- RPC errors
    - TX failed
        - Out-of-gas error: Resend with more gas?
        - Contract error: Redirect to app
    - RPC timed out
        - Check if TX succeeded
            - Yes: Redirect to success page
            - No: Display "The network may be running a bit slow. Please try again!"

Thoughts? @vgrichina @Patrick1904 @marcinbodnar @corwinharrell

marcinbodnar:
I gather some possible situations that are not handled, regarding user experience:

Create Account:
 - if you enter one char accountId, submit button is active and after click it shows that username is taken (issue #95)
 - situation when user enters incorrect char in accountId field, at this moment we are blocking that char, but don't display any information

Protect your Account:
 - if the user enters wrong email or phone number, we can display some sort of info, red frame for input field might be enough?

Send Money:
 - we should inform user when there is no money to Send - at this moment we are limiting to 5 decimal digits, so we can check if there is enough money at component mount
 - if user enters Send Money link with wrong username in URL we are showing info but user name field can be filled:
https://wallet.nearprotocol.com/send-money/marcinx

Login:
 - if there is no account, and we are redirecting user to Create Account page we can inform him that he will be redirected to Login page after account will be created
[...]

PR #443, opened 2020-03-13
Fix account creation when the network is slow, with flow changes

marcinbodnar:
This is the second attempt for #349 changes, with a different flow. 

With these changes when the account exists in local storage but does not exist in the blockchain, we will clear the account (only this one) from local storage. Next:
1. If the user has another account we will load the account, leave him on the same page and show GlobalAlert with info. 
2. If the user does not have another account he will be redirected to Create Account and we will show the same globalAlert.

#349 can be closed.

<a href="https://gitpod.io/#https://github.com/nearprotocol/near-wallet/pull/443"></a>

### DIFF
--- a/src/utils/localStorage.js
+++ b/src/utils/localStorage.js
@@ -1,25 +1,13 @@
 import { keyAccountConfirmed } from './wallet'
 
 export const setAccountConfirmed = (accountId, confirmed) => {
-    try {
-        localStorage.setItem(keyAccountConfirmed(accountId), confirmed)
-    } catch (err) {
-        console.warn(err)
-    }
+    localStorage.setItem(keyAccountConfirmed(accountId), confirmed)
 }
 
 export const getAccountConfirmed = (accountId) => {
-    try {
-        return localStorage.getItem(keyAccountConfirmed(accountId)) === 'true'
-    } catch (err) {
-        console.warn(err)
-    }
+    return localStorage.getItem(keyAccountConfirmed(accountId)) === 'true'
 }
 
 export const removeAccountConfirmed = (accountId) => {
-    try {
-        localStorage.removeItem(keyAccountConfirmed(accountId))
-    } catch (err) {
-        console.warn(err)
-    }
+    localStorage.removeItem(keyAccountConfirmed(accountId))
 }
--- a/src/utils/localStorage.js
+++ b/src/utils/localStorage.js
@@ -1,24 +1,24 @@
-const keyAccountConfirmed = (accountId, networkId) => `wallet.account:${accountId}:${networkId}:confirmed`
+import { keyAccountConfirmed } from './wallet'
 
-export const setAccountConfirmed = (accountId, networkId, confirmed) => {
+export const setAccountConfirmed = (accountId, confirmed) => {
     try {
-        localStorage.setItem(keyAccountConfirmed(accountId, networkId), confirmed)
+        localStorage.setItem(keyAccountConfirmed(accountId), confirmed)
     } catch (err) {
         console.warn(err)
     }
 }
 
-export const getAccountConfirmed = (accountId, networkId) => {
+export const getAccountConfirmed = (accountId) => {
     try {
-        return localStorage.getItem(keyAccountConfirmed(accountId, networkId)) === 'true'
+        return localStorage.getItem(keyAccountConfirmed(accountId)) === 'true'
     } catch (err) {
         console.warn(err)
     }
 }
 
-export const removeAccountConfirmed = (accountId, networkId) => {
+export const removeAccountConfirmed = (accountId) => {
     try {
-        localStorage.removeItem(keyAccountConfirmed(accountId, networkId))
+        localStorage.removeItem(keyAccountConfirmed(accountId))
     } catch (err) {
         console.warn(err)
     }
--- a/src/utils/localStorage.js
+++ b/src/utils/localStorage.js
@@ -1,0 +1,25 @@
+const keyAccountConfirmed = (accountId, networkId) => `wallet.account:${accountId}:${networkId}:confirmed`
+
+export const setAccountConfirmed = (accountId, networkId, confirmed) => {
+    try {
+        localStorage.setItem(keyAccountConfirmed(accountId, networkId), confirmed)
+    } catch (err) {
+        console.warn(err)
+    }
+}
+
+export const getAccountConfirmed = (accountId, networkId) => {
+    try {
+        return localStorage.getItem(keyAccountConfirmed(accountId, networkId)) === 'true'
+    } catch (err) {
+        console.warn(err)
+    }
+}
+
+export const removeAccountConfirmed = (accountId, networkId) => {
+    try {
+        localStorage.removeItem(keyAccountConfirmed(accountId, networkId))
+    } catch (err) {
+        console.warn(err)
+    }
+}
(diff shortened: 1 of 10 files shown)


vgrichina:
> If the user has another account we will load the account, leave him on the same page and show GlobalAlert with info.

@marcinbodnar @kcole16 

I think this is confusing in case user was just creating account. I think it makes sense to keep separate flag in local storage indicating whether given account was ever successfully loaded. If it never was loaded successfully – it makes sense to bring user back to account creation page (with account name already populated).

What do you think?

marcinbodnar:
@vgrichina 
> keep separate flag in local storage indicating whether given account was ever successfully loaded

Sure, sounds good. How it should behave in case if an account was loaded earlier? Stay on the same page and show GlobalAlert with info?

Also, how would you like to indicate it in local storage?

kcole16:
@marcinbodnar 

Agree with @vgrichina, I think we should deal with both cases in the same way (whether user has another account or not):

Redirecting the user to Create Account, and pre-populating the accountName field with the name they attempted to create.

Showing GlobalAlert makes sense as well, though this has some overlap with #289, so we should try to generalize it.

vgrichina:
@kcole16 do you have any comments on:

>  How it should behave in case if an account was loaded earlier? Stay on the same page and show GlobalAlert with info?

In this case basically account got removed from network after creation (basically means that hard fork happened). I assume in this case it actually makes sense to find which of accounts isn't removed and switch to it (or go to create account page if there are no valid accounts).

kcole16:
> In this case basically account got removed from network after creation (basically means that hard fork happened).

Ah, got it. I suppose it depends on the time between initial account creation and hard fork. Normally, this will occur almost immediately, right?

If it occurs within a few seconds of account creation, I think it makes sense to give the user the opportunity to retry. If it's significantly later, then switching the account to a valid one and providing some info with GlobalAlert seems like a good solution.

vgrichina:
> Also, how would you like to indicate it in local storage?

something like this maybe?

`wallet:account:<accountId>:<networkId>:confirmed` -> `true`

marcinbodnar:
@vgrichina @kcole16 
>>I think we should deal with both cases in the same way (whether user has another account or not)

The idea with staying on the same page was predicted mostly for Login situations, if the user will be redirected from Login to Create Account it will not be easily possible to get back to Login page without creating an account, and the user would want to just switch account.

It might also be a situation (in the future) for a user that will remove the account from one device/browser and then use another device/browser to enter Wallet.

I think also important thing is to reload user (or give an opportunity to reload) to another account, without that he will lose the possibility to switch accounts without creating a new one.

We can also resolve it in another way, for example in all situations we can show page with two options (something similar was mentioned somewhere):
1. Show user all available accounts and allow to log in - in this case, user will be redirected back to the previous page.
2. And also show Create Account button - on click user will be redirected to Create Account page with accountId field filled.

kcole16:
@marcinbodnar Ah, got it, was thinking purely about the Create Account flow.

As @vgrichina mentioned, the most likely case where a user will try to load an account that does not exist will be due to a hard fork. In this case, I think the flow of loading another account from localStorage (if one exists) or moving to the Create Account flow (if not), makes sense.

Re: multi-device: users will not be able to remove accounts from devices remotely. They will be able to invalidate keys on other devices (by deleting them from an account), but this is a pretty different scenario (which we will need to handle post #422).

marcinbodnar:
@vgrichina 
Done. The flag is added as `false` while creating an account, next if the account was loaded successfully then the flag will be changed to `true`.

While loading account, If the account doesn't exists in blockchain it will check the flag and if it's:
 - `true` - it will leave the user at the same page and show alert
 - `false` - it will redirect to Create Account page with accountId filled (and checked) and show alert.
In both cases, it will check for next confirmed account and load it.

It was a bit problematic to implement but it's working well.

src/utils/localStorage.js
```diff
@@ -0,0 +1,25 @@
+const keyAccountConfirmed = (accountId, networkId) => `wallet.account:${accountId}:${networkId}:confirmed`
+
+export const setAccountConfirmed = (accountId, networkId, confirmed) => {
+    try {
+        localStorage.setItem(keyAccountConfirmed(accountId, networkId), confirmed)
+    } catch (err) {
```
vgrichina, line 6: I think we shouldn't ignore errors here, makes it hard to notice if things went wrong. If we don't have local storage working – we won't be able to store keys for user anyway.

src/utils/localStorage.js
```diff
@@ -0,0 +1,25 @@
+const keyAccountConfirmed = (accountId, networkId) => `wallet.account:${accountId}:${networkId}:confirmed`
```
vgrichina, line 1: I think's it's a bit misleading to put this into `utils/localStorage.js`, as it is specific to storing wallet accounts and other stuff touching `localStorage` isn't using this module. I feel like it makes sense to keep these in `wallet.js`.

marcinbodnar: Done. My idea was to avoid circular dependencies. But code is simpler and it's working fine.

src/utils/localStorage.js
marcinbodnar: Done. I noticed that if localStorage is disabled the Wallet is not loading at all (only console error is shown), maybe we should handle it by adding a page with info for user?

kcole16: @marcinbodnar Good point! I'd expect the % of users with localStorage disabled to be low, so perhaps using a modal to display "You must enable Cookies in order to use this application" is enough

marcinbodnar: @kcole16 we can, but the thing is that local storage availability should be checked as first thing, in this file (if we don't want to complicate stuff too much):
https://github.com/nearprotocol/near-wallet/blob/master/src/index.js
So there will be not much to show undernith modal element.

Please add issue for this, I will think about it.

vgrichina: @kcole16 @marcinbodnar I think probably makes sense to have static html landing for cases when either some basic tech we need (like localStorage) is disabled or JS altogether is disabled. This also can provide useful content for web crawlers, etc

marcinbodnar: Sounds good to me. Maybe also some basic info/links on how to change it.

marcinbodnar:
@vgrichina ready to merge

vgrichina:
let's wait for https://github.com/nearprotocol/nearcore/issues/2039 to land before this PR

marcinbodnar:
ok

bowenwang1996:
@vgrichina @kcole16 
> In this case basically account got removed from network after creation 

It doesn't necessarily imply a hard fork. Can just be a reorg. So I suggest wallet always query the final state (if it is not doing so already).

frol:
> let's wait for nearprotocol/nearcore#2039 to land before this PR

That is resolved and the RPC APIs are available on testnet, betanet, and devnet.

marcinbodnar:
@vgrichina do we want to merge it now? If so I will rebase.

vgrichina:
@marcinbodnar looks like necessary nearcore changes have lended – let's revisit this PR

marcinbodnar:
@vgrichina Please let me know if we want to merge it, or if any additional changes are needed. if it's ok, I will rebase it.

marcinbodnar:
@vgrichina PR is rebased, also made changes for error types, please check if it's ok:
https://github.com/near/near-wallet/pull/443/commits/22993329bdc5e6354bcf4f2027dbcd74b8a5f013

kcole16:
Seems to be working fine. 

Pretty orthogonal, but noticed we still have the issue of "Checking availability..." popping up when the user clicks "Create".

vgrichina:
@kcole16 going to merge this now to avoid conflicts. Let's pull that into separate issue?

There is also another thing I discussed with @marcinbodnar already – when network is going slow it actually may happen that account still doesn't exist when we check later. So need another change to be robust against it – record timed out transaction and then repeat it.

PR #445, opened 2020-03-13
fix license to be dual license MIT and Apache

mikedotexe:
Discussions have landed that while `nearcore` will have the GNU license, many other repositories will have a dual license of MIT and Apache. This is one of those repositories.


vgrichina:
@mikedotexe friendly ping

mikedotexe:
> @mikedotexe friendly ping

thanks, pushed

PR #454, opened 2020-03-18
Recovery setup re-design

Patrick1904:
https://github.com/nearprotocol/near-wallet/issues/355


kcole16:
@corwinharrell Noticed the submit button is below the fold on my 13 inch Mac. @Patrick1904 And I were thinking to cut the h2, and reduce the margin of the h1.

Screen Shot 2020-03-23 at 6 02 38 PM

Also considering adding a hoverable "?" next to "Basic Security" and "Advanced Security" with additional details around the recovery process and the tradeoffs

vgrichina:
@kcole16 
> Noticed the submit button is below the fold on my 13 inch Mac. 

that is also happening on like all the phones and not solvable with just changes to page heading. I think ultimately it's a non issue as users nowadays are more used to scrolling.

vgrichina:
@kcole16 also let's make sure to test everything on mobile.

I'll also get us licenses of https://sizzy.app to make it easy to check multiple form factors at once

kcole16:
> that is also happening on like all the phones and not solvable with just changes to page heading. I think ultimately it's a non issue as users nowadays are more used to scrolling.

The mobile experience is much better than the desktop one currently, as autofocus makes the submit button clearly visible. Scrolling is fine on mobile, less so on desktop.

Good call on sizzy.app!

corwinharrell:
@kcole16 I'm in general agreement with @vgrichina about content below the fold. Whenever this topic comes up I usually share this: https://www.hugeinc.com/articles/everybody-scrolls. Others might disagree but I think that we are at a point where people operate digital products with an expectation to scroll and that its not something thats going to break an experience.

In our scenario, the main thing that results in such large content height is our typographic scale and use of white space. Our typography on desktop is quite large for a utilitarian UI. It's more appropriate for marketing sites like our own, where the same styles are used I believe.

For a UI, especially one which plays a role of displaying large amounts of activity and financial / blockchain related data, the typography needs to be a bit tighter and contain less variation. A UI is likely to contain a larger number of interactive elements, components, and labels than just a static page with text. The result of having a scale with such large variations between different font-sizes is that everything overall just ends up being bigger and taking up more space than it needs to. This is something I've noted as a possible thing to explorer in a redesign.

Ultimately I think in the context of a UI like this you should be able to display more than just 3 inputs and a button on a page when viewing it on desktop, so just wanted to keep that in for future changes.

corwinharrell:
There's part of me that actually feels like we may have stripped out _too much_ content that may have provided some good context, but I may need to look back at it to be sure.

src/translations/en.global.json
```diff
@@ -184,31 +185,20 @@
...
-        "useInstead": {
-            "email": "Use email instead",
-            "phoneNumber": "Use phone instead"
-        },
-        "switchToManualBackup": {
-            "title": "Switch to Manual Backup",
-            "text": "With this option, you must write down and save a 12 word phrase to recover your account.",
-            "button": "Setup <span>Recovery Phrase</span>"
-        }
+        "subTitleOne": "Select the method you would like to use if you ever need to recover your account.",
```
kcole16: "Select an account recovery method:"

kcole16:
@corwinharrell @vgrichina Fair points. I'm open to continuing with the current version (with some copy tweaks) for now, until we make broader font and spacing changes. 

We can monitor for user behavior changes/user testing feedback on this screen in the meantime. 

@Patrick1904 Thoughts?

vgrichina:
@Patrick1904 please resolve conflicts

Patrick1904:
> @corwinharrell @vgrichina Fair points. I'm open to continuing with the current version (with some copy tweaks) for now, until we make broader font and spacing changes.
> 
> We can monitor for user behavior changes/user testing feedback on this screen in the meantime.
> 
> @Patrick1904 Thoughts?

I don't have a strong opinion on this as I see pro's and con's with both approaches. Let's go with current and get a feel for it and hopefully some user feedback and we can adjust accordingly.

PR #466, opened 2020-03-23
Change actions query yf

### LINKED ISSUE #492: Activity on homepage doesn't match explorer
**Problem**

Activity on homepage doesn't match explorer for account created through linkdrop:
https://explorer.nearprotocol.com/accounts/test-linkdrop

**Expected Behavior**

Activity in explorer and wallet is the same

**Steps to reproduce**

1. Create new account using linkdrop URL.
2. Observe absence of activity on homepage.
3. Click "View All" to see activity in explorer (it's not empty now)

icerove:
actions is separated from transactions and make it a separate table already. Update actions query from explorer backend database and wait for some other DB changes for explorer. After Explorer ready, make frontend UI proper

fix #492


src/components/dashboard/DashboardDetail.js
```diff
@@ -83,24 +83,24 @@ class DashboardDetail extends Component {
                     <DashboardActivity
                         loader={loader}
                         image={activityGreyImage}
-                        title=<Translate id='dashboard.activity' />
+                        title={<Translate id='dashboard.activity' />}
```
vgrichina: Why do we need changes in this file? Seem unrelated to main topic of PR.

src/utils/explorer-api.js
```diff
@@ -39,8 +38,31 @@ export async function getTransactions(accountId = '') {
...
+                `SELECT actions.action_type as kind, 
+                        actions.action_args as args
+                FROM actions
+                WHERE actions.transaction_hash = :hash
+                ORDER BY actions.action_index`,
+                  {
+                    hash: t.hash
+                  }
+            ]
+        ))
```
vgrichina: This doesn’t call resolve/reject so cannot work properly. I suggest wrapping wamp.call into helper function which handles promise creation automatically.

src/utils/explorer-api.js
```diff
@@ -39,8 +38,31 @@ export async function getTransactions(accountId = '') {
...
+        tx.actions = actions.map(action => {
+              if (typeof action === "string") {
+                return { [action]: {} };
+              } else {
```
vgrichina: Nit: else isn’t really needed given that previous statement uses return.

src/components/dashboard/DashboardDetail.js
icerove: touch this file to change postprocessSerdeStruct function

vgrichina: I don't understand. Are you committing this because you needed to trigger auto-compile locally?

icerove: I will change back. when I first search where actions display, I fount this file. it gives the error if without curly braces. so I change them.

vgrichina:
1) why do you remove `yarn.lock`?
2) my understanding is that with this change retrieving user activity requires N+1 queries to server (where N is number of transactions). This sounds like super inefficient from both client and server POV. Why is it necessary? Why cannot data be joined on server instead?

icerove:
1. remove it accidentally and change back
2. we now separate action table from transaction table and action has to be query from transaction hash. This is for explorer backend improvement.

vgrichina:
> we now separate action table from transaction table and action has to be query from transaction hash. This is for explorer backend improvement.

@icerove /cc @frol 

This doesn't make sense to me. Backend now has to process 10x requests and make 10x more DB queries to get info necessary to display 10 transactions. How does it improve backend performance?

frol:
@icerove You have an [N+1 query problem](https://stackoverflow.com/questions/97197/what-is-the-n1-selects-problem-in-orm-object-relational-mapping) here. It is fine to use a single query with a JOIN here (select transactions, order them by block_timestamp, limit by a number, and join their actions - as you can see, since ordering and limiting happen on the transactions table, it only needs to find and join actions to a dozen transactions, which is fast), or alternatively do two queries: select transactions and then select actions for the given transaction hashes (`WHERE transaction_hash IN ( list of transaction hashes )`).

vgrichina:
@frol  Just noticed `[action.kind]: JSON.parse(action.args)` BTW, another improvement possible is to use `JSONB` in Postgres for storing JSON. This is both more compact and can be faster to query (you can even index paths in JSON).

frol:
@vgrichina Postgres turned out to be slower for our common queries (well, in the previous DB structure, where we had excessive ORDER BY over JOINed columns), and it has a track of [slow COUNT queries](https://wiki.postgresql.org/wiki/Slow_Counting). We still have it in our Icebox, but I hope we won't need to do this (meaning that the new indexer https://github.com/nearprotocol/near-explorer/issues/236 can hopefully resolve this before we hit the ceiling)

icerove:
@frol  @vgrichina 
I change query to this:
image

vgrichina:
@icerove when I try using it with newly created account at https://near-wallet-pr-466.onrender.com/ I get blank screen and following error:

Uncaught (in promise) Error: amount property should not be null

Same error is reproduced on https://betanet-wallet-pr-466.onrender.com

Please test your changes in production settings, that's why we have auto-deploy for every PR – stuff can work differently that in your local setup.

src/utils/explorer-api.js
```diff
@@ -12,22 +12,16 @@ export async function getTransactions(accountId = '') {
...
-                LIMIT :offset, :count
+            `SELECT 
+                    hash, signer_id, receiver_id, block_hash, 
+                    block_timestamp, action_type as kind, action_args as args
+            from 
+                (SELECT hash, signer_id,  receiver_id, block_hash,  block_timestamp
+                    FROM  transactions
+                    WHERE  signer_id = :accountId OR receiver_id = :accountId
+                    ORDER BY block_timestamp DESC
+                    LIMIT :offset ,:count) as transactions
```
vgrichina, line 23: nit: `:offset, :count`

src/utils/explorer-api.js
```diff
@@ -12,22 +12,16 @@ export async function getTransactions(accountId = '') {
...
+                    LEFT JOIN actions ON actions.transaction_hash = transactions.hash
```
vgrichina, line 24: nit: needs to be pulled back in indentation IMO as now it looks like this join is part of inner `(SELECT`, but in fact it is part of outer `SELECT`.

src/utils/explorer-api.js
```diff
@@ -12,22 +12,16 @@ export async function getTransactions(accountId = '') {
...
-                    signer_id = :accountId 
-                    OR receiver_id = :accountId
-                ORDER BY blocks.height DESC
-                LIMIT :offset, :count
+            `SELECT 
+                    hash, signer_id, receiver_id, block_hash, 
+                    block_timestamp, action_type as kind, action_args as args
+            from 
+                (SELECT hash, signer_id,  receiver_id, block_hash,  block_timestamp
+                    FROM  transactions
```
vgrichina, line 20: nit: `FROM transactions`

src/utils/explorer-api.js
```diff
@@ -12,22 +12,16 @@ export async function getTransactions(accountId = '') {
...
-                LEFT JOIN blocks ON blocks.hash = transactions.block_hash
-                WHERE 
-                    signer_id = :accountId 
-                    OR receiver_id = :accountId
-                ORDER BY blocks.height DESC
-                LIMIT :offset, :count
+            `SELECT 
+                    hash, signer_id, receiver_id, block_hash, 
+                    block_timestamp, action_type as kind, action_args as args
+            from
```
vgrichina, line 18: nit: `FROM`

src/utils/explorer-api.js
```diff
@@ -12,22 +12,16 @@ export async function getTransactions(accountId = '') {
...
-                    blocks.timestamp as blockTimestamp
-                FROM 
-                    transactions
-                LEFT JOIN blocks ON blocks.hash = transactions.block_hash
-                WHERE 
-                    signer_id = :accountId 
-                    OR receiver_id = :accountId
-                ORDER BY blocks.height DESC
-                LIMIT :offset, :count
+            `SELECT
```
frol, line 15: Please, test this query:

```SQL
SELECT 
    transactions.hash, transactions.signer_id, transactions.receiver_id, transactions.block_hash, 
    transactions.block_timestamp, actions.action_type as kind, actions.action_args as args
FROM transactions
LEFT JOIN actions ON actions.transaction_hash = transactions.hash
WHERE transactions.signer_id = :accountId OR transactions.receiver_id = :accountId
ORDER BY block_timestamp DESC
LIMIT :offset, :count
```

src/components/dashboard/DashboardDetail.js
```diff
@@ -115,14 +115,14 @@ const mapDispatchToProps = {
 }
 
 // make sure that an action is an object, for UI purpose
-const postprocessSerdeStruct = (action) => typeof action == 'object' ? [action] : [{[action]: {}}]
+const postprocessSerdeStruct = (action) => typeof action == 'object' ? [{[action.kind]: [action.args]}] : [{[action]: {}}]
 
 const mapStateToProps = ({ account }) => {
     const transactions = account.transactions 
-        ? account.transactions.flatMap(t => t.actions.map((a) => ({
+        ? account.transactions.map(t => ({
```
frol, line 122: @vgrichina Interesting, so we don't care if those actions were a part of a single transaction; is this intended?

src/components/dashboard/DashboardDetail.js
```diff
@@ -115,14 +115,14 @@ const mapDispatchToProps = {
 }
 
 // make sure that an action is an object, for UI purpose
-const postprocessSerdeStruct = (action) => typeof action == 'object' ? [action] : [{[action]: {}}]
+const postprocessSerdeStruct = (action) => typeof action == 'object' ? [{[action.kind]: [action.args]}] : [{[action]: {}}]
```
frol, line 118: I guess, this method is useless since we normalize actions with the DB, and there is only a single format now.

src/components/dashboard/DashboardDetail.js
vgrichina: @frol any reason for us to care? It's displayed as a flat list anyway. Batch transaction concept as we have now generally isn't very useful, as it only works for same receiver (while in apps it's most interesting to batch calls to different receivers).

PR #467, opened 2020-03-24
Fix /profile links to use accountId everywhere

vgrichina:


src/components/sign/Sign.js
```diff
@@ -23,7 +23,8 @@ class Sign extends Component {
     }
 
     handleAddFunds = () => {
```
vgrichina: @Patrick1904 do you have any suggestion on what's best way to test this? See here how I tried to trigger:
https://github.com/nearprotocol/near-wallet/pull/467/commits/a4b610743bd4c6b8cde05ebed9f539cdb9d19538

src/components/accounts/RecoverWithLink.js
```diff
@@ -146,7 +146,8 @@ class RecoverWithLink extends Component {
                     this.setState({ successView: false });
                 } else {
                     this.props.refreshAccount();
-                    this.props.history.push('/profile');
+                    // TODO: Should this use Redux action to navigate?
+                    this.props.history.push(`/profile/${this.state.accountId}`);
```
Patrick1904: I'm not sure why we'd use Redux here? Can you please provide more info.

src/components/sign/Sign.js
Patrick1904: Not off the top of my head. Some bugs I fix could have been prevented with proptypes. Not this one though.

src/components/accounts/RecoverWithLink.js
vgrichina: @Patrick1904 maybe it already works fine, but I think it might be needed so that redux dev tools don't conflict with history changes (but instead record them as actions).

In any case it's kinda trivial:
https://github.com/supasate/connected-react-router/blob/master/FAQ.md#with-storedispatch

src/components/sign/Sign.js
vgrichina: @Patrick1904 I'm more concerned on how to actually trigger the related Add Funds UI (as I failed to do it) vs how to prevent bugs like this from happening.

PR #471, opened 2020-03-24
Fix navbar username/balance overflow

Patrick1904:


vgrichina:
so when I make window smaller – everything fits:
Screen Shot 2020-03-24 at 12 00 46 PM

when I make window bigger – username doesn't fit, but there is a lot of whitespace
Screen Shot 2020-03-24 at 12 01 05 PM

I think we need to expand rightmost panel with username when window is bigger.

vgrichina:
note that most usernames would be longer than what seems to be currently expected. E.g. `vgrichina.testnet` is an example of relatively short name, there's going to be longer names for sure (think of them as domains, not Twitter usernames).

Patrick1904:
> so when I make window smaller – everything fits:
> Screen Shot 2020-03-24 at 12 00 46 PM
> 
> when I make window bigger – username doesn't fit, but there is a lot of whitespace
> Screen Shot 2020-03-24 at 12 01 05 PM
> 
> I think we need to expand rightmost panel with username when window is bigger.

Yes I'll increase the username 'max-width' for larger screens.

@corwinharrell , maybe we should consider re-designing the username/user balance layout to work better with long usernames and balances? It doesn't seem optimal to apply 'ellipsis' treatment on this information.

Patrick1904:
> note that most usernames would be longer than what seems to be currently expected. E.g. `vgrichina.testnet` is an example of relatively short name, there's going to be longer names for sure (think of them as domains, not Twitter usernames).

@vgrichina I've added adjustments based on screen width and seems to handle long usernames and balances well. Please check it out.

vgrichina:
Screen Shot 2020-03-24 at 1 27 40 PM

wonder if we should switch to mobile layout for navbar earlier

Patrick1904:
> Screen Shot 2020-03-24 at 1 27 40 PM
> 
> wonder if we should switch to mobile layout for navbar earlier

@vgrichina I changed to show mobile at 991px. I think we should consider making adjustments to design/layout if we need to switch higher than that.

PR #475, opened 2020-03-26
Append network name to username

Patrick1904:
https://github.com/nearprotocol/near-wallet/issues/407

<a href="https://gitpod.io/#https://github.com/nearprotocol/near-wallet/pull/475"></a>


vgrichina:
Screen Shot 2020-03-25 at 10 39 14 PM

I feel like `.test` is too far from username, I'm likely to not even notice it. 

Also this shouldn't be happening when e.g. when choosing account to send money to (as it's not like all accounts will be forced to have same suffix, only the ones created through our wallet flow).

@kcole16 @corwinharrell wonder what's your thoughts on this?

Patrick1904:
> Screen Shot 2020-03-25 at 10 39 14 PM
> 
> I feel like `.test` is too far from username, I'm likely to not even notice it.
> 
> Also this shouldn't be happening when choosing account to send money to (as it's not like all accounts will be forced to have same suffix, only the ones created through our wallet flow).
> 
> @kcole16 @corwinharrell wonder what's your thoughts on this?

Good catch @vgrichina. Now showing network on account creation only.

As for '.test' being barely visible: I find it quite clear even though it's subtle. I got the inspiration [here](https://github.com/nearprotocol/near-wallet/issues/407#issuecomment-590535940). But it depends how important it is to convey network name when creating an account. I'm assuming that in the mainnet (.near) case, conveying network is less important since it's the 'default', but I might be wrong? For testnet, betanet and devnet cases, I see it being more important to convey network name so that users are not under the impression that they're creating a 'real' account. If those assumptions are correct, maybe non-mainnet cases deserve more explicit network status by adding color to the indicator or/and explicit disclaimer text around text input.

kcole16:
@Patrick1904 I do think it's worth making the suffix more noticeable. Mainly because we want to clearly communicate that the suffix is part of the account name (and partially to indicate network, though the other indicators will help as well).

Simple way to do this might be making the suffix font size slightly larger, and in black.

Also, this is definitely a corner (possibly edge) case, but shows like this for me (same when resized).

Screen Shot 2020-03-26 at 1 34 18 PM

Patrick1904:
Screen Shot 2020-03-26 at 2 45 54 PM

Patrick1904:
Screen Shot 2020-03-26 at 3 50 09 PM

Should we still keep icon/checkmark for non account creation cases?

Patrick1904:
@kcole16 @corwinharrell should we still show the checkmark/error icons on the right side of the text input for non account creation cases (when we don't have to show network)? Currently that's on /send and /recovery-seed-phrase pages

kcole16:
> @kcole16 @corwinharrell should we still show the checkmark/error icons on the right side of the text input for non account creation cases (when we don't have to show network)? Currently that's on /send and /recovery-seed-phrase pages

It's a nice touch, though highlighting the input outline is enough IMO. I'd say go with the more straightforward implementation.

vgrichina:
@kcole16 @Patrick1904 this works fine, but I really have trouble accepting how it looks. 
I don't think we make it clear that `.test` is a part of account name (it's way too far on the right, can be really anything, there is no explanation where we explain how names work).

Also loader spinner is now overlapping `.test` which looks ugly. I think at the bare minimum it needs to be hidden (button already shows spinner too).


> when do you think we can change to '.testnet'?

someone needs to create `testnet` accounts with boatload of tokens and use it as account creator in contract-helper

Patrick1904:
> @kcole16 @Patrick1904 this works fine, but I really have trouble accepting how it looks.
> I don't think we make it clear that `.test` is a part of account name (it's way too far on the right, can be really anything, there is no explanation where we explain how names work).
> 
> Also loader spinner is now overlapping `.test` which looks ugly. I think at the bare minimum it needs to be hidden (button already shows spinner too).
> 
> > when do you think we can change to '.testnet'?
> 
> someone needs to create `testnet` accounts with boatload of tokens and use it as account creator in contract-helper

I agree reg. spinner - we can hide it.

It seems the other potential issues are:
1. It’s still not clear that ‘.test’ is part of account name
2. It’s far to the right and ‘easy to miss’
3. There’s no explanation around ‘how names work’

I’ve mocked up two options:
[Option 1](https://www.figma.com/proto/NqcW900dzrsIqqTHwlyDPl/Untitled?node-id=1%3A169&viewport=-85%2C-864%2C0.508428156375885&scaling=min-zoom):
- Make ‘.test’ more visible with gray background and icon
- Make it more clear that ’.test’ is part of username by showing it as part of success label
- Provide easily accessible info around names

[Option 2](https://www.figma.com/proto/NqcW900dzrsIqqTHwlyDPl/Untitled?node-id=1%3A169&viewport=-85%2C-864%2C0.508428156375885&scaling=min-zoom):
- Make it clear that ‘.test’ is part of username by appending it to username when typing and also show as part of success label
- No more info around names. How important is this information?

[Full Figma sheet](https://www.figma.com/file/NqcW900dzrsIqqTHwlyDPl/Untitled?node-id=0%3A1)

kcole16:
> No more info around names. How important is this information?

TLDs are a more advanced topic, which aren't critical for every user to understand (info bubble works well for this).

Option 2 is the most direct. I prefer the look and feel of 1 (and the info bubble), but it's very important that users understand `.near`, etc are part of their account name.

Patrick1904:
@vgrichina, @kcole16 and I discussed showing username on recovery setup step for added clarity. My POV is that there should be no doubt about TLD if we show it on text input, as part of success label, and on recovery setup step. Thoughts?
Screen Shot 2020-04-07 at 3 41 23 PM

vgrichina:
> @vgrichina, @kcole16 and I discussed showing username on recovery setup step for added clarity.

Looks reasonable to me. As for info bubble – please make sure mobile is covered (I feel like current solution won't work as is).

My understanding is that "Option 2" is hard to implement cleanly (i.e. with `.testnet` not getting in the way of typing), correct?

Patrick1904:
> > @vgrichina, @kcole16 and I discussed showing username on recovery setup step for added clarity.
> 
> Looks reasonable to me. As for info bubble – please make sure mobile is covered (I feel like current solution won't work as is).
> 
> My understanding is that "Option 2" is hard to implement cleanly (i.e. with `.testnet` not getting in the way of typing), correct?

@vgrichina @kcole16 

I agree that tooltip isn't perfect mobile UX, we could do modal (like banner) for all devices instead - more mobile first.

As for option 2: It's not easy to make it smooth and cover all cases, but it can be done if we think it's the best option and worth the effort. It's clear that TLD is part of name but some users may find it confusing/annoying IMO as you're used to controlling the value and having the cursor on the right side of text. It would've helped if we could change the color for the suffix part of the text input value but nope.

My POV is that option 1 wins overall while option 2 wins on clarity around TLD. My vote is on option 1.

Patrick1904:
@kcole16 can you please provide some copy for TLD info modal?

kcole16:
**Top Level Accounts**
Account names are similar to domain names. Only the @testnet account can create accounts such as @yourname.testnet, and only @yourname.testnet can create @app.yourname.testnet. All accounts created in this wallet use the .testnet Top Level Account (TLA). To learn more about account names and creating your own TLA, visit the [docs](https://docs.nearprotocol.com/docs/concepts/account).

@Patrick1904 Does require a bit of variable text (for the TLA). Thoughts?

Patrick1904:
Now while implementing I feel like too much is going on with the text input (esp on mobile). Busy. What do you guys think about skipping the '@' now? It seems TLD should get more eyes. I like the cleaner look. Should we also adjust the placeholder a bit? I like 'E.g. xxxxx' - any thoughts on x?
@kcole16 @vgrichina

[Side-by-side comparison on mobile
](https://www.figma.com/file/A0Oaw3gHFNNMsAvzBYFZvE?node-id=0%3A1)

kcole16:
> Now while implementing I feel like too much is going on with the text input (esp on mobile). Busy. What do you guys think about skipping the '@' now? It seems TLD should get more eyes. I like the cleaner look. Should we also adjust the placeholder a bit? I like 'E.g. xxxxx' - any thoughts on placeholder?

I think this is what we had before? From what I remember, we changed it to make it clear the "@" isn't actually part of the username, but is a decorator

Patrick1904:
> > Now while implementing I feel like too much is going on with the text input (esp on mobile). Busy. What do you guys think about skipping the '@' now? It seems TLD should get more eyes. I like the cleaner look. Should we also adjust the placeholder a bit? I like 'E.g. xxxxx' - any thoughts on placeholder?
> 
> I think this is what we had before? From what I remember, we changed it to make it clear the "@" isn't actually part of the username, but is a decorator

Yes before there was no '@' icon. I think it provides minimal value in terms of clarifying that it's not part of the account name. More visual design value. But now with TLD it becomes too much IMO.

vgrichina:
@Patrick1904 "i" icon for some reason isn't clickable (at least on Mac Safari). Maybe just hide it for now and then add in later PR if there is problem with doing it compatible with both mobile and desktop?

I'm ok with hiding `@` icon as well.

This been out for a bit too long in Draft stage. Let's finish up shippable version and then iterate on improvements later,

Patrick1904:
> @Patrick1904 "i" icon for some reason isn't clickable (at least on Mac Safari). Maybe just hide it for now and then add in later PR if there is problem with doing it compatible with both mobile and desktop?
> 
> I'm ok with hiding `@` icon as well.
> 
> This been out for a bit too long in Draft stage. Let's finish up shippable version and then iterate on improvements later,

To be fair we weren't happy with initial version so I changed it back to draft and provided alternative design and now implemented. Modal should work now ;)

vgrichina:
@Patrick1904 
> To be fair we weren't happy with initial version so I changed it back to draft and provided alternative design and now implemented. Modal should work now ;)

I didn't mean to imply your fault here. More like we've been iterating on it sufficiently long, time to ship imperfect solution and improve later if needed.

marcinbodnar:
@vgrichina @Patrick1904 @kcole16 I'm working on #511 which includes changes for `AccountFormAccountId` component. I saw that there are many changes included and I have a few comments:
 - loader (for checking accountId availability) is not showing now, it might be a problem for a user with a slow network - a user might wait for a long time for error/success indicator - the loader was giving information that something is happening. It might be fixed by moving loader left so it will be visible next to `.test`
 - on `Send Money` or `Recover using Seed Phrase` error/success icons ware displayed inside the input element - not showing now.
 - missing localized string for modal texts
 - also, a minor thing: before error/success information was displayed on the right side of input for desktop: 
https://projects.invisionapp.com/share/Y5QIVUGTVER#/screens/356316645

Want me to fix this while working on #511?

PR #486, opened 2020-03-28
Add confirm accountId functionality to deauthorize full access key and additional changes for Full Access Keys components

marcinbodnar:
@vgrichina @kcole16 
Question: is Confirm functionality should be available for all full access keys?

I also added changes regarding TODO comments in code, and some additional:
 - fixed publickey display in details box - there was a problem with displaying it for different screen sizes
 - added loading effect for Deauthorize and Confirm buttons, because it takes some time to proceed deauthorize
 - refactor `AccessKeysDeauthorize` component and ContactsRemove since they based on the same CSS from `PaginationSubBox` file.

### DIFF
--- a/src/components/access-keys/AccessKeys.js
+++ b/src/components/access-keys/AccessKeys.js
@@ -27,7 +27,8 @@ class AccessKeys extends Component {
             { img: '', name: 'ALL' },
             { img: '', name: 'ALL' },
             { img: '', name: 'ALL' }
-        ]
+        ],
+        buttonLoader: false
     }
 
     handleConfirm = () => {
@@ -64,6 +65,7 @@ class AccessKeys extends Component {
             accountId: '',
             confirm: false,
             confirmStatus: '',
+            buttonLoader: false
         }))
     }
 
@@ -93,7 +95,8 @@ class AccessKeys extends Component {
         const publicKey = this.state.showSubData.public_key
 
         this.setState(() => ({
-            loader: true
+            loader: true,
+            buttonLoader: true
         }))
 
         this.props.removeAccessKey(publicKey).then(() => {
@@ -126,7 +129,8 @@ class AccessKeys extends Component {
             showSubData,
             accountId,
             confirm,
-            confirmStatus
+            confirmStatus,
+            buttonLoader
         } = this.state
 
         const { authorizedApps, title } = this.props
@@ -156,6 +160,7 @@ class AccessKeys extends Component {
                     accountId={accountId}
                     confirm={confirm}
                     confirmStatus={confirmStatus}
+                    buttonLoader={buttonLoader}
                 >
                     {authorizedApps && (authorizedApps.length 
                         ? authorizedApps.map((accessKey, i) => (
--- a/src/components/access-keys/AccessKeys.js
+++ b/src/components/access-keys/AccessKeys.js
@@ -18,6 +18,9 @@ class AccessKeys extends Component {
         showSub: false,
         showSubOpen: 0,
         showSubData: null,
+        accountId: '',
+        confirm: false,
+        confirmStatus: '',
         authorizedApps: [],
         filterTypes: [
             { img: '', name: 'ALL' },
@@ -25,6 +28,43 @@ class AccessKeys extends Component {
             { img: '', name: 'ALL' },
             { img: '', name: 'ALL' }
         ]
+    }
+
+    handleConfirm = () => {
+        this.setState((state) => ({
+            confirm: !state.confirm
+        }))
+    }
+
+    handleConfirmSubmit = (e) => {
+        e.preventDefault()
+
+        if (this.state.accountId === this.props.accountId) {
+            this.setState(() => ({
+                confirmStatus: 'success'
+            }))
+            this.handleDeauthorize()
+        }
+        else {
+            this.setState(() => ({
+                confirmStatus: 'problem'
+            }))
+        }
+    }
+
+    handleChange = (e, { name, value }) => {
+        this.setState(() => ({
+            [name]: value,
+            confirmStatus: ''
+        }))
+    }
+
+    handleConfirmClear = () => {
+        this.setState(() => ({
+            accountId: '',
+            confirm: false,
+            confirmStatus: '',
+        }))
     }
 
     toggleShowSub = (i, accessKey) => {
@@ -35,6 +75,8 @@ class AccessKeys extends Component {
             showSubOpen: i,
             showSubData: accessKey
         }))
+
+        this.handleConfirmClear()
     }
 
     toggleCloseSub = () => {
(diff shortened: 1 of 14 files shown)


src/components/access-keys/AccessKeys.js
```diff
@@ -43,13 +87,16 @@ class AccessKeys extends Component {
...
+
+        this.handleConfirmClear()
     }
 
     handleDeauthorize = () => {
         const publicKey = this.state.showSubData.public_key
 
         this.setState(() => ({
-            loader: true
+            loader: true,
```
vgrichina, line 98: any reason why 2 different loader states are needed? Shouldn't global loader state be enough here?

vgrichina:
@Patrick1904 PTAL, as I have too much stuff in pipeline to do thorough review

src/components/contacts/ContactsRemove.js
```diff
@@ -2,62 +2,71 @@ import React from 'react'
...
+                <b className='color-black'>Alex sent you 1020 Ⓝ</b>
+                <div>3d ago</div>
+            </div>
+            <div className='row border-top'>
+                <b className='color-black'>You and Alex played NEAR Chess</b>
+                <div>1w ago</div>
+            </div>
+        </div>
+        <div className='send'>
+            <Link to='/send-money/marcin'>
```
Patrick1904, line 60: Minor: Probably better to style `Link` component. Button inside of anchor tag is considered [invalid html5](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element).

src/components/pagination/PaginationSubBox.js
```diff
@@ -12,191 +12,250 @@ import CheckBlueImage from '../../images/icon-check-blue.svg'
...
+                    }
+
+                    h2 {
+                        font-size: 32px !important;
+                    }
+                    .title {
+                        padding-bottom: 12px;
+                        .image {
+                            > div {
+                                margin: 0 auto;
```
Patrick1904, line 191: For future re-factor efforts: Let's utilize styled-components when appropriate to improve readability and make things more modular. Many levels of nesting makes it hard to follow.

src/components/access-keys/AccessKeysDeauthorizeConfirm.js
```diff
@@ -0,0 +1,55 @@
...
+                    autoCorrect='off'
+                    autoCapitalize='off'
+                    spellCheck='false'
+                    tabIndex='1'
+                    autoFocus={true}
+                />
+            )}
+        </Translate>
+        <div className='alert-info'>
+            {confirmStatus === 'problem' && `Account name doesn't match`}
```
Patrick1904, line 30: Minor: Seems string is adequate since there's no variable

src/components/pagination/PaginationSubBox.js
```diff
@@ -12,191 +12,250 @@ import CheckBlueImage from '../../images/icon-check-blue.svg'
...
-
-            :hover {
-               background: #fff;
-               color: #ff585d;
-            }
-         }
-      }
-      .recent-transactions {
-         background-color: #f8f8f8;
-         margin: 0 0 0 0;
```
Patrick1904, line 67: Minor: `margin: 0` should suffice. There are a few occurrences. Same for `padding` cases.

src/components/access-keys/AccessKeys.js
marcinbodnar: Good point

src/components/contacts/ContactsRemove.js
marcinbodnar: Yes, good point, this part of code is not visible in the project, but the same problem appears in a few other places. My idea was to use <FormButton> component for all form buttons in the project after that linked button was needed and I used it forgetting about that.

The idea with <FormButton> component is good I think, it will be easy to change all buttons if the design will change, also resolves an issue with repeating css.

My proposition is to instead of styling Link, I will prepare component that will be a button with link option (it will be possible to add `to` attr, etc) and we can use it with all buttons. 

I think it's a good idea with having all buttons in one component. What do you think?

src/components/pagination/PaginationSubBox.js
marcinbodnar: I know what you mean, it's problematic. But also I think that using styled component for small items is problematic to debug too because you end up with hashed classes in dev tools, which makes it harder to find in code.

I think the ideal way to use style components is to make as many reusable components as possible and define more things in GlobalStyles. Personally I don't like when css is defined inside react component because it's a bit messy, that's why in some cases I did put all css in one parent component (for example `SendMoneyContainer.js`), define css for child components there - but of course, it's not a good idea for some cases. 

At this moment css structure is a bit messy because at early stages the design for all pages was changing a lot, many new elements were added. But in the future, it would be best to rebuild the whole css structure, makes more reusable components and have more things in GlobalStyles. Maybe we can also export style components to separate files for every folder or something like that?

Long reply, but I think it's a big issue and we should discuss it in the future.

src/components/contacts/ContactsRemove.js
Patrick1904: Sounds good to me

src/components/pagination/PaginationSubBox.js
vgrichina: Wonder if we should use this macro to make debugging style-components easier? https://stackoverflow.com/a/60887852/341267

I think I also prefer styles to be encapsulated in components, this seems to scale far better than all big previous projects I worked on which used global styles – they ultimately became very hard to manage mess and with no clear scope which styles impact which component.

So likely best not to use global styles except for basic stuff – like making default styles for HTML tags like `h1`, `h2`, links, etc

marcinbodnar: > So likely best not to use global styles except for basic stuff

I agree, what I mean about "have more things in GlobalStyles" was to put in GlobalStyles most of CSS code that is repeated in files and that cannot be put (or it's unnecessary/overdone) to a reusable component. For example all font styles (like you mentioned), border styles, font colors (as it might be mixed for different font styles) etc.

> Not having to think about classNames is a major benefit IMO.

What I meant, showing on examples, is that having separate styled components for Header and Title here:
https://github.com/nearprotocol/near-wallet/blob/master/src/components/accounts/recovery_setup/RecoveryOption.js#L110
seems a bit unnecessary, but I start to think it might be a better approach - JSX structure looks much better for sure.

vgrichina: > What I meant, showing on examples, is that having separate styled components for Header and Title here:

agree that it's less clear trade-off in this case. I think biggest advantage of having it as styled component is just that it's easier to extract it into separate reusable one when needed.

src/components/access-keys/AccessKeysDeauthorizeConfirm.js
vgrichina: @marcinbodnar shouldn't this use localized string too?

marcinbodnar: @vgrichina Good point. Done in #514

PR #491, opened 2020-03-31
Add test for seed phrase recovery

Patrick1904:


cypress/integration/recover_seed_phrase.js
```diff
@@ -0,0 +1,12 @@
+describe('Recover w/ seed phrase', function() {
+    it('Enters username and phrase', function() {
+        cy.visit('https://wallet.nearprotocol.com')
+        cy.get('a[href="/recover-account"]').click()
+        cy.get('a[href="/recover-seed-phrase"]').click()
+        cy.get('input').first().type('testing-account-cy')
```
Patrick1904: @vgrichina I'm not sure how test seed phrase recovery since I need to add username/phrase in the code? Unless maybe if I create an account and setup seed phrase. Then save that info and recover the account. The drawback with that approach is that I need to create an account just to test recovery.

Patrick1904:
@vgrichina should we keep the examples or unnecessary code?

cypress/integration/recover_seed_phrase.js
vgrichina: I think it makes sense to create account to test recovery. I guess it might also make sense grouping together with other test cases that require new account (so that tests can run faster).

vgrichina:
> @vgrichina should we keep the examples or unnecessary code?

I think we shouldn't keep examples, but if there are some useful utility functions – makes sense to keep them.

Patrick1904:
Closing this PR for now as I have meaningful amount of work left for it to be shippable.

PR #500, opened 2020-04-03
Add banner/modal to convey current network

Patrick1904:
https://github.com/nearprotocol/near-wallet/issues/208


src/utils/wallet.js
```diff
@@ -11,7 +11,10 @@ export const WALLET_CREATE_NEW_ACCOUNT_URL = 'create'
 export const WALLET_CREATE_NEW_ACCOUNT_FLOW_URLS = ['create', 'set-recovery', 'setup-seed-phrase', 'recover-account', 'recover-seed-phrase']
 export const WALLET_LOGIN_URL = 'login'
 
-const NETWORK_ID = process.env.REACT_APP_NETWORK_ID || 'default'
+export const NETWORK_ID = process.env.REACT_APP_NETWORK_ID || 'testnet'
```
vgrichina, line 14: Doesn't this break our wallet (all keys lost)?

kcole16: This might be due to my misunderstanding. 

Do we set this environment variable in each production environment (testnet/default, betanet, devnet)? 

Also, do you know if we are planning to change `default` from signaling Testnet to signaling Mainnet?

vgrichina: I don't think we currently set it. I am not sure it is very meaningful in wallet context, as each wallet instance is only holding keys for single network.

`default` doesn't really signal anything besides which namespace is used in key store to store keys.

Patrick1904: @vgrichina I created a new variable `export const CURRENT_NETWORK = process.env.REACT_APP_CURRENT_NETWORK || 'testnet'`

vgrichina: @kcole16 @Patrick1904 I've been thinking on it, and I feel that given that main consumer of alternate wallets going to be developers, maybe we should do this:
1) Display actual RPC node URL in banner (this is what really matters to devs).
2) Simply pass `IS_MAINNET` as an external flag.

kcole16: That's certainly true of betanet and devnet. 

For Testnet, I think it's reasonably likely some apps will release as "beta", and promote the testnet app to end users (common on Ethereum). Generally these are savvier users, but still worth making clear which network they're using.

Agree it makes sense to add the RPC node to the banner as well.

vgrichina: Maybe display smth like "Test-only Wallet (rpc.betanet.nearprotocol.com)"?

vgrichina: @kcole16 @Patrick1904

Patrick1904: @vgrichina 

To reference RPC URL, should I use `const NODE_URL = process.env.REACT_APP_NODE_URL || 'https://rpc.nearprotocol.com'`?

What should we base `IS_MAINNET` flag on?
I'm currently doing this:
`export const CURRENT_NETWORK = process.env.REACT_APP_CURRENT_NETWORK || 'testnet'
export const IS_MAINNET = CURRENT_NETWORK === 'mainnet'
`

vgrichina: > What should we base IS_MAINNET flag on?

I mean just pass it in explicitly as env variable, no need to base on smth.

> To reference RPC URL, should I use const NODE_URL = process.env.REACT_APP_NODE_URL || 'https://rpc.nearprotocol.com'?

yeah

vgrichina: I would display URL with domain only though (no `https://`) and make it clickable to go to `/status` endpoint.

src/components/common/NetworkBanner.js
```diff
@@ -0,0 +1,75 @@
...
+                            Test-only Wallet&nbsp;
+                            (<a href={`${NODE_URL}/status`} target='_blank' rel='noopener noreferrer' onClick={e => e.stopPropagation()}>
+                                {NODE_URL.replace(/^(?:https?:\/\/)?(?:www\.)?/i, "").split('/')[0]}
+                            </a>)
+                        </div>
+                    }
+                    closeIcon
+                >
+                    <Header>NEAR test Wallet</Header>
+                    NEAR test Wallet is for testing purposes only. Tokens and other assets have no value. Accounts created do not transfer to Mainnet.
```
kcole16, line 67: "This wallet (and network) are for testing purposes only. Tokens and other assets have no value. Accounts created do not transfer to Mainnet."

Patrick1904: @kcole16 updated copy

src/utils/wallet.js
```diff
@@ -11,18 +11,20 @@ export const WALLET_CREATE_NEW_ACCOUNT_URL = 'create'
 export const WALLET_CREATE_NEW_ACCOUNT_FLOW_URLS = ['create', 'set-recovery', 'setup-seed-phrase', 'recover-account', 'recover-seed-phrase']
 export const WALLET_LOGIN_URL = 'login'
 
+export const IS_MAINNET = process.env.REACT_APP_IS_MAINNET || false
```
vgrichina, line 14: @Patrick1904 "false" will be treated as true here, best to check whether `process.env.REACT_APP_IS_MAINNET` is either "true" or "yes".

Patrick1904: @vgrichina please check my [update](https://github.com/nearprotocol/near-wallet/pull/500/commits/5a16b6f20a99fb20facdb6c1ed13e4b3968103a0)

PR #507, opened 2020-04-08
Add Link functionality to FormButton component

marcinbodnar:


vgrichina:
@marcinbodnar make sure to resolve conflicts

marcinbodnar:
@vgrichina Done. I always rebase before PR. Is there a way to receive an email notification if conflicts appear after PR was opened?

src/components/dashboard/DashboardKeys.js
```diff
@@ -70,11 +69,13 @@ const DashboardKeys = ({ image, title, to, accessKeys, empty }) => (
         <Grid.Row>
             <Grid.Column textAlign='left' width={16}>
                 {accessKeys && accessKeys.length !== 0 ? (
-                    <Link to={to}>
-                        <FormButton color='gray-blue' size='small'>
-                            <Translate id='button.viewAll' />
-                        </FormButton>
-                    </Link>
+                    <FormButton
+                        to={to}
```
Patrick1904, line 73: Would it make sense to name URL prop something else? `to` doesn't help me understand what it is outside of FormButton context.

Patrick1904:
This is a big out of context but does your IDE complain about lack of braces too?
Screen Shot 2020-04-08 at 1 10 51 PM

src/components/common/FormButton.js
```diff
@@ -321,13 +322,19 @@ const FormButton = ({
...
-    size = ''
+    size = '',
+    to,
+    history,
+    className
 }) => (
     <CustomButton
         type={type}
-        className={`${color} ${size} ${sending ? `dots` : ``}`}
+        className={`${color} ${size} ${sending ? `dots` : ``} ${className}`}
```
Patrick1904, line 332: Screen Shot 2020-04-08 at 1 23 42 PM

I think we should avoid `undefined` if no className?

vgrichina:
> @vgrichina Done. I always rebase before PR. Is there a way to receive an email notification if conflicts appear after PR was opened?

not sure if you can receive e-mail notification, but I think you can get notification in Slack (https://github.com/pullreminders/backlog/issues/42) from http://pullreminders.com

src/components/common/FormButton.js
vgrichina: also maybe we should also remove  dependency on `import { Button } from 'semantic-ui-react'` and avoid dealing with `className` here altogether? 

otherwise looks like good use for `utils/classNames`

src/components/dashboard/DashboardKeys.js
vgrichina: `linkTo`?

src/components/common/FormButton.js
Patrick1904: Agree. Let's just make sure we test a lot if we remove semantic-ui since there are a lot of variations of the `FormButton` that may be affected.

Good point on `classNames`

marcinbodnar:
@Patrick1904 Done.

> This is a bit out of context but does your IDE complain about lack of braces too?

It does, but it was fixed at the same time so I left it:
https://github.com/nearprotocol/near-wallet/pull/466/commits/74060e214e98d81c79132240ed75870d4499e1d5

Also added small change for `ClassNames`, so it will handle undefined `names`.

src/components/common/FormButton.js
marcinbodnar: >avoid dealing with className here altogether?

@vgrichina what do you mean?

vgrichina: I mean that classes need to be set only because we rely on semantic-ui-react.  If not using semantic-ui-react (creating our custom component from scratch) we can for example use props directly in styled-components CSS (or otherwise apply styles programmatically).

src/utils/classNames.js
```diff
@@ -3,6 +3,10 @@
     <div className={classNames(['base-class', {'second' : someFlag, 'third' : true}])}></div>
 */
 const classNames = (names) => {
+    if (!names) {
+        return false
```
vgrichina, line 7: nit: seems like should be returning '' instead?

src/utils/classNames.js
```diff
@@ -3,6 +3,10 @@
     <div className={classNames(['base-class', {'second' : someFlag, 'third' : true}])}></div>
 */
 const classNames = (names) => {
+    if (!names) {
```
vgrichina, line 6: nit: @marcinbodnar makes sense to add test if you find some bug with edge case (given this already has tests)

PR #513, opened 2020-04-10
create account: include TLA when checking availability

Patrick1904:
<a href="https://gitpod.io/#https://github.com/nearprotocol/near-wallet/pull/513"></a>

### DIFF
--- a/src/components/accounts/AccountFormAccountId.js
+++ b/src/components/accounts/AccountFormAccountId.js
@@ -63,7 +63,7 @@ class AccountFormAccountId extends Component {
         this.timeout && clearTimeout(this.timeout)
 
         this.timeout = setTimeout(() => {
-            checkAvailability(value)
+            checkAvailability(this.props.accountId)
         }, ACCOUNT_CHECK_TIMEOUT)
     }
 


Patrick1904:
This fixes false positives when checking account availability.

marcinbodnar:
@vgrichina @Patrick1904 
There is a problem with this change because now it's not checking account availability properly on Send Money and Recover using Seed Phrase pages. You can fix it by passing 'accountId' to `SendMoneyFirstStep` and `RecoverAccountSeedPhraseForm` component, but the idea for `AccountFormAccountId` was that it will check for both:
 - if accountId is available for creating
 - it accountId exists
without additional changes for parent components.

So I think here:
https://github.com/nearprotocol/near-wallet/blob/master/src/components/accounts/AccountFormAccountId.js#L66
it should still check account for `value`, and also this change:
https://github.com/nearprotocol/near-wallet/blob/master/src/components/accounts/CreateAccount.js#L37
should be moved to `handleChangeAccountId` in `AccountFormAccountId` component. If you need to handle it for Create Account situation only you can use `type` prop.

marcinbodnar:
@vgrichina @Patrick1904 
Or maybe adding `ACCOUNT_ID_SUFFIX` should be moved to `checkNewAccount` and `createNewAccount` in `wallet.js`?

Patrick1904:
@marcinbodnar great catch! Agree that we can re-think. Added a [temp fix](https://github.com/nearprotocol/near-wallet/pull/516) now.

Issue #511, opened 2020-04-09
Validate account name input length


kcole16:
### Overview
The minimum length for account names is 2, and the max is 64. We currently allow users to enter names of 1 character and >64, which then returns an error. The "Create" button should be hidden and an error message displayed when the length is invalid.

### Acceptance Criteria
When input is < 2 and > 64 characters
- [ ] "Create" button is disabled 
- [ ]  Displays "Username must be between 2 and 64 characters"

PR #523, opened 2020-04-15
Validate account name input length (#511)

marcinbodnar:
<a href="https://gitpod.io/#https://github.com/nearprotocol/near-wallet/pull/523"></a>

### DIFF
--- a/src/actions/account.js
+++ b/src/actions/account.js
@@ -203,12 +203,13 @@ export const { signAndSendTransactions } = createActions({
     ]
 })
 
-export const { switchAccount, refreshAccount, resetAccounts, refreshUrl } = createActions({
+export const { switchAccount, refreshAccount, resetAccounts, refreshUrl, setFormLoader } = createActions({
     SWITCH_ACCOUNT: wallet.selectAccount.bind(wallet),
     REFRESH_ACCOUNT: [
         wallet.loadAccount.bind(wallet),
         () => ({ accountId: wallet.getAccountId(), })
     ],
     RESET_ACCOUNTS: wallet.clearState.bind(wallet),
-    REFRESH_URL: null
+    REFRESH_URL: null,
+    SET_FORM_LOADER: null
 })
--- a/src/components/GlobalStyle.js
+++ b/src/components/GlobalStyle.js
@@ -1,7 +1,5 @@
 import { createGlobalStyle } from 'styled-components'
 
-import ProblemsImage from '../images/icon-problems.svg'
-import CheckBlueImage from '../images/icon-check-blue.svg'
 import EmailIconGray from '../images/email-icon-gray.svg'
 import CloseBtn from '../images/close-btn.svg'
 
--- a/src/components/accounts/AccountFormAccountId.js
+++ b/src/components/accounts/AccountFormAccountId.js
@@ -63,19 +63,21 @@ class AccountFormAccountId extends Component {
 
         this.timeout && clearTimeout(this.timeout)
 
-        this.timeout = setTimeout(() => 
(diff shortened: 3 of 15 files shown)


Patrick1904:
LGTM.

Might just be me, but feels a bit jarring to get the 'red/error' treatment after entering just one character? It's a small detail but essentially every user will see it and it's a first impression kind of thing.

Thoughts @marcinbodnar @kcole16 @vgrichina ?

vgrichina:
> but feels a bit jarring to get the 'red/error' treatment after entering just one character? It's a small detail but essentially every user will see it and it's a first impression kind of thing.

I agree

marcinbodnar:
agree, it seems a bit unnecessary

kcole16:
@marcinbodnar @Patrick1904  Can we add a short delay to the validation? Alternative is onBlur, but I'm wondering how well this will work with a single input and disabled "Submit" button.

Patrick1904:
> @marcinbodnar @Patrick1904 Is there a good way to solve this? Alternative is onBlur, but I'm wondering how well this will work with a single input and disabled "Submit" button.
> 
> Looks good other than that though!

We could hide the label while typing?

marcinbodnar:
I will check hiding the label and delay options, see what works best.

marcinbodnar:
@kcole16 @Patrick1904 ready, with additional changes that we were talking about on Slack

Patrick1904:
Screen Shot 2020-05-01 at 4 23 38 PM
Screen Shot 2020-05-01 at 4 15 29 PM

These improvements are really great. Some minor things:
- Seems it easily gets stuck at ‘checking availability…’ (can we also make C uppercase) when only one character. Can probably default to ‘Username must be between 2 and 64 characters’ when value length is one.
- ‘Checking availability…’ doesn’t make much sense for non-account-creation cases. Maybe ‘Checking…’ for all other cases?
- Can we also add some spacing between label and input on ‘send’ page please?

Thoughts @marcinbodnar @kcole16

kcole16:
@Patrick1904 @marcinbodnar Agree with those points.

Re: getting stuck on `checking availability...`, seems that the error shows correctly if you enter one character into the empty input, but will get stuck if you enter 2+ characters then delete back to just 1.

marcinbodnar:
@kcole16 @Patrick1904 
Done.

Additionally, I refactored requestStatusBox component (get rid of semantic-ui, simplify css - it was causing troubles) and made a proposition for showing requestStatus on Send Money and Create Account so the bottom content will not jump. On Create account I change:
 - margins - now the gap between small title and input is the same as between input and recaptcha text - I feel like recaptcha text belongs more to button, additionally input will be better visible
 - font size - changed to 12px - we have color on border that indicates status, so I feel like big text is not necessary
 - requestStatus text is now closer to the input.

Let me know what you think. It can be easily removed with the last commit.

kcole16:
@marcinbodnar Changes look good! 

Last nit is that the text flips back to "Checking availability..." while waiting for the account creation transaction to confirm.

Patrick1904:
Great call on adding space between input and button to fix the jumping.

In regards to the label, I suggest we keep the 14px font size to maximize legibility since we expect user to actually read this :) The light blue color makes it especially hard to read with 12px font size. I’d also add a bit more space between the label and input: margin-top: -20px —> margin-top: -15px (seems to fix it).

vgrichina:
@Patrick1904 @marcinbodnar @kcole16 any reason we aren't merging this?

Patrick1904:
> @Patrick1904 @marcinbodnar @kcole16 any reason we aren't merging this?

My bad. I was thought this was merged already.

Issue #519, opened 2020-04-14
TXs in Activity Feed displaying incorrectly


kcole16:
**Problem**
The activity feed appears to have a few issues (possibly a result of #466). 
- The timestamp always displays "Just now", even for TXs confirmed days/weeks ago
- Failed TXs are displayed

**Expected Behavior**
- Timestamps should reflect the time since the TX confirmed
- Only successful TXs should show in Activity Feed (for now)

**Steps to Reproduce**
1. Go to wallet.nearprotocol.com (use an existing account)
2. Check the timestamp
3. Click "View All" and compare timestamps between Explorer and Wallet

kcole16:
@vgrichina @Patrick1904 @marcinbodnar 

What are your thoughts re: displaying Failed TXs in the wallet? Currently, we do not show the status, so all TXs in the Activity Feed are implied to be successful.

I think it's ideal to display all TXs, including failed ones. Though until we have more info to display re: the reason for failure (and considering some history may be missing when a TX fails and isn't included in a block), might make more sense to include only successful and pending TXs.

Patrick1904:
> might make more sense to include only successful and pending TXs
I like this. I think we should consider a filter dropdown btn. Default can be to hide failed TXs but user can click the dropdown and select to include failed TXs, etc.

marcinbodnar:
IMO we should display exactly the same transactions that are displayed in explorer, it might be misleading for a user, if he clicks `VIEW ALL` and sees a different list of transactions. 

If we would like to hide failed transactions we can use switcher, instead of a dropdown to show all/succeded. It was designed for explorer (https://projects.invisionapp.com/share/NHQIS41UW62#/screens/354347466) and it's available in Wallet in `ShowingSwitcher` component.

marcinbodnar:
@kcole16 At this moment transactions are not displayed in explorer and wallet, some bug?

kcole16:
@marcinbodnar Explorer was down yesterday due to testnet upgrade. Back up now.

The switcher is a good potential option. We should consider a redesign of the Activity Feed before implementing #456 (tokens), and toggling between TX statuses might fit well there.

In the meantime, should at least add the TX status before the timestamp, as currently users will be unable to distinguish between successful and failed TXs.

**Explorer example:**
[...]

PR #526, opened 2020-04-17
Transactions status (#519)

marcinbodnar:
<a href="https://gitpod.io/#https://github.com/nearprotocol/near-wallet/pull/526"></a>

### DIFF
--- a/src/components/dashboard/ActionsList.js
+++ b/src/components/dashboard/ActionsList.js
@@ -115,17 +115,18 @@ const CustomGridRow = styled(Grid.Row)`
     }
 `
 
-const ActionsList = ({ transaction, wide, accountId }) => (
+const ActionsList = ({ transaction, wide, accountId, getTransactionStatus }) => (
     <ActionRow 
         transaction={transaction} 
         actionArgs={JSON.parse(transaction.args)} 
         actionKind={transaction.kind}  
         wide={wide}
         accountId={accountId}
+        getTransactionStatus={getTransactionStatus}
     />
 )
 
-const ActionRow = ({ transaction, actionArgs, actionKind, wide, showSub = false, accountId }) => (
+const ActionRow = ({ transaction, actionArgs, actionKind, wide, showSub = false, accountId, getTransactionStatus }) => {
     <CustomGridRow
         verticalAlign='middle'
         className={`${wide ? `wide` : ``} ${
--- a/src/components/dashboard/ActionsList.js
+++ b/src/components/dashboard/ActionsList.js
@@ -50,8 +50,7 @@ const CustomGridRow = styled(Grid.Row)`
         .main-row-title {
             font-weight: 600;
             width: auto;
-            padding-right: 0px;
-            padding-left: 0px;
+            padding: 8px 0;
             flex: 1;
             word-break: break-all;
             display: flex;
--- a/src/components/dashboard/ActionsList.js
+++ b/src/components/dashboard/ActionsList.js
@@ -130,14 +130,16 @@ const ActionsList = ({ transaction, wide, accountId, getTransactionStatus }) => 
 
 const ActionRow = ({ transaction, actionArgs, actionKind, wide, showSub = false, accountId, getTransactionStatus }) => {
     const { checkStatus, status, hash, signer_id, block_timestamp } = transaction
+    const getTransactionStatusConditions = () => checkStatus && !document.hidden && getTransactionStatus(hash, signer_id)
 
     useEffect(() => {
-        const interval = (checkStatus) && setInterval((() => {
-            getTransactionStatus(hash, signer_id)
-        })(), TRANSACTIONS_REFRESH_INTERVAL)
+        getTransactionStatusConditions()
+        const interval = setInterval(() => {
+            getTransactionStatusConditions()
+        }, TRANSACTIONS_REFRESH_INTERVAL)
 
         return () => clearInterval(interval)
-    }, [hash])
+    }, [hash, checkStatus])
 
     return (
         <CustomGridRow
--- a/src/components/dashboard/ActionsList.js
+++ b/src/components/dashboard/ActionsList.js
@@ -247,10 +247,8 @@ const TX_STATUS_COLOR = {
 }
 
 const ActionStatus = ({ status }) => (
-    <div className={classNames(['font-small', status && TX_STATUS_COLOR[status]])}>
-        {status
-            ? <b><Translate id={`transaction.status.${status}`} /></b>
-            : <div className='dots'><Translate id={'transaction.status.checkingStatus'} /></div>}
+    <div className={classNames(['font-small', {[TX_STATUS_COLOR[status]]: status, 'dots': !status}])}>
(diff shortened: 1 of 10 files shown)


src/components/dashboard/ActionsList.js
```diff
@@ -216,7 +222,32 @@ const ActionIcon = ({ actionKind }) => (
 )
 
 const ActionTimeStamp = ({ timeStamp }) => (
-    <div className='font-small time-stamp'>{format(timeStamp)}</div>
+    <div className='font-small'>{format(timeStamp)}</div>
+)
+
+const TX_STATUS = {
+    NotStarted: {
+        textId: 'notStarted',
```
vgrichina, line 230: why introduce separate `textId` instead of just using tx status name as is?

src/utils/explorer-api.js
```diff
@@ -33,6 +45,10 @@ export async function getTransactions(accountId = '') {
         }
     ));
     
-    return tx.map(t => ({
-        ...t}))
+    return Promise.all(tx.map(async t => ({
+        ...t,
+        status: Object.keys((await transactionExtraInfo(t)).status)[0]
```
vgrichina, line 50: this looks super inefficient in terms of number of queries needed to receive info.

@marcinbodnar any reason we have to do it that way vs have explorer API return status? /cc @frol

src/components/dashboard/ActionsList.js
marcinbodnar: I thought about it but figured that if status name will be changed, it will be easier to change it in one place, than in a few places in translation files.

kcole16:
@marcinbodnar Status is looking good! Re: the loading animation, let's use a similar one to this (using for Recovery Options loading): https://github.com/nearprotocol/near-wallet/pull/381/commits/b6b689f8b5126897f1f1cfc8bb2066478ba39042

src/utils/explorer-api.js
marcinbodnar: I started doing it using `Wampy` with `nearcore-tx`, but it doesn't work. As far as I figured out, the explorer is not storing `status` in db - it does the same thing queries rpc for every transaction separately:
https://github.com/nearprotocol/near-explorer/blob/master/frontend/src/libraries/explorer-wamp/transactions.ts#L180

It would be better to store it in db, and received with `SELECT` query that already is executed in the wallet, but in this case, the explorer would have to check for tx update even if it's not loaded.

We can make it a little bit more efficient by checking status for every transaction, not action - in this case, we can check if `hash` is the same and apply the same value as in action before.

marcinbodnar:
> Re: the loading animation, let's use a similar one to this (using for Recovery Options loading): [b6b689f](https://github.com/nearprotocol/near-wallet/commit/b6b689f8b5126897f1f1cfc8bb2066478ba39042)

@kcole16 I'm afraid will not look good, because we are reloading this content every 10s. The idea with dots is that the content is not blinking every reload, just values are changing - it's smooth and discreet. Also, I feel this background animation works for small spaces only.

Also noticed, that going directly to profile doesn't work:
https://wallet.nearprotocol.com/profile/marcink56.test

kcole16:
> @kcole16 I'm afraid will not look good, because we are reloading this content every 10s.

Ah, missed the polling. Ok, that makes sense. 

> Also noticed, that going directly to profile doesn't work:

Yes, looks like this might have to do with the `.test` suffix. Notice how https://wallet.nearprotocol.com/profile/emailagain should work. Will open a separate issue.

marcinbodnar:
> Yes, looks like this might have to do with the .test suffix. Notice how

Yes, it was working fine before.

src/components/dashboard/ActionsList.js
vgrichina: >  if status name will be changed

blockchain API has to be frozen with mainnet release, any changes would be complex multi-step process

> than in a few places in translation files.

seems like simple search&replace in multiple files, no need for extra mapping

src/utils/explorer-api.js
vgrichina: This for sure shouldn't be combined as is with polling every 10 sec. Especially given that there seems to be no machinery to cancel outstanding requests.  One open wallet instance will be easily generating 1 request per second with current implementation.

If @frol doesn't have a solution to avoid making extra queries, I think we should disable polling for now.
Polling can be enabled either after changes on explorer side or with such optimizations:
- don't poll when window is not active in browser
- don't make requests per action (make unique request per tx hash)
- only poll status on the transactions which can still change status (if it either succeeded or failed – it cannot change status anymore)

Also if we have to ship multi-query solution – it's best not to block loading of list before all the tx statuses arrive. We can start displaying list of transactions long before all statuses arrive. Status query for transaction in flight can easily be blocking for 10 seconds as for now. Also same goes for error handling – we shouldn't fail loading of the transaction list just because loading status of one of transactions has failed.

vgrichina:
@kcole16 
> Yes, looks like this might have to do with the .test suffix. Notice how https://wallet.nearprotocol.com/profile/emailagain should work. Will open a separate issue.

I've noticed it before and AFAIK it's caused by how URL rewrites work on Render. We can workaround by running own HTTP server vs relying on whatever they provide for free, but they also promised to fix it (just will take some time).

src/utils/explorer-api.js
frol: It is unfortunate that nearcore RPC does not expose the transaction status in the `chunks` API: https://github.com/nearprotocol/nearcore/issues/1880, so we also struggle on the Explorer side. I will take a look in the near future, but it might be a good reason not to expose that as it might require fetching all the receipts for the transaction to know the status (I have not investigated it yet). In any case, we are going to resolve this on Explorer side once we sync all the receipts (https://github.com/nearprotocol/near-explorer/issues/218)

frol:
> need to figure out with @frol how to get rid of N+1 query problem

I am sorry to say this, but there is no way to solve this at the moment (see my comment above). Explorer operates the same way at the moment.

src/utils/explorer-api.js
vgrichina: > that as it might require fetching all the receipts for the transaction to know the status (

@frol won't us constantly polling it from client-side be even worse? Do you have ETA for https://github.com/nearprotocol/near-explorer/issues/218?

frol: > won't us constantly polling it from client-side be even worse? 

Well, it is worse, you are right.

nearprotocol/near-explorer#218 is the next big item on the Explorer development queue, but it is currently estimated to take a couple of weeks to land it properly.

marcinbodnar: > or with such optimizations

@vgrichina ok, np

marcinbodnar: @vgrichina We can also consider adding `check status` option - add it instead of status for every action, or near `Activity` header so it will check status for all actions in activity block?

marcinbodnar: @vgrichina Done. There is just one problem, I'm not sure why a request for tx status is not working here - it's working fine locally. Do you know what is the reason?

PR #533, opened 2020-04-22
Remove Amplitude, add mixpanel, track app titles

kcole16:
<a href="https://gitpod.io/#https://github.com/nearprotocol/near-wallet/pull/533"></a>


vgrichina:
@kcole16 is this ready to merge or we are waiting for smth?

src/middleware/index.js
```diff
@@ -43,14 +43,17 @@ const analyticsMiddleware = store => next => action => {
     let details = {
         pathname: window.location.pathname,
     }
+    if (action.type === 'ADD_ACCESS_KEY') {
+        details['appTitle'] = action.meta.data.title
```
vgrichina, line 47: @kcole16 looks like long-term we should log contract ID here, generally we should stop using titles supplied by user.

PR #534, opened 2020-04-22
profile: display balance breakdown

Patrick1904:
https://github.com/nearprotocol/near-wallet/issues/481

Covered in this PR:
- [x] - Show balance breakdown on profile
- [x] - Calculate actual state stake (storageUsage (bytes) * cost per byte)
- [x] - Prevent user from sending/signing above available balance
- [x] - Make total balance equal `amount + locked` instead of `amount`
- [x] - Explain minimum balance (icon + tooltip)
- [x] - Explain available balance (icon + tooltip)
- [x] - Show available balance on: send, sign
- [x] - Make `Balance` component return 0 when it's actually 0

Not covered in this PR:
- Vested balance
- Delegated balance

<a href="https://gitpod.io/#https://github.com/nearprotocol/near-wallet/pull/534"></a>


Patrick1904:
@kcole16 @vgrichina for state stake: I have the storage size (`storageUsage`) but I'm not sure how to get the price per byte. Any idea?

vgrichina:
> I have the storage size (storageUsage) but I'm not sure how to get the price per byte. Any idea?

@Patrick1904 no idea either. @nearmax how do we get this?

Patrick1904:
> > I have the storage size (storageUsage) but I'm not sure how to get the price per byte. Any idea?
> 
> @Patrick1904 no idea either. @nearmax how do we get this?

@vgrichina I got info from @kcole16 that cost per byte currently is `909 * 100_000_000_000_000_000`

I set this in `wallet.js`
`const ACCOUNT_COST_PER_BYTE = process.env.REACT_APP_ACCOUNT_COST_PER_BYTE || '90900000000000000000'`

`const costPerByte = new BN(ACCOUNT_COST_PER_BYTE)`
`const stateStaked = new BN(account.storageUsage).mul(costPerByte)`

Here's result for my account with 500 NEAR
Screen Shot 2020-04-24 at 5 18 32 PM

vgrichina:
@Patrick1904 we for sure shouldn't be hardcoding this (even in env variable) as it's going to be dynamic depending on system load.

@frol maybe you know if there is any way to retrieve it from nearcore? I think if there is no way right no – probably best to add endpoint with hardcoded response vs hardcoding in wallet.

frol:
@Patrick1904 @vgrichina you can query [RPC `method=EXPERIMENTAL_genesis_config`](https://docs.nearprotocol.com/docs/api/rpc-experimental#genesis-config), which exposes `runtime_config.storage_amount_per_byte`

nearmax:
@Patrick1904 I suggest we do not hardcode nearcore constants in the Applayer code. They are subject to change. Instead the constants should be retrieved by querying the config from the RPC.

Patrick1904:
@frol @vgrichina can we avoid calculating this stuff in the wallet/frontend and make `balance` data part of current `account` object returned from RPC? Seems like much better DevX and less potential bugs and confusion.

```
balance: {
	total: '1000',
	state-staked: '1',
	staked: '100',
	available: ‘899’
}
```

frol:
@Patrick1904 I believe, it belongs to near-api-js (+ some caching should be in place to avoid querying genesis config over and over again) since we avoid exposing data that can be computed.

kcole16:
The available balance (and locked balance breakdown) is something we will need to access elsewhere (Explorer, for example) and could be useful to app devs as well (especially displaying available balance in a UI, and determining whether user has sufficient balance for an action).

@vgrichina Do you think near-api-js is a good place for this?

kcole16:
@frol For fetching the locked account balances that are dependent on contracts (Vesting and Delegation), where do you think this data should live?

vgrichina:
> @vgrichina Do you think near-api-js is a good place for this?

@kcole16 @Patrick1904 yeah, it's good call to move `getAccountBalance` to `near-api-js`. I'd also change it to just return object with all different "types" of balance instead of taking `type` argument.

frol:
@kcole16  Ideally, we should have separate npm package(s) for those contracts, and integrated into the smart contracts CI to test that they match the contract implementation. @evgenykuzyakov what do you think?

marcinbodnar:
@Patrick1904 
I guess we can remove `nearlib` dependency and change it to `near-api-js` everywhere? Maybe also change name in wallet from `nearlib` to `nearApiJs`?

For Profile page, I think we can move `wallet.getBalance()` to `loadAccount` action in `actions/account.js`?

Also, not related to these changes, but when it comes to Profile page I think it might be a bit more efficient if we would dispatch `loadAccount` action only if this is a different account than the current user. Now, if this is the same user, we are dispatching `loadAccount` action when user enters Profile page, but at the same time account state is refreshing by `refreshAccount` action that is dispatched every page change. What do you thing?

Patrick1904:
> @Patrick1904
> I guess we can remove `nearlib` dependency and change it to `near-api-js` everywhere? Maybe also change name in wallet from `nearlib` to `nearApiJs`?
> 
> For Profile page, I think we can move `wallet.getBalance()` to `loadAccount` action in `actions/account.js`?
> 
> Also, not related to these changes, but when it comes to Profile page I think it might be a bit more efficient if we would dispatch `loadAccount` action only if this is a different account than the current user. Now, if this is the same user, we are dispatching `loadAccount` action when user enters Profile page, but at the same time account state is refreshing by `refreshAccount` action that is dispatched every page change. What do you thing?

I agree :)

For other participants: @marcinbodnar is going to fix the last touches, which includes [recovery method bug (580)](https://github.com/near/near-wallet/issues/580) and get this shippable (as discussed over Slack. thank you).

marcinbodnar:
@Patrick1904 I also changed so `Available Balance` info will stay while typing, I think it might be handy if someone will like to send the max amount. It will prevent content from additional jumping so double plus.

package.json
```diff
@@ -7,6 +7,7 @@
     "connected-react-router": "^6.6.0",
     "fetch-send-json": "0.0.2",
     "js-sha256": "^0.9.0",
+    "near-api-js": "^0.25.0",
     "near-ledger-js": "^0.0.4",
     "near-seed-phrase": "^0.0.2",
     "nearlib": "^0.22.0",
```
vgrichina, line 13: `nearlib` should be removed

src/components/profile/Recovery/RecoveryContainer.js
```diff
@@ -89,27 +88,31 @@ class RecoveryContainer extends Component {
...
 
     handleDeleteMethod = (method) => {
-        const { deleteRecoveryMethod, loadRecoveryMethods, accountId } = this.props;
+        const { deleteRecoveryMethod, loadRecoveryMethods } = this.props;
 
         this.setState({ deletingMethod: method.kind })
         deleteRecoveryMethod(method)
             .then(({ error }) => {
                 if (error) return
-                loadRecoveryMethods(accountId);
```
vgrichina, line 105: did this get changed to load recovery methods for current account? not sure how will this play with https://github.com/near/near-wallet/issues/580

vgrichina:
Looks like balance gets shortened in unfortunate ways sometimes (significant digits not displayed):

Screen Shot 2020-05-18 at 6 42 50 PM

Patrick1904:
> Looks like balance gets shortened in unfortunate ways sometimes (significant digits not displayed):
> 
> Screen Shot 2020-05-18 at 6 42 50 PM

@corwinharrell do you have any ideas on how to adjust nav layout so that we can fit long account names and balances on desktop? Omitting user icon doesn't do much. Could stack the two again but think we need to adjust the design a bit then.

marcinbodnar:
> Looks like balance gets shortened in unfortunate ways sometimes (significant digits not displayed):
@Patrick1904 actually, the design for this part was different, it should look exactly the same as a mobile version (that's why it was the same component earlier), because we resign of image icons on every page at some point. I think that while you ware changing this part, you used old designs, it was looking like this before your changes:

kcole16:
> Looks like balance gets shortened in unfortunate ways sometimes (significant digits not displayed)

This is a good point, and also a bit orthogonal to this PR.

I also hope very few users with that much NEAR ever use this wallet 😅 (and then only with Ledger)

src/components/profile/Recovery/RecoveryContainer.js
marcinbodnar: `accountId` param was unnecessary, `loadRecoveryMethods` was using current logged account always:
https://github.com/near/near-wallet/pull/534/files/21a8e48a3f8fac0ad033d5640c34a5e72bfe8c21#diff-b3e6d1b5726af892a08cb5d16f807a6cL17

Patrick1904:
@marcinbodnar yes the desktop stacking layout looks suboptimal IMO (as mentioned). I believe @corwinharrell has been tinkering with a new account/profile component design for the nav that solves these issues and looks great.

corwinharrell:
One thing I want to question/challenge is the persistent display of a user's balance in the application header. I can understand that the idea behind this is to allow the user to have an idea of their balance at any given moment.

That being said, I don't think this makes sense from a security perspective. At any point that you have the application open, anyone can see your balance with a simple glance at your screen. In addition, when I want to `receive` funds using the QR code method, I need to show my screen to another individual and have them scan my QR code with their camera. This instance also exposes my balance when I share my screen. If I'm concerned about security, this renders the QR code function useless. In future iterations, I'd like to suggest that we keep the balance one level deeper, (user must open the navigation to see available balance) in order to account for this. Thoughts?

Patrick1904:
> In future iterations, I'd like to suggest that we keep the balance one level deeper, (user must open the navigation to see available balance) in order to account for this. Thoughts?

100%

kcole16:
@vgrichina @marcinbodnar Outside of navbar balance concerns (will make separate issue), is this ready to merge?

marcinbodnar:
@kcole16 yes

src/utils/wallet.js
```diff
@@ -43,8 +42,8 @@ async function getKeyMeta(publicKey) {
 
 class Wallet {
     constructor() {
-        this.keyStore = new nearlib.keyStores.BrowserLocalStorageKeyStore(window.localStorage, 'nearlib:keystore:')
-        const inMemorySigner = new nearlib.InMemorySigner(this.keyStore)
+        this.keyStore = new nearApiJs.keyStores.BrowserLocalStorageKeyStore(window.localStorage, 'nearlib:keystore:')
```
vgrichina, line 45: nit: everything here is JS, just `nearApi` would be enough )

corwinharrell:
> > Looks like balance gets shortened in unfortunate ways sometimes (significant digits not displayed):
> > Screen Shot 2020-05-18 at 6 42 50 PM
> 
> @corwinharrell do you have any ideas on how to adjust nav layout so that we can fit long account names and balances on desktop? Omitting user icon doesn't do much. Could stack the two again but think we need to adjust the design a bit then.

@Patrick1904 apologies, I think I missed this when you originally posted it. Will continue to think about this while designing for in-progress issue and perhaps there's a low-lift option that offers a better solution.

PR #535, opened 2020-04-24
Make all info about TLA use env var

chadoh:
The modal that popped up did not match the suffix shown in the account creation form. This makes them match, and moves the strings of text to the translations file so we don't miss them when translating the interface to new languages.

<a href="https://gitpod.io/#https://github.com/nearprotocol/near-wallet/pull/535"></a>


chadoh:
You can see that this works at https://near-wallet-pr-535.onrender.com/create/, where I've set the account suffix to `lolnet`. Click on the ⓘ to see the matching modal.

PR #541, opened 2020-04-28
Clear transactions for new accountId (#519)

marcinbodnar:
Thanks to this change, the transactions list will be cleared (before fetching a new list) if the user will create a new account, recover account, or switch account.

<a href="https://gitpod.io/#https://github.com/nearprotocol/near-wallet/pull/541"></a>

### DIFF
--- a/src/reducers/account/index.js
+++ b/src/reducers/account/index.js
@@ -92,7 +92,16 @@ const accessKeys = handleActions({
 
 const transactions = handleActions({
     [getTransactions]: (state, { error, payload, ready }) => {
-        const hash = state.transactions && state.transactions.reduce((h, t) => ({
+        const { transactions } = state
+
+        if (!ready && transactions && transactions[0].query_account_id !== state.accountId) {
+            return {
+                ...state,
+                transactions: undefined
+            }
+        }
+
+        const hash = transactions && transactions.reduce((h, t) => ({
             ...h,
             [t.hash_with_index]: t
         }), {})
@@ -109,8 +118,7 @@ const transactions = handleActions({
                         } 
                         : t
                 ))
-                : state.transactions
-
+                : transactions
         })
     },
     [getTransactionStatus]: (state, { error, payload, ready, meta }) => ({
--- a/src/reducers/account/index.js
+++ b/src/reducers/account/index.js
@@ -91,15 +91,8 @@ const accessKeys = handleActions({
 }, initialState)
 
 const transactions = handleActions({
-    [getTransactions]: (state, { error, payload, ready }) => {
-        const { transactions } = state
-
-        if (!ready && transactions && transactions[0].query_account_id !== state.accountId) {
-            return {
-                ...state,
-                transactions: undefined
-            }
-        }
+    [getTransactions]: (state, { error, payload, ready, meta }) => {
+        const transactions = state.transactions ? state.transactions[meta.accountId] : undefined
 
         const hash = transactions && transactions.reduce((h, t) => ({
             ...h,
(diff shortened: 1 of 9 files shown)


src/utils/explorer-api.js
```diff
@@ -22,7 +22,8 @@ export async function getTransactions(accountId = '') {
                     transactions.block_timestamp, 
                     actions.action_type as kind, 
                     actions.action_args as args,
-                    actions.action_index || ':' || transactions.hash as hash_with_index
+                    actions.action_index || ':' || transactions.hash as hash_with_index,
+                    :accountId as query_account_id
```
vgrichina, line 26: why not just add this as a field to returned object? No need to have it in every row.

vgrichina: also given that this is info only useful for reducer, it makes sense to add it in action implementation vs here.

src/reducers/account/index.js
```diff
@@ -91,7 +91,16 @@ const accessKeys = handleActions({
 
 const transactions = handleActions({
     [getTransactions]: (state, { error, payload, ready }) => {
-        const hash = state.transactions && state.transactions.reduce((h, t) => ({
+        const { transactions } = state
+
+        if (!ready && transactions && transactions[0].query_account_id !== state.accountId) {
+            return {
+                ...state,
+                transactions: undefined
```
vgrichina, line 99: why `undefined` specifically vs e.g. empty array?

src/utils/explorer-api.js
marcinbodnar: > why not just add this as a field to returned object?

You mean in `map` below? We can, the result will be exactly the same - there will be accountId prop for every tx.

> also given that this is info only useful for reducer, it makes sense to add it in action implementation vs here.

We can put it in `reducer` but we would have to check if accountId is  signer_id or receiver_id, which is already checked in SQL query.

Also, my intention was to make it more generic, `getTransactions` has optional `accountId` - adding it here allows us to use this function for example on Profile page, for another user than already logged in.

src/reducers/account/index.js
marcinbodnar: At this moment if `getTransactions` is launched first time, for `!ready` we have `transactions` undefined (not empty), so I implement it the same here - thanks to that we would not have to check `transactions.length` additionally.

src/utils/explorer-api.js
vgrichina: >  We can, the result will be exactly the same - there will be accountId prop for every tx.

no, I mean just return an object with `{ transactions: [...], accountId: ... }` (i.e. I don't see a need to repeat this for every row). 

the result won't be the same BTW, we won't be transferring unnecessary data over wire (first from DB to explorer, then from explorer to wallet).

> getTransactions has optional accountId

Does it need to be optional vs always required? Seems like `explorer-api` doesn't have to know about concept of current account.

src/reducers/account/index.js
vgrichina: should it be `if (!ready || transactions && transactions[0].query_account_id !== state.accountId)`? 

also given that we can be querying for multiple account IDs, maybe it makes sense to restructure store around this? Smth like:

```
{ transactions: { [accountId]: [ ... transaction list goes here ] } }
```

src/utils/explorer-api.js
marcinbodnar: > no, I mean just return an object with { transactions: [...], accountId: ... } (i.e. I don't see a need to repeat this for every row).

ok, np

marcinbodnar: > Does it need to be optional vs always required?

good point - we can change that, in this case we would only have to add accountId for `getTransactions` action dispatched in `DashboardDetail.js`

src/reducers/account/index.js
marcinbodnar: > should it be if (!ready || transactions && transactions[0].query_account_id !== state.accountId)?

not really, we want to lunch it only for `!ready` and only if transactions list is available (transaction clear will not proceed if for example you just created a new account). The problem was that after user switch account, transactions list for the old account was visible for a short while, till the new transactions list was received (for new accountId). 

> also given that we can be querying for multiple account IDs

what do you mean?

vgrichina: I mean that we can be querying different accounts, so it makes sense to organize transaction data in Redux store keyed by account. That way when you switch to different account – we can show cached data before new transaction list is received.

marcinbodnar: Sure, so maybe we should export all transaction state to separate reducer, instead of keeping it in `account` state?

vgrichina: > maybe we should export all transaction state to separate reducer, instead of keeping it in account state?

sure

marcinbodnar:
@vgrichina ready

marcinbodnar:
@vgrichina can we merge it?

vgrichina:
@marcinbodnar please merge with / rebase on `master`. Looks good overall.

marcinbodnar:
@vgrichina Done.

PR #559, opened 2020-05-05
require recovery method verification as part of setup

Patrick1904:
**This PR is blocked by near-contract-helper changes:**
- [migrations and model changes for recovery methods (135)](https://github.com/near/near-contract-helper/pull/135)
- [add recovery method verification #132](https://github.com/near/near-contract-helper/pull/132)

**Covered in this PR:**

Setup recovery:

- initializeRecoveryMethod
- validateSecurityCode
- setupRecoveryMessage

Re-send recovery link:
- sendNewRecoveryLink
- replaceAccessKey


vgrichina:
@Patrick1904 tested this together with contract helper, found some issues:

- Looks like even if we get error from contract helper – form to enter code still appears. 
- When I press "Enable" for phone number as verification – form still opens with "Email Recovery" selected

Patrick1904:
> * Looks like even if we get error from contract helper – form to enter code still appears.

@vgrichina great catch I'll have a look

> * When I press "Enable" for phone number as verification – form still opens with "Email Recovery" selected

I made those changes [separately in 557](https://github.com/near/near-wallet/pull/577) and will get it ready for review now that this is on master

kcole16:
I'd also suggest we add back the "Confirm" page (after the "Verify Code" page), as it feels a bit jarring to jump from Verify to dashboard, etc with no explanation of the recovery process.

Patrick1904:
> I'd also suggest we add back the "Confirm" page (after the "Verify Code" page), as it feels a bit jarring to jump from Verify to dashboard, etc with no explanation of the recovery process.

I agree. We could use the alert/banner that we're using for seed phrase, or success page. Think it'd be a good idea to align the success experiences if possible.

marcinbodnar:
@Patrick1904 Works well.
Looks like `requestCode` can be removed from `reducers/account/index.js` .

Also, I can see that in the email we write `We cannot resend this email.`, but it's possible from the Profile page.

PR #577, opened 2020-05-13
disable ability to add existing recovery method

Patrick1904:
Includes:
- Disable ability to add same recovery method more than once
- Focus on corresponding method when coming from `/profile`
- Pass existing recovery methods when navigating from `/profile` to avoid loading recovery methods (slow) on `/set-recovery`
- Use globalAlert for email/phone recovery setup success
- Redirect to `/profile` post recovery setup if there's no success url, etc.


Patrick1904:
Holding off on this until [559](https://github.com/near/near-wallet/pull/559) is merged

src/components/accounts/recovery_setup/SetupRecoveryMethod.js
```diff
@@ -53,7 +53,27 @@ class SetupRecoveryMethod extends Component {
...
         success: false,
         emailInvalid: false,
-        phoneInvalid: false
+        phoneInvalid: false,
+        activeMethods: []
+    }
+
+    componentDidMount() {
+        const {loadRecoveryMethods, accountId, router } = this.props;
+        const { activeMethods, method } = router.location;
```
vgrichina, line 62: not sure I understand this, why are we passing `activeMethods` through `router.location`?

vgrichina: (I mean this seems like something that should be taken from Redux store)

Patrick1904: I passed `activeMethods` through router to avoid having to 'check' which methods are active when user lands on `/setup-recovery`. Just to speed up the UX and avoid loading.

kcole16:
Looks good. Did we decide to fix the recovery options list in `/profile` in this PR?

Patrick1904:
> Looks good. Did we decide to fix the recovery options list in `/profile` in this PR? (Email, Phone, Seed: regardless of enabled/disabled)

I'll fix that separately.

> Also, would be good to redirect to `/profile` when setting up recovery methods after account creation (can make separate issue for this though).

I agree. @corwinharrell ?

corwinharrell:
> Also, would be good to redirect to /profile when setting up recovery methods after account creation (can make separate issue for this though).

@Patrick1904 @kcole16 yeah I think that makes sense as opposed to redirecting to dashboard 👍

src/components/accounts/recovery_setup/SetupRecoveryMethod.js
vgrichina: Why not cache it using Redux store instead?

Patrick1904:
@vgrichina now referencing Redux for existing recovery methods. Also added re-direct to `/profile` if there's no success_url, etc.

src/components/accounts/recovery_setup/SetupRecoveryMethod.js
```diff
@@ -104,13 +124,17 @@ class SetupRecoveryMethod extends Component {
...
     handleSetupRecoveryMethod = (securityCode) => {
-        const  { accountId, setupRecoveryMessage, redirectToApp } = this.props;
+        const  { accountId, setupRecoveryMessage, redirectToApp, url, history } = this.props;
 
         setupRecoveryMessage(accountId, this.method, securityCode)
             .then(({ error }) => {
                 if (error) return;
 
-                redirectToApp();
+                if (url && Object.entries(url).length !== 0) {
```
vgrichina, line 133: @Patrick1904 any reason not to encapsulate this into action? Seems like get copy-pasted between different places?

Patrick1904: @vgrichina good point. I added a `fallBack` `url` to `redirectApp`

PR #579, opened 2020-05-14
Disable autocomplete when removing recovery method

vgrichina:


vgrichina:
@Patrick1904 I think we should also do it for other similar cases where we want user to think twice. Also at least Safari was getting confused and proposing to autocomplete e-mail in this input.

Patrick1904:
> @Patrick1904 I think we should also do it for other similar cases where we want user to think twice. Also at least Safari was getting confused and proposing to autocomplete e-mail in this input.

@vgrichina 100%. I'll do an audit and fix other similar cases.

Issue #570, opened 2020-05-08
Success page for Near Shell login


kcole16:
### Overview

To handle the near-shell `/login`, we currently provide a `success_url` with a localhost route. This is failing with some regularity on Linux and Windows machine. 

We should add a page and route `/cli-login-success` to handle this.

### Content
```
You've successfully authorized NEAR Shell!

Please close this window and follow the instructions in the terminal.
```

PR #584, opened 2020-05-15
Success page for Near Shell login (#570)

marcinbodnar:


vgrichina:
@kcole16 @marcinbodnar this more a meta-common on https://github.com/near/near-wallet/issues/570 than this PR in particular:
- I think we actually need to figure out root reasons for problems on Windows/Linux before patching away. Pretty sure it doesn't have to be smth big, e.g. in case of GitPod it was just wrong type of redirect: https://github.com/near/near-shell/pull/363
- We can get both of best worlds if we provide a landing page in wallet (like done here), but also make a way for it to call back localhost (if port was opened successfully). We probably can just try loading 1x1 gif from certain URL or smth like that.

Basically need a bit more time to understand what's best way to integrate this into near-shell experience.

marcinbodnar:
@vgrichina what is the root reason here, it's about this thread?
https://near-protocol.slack.com/archives/C012NUGRDQ9/p1588765049028700

vgrichina:
@marcinbodnar 
> @vgrichina what is the root reason here, it's about this thread?
> https://near-protocol.slack.com/archives/C012NUGRDQ9/p1588765049028700

root reason for the thread has been fixed here https://github.com/near/near-shell/pull/363 AFAIK

however we also have different issue where if somebody runs near-shell on remote machine – redirect after login won't work. I guess this change is for that case? However it will still need corresponding near-shell change.

Looks like we can merge this safely though and integrate with near-shell later.

Issue #548, opened 2020-04-30
Add Ledger to account


kcole16:
### Story

As a user with a Ledger Nano S, I want to use the Ledger to custody my NEAR account key, so I can more securely store up to $10k worth of assets.

### Proposed User Flow
1. User selects "Add Ledger Nano S to Account" in `/profile`
2. Redirected to `/setup-ledger`
3. Views instructions (see below) for installing Ledger app, completes installation (on Ledger)
4. User is prompted to connect Ledger via USB, open the NEAR app
  i. If error, "Connection failed. Please check that you have the NEAR app open"
5. User is asked "Would you like to remove your other Full Access Keys?"
  i. If yes, add Ledger key, remove other Full Access Keys
  ii. If no, add Ledger key
6. Redirected to `/profile`

### Notes
Pretty good example of a Ledger flow from MyCrypto: https://beta.mycrypto.com/add-account/ledger_nano_s

**Likely instructions to install Ledger App**
1. Open Ledger Live and go to Settings (gear icon on the right):

![](https://raw.githubusercontent.com/cosmos/ledger-cosmos/master/docs/img/cosmos_app1.png)

2. Enable developer mode (last option):

![](https://raw.githubusercontent.com/cosmos/ledger-cosmos/master/docs/img/cosmos_app2.png)

3. Now go back to manager and search for NEAR:

![](https://raw.githubusercontent.com/cosmos/ledger-cosmos/master/docs/img/cosmos_app1.png)

corwinharrell:
@kcole16 When do we plan to support other hardware wallets besides Ledger? I ask incase it makes sense to use language like "Connect Hardware Wallet" vs "Add Ledger S" to make things a little bit more flexible and future proof.

corwinharrell:
Also, do we anticipate ever wanting to offer this option as a part of onboarding rather than after the fact?

kcole16:
@corwinharrell 
> When do we plan to support other hardware wallets besides Ledger?
We do not have any current plans for this, and realistically will be a at least a few months before we get to it. We should use Ledger-specific language for now.

> Also, do we anticipate ever wanting to offer this option as a part of onboarding rather than after the fact?
Yes, we should add it to onboarding pretty much immediately after we finish this flow.

corwinharrell:
[...]

PR #594, opened 2020-05-22
Enable Ledger setup

Patrick1904:
**WIP** [548](https://github.com/near/near-wallet/issues/548)


Patrick1904:
@kcole16 @corwinharrell realized we may need user to confirm username to disable Ledger?

kcole16:
You mean in the UI? We should have a record of it locally, and can also get it (or at least infer it) from the Ledger device

corwinharrell:
> I mean for other recovery methods we require user to enter username before disabling the method.

@Patrick1904 @kcole16 yeah IMO let's keep this a consistent pattern across the recovery setup experience.

vgrichina:
@Patrick1904 getting such error in Safari:

```
id: "U2FNotSupported"

message: "U2F browser support is needed for Ledger. Please use Chrome, Opera or Firefox with a U2F extension. Also make sure you're on an HTTP…"
name: "TransportError"
```

and "Connecting..." button is stuck.

Error is expected for Safari, but I think we need to surface it to user and not get stuck in in progress state. Looks like we can rely on Ledger connector to provide distinct `id` for different types of errors.

vgrichina:
> we {name: "TransportError", message: "Failed to sign with Ledger device: U2F DEVICE_INELIGIBLE", stack: "Error↵    at new we (https://near-wallet-pr-594.on…allet-pr-594.onrender.com/src.ec0de8f6.js:924:964", id: "U2F_4", originalError: Error: Sign failed
@Patrick1904 similar thing happens when Ledger is not connected in Chrome (different error), but  also stuck in "Connecting..."

Patrick1904:
> Error is expected for Safari, but I think we need to surface it to user and not get stuck in in progress state. Looks like we can rely on Ledger connector to provide distinct `id` for different types of errors.

@vgrichina should we pull errorMessage directly (since it can change?) or use the `id` to create a custom/translatable string?

vgrichina:
@Patrick1904 I think we need to use `id` to create translatable string.

src/translations/en.global.json
```diff
@@ -568,5 +568,8 @@
                 "remove": "If you lose access to your seed phrase, NEAR Inc. will be unable to assist you in recovery of your account and its funds."
             }
         }
-    }
+    },
+
+    "U2FNotSupported": "U2F browser support is needed for Ledger. Please use Chrome, Opera or Firefox with a U2F extension. Also make sure you're on an HTTPS connection.",
```
Patrick1904, line 573: @vgrichina I added support for two possible 'setup ledger' errors. I'm not sure how we'd add a 'fallback' error string for other possible errors though.

vgrichina: @Patrick1904 here's how to handle missing translations in general, looks like can be overridden for individual `<Translate>` usage: https://ryandrewjohnson.github.io/react-localize-redux-docs/#handle-missing-translations

Another option is to check explicitly whether error `id` is in list of recognized error ids. Something like: 

```js
if (['U2FNotSupported', 'U2F_4'].indexOf(error.id) == -1) {
   // ... do whatever needed for missing translation
}
``` 

P.S. I think we shouldn't just use Ledger error codes in top level of translations. Makes sense to group them, e.g. under `errors.ledger` or smth like that.

Patrick1904: @vgrichina thank you I’ll fix that

src/components/responsive/GlobalAlert.js
```diff
@@ -113,29 +113,40 @@ const CustomMessage = styled(Message)`
...
-                </Message.Content>
-            </CustomMessage>
-        </Container>
-    : null
-)
+const GlobalAlert = ({ globalAlert, clearAlert, closeIcon = true }) => {
+
+    const onMissingTranslation = () => {
+        if (!globalAlert.success) {
+            return 'Sorry an error has occurred. You may want to try again.';
```
Patrick1904, line 120: @vgrichina do you have any idea what I can do in order to pass `<Translate id='errors.ledger.defaultError'/>` here instead? `TypeError: strings.split is not a function`

Patrick1904:
@vgrichina I did some testing and it seems 'ledger-only' mode is the main thing causing errors. Here are a the errors I encountered:

In Ledger-only mode (post deleting all other keys):

1.  Setup seed phrase: `Ledger device: UNKNOWN_ERROR (0x6990)`. Navigating to profile after shows that seed phrase setup was successful (from DB). So seems chain failed, but was still saved in DB.
2. Setup email/phone: get stuck at code verification stage with `Error: Transaction is not signed with the given public key`. Trying a second time yields `Ledger device: UNKNOWN_ERROR (0x6990)`
3. Tried sending NEAR and confirming TX on Ledger device and it worked the first time. Second time I got: `Transaction is not signed with the given public key`. Third time it worked.
4. Disabling Ledger and without any other recovery methods setup, trying to setup email/phone/phrase or Ledger recovery gives error: `Error: Access Key {account_id}:{public_key} does not have enough balance 0 for transaction costing 2098246250000000`. 
5. Creating account -> set up email recovery -> setup Ledger, but keep other keys. Then recover account in an Incognito window using the email link. Now the Ledger shows as disabled. And when I try to enable it, I get error: `Error: The public key ed25519:216LpdHooz8c1Sqds5ynYUyd58Gc5zrhzhzKzScEPsVC is already used for an existing access key`. Then navigating to profile after and it shows it as enabled, and Ledger key shows in Full Access Keys.
6. Enabling Ledger, but keeping other keys. Then disabling Ledger. Then enable Ledger again and delete all other keys. Deleting is successful but Ledger device asks me to sign (post success), and confirming on Ledger gives error: `Error: Transaction is not signed with the given public key`

vgrichina:
> Ledger device: UNKNOWN_ERROR (0x6990)
> Error: Transaction is not signed with the given public key

@Patrick1904 did you pull latest near-ledger-app code? It should say Version 1.0.1 in About screen on device

Patrick1904:
> @Patrick1904 did you pull latest near-ledger-app code? It should say Version 1.0.1 in About screen on device

Ah mine still says 1.0.0. Only tested using onrender link. Will update it.

vgrichina:
> Now the Ledger shows as disabled. 

This is somewhat expected limitation for now, as long as enabling it again works out alright (going to check on this).

> Disabled Ledger and without any other recovery methods setup, then trying to setup email/phone/phrase or Ledger recovery gives error: Error: Access Key {account_id}:{public_key} does not have enough balance 0 for transaction costing 2098246250000000.

@Patrick1904 hmm, did we just lose all the keys in the account here? Probably just shouldn't allow disabling Ledger without any other recovery methods set up?

Patrick1904:
> Disabled Ledger and without any other recovery methods setup, then trying to setup email/phone/phrase or Ledger recovery gives error: Error: Access Key {account_id}:{public_key} does not have enough balance 0 for transaction costing 2098246250000000.
> 
> @Patrick1904 hmm, did we just lose all the keys in the account here? Probably just shouldn't allow disabling Ledger without any other recovery methods set up?

@vgrichina yes, I think it's because I was able to 'setup' phrase recovery before disabling Ledger, but phrase wasn't saved on Chain, only in DB so it allowed me to delete Ledger

vgrichina:
Going to merge this as works ok for happy path with latest Ledger app, will make more robust in follow up.

Issue #566, opened 2020-05-07
New user trying to follow the registration flow ends up with "Unknown App is requesting authorization"


frol:
**Problem**

Following Wallet redirects I ended up with the following URL:

https://wallet.nearprotocol.com/sign?app_url=https%3A%2F%2Fkhorolets.github.io&contract_id=nearoverflow.testnet&failure_url=https%3A%2F%2Fkhorolets.github.io%2Fnearoverflow-demo%2F1&public_key=ed25519%3ADdioPWCgFh2c1MTtYKpXhLVDpRusAg4j5HUcPMWCwLwD&redirect_url=%2Fsign&referrer=https%3A%2F%2Fkhorolets.github.io%2Fnearoverflow-demo%2F&success_url=https%3A%2F%2Fkhorolets.github.io%2Fnearoverflow-demo%2F1&title=NEAROverflow

![image](https://user-images.githubusercontent.com/304265/81327674-01daa800-9072-11ea-93d8-90434510bf0c.png)

The "More information" button leads to a completely blank page

**Expected Behavior**

I expected the page to display "NEARoverflow" instead of "Unknown App".

**Steps to reproduce**

(I did not try to minimize the flow, so some of the steps might not be useful)

1. I landed on the page suggesting me to transfer some tokens, but my frol.near account did not have enough tokens, so I decided to create a new account
2. Creation of `frol.testnet` failed due to timeout, so I had to pick a new account id `frol2.testnet`, followed the pass-phrase recovery process
3. Once the account creation was completed, I was redirected to the page I described above

marcinbodnar:
There are several problems here, it was quite problematic to find. The best way to solve it all would be to rebuild it and apply the same approach as in `/login` where all data from URL is stored in Session Storage. I will implement it.

@kcole16 @vgrichina Also one question. Now, as app title, we are showing referrer domain, we want to change it to app title, or show app title only if it's available, if not available show referrer domain?

PR #595, opened 2020-05-22
Fixes for signing transactions (#566)

marcinbodnar:


vgrichina:
@marcinbodnar we cannot be trusting `title` passed by the caller (they can pass anything to scam user), that's why we need to show Referrer domain instead.

marcinbodnar:
@vgrichina got it. It's not using title. Want to merge it?

vgrichina:
@marcinbodnar  may you resolve conflicts?

marcinbodnar:
@vgrichina Done.

Issue #506, opened 2020-04-08
Setup bug reporting software (BugSnag, Sentry)


kcole16:
### Overview
Using bug tracking/log reporting services such as BugSnag/Sentry will help us catch production issues quickly, and provide user logs and context to help debug.

### Acceptance Criteria
- [x] Logs HTTP errors
- [x] Logs client-side errors (JS, React, etc)
- [x] Does not log the following data
  - [ ] Account name
  - [ ] Email
  - [ ] Phone
  - [x] Seed phrase

vgrichina:
This is not done, checklist not complete. Still need to filter data out client side (seed phrase in url gets logged).

PR #602, opened 2020-05-27
Integrate Sentry for error reporting

vgrichina:
We need to start tracking errors as the goal is stabilize as much as possible for mainnet. Some flows anecdotally have a chance to fail with error and get stuck (e.g. Sasha got stuck creating account).

Sentry chosen because it has decent open-source policy: https://sentry.io/_/open-source/  and we can self-host it if needed.

Fixes https://github.com/near/near-wallet/issues/506

WIP: not configured data scrubbing yet

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@sentry/browser": "^5.15.5",
     "connected-react-router": "^6.6.0",
     "fetch-send-json": "0.0.2",
     "js-sha256": "^0.9.0",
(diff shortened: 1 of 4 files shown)


kcole16:
@vgrichina Do you think we should do proactive filtering of potentially sensitive data (especially worried about recovery phrases)? I don't think we'll leak that via errors, but not 100% sure.

https://docs.sentry.io/error-reporting/configuration/filtering/?platform=javascript

vgrichina:
> @vgrichina Do you think we should do proactive filtering of potentially sensitive data (especially worried about recovery phrases)? I don't think we'll leak that via errors, but not 100% sure.
> 
> https://docs.sentry.io/error-reporting/configuration/filtering/?platform=javascript

yes, that's what I meant in:

> WIP: not configured data scrubbing yet

need to tinker with it a little bit more

PR #616, opened 2020-05-28
Hide spinner on number input

Patrick1904:


Patrick1904:
@vgrichina @kcole16 changed to `tel` type instead so that you get the number keyboard on mobile. Let me know if you have objections.

vgrichina:
@Patrick1904 let's not use `tel`, previous version worked alright to suggest pasting security code from SMS.

There is also specific way to control keyboard: https://css-tricks.com/finger-friendly-numerical-inputs-with-inputmode/

`tel` might confuse auto-completion I assume

Patrick1904:
@vgrichina Thank you I didn't know

Issue #612, opened 2020-05-27
Can't create one letter prefix account


evgenykuzyakov:
**Problem**
The wallet UI fails to create 1 letter prefix on betanet. E.g. `f.betanet`. It's a valid account ID name. 

**Expected Behavior**
You should be able to create `f.betanet`

Correct Regex is `/^(([a-z\d]+[-_])*[a-z\d]+\.)*([a-z\d]+[-_])*[a-z\d]+$/`

**Steps to reproduce**

1. Go to https://wallet.betanet.near.org/create/ 
2. Try to create `f.betanet`

kcole16:
This is intended behavior. I thought names needed to be 2-64 characters?

evgenykuzyakov:
But the name `f.betanet` is longer than 2 characters, so it's legit. While simple `f` will not be a legal account ID

PR #625, opened 2020-05-29
Check accountId length with suffix, fix Send Money functionality when accountId is defined in URL, additional changes

marcinbodnar:


marcinbodnar:
@kcole16 @Patrick1904 
I resolved a few issues from this comment https://github.com/near/near-wallet/issues/289#issuecomment-604047438 

1. Added small animation when the user enters wrong char into accountId input field - it will blink red, I think it looks good and gives proper feedback to the user.
2. Added tabindex to Protect your Account form.
3. Added indicate of wrong email and phone number that will be shown onBlur - I added it to box (instead of input) so it will look better.
4. Fixed Send Money functionality because it's not working now in case you have accountId in URL.
5. Also fixed #612 .

Patrick1904:
Great changes @marcinbodnar. The invalid character indicator is a nice detail.

PR #636, opened 2020-06-03
Mobile signer support

vgrichina:
**Not mergable in current state; here for reference.**
PoC of support for iOS wallet wrapper @ https://github.com/nearprotocol/near-wallet-ios


agileurbanite:
Closing out this draft as it's not relevant atm. Might be relevant when we look into wallet connect though.

Issue #589, opened 2020-05-19
Login with Ledger


kcole16:
### Stories

As a user with a Ledger-only account, I want to login with my Ledger and view my account, so I can view my balances, history, and manage my account.

As a user with a Ledger-only account, I want to connect my Ledger to sign transactions, so I can use Wallet features and easily interact with apps.

As a user with multiple accounts using the same Ledger, I want to load all of my accounts when I login with Ledger, so I can choose to use any of them.

### Acceptance Criteria

- [ ] User can "Login with Ledger"
  - [ ] All accounts with the Ledger public key are recovered (whether one or multiple)
- [ ] User can view account without Ledger connected after login
- [ ] Mobile ok
- [ ] Tested

### Login Flow
1. User selects "Login with Ledger" on /create
2. Prompted to "Connect your Ledger and open the NEAR app"
  a. If successful, user is redirected to Dashboard (/)
  b. If unsuccessful, user sees "Connection failed. Please reconnect your Ledger and try again"

### Design
https://www.figma.com/file/BhrKSEYsyegyZ0jKsFJ3QR/NEAR-Wallet?node-id=439%3A4892

### Loading accounts
Use this Explorer API call to lookup all accounts by public key: https://gist.github.com/kcole16/c3b088e85d42c0aa6e7612aa66d17f5b

kcole16:
@corwinharrell @Patrick1904 Thoughts on using a modal for the "Sign with Ledger" communication, rather than redirecting them to a page for the signing flow?

Patrick1904:
> @corwinharrell @Patrick1904 Thoughts on using a modal for the "Sign with Ledger" communication, rather than redirecting them to a page for the signing flow?

Could do a modal for quick implementation but an inline Ledger element confirming that the user is in a 'Ledger-only mode' and saying 'Have your Ledger ready to sign' would be slightly better UX IMO. More consistent signing UX = better.

corwinharrell:
- [ ] Design Complete: https://www.figma.com/file/BhrKSEYsyegyZ0jKsFJ3QR/NEAR-Wallet?node-id=439%3A4892

PR #645, opened 2020-06-04
Ledger sign in

Patrick1904:
WIP [589](https://github.com/near/near-wallet/issues/589)


src/components/accounts/ledger/SignInLedger.js
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    return (
+        <Theme>
+            <h1><Translate id='signInLedger.header'/></h1>
+            <LedgerImage/>
+            <p><Translate id='signInLedger.one'/></p>
+            <FormButton
+                onClick={handleSignIn}
+                sending={signingIn === 'true'}
```
vgrichina, line 32: let's not use `true` as one of the states as it's super confusing, especially given that opposite state seems to be `fail`? Are there any other possible states?  Should we just use latest action's request status from Redux state?

Patrick1904: @vgrichina good point. I'll use Redux state instead.

kcole16:
@Patrick1904 @vgrichina What's the latest on this PR?

kcole16:
@Patrick1904 Noticing a few issues here:
1. Seed phrase copy isn't working correctly: The words do not have spaces
2. Setting up Ledger is acting strangely when setting it up and deleting keys
  - Ledger is asking you to sign a TX "confirming multiple actions", which I think might be related to the TX deleting the keys (shouldn't require Ledger here, as still have the localStorage key)

Patrick1904:
> 1. Seed phrase copy/paste isn't working correctly: The words do not have spaces

It seems the layout change, specifically 'display: flex' removes the spaces (not all browsers). Not sure if this is a bug since you're not actually copying the phrase, and if we want to allow that, we should put the copy button there, like mobile? Actually copying works fine.


> 2. Setting up Ledger is acting strangely when setting it up and deleting keys

Is the Ledger acting strangely or what part specifically?

> * Ledger is asking you to sign a TX "confirming multiple actions", which I think might be related to the TX deleting the keys (shouldn't require Ledger here, as still have the localStorage key)

@vgrichina any idea why this is happening?

> 1. Looks like the modal isn't popping up to sign a TX with Ledger (though I assume this isn't implemented yet)

Yes have been holding off on that until the basic 'Ledger sign in' works.

----------

The remaining item left (aside from notes above) is 'sign in':
- Get key from Ledger and lookup accountId(s) using the explorer function @marcinbodnar wrote, and then add those to localStorage to sign in the user. @vgrichina are there any new requirements now after you've updated the Ledger app? If not, can @marcinbodnar help with this part?

kcole16:
> It seems the layout change, specifically 'display: flex' removes the spaces (not all browsers). Not sure if this is a bug since you're not actually copying the phrase, and if we want to allow that, we should put the copy button there, like mobile? Actually copying works fine.

We decided to allow users to copy the phrase, but not encourage it on desktop by offering the copy button (though decided to do so on mobile). The reality is most users will want to copy it, and will view the current behavior as a bug. We can take the stance that we don't support this behavior, but I expect we will get persistent bug reports about this. Seems like the best option is to add spaces to the copied text (as we had before).

> Is the Ledger acting strangely or what part specifically?

Ah sorry, the bullet point under #2 is the details. Updated to make more clear.

> Yes have been holding off on that until the basic 'Ledger sign in' works.

Makes sense. I'll split out the issue into "Login/Use in Ledger only mode" and "Sign with Ledger". Fits a bit better with current v0 plans.

Patrick1904:
> We decided to allow users to copy the phrase, but not encourage it on desktop by offering the copy button (though decided to do so on mobile). The reality is most users will want to copy it, and will view the current behavior as a bug. We can take the stance that we don't support this behavior, but I expect we will get persistent bug reports about this. Seems like the best option is to add spaces to the copied text (as we had before).

Yes we can fix so that you can still highlight and copy the text and retain the spaces in Chrome. Seems not a quick fix because it requires writing the layout without flexbox. My POV is still that we should put an actual 'copy' button on desktop, since we're already doing it on mobile. Who has pen and paper anymore :)

vgrichina:
> It seems the layout change, specifically 'display: flex' removes the spaces (not all browsers). Not sure if this is a bug

copying is discouraged, but also we cannot prevent it from happening

maybe we actually should also duplicate seed phrase in password field so that you can save it in password manager easily. @Patrick1904  @corwinharrell  @kcole16 wonder what you think on this.

vgrichina:
@marcinbodnar @Patrick1904 given I'm back I can continue to work on this. 

> Ledger is asking you to sign a TX "confirming multiple actions", which I think might be related to the TX deleting the keys (shouldn't require Ledger here, as still have the localStorage key)

I think there used to be a bug with Ledger app where it will keep old transaction after signing it once. I think I fixed it in latest updates to Ledger app (in fact this was one of the things delaying direct work on this PR), I can test it all together with latest app to make sure stuff works / fix Ledger app as necessary.

Patrick1904:
> @marcinbodnar @Patrick1904 given I'm back I can continue to work on this.
> 
> > Ledger is asking you to sign a TX "confirming multiple actions", which I think might be related to the TX deleting the keys (shouldn't require Ledger here, as still have the localStorage key)
> 
> I think there used to be a bug with Ledger app where it will keep old transaction after signing it once. I think I fixed it in latest updates to Ledger app (in fact this was one of the things delaying direct work on this PR), I can test it all together with latest app to make sure stuff works / fix Ledger app as necessary.

Sounds good to me. @vgrichina I've rebased the branch so it's up-to-date with master.

Patrick1904:
> > It seems the layout change, specifically 'display: flex' removes the spaces (not all browsers). Not sure if this is a bug
> 
> copying is discouraged, but also we cannot prevent it from happening
> 
> maybe we actually should also duplicate seed phrase in password field so that you can save it in password manager easily. @Patrick1904 @corwinharrell @kcole16 wonder what you think on this.

IMO we should always aim to align the desktop and mobile UX. We offer a 'copy phrase' button on mobile, but we don't on desktop. But we want to make it easy for users to highlight the full phrase on desktop and paste it in a doc etc (and preserve spaces between words). I think we should decide on one policy. Either we allow all users to copy the seed phrase, or we don't. It's not any easier for me to 'write down' the phrase while on desktop, than on mobile.

My perception is that users want to be able to copy the phrase, and will attempt to copy it (by highlighting) even if there's no explicit copy button. My recommendation is therefore that we show the 'copy phrase' button on both desktop and mobile.

vgrichina:
> My recommendation is therefore that we show the 'copy phrase' button on both desktop and mobile.

if it's easiest fix I'm ok on this

marcinbodnar:
@vgrichina is `createLedgerU2FClient` working for you locally?

vgrichina:
@marcinbodnar it used to work, you need to serve website using HTTPS though and use browser with U2F support (e.g. Chrome)

vgrichina:
@Patrick1904 @kcole16 should work now. Copy phrase also added back.

kcole16:
Looks like I'm being "logged in" in that the account is showing in the navbar and on the dashboard, but I'm still redirected to the `/create` page and see this error in the console.

`Error loading account: No account.`

I also noticed that the user will need to confirm one `addKey` action per account that has used the Ledger key. I assumed it wasn't working for a while, then realized it was each of the accounts I'd used Ledger for. Will add some copy to `/sign-in-ledger` to give the user some guidance here.

vgrichina:
>  but I'm still redirected to the /create page 

This seems like a general pattern where we "remember" some internal pages to redirect to.
@marcinbodnar can you look into this? We only should remember pages which aren't a part of create account / recover / etc flows. Maybe we should only redirect to external pages to simplify this?

> I also noticed that the user will need to confirm one addKey action per account that has used the Ledger key. I assumed it wasn't working for a while, then realized it was each of the accounts I'd used Ledger for. Will add some copy to /sign-in-ledger to give the user some guidance here.

@kcole16 let's have a tracking issue to update this in UI as well. I think we should indicate clearly when we are actually waiting for user to confirm on Ledger and when just some other stuff is going on. So needs some progress indicator reflecting this. /cc @corwinharrell @Patrick1904 

Other than that I think it's best to merge this already and improve in follow-up PRs.

marcinbodnar:
@vgrichina We need to add `/sign-in-ledger` to `WALLET_CREATE_NEW_ACCOUNT_FLOW_URLS`. Because now `redirect_url` is cleared when entering `/sign-in-ledger`. But I'm not sure if it would solve the problem. 

I'm finishing for today, but I will check this out tomorrow.

> We only should remember pages which aren't a part of create account / recover / etc flows. Maybe we should only redirect to external pages to simplify this?

It can only happen if for example user enters `https://wallet.nearprotocol.com/create` in a browser and don't have an account or if the user clicks Create Account button while being on any of account flow pages. But we can add a simple condition, that `redirect_url` will not be updated if the URL is one of account flow.

> I also noticed that the user will need to confirm one addKey action per account that has used the Ledger key.

@kcole16 It would be nice if in the middle of the process we could show list of found accounts with additional information that says it is required to confirm for all of them separately in the device. We can also add an icon next to the every account showing if it's the account is already confirmed.

Issue #656, opened 2020-06-08
Improve "Send" Flow


kcole16:
### Overview
The current "Send NEAR" flow is sub-optimal, especially on mobile. cc @corwinharrell @Patrick1904 who are working on improvements.

Patrick1904:
Closing as this issue is now outdated.

PR #657, opened 2020-06-08
Send page fixes

Patrick1904:
Mostly Behavioral bugs. Will address UI improvements in [656](https://github.com/near/near-wallet/issues/656)

### DIFF
--- a/src/components/send-money/SendMoneyAmountInput.js
+++ b/src/components/send-money/SendMoneyAmountInput.js
@@ -135,9 +135,9 @@ class SendMoneyAmountInput extends Component {
                         <Translate id={amountStatusId} />
                     </Segment>)}
                 {amountDisplay ? (
-                    <Balance amount={amountDisplay} /> 
+                    <><Translate id='sendMoney.amountStatusId.sending'/>&nbsp;<Balance amount={amountDisplay}/></> 
                 ) : (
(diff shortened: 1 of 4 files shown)


vgrichina:
@Patrick1904 

> scroll to top when moving forward in send flow

Do we have this problem when moving between different pages in general or does it only apply to flows which are encapsulated inside single component (i.e. not changing URL between steps)?

Seems like we need generic solution, not just for send flow.

Patrick1904:
> Seems like we need generic solution, not just for send flow.

@vgrichina yes exactly, since these are 'views', but don't change route in the flow. The generic route-based scroll-to-top is [here](https://github.com/near/near-wallet/pull/647).

src/components/send-money/SendMoneyAmountInput.js
```diff
@@ -135,9 +135,9 @@ class SendMoneyAmountInput extends Component {
                         <Translate id={amountStatusId} />
                     </Segment>)}
                 {amountDisplay ? (
-                    <Balance amount={amountDisplay} /> 
+                    <><Translate id='sendMoney.amountStatusId.sending'/>&nbsp;<Balance amount={amountDisplay}/></>
```
vgrichina, line 138: @Patrick1904 diamond here (`<>`) looks a bit suspicious

PR #660, opened 2020-06-09
verify code: scroll to top upon landing

Patrick1904:


src/components/accounts/recovery_setup/SetupRecoveryMethod.js
```diff
@@ -71,6 +71,12 @@ class SetupRecoveryMethod extends Component {
         }
     }
 
+    componentDidUpdate(prevProps, prevState) {
+        if (prevState.success !== this.state.success) {
```
vgrichina, line 75: @Patrick1904 any reason not to do it in `handleNext`?  Seems more straightforward.

Patrick1904: Good idea. Fixed.

PR #671, opened 2020-06-12
Fix promise middleware return value

vgrichina:
This allows to handle errors using `try` / `catch`, etc
Especially nice when it's necessary to chain multiple Redux actions.


vgrichina:
@kcole16 this basically should be no-op, but would be helpful to have some smoke check

kcole16:
Tried out create account, send money, login, add recovery. Looks good

PR #673, opened 2020-06-12
Scrub Mixpanel data

kcole16:
Removes keys from Mixpanel tracking


src/middleware/index.js
```diff
@@ -44,12 +44,13 @@ const ACTIONS_TO_TRACK = ['CREATE_NEW_ACCOUNT','ADD_ACCESS_KEY',
...
 
 if (process.env.MIXPANEL_TOKEN) {
-    mixpanel.init(process.env.MIXPANEL_TOKEN)
+    mixpanel.init(process.env.MIXPANEL_TOKEN, {'property_blacklist': ['$current_url']})
 }
 
 const analyticsMiddleware = store => next => action => {
     let details = {
-        pathname: window.location.pathname,
+        pathname: !window.location.pathname.includes('recover-with-link') ?
```
marcinbodnar, line 52: We can store `recover-with-link` as const, we try to do that with all path related functionalities, like here:
https://github.com/near/near-wallet/blob/master/src/utils/wallet.js#L12

Not relevant to this PR, but I think generally if we want to relay on URL's we should store all routes in a separate file as consts, and additionally an array of objects with properties: `path`, `component`, `exact`, so we could map through it in `Routing.js`. What do you think @vgrichina @Patrick1904 ?

yarn.lock
```diff
@@ -9814,4 +9814,4 @@ yargs@^15.3.1:
     string-width "^4.2.0"
     which-module "^2.0.0"
     y18n "^4.0.0"
-    yargs-parser "^18.1.1"
+    yargs-parser "^18.1.1"
```
marcinbodnar, line 9817: nit: newline at the end is missing.

src/middleware/index.js
Patrick1904: I agree, it makes sense to organize this a bit more as number of routes/urls grows

marcinbodnar: OK, I added an issue, to be handled in the near future.

PR #674, opened 2020-06-12
Fixes for Recovery using Seed Phrase page

marcinbodnar:


Patrick1904:
Screen Shot 2020-06-12 at 5 53 44 PM

@marcinbodnar what do you think about putting seed phrase error below the seed phrase input instead? Could create a seed phrase input component, if needed. Seems `AccountFormAccountId` is getting quite big and trying to handle things that aren't directly related to `accountId`. Reusability is great, as long as it's also flexible.

src/components/accounts/RecoverAccountSeedPhraseForm.js
```diff
@@ -17,11 +17,12 @@ const RecoverAccountSeedPhraseForm = ({
     checkAvailability,
     requestStatus,
     clearRequestStatus,
-    setFormLoader
+    setFormLoader,
```
vgrichina, line 20: Any reason to not use just global loader status here? Is this because of account ID verification?

marcinbodnar: Yes, it's for showing `SENDING...` properly, loader was appearing in the button while typing accountId.

Patrick1904:
Yes, there are a few solutions and might require some refactoring. As long as we show the seed phrase error below the seed phrase input.

> `requestStatus` used for all actions is a good idea because we can always show a proper error, but we should have separate solutions for showing inputs errors or form errors.

>create separate error handler for accountId only and for other form errors.

Yes agree, something more scalable.

kcole16:
Agree with @Patrick1904 that the error is more likely to be due to incorrect seed phrase than incorrect account ID (though that's still possible, and we technically can't tell). 

If it's straightforward to move the error to the seed phrase input, we should do that. If it's a large refactor, more open to doing this at another time.

marcinbodnar:
> If it's straightforward

not really, but it's also not a big thing. If we choosing to refactor it I would suggest checking if showing Global Alert wouldn't be a big work, because it would be much more sensible to show it as general alert, because, as you said, we don't know if it's wrong accountId or seed phrase.

kcole16:
@marcinbodnar How about highlighting both inputs in red? I think it's ok to have the message where it currently is if we do that. Thoughts @Patrick1904 ?

marcinbodnar:
@kcole16 OK

vgrichina:
@kcole16 @Patrick1904 @marcinbodnar keep in mind that AFAIK we also can avoid need to enter `accountId` here altogether (by querying explorer API using public key derived from seed phrase).

src/components/accounts/RecoverAccountSeedPhraseForm.js
vgrichina: @marcinbodnar should we make it easy to create actions that opt-out of loader? Or just not use Redux action for transient stuff like looking up account existence?  We can be passing some flag in action `meta` field indicating this.

Patrick1904:
> @kcole16 @Patrick1904 @marcinbodnar keep in mind that AFAIK we also can avoid need to enter `accountId` here altogether (by querying explorer API using public key derived from seed phrase).

Great idea. I'm all for that!

kcole16:
@Patrick1904 Are we using the key -> account lookup anywhere in the wallet yet? (e.g. for Ledger)

Patrick1904:
> @Patrick1904 Are we using the key -> account lookup anywhere in the wallet yet? (e.g. for Ledger)

Not yet, looking at it now

marcinbodnar:
@kcole16 @vgrichina I discussed it with @Patrick1904 I will prepare this functionality, so it will be possible to implement it in recovery from Ledger

marcinbodnar:
@kcole16 @vgrichina @Patrick1904 Are we using this kind of route (with accountId) `/recover-seed-phrase/:accountId?/:seedPhrase?` anywhere? `accountId` is not needed anymore so I change it to `/recover-seed-phrase/:seedPhrase?`, so the old routes will not work anymore.

@kcole16 I changed the error message from `Seed phrase is invalid for this account` to `Seed phrase is invalid` so it will fit form without accountId.

src/components/accounts/RecoverAccountSeedPhraseForm.js
marcinbodnar: > We can be passing some flag in action meta field indicating this.
That is a good idea, I will implement it for creating account.

vgrichina:
> @kcole16 @vgrichina @Patrick1904 Are we using this kind of route (with accountId) `/recover-seed-phrase/:accountId?/:seedPhrase?` anywhere? `accountId` is not needed anymore so I change it to `/recover-seed-phrase/:seedPhrase?`, so the old routes will not work anymore.
> 

I think at current point it should be safe to remote this route altogether. Note that we need to keep `/recover-with-link` route working (with both account and seed phrase) to avoid breaking links though.

marcinbodnar:
@vgrichina OK, I added two routes (with and without accountId) so the old links will work, and additionally we can use a new one.

vgrichina:
> @vgrichina OK, I added two routes (with and without accountId) so the old links will work, and additionally we can use a new one.

I think you misunderstood me. What I meant is that we just need to keep                                     `path='/recover-with-link/:accountId?/:seedPhrase?'` route for recovery links, `recover-seed-phrase` doesn't have to accept params at all.

It's obviously not a blocker though, should work fine as is.

src/utils/explorer-api.js
```diff
@@ -54,3 +54,32 @@ export async function getTransactions(accountId) {
...
+            onSuccess(dataArr) {
+                resolve(dataArr[0][0] && dataArr[0][0].account_id)
+            },
+            onError(err) {
+                reject(err);
+            }
+        }
+    ));
+
+    return accountId
```
vgrichina, line 84: @marcinbodnar there can be multiple matching accounts. We'll need to recover all of them, but only select one. I guess selecting most recent one makes sense if we can order by creation time.

src/utils/explorer-api.js
```diff
@@ -54,3 +54,32 @@ export async function getTransactions(accountId) {
...
+                    account_id
+                FROM 
+                    access_keys
+                WHERE 
+                    public_key = :publicKey
+            `,
+            { publicKey }
+        ],
+        {
+            onSuccess(dataArr) {
```
vgrichina, line 75: nit: looks like it makes sense to make a wrapper function for making `wamp.call` with all the boilerplate related to `Promise` and `onSuccess`, etc. E.g. I used this in other project:

```js
    const queryExplorer = (sql, params) => new Promise((resolve, reject) => wamp.call(
        `${WAMP_NEAR_EXPLORER_TOPIC_PREFIX}.select`,
        [ sql, params ],
        {
            onSuccess(dataArr) {
                resolve(dataArr[0]);
            },
            onError(err) {
                reject(err);
            }
        }
    ));
```

src/components/accounts/RecoverAccountSeedPhrase.js
```diff
@@ -3,21 +3,20 @@ import { withRouter } from 'react-router-dom'
...
 class RecoverAccountSeedPhrase extends Component {
     state = {
-        accountId: this.props.accountId,
         seedPhrase: this.props.seedPhrase
     }
 
     // TODO: Use some validation framework?
     validators = {
-        seedPhrase: value => true // TODO validate seed phrase
+        seedPhrase: value => value.length // TODO validate seed phrase
```
vgrichina, line 19: nit: I prefer `!!value.length` here to make sure it's coerced to boolean

src/components/accounts/RecoverAccountSeedPhrase.js
```diff
@@ -26,36 +25,48 @@ class RecoverAccountSeedPhrase extends Component {
...
-        this.props.recoverAccountSeedPhrase(this.state.seedPhrase, accountId)
+        this.props.recoverAccountSeedPhrase(this.state.seedPhrase)
             .then(({ error }) => {
                 if (error) return
                 this.props.refreshAccount()
                 this.props.redirectToApp()
             })
     }
 
+    checkInvalidSeedPhrase = () => this.props.requestStatus.errorMessage && this.props.requestStatus.errorMessage.includes('Cannot find matching public key')
```
vgrichina, line 49: nit: let's name this `isSeedPhraseValid` (and reverse meaning) as it's not clear from `check` prefix whether method will do something or just return `true` / `false`. 

Is there any way to avoid depending on specific wording of message in thrown error?

marcinbodnar:
> I think you misunderstood me.

@vgrichina Right, now I see that you wrote `recover-with-link`. OK, I will remove params for `recover-seed-phrase` if we are not using it.

src/components/accounts/RecoverAccountSeedPhrase.js
```diff
@@ -26,36 +25,48 @@ class RecoverAccountSeedPhrase extends Component {
...
+                ? this.invalidSeedPhraseRequestStatus
+                : this.props.requestStatus
+            : null
+    )
+
+    get invalidSeedPhraseRequestStatus() {
+        return {
+            success: false,
+            messageCode: 'account.recoverAccount.errorInvalidSeedPhrase'
+    }}
```
vgrichina, line 63: nit: look like typo with formatting

src/components/accounts/RecoverAccountSeedPhrase.js
```diff
@@ -26,36 +25,48 @@ class RecoverAccountSeedPhrase extends Component {
...
             .then(({ error }) => {
                 if (error) return
                 this.props.refreshAccount()
                 this.props.redirectToApp()
             })
     }
 
+    checkInvalidSeedPhrase = () => this.props.requestStatus.errorMessage && this.props.requestStatus.errorMessage.includes('Cannot find matching public key')
+
+    handleRequestStatus = () => (
```
vgrichina, line 51: nit: `handle` assumes method would actually have some side-effect. Seems like here it is more like just a getter for adjusted definition of request status. So we can be more explicit about it:

```js
get requestStatusWithFormValidation() {
    ...
}
```

src/utils/explorer-api.js
marcinbodnar: It seems like `access_keys` table stores  `access_key_type`, `account_id` and `public_key` only. At this moment we are taking the first one:
https://github.com/near/near-wallet/pull/674/files#diff-8240df5baf1b5b5a5233c10523e59958R76
, but I can change it to the last one (more likely the last one will be the most recent one)?

src/components/accounts/RecoverAccountSeedPhrase.js
marcinbodnar: > Is there any way to avoid depending on specific wording of message in thrown error?

We can add error code to error and change `readyStatePromise` to return this code as well - then we would be able to pass it through `requestResultReducer`, and check code instead of a message. This code can be Translate id, so it will be easy to show it.

Let me know what you think. If you are OK with that let me handle it in separate PR.

vgrichina: @marcinbodnar I like this idea, but I don't understand how do you want to modify `readyStatePromise`. Shouldn't it just passthrough `Error` that was thrown, so error code just needs to be part of `Error`?

marcinbodnar: >Shouldn't it just passthrough Error that was thrown, so error code just needs to be part of Error?

@vgrichina you're right, we don't need to modify `readyStatePromise`.

PR #681, opened 2020-06-17
keep order of recovery methods fixed on profile

Patrick1904:
- Keep order of recovery methods fixed
- Refactor `RecoveryContainer` to use the `useRecoveryMethods` hook
- Replace Snackbar with GlobalAlert for recovery link resend success/failure to keep UX consistent with initial recovery method setup.
- Show Ledger module on profile only if you're logged in to account

cc @corwinharrell

### DIFF
--- a/src/utils/wallet.js
+++ b/src/utils/wallet.js
@@ -406,7 +406,7 @@ class Wallet {
     async replaceAccessKey(oldKey, newKey) {
         const accountId = this.accountId;
         await this.getAccount(accountId).addKey(newKey)
-        return await this.removeAccessKey(oldKey)
+        await this.removeAccessKey(oldKey)
     }
 
     async sendNewRecoveryLink(method) {
@@ -418,9 +418,8 @@ class Wallet {
             method,
             seedPhrase,
             publicKey
-        })
-        return await this.replaceAccessKey(method.publicKey, publicKey)
-
+        });
+        await this.replaceAccessKey(method.publicKey, publicKey)
     }
 
     async deleteRecoveryMethod({ kind, publicKey }) {
--- a/src/utils/wallet.js
+++ b/src/utils/wallet.js
@@ -413,13 +413,15 @@ class Wallet {
         const accountId = this.accountId;
         const { seedPhrase, publicKey } = generateSeedPhrase();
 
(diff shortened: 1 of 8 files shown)


src/utils/wallet.js
```diff
@@ -413,13 +413,15 @@ class Wallet {
...
         const { seedPhrase, publicKey } = generateSeedPhrase();
 
-        await this.postSignedJson('/account/resendRecoveryLink', {
-            accountId,
-            method,
-            seedPhrase,
-            publicKey
-        });
-        await this.replaceAccessKey(method.publicKey, publicKey);
+        return await Promise.all([
```
vgrichina, line 416: @Patrick1904 what is the reason for the change? I think there might be a problem here potentially:
1) `Promise.all` means these actions will happen in parallel.
2) It can then happen that `.postSignedJson` fails, but `.replaceAccessKey` succeeds. In that case user will get locked out of given recovery method.

Patrick1904: @vgrichina Good point, didn't think of that. Fixed it [here](https://github.com/near/near-wallet/pull/681/commits/013baaa8be35feb194b39e4ecf7f8ca7e8d4f345).

src/utils/wallet.js
```diff
@@ -406,22 +406,21 @@ class Wallet {
     async replaceAccessKey(oldKey, newKey) {
         const accountId = this.accountId;
         await this.getAccount(accountId).addKey(newKey)
-        await this.removeAccessKey(oldKey)
+        return await this.removeAccessKey(oldKey)
```
vgrichina, line 409: @Patrick1904 why `return`? Not a blocker, just curious.

Patrick1904: Without it I don't get the success triggered when calling `sendNewRecoveryLink`. Maybe there's a different way?

Screen Shot 2020-06-19 at 12 20 33 PM

vgrichina: Got it – yeah, I think we need to change reducers to deal with `Promise<void>` results properly. I'll make a change and merge.

vgrichina: Done here https://github.com/near/near-wallet/pull/681/commits/ac5970a0d89abc736b0401e8af231ff992be4546

Issue #683, opened 2020-06-18
Support recovery of multiple accounts with the same seed phrase


vgrichina:
**Problem**

Multiple accounts can reuse same public key (and so seed phrase). Currently only one of the will be recovered.

**Expected Behavior**

All of the accounts are recovered, but only last one selected.

**Steps to reproduce**

1. Setup account with seed phrase recovery.
2. Add same full access key as used for seed phrase to another account (can be done using `near repl`).
3. Try recovering account using seed phrase.

vgrichina:
Note that this is not completely theoretical problem, as I think it may happen if somebody used Ledger before for multiple accounts and then wants to recover with the same seed phrase.

marcinbodnar:
@vgrichina OK, got it.

kcole16:
@vgrichina Completely re: Ledger and the need to support that, cc @Patrick1904, but that shouldn't affect seed phrase recovery?

PR #691, opened 2020-06-23
Support recovery of multiple accounts with the same seed phrase (#683)

marcinbodnar:


marcinbodnar:
@vgrichina @kcole16 
Additionally, added Global Alert that will appear after redirection, which will show a number and list of recovered accounts. So the user will be aware of what actually happened - it might be important mostly if more then one account was recovered. Please check if the global alert message sounds ok.

Seed phrase for testing: `arrest water task base wrap snap coil cherry era potato whisper spoon`

Additional question. At this moment while recovery (with seed phrase or with a link) we are not checking if the account is already added. It's a small thing but it might be a little bit more efficient to check first if an account is already added and if so generate neither set new key.

vgrichina:
@marcinbodnar 

> Additional question. At this moment while recovery (with seed phrase or with a link) we are not checking if the account is already added. It's a small thing but it might be a little bit more efficient to check first if an account is already added and if so generate neither set new key.

yeah, it's a good idea to add this check.

PR #694, opened 2020-06-24
Improvements for Recover using Seed Phrase 

marcinbodnar:


src/utils/wallet.js
```diff
@@ -440,7 +440,9 @@ class Wallet {
         const accountsIds = await getAccountId(publicKey)
 
         if (!accountsIds.length) {
-            throw new Error(`Cannot find matching public key`);
+            const error = new Error(`Cannot find matching public key`)
```
vgrichina, line 443: I think it makes sense to define some custom type like `WalletError`, so that you can do it like:

```
throw new WalletError('account.recoverAccount.errorInvalidSeedPhrase', { publicKey });
```

and it would look up message automatically (and also include relevant parameters like `publicKey`)

marcinbodnar:
@vgrichina 
Additionally I would propose to refactor `AccountFormAccountId` component by moving some logic to `checkNewAccount`/`checkAccountAvailable` methods in `wallet.js`. With this change `AccountFormAccountId` component will be much simpler, and `checkNewAccount`/`checkAccountAvailable` methods will contain all checks, like: 
 - accountId length
 - same accountId check, for Send Money
 - `checking...` will be handled by formLoader only.

The only problematic thing is a timeout while typing, it will be also moved to `checkNewAccount`/`checkAccountAvailable` methods but it can be optional, so the immediate check will also be possible.

What do you think?

src/utils/wallet.js
```diff
@@ -440,7 +441,7 @@ class Wallet {
         const accountsIds = await getAccountId(publicKey)
 
         if (!accountsIds.length) {
-            throw new Error(`Cannot find matching public key`);
+            throw new WalletError('Cannot find matching public key', 'account.recoverAccount.errorInvalidSeedPhrase', { aaa: 'bbb' })
```
vgrichina, line 444: @marcinbodnar 
`, { aaa: 'bbb' }` this doesn't look right. Any reason not to pass `publicKey` as in original suggestion?

marcinbodnar: My bad. Fixed.

vgrichina:
@marcinbodnar 

> moving some logic to checkNewAccount/checkAccountAvailable methods in wallet.js. 

Makes sense to me

> same accountId check, for Send Money

Do we need to have different `check` method for cases when receiver account is selected?

> checking... will be handled by formLoader only.

Not sure how it will work, however something I've been thinking about is that we can add some `meta` to Redux actions to indicate how request status should be processed. So that we can clearly separate actions which affect global "readiness" state (something that user initiated like sending money / adding key / etc) and actions that are more or less "background" (e.g. checking if account is available, re-loading activity, etc).

> The only problematic thing is a timeout while typing

It makes sense to keep it in UI component. I think it makes sense extract debounce function into utility module though:
https://levelup.gitconnected.com/debounce-in-javascript-improve-your-applications-performance-5b01855e086

marcinbodnar:
@vgrichina 
> Do we need to have different check method for cases when receiver account is selected?

Only because on Send Money we show proper `requestStatus` for this situation. But we can move check to `checkAccountAvailable` and return error with proper `messageCode`.

> however something I've been thinking about is that we can add some meta to Redux actions to indicate how request status should be processed.

OK, got it. 

> I think it makes sense extract debounce function into utility module though:

Sure thing.

> It makes sense to keep it in UI component.

My idea was that on keypress we immediately run `handleCheckAvailability` action, which will be indicated on redux state and it would be possible to show `checking...` without any additional actions. But it can be handled in another way, I will think about the best solution.

Issue #709, opened 2020-06-30
Remove ReCaptcha


kcole16:
### Overview
Given we will not have a faucet for the mainnet wallet (at least for now), we do not really need ReCaptcha. And as @abacabadabacaba pointed out, it is not privacy friendly. 

At the very least, we should consider commenting out/flag the code for now.

cc @vgrichina @Patrick1904 @marcinbodnar

PR #710, opened 2020-06-30
Remove reCAPTCHA

Patrick1904:
https://github.com/near/near-wallet/issues/709


vgrichina:
What is long term plan? Just remove it altogether?

kcole16:
@vgrichina I’d say yes, we should just remove it.

Especially given we no longer have plans for a faucet, and would need a better Sybil resistance/lightweight KYC solution anyway (since the NEaR required to create an account is higher than when we decided to add recaptcha)

PR #718, opened 2020-07-06
remove 2fa methods from recovery methods list

Patrick1904:
Bug: List of recovery methods was showing 2FA methods too. Updated filter to exclude any method that doesn't have a publicKey.

### DIFF
--- a/src/components/profile/Recovery/RecoveryContainer.js
+++ b/src/components/profile/Recovery/RecoveryContainer.js
@@ -82,7 +82,7 @@ const RecoveryContainer = () => {
     const dispatch = useDispatch();
     const account = useSelector(({ account }) => account);
     const accountId = account.accountId;
-    const activeMethods = useRecoveryMethods(accountId).filter(method => method.confirmed);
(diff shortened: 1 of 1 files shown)


src/components/profile/Recovery/RecoveryContainer.js
```diff
@@ -82,7 +82,7 @@ const RecoveryContainer = () => {
     const dispatch = useDispatch();
     const account = useSelector(({ account }) => account);
     const accountId = account.accountId;
-    const activeMethods = useRecoveryMethods(accountId).filter(method => method.confirmed);
+    const activeMethods = useRecoveryMethods(accountId).filter(method => method.publicKey);
```
vgrichina, line 85: @Patrick1904 shouldn't we check `method.confirmed && method.publicKey`? Or am I misunderstanding smth?

Patrick1904: I checked `method.publicKey` since we only add a `publicKey` once a method is confirmed, but that might change. I updated to check `method.confirmed` too.

PR #724, opened 2020-07-08
Language localStorage support and new UI

picturepan2:
- [ ] Support saving language selection to localStorage. 
- [ ] Implement new design.
- [ ] Compare and complete new translations for both Chinese files.
- [ ] Add simple onMissingTranslation fallback. 
- [ ] Add help and lang switcher to all login/recover pages (when user is not logged in).

2 features need other solutions which we can do in future version:

1) Parcel doesn't support [Dynamic Imports with variables](https://github.com/parcel-bundler/parcel/issues/4413) which should be supported to enable loading translation files dynamically. One solution is switching to Webpack.

2) For onMissingTranslation, because the translations will be loaded dynamically, it might make the logic complex to load different translations multiple times even users don't change languages. For now, the missing translations will just show translationID, e.g. `setupLedgerSteps.header` which is better than `translations for xxx in Chinese not found`. 

One possible solution is using a `build script` to copy missing translations from English to other languages to make sure all strings have at least one translation.


vgrichina:
> Parcel doesn't support Dynamic Imports with variables which should be supported to enable loading translation files dynamically. One solution is switching to Webpack.

There are few different ways to workaround, let's not switch to webpack.
1) There can be wrapper which is something like this:

```
function getTranslations(lang) {
   switch(lang) {
   case 'en': return import('translations/en.json');
   case 'ru': return import('translations/ru.json');
   ...
   }
}
```
2) Nothing prevents us from doing explicit HTTP fetch to load translations instead of relying on bundler. 

> For `onMissingTranslation`, because the translations will be loaded dynamically, it might make the logic complex to load different translations multiple times even users don't change languages. For now, the missing translations will just show translationID, e.g. setupLedgerSteps.header which is better than translations for xxx in Chinese not found.

I think it's best to show English translation for missing translations.

picturepan2:
Ok. I will try in another PR. At least this PR is feature and design completed, also translations completed. 🤗

Patrick1904:
@picturepan2 these changes look great! A few notes:

1. Can you remove the `<select>` focus outline color (white and blue)?
2. I like that you're showing the lang switcher in guest mode on desktop. Can you add it for mobile too? You can place it in the same spot (since there's no hamburger menu in guest mode).
3. Can you add a `cursor: pointer` to the lang switcher?
4. On mobile, can you please make the chevron icon on the right-side of the switcher a bit bigger? Please see design for reference.

PR #725, opened 2020-07-08
Display referrer domain instead of app title on /login

marcinbodnar:


src/components/login/Login.js
```diff
@@ -123,9 +123,19 @@ const mapDispatchToProps = {
...
-const mapStateToProps = ({ account }) => ({
-    account
-})
+const mapStateToProps = ({ account }) => {
+    const { referrer } = account.url
+    let referrerDomain;
+    if (referrer) {
+        const referrerUrl = new URL(account.url.referrer)
+        referrerDomain = referrerUrl.hostname
+    }
```
vgrichina, line 132: seems like makes sense to move above logic to src/actions/account.js:

```
40-    if ([...WALLET_CREATE_NEW_ACCOUNT_FLOW_URLS, WALLET_LOGIN_URL, WALLET_SIGN_URL].includes(currentPage)) {
41-        const parsedUrl = {
42:            referrer: document.referrer,
43-            ...parse(search)
44-        }
```

as otherwise it get's duplicated between Login and Sign

marcinbodnar: Sure, in this case, we should save it separately, not to lose full referrer?

vgrichina: we don't use it anywhere else AFAIK

PR #728, opened 2020-07-09
Fix problem with local access key being selected instead of Ledger

vgrichina:


vgrichina:
Merging this hot fix as we didn't notice that transactions won't go through after signing in with Ledger.

PR #729, opened 2020-07-09
Fix Sign in with Ledger redirection

marcinbodnar:


vgrichina:
Screen Shot 2020-07-09 at 4 08 27 PM

@marcinbodnar I tried signing in with Ledger when logging into guest book app (check out https://gitpod.io/#snapshot/d0c6dc02-e8ec-4d9d-a82e-9ca9baa3ff7b) and looks like I got stuck on Ledger sign in screen instead of redirect (see screenshot)

marcinbodnar:
@vgrichina looks like the problem here is with `nearConfig.walletUrl = 'https://near-wallet-pr-729.onrender.com/'` - `/` at the end. Works fine if remove it. In this case, if you would have an account you would be redirected to dashboard instead of `/login`. 

Maybe we should handle this situation? When the user enters `http://wallet.nearprotocol.com//login` maybe detect it and redirect to proper `http://wallet.nearprotocol.com/login`?

vgrichina:
> Maybe we should handle this situation? When the user enters http://wallet.nearprotocol.com//login maybe detect it and redirect to proper http://wallet.nearprotocol.com/login?

@marcinbodnar yeah, given how often I hit this, probably makes sense to normalize URL somehow :)

marcinbodnar:
> @marcinbodnar yeah, given how often I hit this, probably makes sense to normalize URL somehow :)

will do

PR #739, opened 2020-07-16
Login with Ledger - improvements

marcinbodnar:


marcinbodnar:
@kcole16 @corwinharrell 
With these changes we have two modal views / two steps:
1. We are showing info `Please confirm public key on your device`, at this step users have to confirm public key od Ledger device.
2. Next, we are showing found accounts as list with status. Every account might be in three states: 
 - pending - account is waiting for the previous account to be confirmed
 - confirm - user needs to confirm this account on Ledger
 - success

There are very simple texts, please check and send me better versions. 
Additionally, there are error messages if the user reject on Ledger at every step. There is one message if the user rejects the public key and another if the user rejects one of the accounts.

@corwinharrell Please propose a design. There is not much to do but I'm sure you can propose something that will look nice.

corwinharrell:
@kcole16 what's the priority of this in relation to mainnet launch?

kcole16:
@corwinharrell @marcinbodnar If this is a fix for the issue @Patrick1904 mentioned in Slack, it's pretty high priority. Though we could potentially split that into a separate PR.

Patrick1904:
@kcole16 @marcinbodnar  this fixes the bug where recovery/sign in fails if there's an account without a key: https://github.com/near/near-wallet/pull/739/commits/f0353705e15e3f77657ccdc81c4a1c1a1bbb694f. And it may be worth pushing this fix to master with priority. However, the other changes on this PR aren't ready for master yet. Should we push separately?

kcole16:
Ah yes, I think we should push that separately. That's rather important for mainnet.

src/utils/wallet.js
```diff
@@ -328,27 +326,46 @@ class Wallet {
...
-    async signInWithLedger() {
+    async getLedgerAccountIds() {
         const publicKey = await this.getLedgerPublicKey()
         await setKeyMeta(publicKey, { type: 'ledger' })
-        const accountIds = await getAccountIds(publicKey.toString())
+        return (
+            (await Promise.all(
+                (await getAccountIds(publicKey.toString()))
+                    .map(async (accountId) => 
+                        await this.getAccount(accountId).findAccessKey()
```
vgrichina, line 336: it's best to check for specific `publicKey` above explicitly vs using `.findAccessKey()` (which isn't public and so not even guaranteed to stay as is)

src/utils/wallet.js
```diff
@@ -328,27 +326,46 @@ class Wallet {
...
         for (let i = 0; i < accountIds.length; i++) {
             const accountId = accountIds[i]
-            const accessKeys =  await this.getAccessKeys(accountId)
-            const localAccessKey = await this.getLocalAccessKey(accountId, accessKeys)
-            let newKeyPair = await this.addWalletMetadataAccessKeyIfNeeded(accountId, localAccessKey)
-
+            const newKeyPair = newKeyPairs[i]
             if (i === accountIds.length - 1) {
-                await this.saveAndSelectAccount(accountId, newKeyPair)
+                await wallet.saveAndSelectAccount(accountId, newKeyPair)
```
vgrichina, line 360: why change from `this.` to `wallet.` here?

src/reducers/account/index.js
```diff
@@ -146,6 +148,51 @@ const account = handleActions({
     })
 }, initialState)
 
+const ledger = handleActions({
```
vgrichina, line 151: should this go into a separate reducer module? Doesn't seem to be related to single account.

src/components/accounts/ledger/SignInLedger.js
```diff
@@ -6,14 +6,27 @@ import FormButton from '../../common/FormButton';
...
 
-    const handleSignIn = async () => {
+    const gettingAccounts = account.actionsPending.includes('GET_LEDGER_ACCOUNT_IDS')
+    const addingAccounts = account.actionsPending.includes('ADD_LEDGER_ACCOUNT_ID')
+    const savingAccounts = account.actionsPending.includes('SAVE_AND_SELECT_LEDGER_ACCOUNTS')
+    const signingIn = gettingAccounts || addingAccounts || savingAccounts
+
+    const ledgerAccounts = addingAccounts && Object.keys(account.signInWithLedger).map((accountId) => ({
+        accountId,
+        status: account.signInWithLedger[accountId] === 'waiting'
```
vgrichina, line 22: not sure about this structure. Isn't it easier to just always have status in `signInWithLedger[accountId].status` and avoid having to use `typeof` and this nested ternary?

vgrichina:
> @kcole16 @marcinbodnar this fixes the bug where recovery/sign in fails if there's an account without a key: [f035370](https://github.com/near/near-wallet/commit/f0353705e15e3f77657ccdc81c4a1c1a1bbb694f). And it may be worth pushing this fix to master with priority. However, the other changes on this PR aren't ready for master yet. Should we push separately?

I don't think this is

src/utils/wallet.js
vgrichina: Also on a second thought – it's likely better not to do filtering here at all. Instead just handle errors when recovering accounts and skip the ones which fail. There can be other reasons for failure (e.g. not enough tokens for storage) that will benefit from this.

marcinbodnar: In this particular case user disabled Ledger for this account earlier, that's why it is checked here, so the account will not be shown in the accounts list in modal - it might be confusing for user to see account that was already disabled.

src/utils/wallet.js
marcinbodnar: mistake

src/reducers/account/index.js
marcinbodnar: sure

src/components/accounts/ledger/SignInLedger.js
marcinbodnar: Good point. Done.

marcinbodnar:
Thanks @kcole16 
Please, check also error messages (changes in `en.global.json`)

corwinharrell:
Here is a demo of the proposed changes to the confirmation modal when adding access keys to each account upon sign-in:

Kapture 2020-07-24 at 12 22 42

@marcinbodnar for the success message that displays upon sign in, let's avoid listing out the accounts that were recovered, since we will have previously displayed those to the user both in the modal and on their ledger device, and since it's possible that there could be a large number of accounts that get recovered during a single sign in. Instead, let's simplify and display the following message:

```
**4** accounts were recovered. Your most recently created account has been set to active.
```

kcole16:
@marcinbodnar Looks good!

Small thing, noticed the modal reloads/flickers after each "add key" action is confirmed. If it's a quick fix, that's worth adjusting.

src/utils/wallet.js
vgrichina: @marcinbodnar still need to change to not use `.findAccessKey()`

vgrichina:
> Access to one of the accounts was rejected. No accounts were recovered.
To continue, you must authorize read-only access to each account on your Ledger device. Please try again.

@marcinbodnar when I don't confirm one of the accounts (times out) I get such message. Should we still recover the accounts which got confirmed?

vgrichina:
@marcinbodnar looks like we also don't save previously confirmed keys and user has to reconfirm every account.

vgrichina:
@marcinbodnar same problem happens when I reject for one of the accounts.

src/utils/wallet.js
marcinbodnar: @vgrichina done

marcinbodnar:
@vgrichina Yes, now it will recover all accounts or nothing. I can change it, no problem. But the question is - if something happens should we:
1. Recover accounts that were confirmed and select the last one as active - skip other accounts.
2. Only save keys, so the user will not have to confirm them again.

I think 2. is a better idea. Version 1. will be a bit more complicated, because we would have to inform the user, which accounts have not been recovered - it might be confusing.

kcole16:
@marcinbodnar Users should be able to choose which accounts to recover (at least by rejecting some on the Ledger). Agree with 2) for the case when something happens during the recovery.

I don’t think I’d consider this blocking for this specific PR though, as it’s useful to get the PR as it stands into production. 

Thoughts @vgrichina? Can create an issue for selective account recovery.

vgrichina:
@kcole16 @marcinbodnar  I think it's best to implement 2) if it's quick, but yeah we should still merge if it takes more than few hours.

I think what is somewhat confusing is that current UI shows given account as already succeeded and then it resets back on any failure, which is unexpected.

marcinbodnar:
@vgrichina @kcole16 
2. is done

>Users should be able to choose which accounts to recover (at least by rejecting some on the Ledger).

@kcole16 It was discussed on Slack in this thread:
https://near-protocol.slack.com/archives/CQA90EC3W/p1594327743251500
I think it would be a nice improvement, mostly because user will not see an error about rejecting the account and would be able to continue, but I'm not sure if it would be used by users. I can implement it with no problems, let me know.

vgrichina:
@marcinbodnar  If I dismiss dialog showing progress – it still says "signing in..." for a while before timeout. Not a blocker, but I think it should be cancelling operation immediately.

marcinbodnar:
@vgrichina you mean if you close modal while checking accounts?

PR #735, opened 2020-07-14
Lockup balance

vgrichina:
Fixes https://github.com/near/near-wallet/issues/670

PR #740, opened 2020-07-16
Update Mainnet wallet

vgrichina:


kcole16:
@vgrichina We should add #735 into this as well

PR #743, opened 2020-07-20
disable full access keys page

Patrick1904:


vgrichina:
Let's hide it behind a flag (e.g. `process.env.ENABLE_FULL_ACCESS_KEYS`) so that we can iterate on it on testnet while disabling on mainnet. @Patrick1904 

Also generally I think it's enough to just remove a link. If somebody is knowledgable enough to go to specific route manually,  why prevent it?

Patrick1904:
> Let's hide it behind a flag (e.g. `process.env.ENABLE_FULL_ACCESS_KEYS`) so that we can iterate on it on testnet while disabling on mainnet. @Patrick1904
> 
> Also generally I think it's enough to just remove a link. If somebody is knowledgable enough to go to specific route manually, why prevent it?

@vgrichina makes sense, I'll make it an env flag.

@kcole16 are you onboard with the route still being active? Let me know and I'll fix.

kcole16:
@Patrick1904 Ya, I'm ok with the link remaining active.

PR #769, opened 2020-07-30
Sign with Ledger

marcinbodnar:


marcinbodnar:
@kcole16 @vgrichina 
Few things:
1. At this moment we are showing `CANCEL OPERATION` button on Ledger modals, but we are not really able to cancel operation - user can deny it on Ledger only. Maybe we should remove this button?
2. @vgrichina I added checking if the account has ledger access key into `refreshAccount` method and stored it in redux state, so we don't have to check it separately to show modal for every action. But maybe it would be easier to also store 'fullAccessKeys` in redux (on `refresAccount`), so we don't have to request it separatelly it in many places.
3. @vgrichina Maybe we should add `LedgerConfirmActionModal` component to `Routing.js` and make a decision to show it based on redux state? Now it is a bit messy.

kcole16:
@marcinbodnar Good point re: "Cancel Operation". I'm in favor of removing it for now. Thoughts @corwinharrell @Patrick1904 ?

vgrichina:
@marcinbodnar 

I thought we already stored fullAccessKeys in Redux store? Maybe I mistake. Something I think we need to make sure that new account-specific state should better be keyed by account (vs stored only for current account), so that we have more smooth experience when switching accounts.

> Maybe we should add LedgerConfirmActionModal component to Routing.js and make a decision to show it based on redux state? Now it is a bit messy.

Not exactly sure about how it would interact with Routing, but it makes sense to me that make sense to orchestrate everything global through Redux state. This includes both Ledger and 2FA UI most likely – makes sense to use common pattern cc @mattlockyer .

marcinbodnar:
@vgrichina 
>  thought we already stored fullAccessKeys in Redux store?

We are, but we request it in several places separately, but since `getAccessKeys` method is called while refreshing account now, we might just add `fullAccessKeys` to state there and don't request it in different places.

> I think we need to make sure that new account-specific state should better be keyed by account (vs stored only for current account), so that we have more smooth experience when switching accounts

Agree, I already have some thoughts on how to refactor redux state, but there was no time to work on it so far.

> Not exactly sure about how it would interact with Routing

I mention placing it in `Routing` so it will be available for every route component, like `GlobalAlert`

vgrichina:
@marcinbodnar why does Ledger modal need to be plumbed in every component vs once globally?

vgrichina:
> I mention placing it in Routing so it will be available for every route component, like GlobalAlert

@marcinbodnar yeah, this makes sense.

marcinbodnar:
@vgrichina @kcole16 ready for review

vgrichina:
@marcinbodnar I think after my merge such things are off:
- errors.ledger.modal (or smth like this) string not translated somehow
- Ledger sign in seems to not work on betanet wallet (did it work before though?). I guess pretty likely this is related to explorer/indexer API? cc @khorolets @frol

marcinbodnar:
@vgrichina 
Translation fixed.

I'm not able to sign in with ledger even locally, maybe it's related to 2fa changes? I will check it after finish with #786 .

vgrichina:
@marcinbodnar @kcole16  there is minor UX problem (probably best to fix in follow-up) – looks like when transaction takes long time to submit (after already being signed) we still show "Please confirm the operation on your device..." UI, even though it's already confirmed on hardware wallet.

vgrichina:
Resulting error is also "Please check that your device is connected and the NEAR app is open.", even though transaction has exceeded retries.

marcinbodnar:
@vgrichina got it, I will take care of it tomorrow.

PR #792, opened 2020-08-09
Add more gas to linkdrop call

ilblackdragon:
Fixes an issue with not enough gas attached: https://explorer.near.org/transactions/2BUbseT5tzG56kNhzHumQFaPTKc4PfdHGC2TuobGgFBC

### DIFF
--- a/src/utils/wallet.js
+++ b/src/utils/wallet.js
@@ -330,7 +330,7 @@ class Wallet {
         await contract.create_account_and_claim({
             new_account_id: accountId,
             new_public_key: publicKey
-        });
+        }, "100000000000000");
     }
 
     async saveAndSelectAccount(accountId, keyPair) {


kcole16:
@vgrichina Can we merge this as a hot fix to ‘stable’ as well?

vgrichina:
@kcole16 @bowenwang1996 @mattlockyer  we need to understand why this fix works, near drops work ok on testnet without any change.

vgrichina:
I'm cherry-picking it into `stable` branch so that mainnet wallet is fixed, but we still need to understand issue better before merging into `master`

vgrichina:
Implemented with a way to override in config here:
https://github.com/near/near-wallet/commit/70542cb221384d9d8d8f9c37087a559a54937aea

PR #795, opened 2020-08-10
Check near drop amount before prompting 2fa setup

Patrick1904:


kcole16:
Looks good re: setup flow (only shows if > 100).

I think we should also hide the option to enable on `/profile` if it's below the threshold.

vgrichina:
> I think we should also hide the option to enable on /profile if it's below the threshold.
 Revert

@Patrick1904 @kcole16 let's do it in follow up

Patrick1904:
> > I think we should also hide the option to enable on /profile if it's below the threshold.
> > Revert
> 
> @Patrick1904 @kcole16 let's do it in follow up

Sounds good I'll fix in follow-up.

PR #801, opened 2020-08-11
2fa update 2

mattlockyer:
This PR has:
1. Update multisig contract with minor improvements to cleaning up storage when keys are deleted and allowing other keys to delete requests after 15m cooldown period in the case of a key rotation.
2. Prevent the transferring animation when using 2fa for app signing txs - it doesn't make sense to show transferring and then show a modal, because not transferring tx yet. A new action was added to update the status in `sign.js`.
3. Animate confirm button to show "Sending..." when using 2fa for app signing txs.


src/utils/wallet.js
```diff
@@ -36,7 +38,7 @@ const KEY_UNIQUE_PREFIX = '_4:'
 const KEY_WALLET_ACCOUNTS = KEY_UNIQUE_PREFIX + 'wallet:accounts_v2'
 const KEY_ACTIVE_ACCOUNT_ID = KEY_UNIQUE_PREFIX + 'wallet:active_account_id_v2'
 const ACCOUNT_ID_REGEX = /^(([a-z\d]+[-_])*[a-z\d]+\.)*([a-z\d]+[-_])*[a-z\d]+$/
-const MULTISIG_CONTRACT_HASHES = process.env.MULTISIG_CONTRACT_HASHES || ['7GQStUCd8bmCK43bzD8PRh7sD2uyyeMJU5h8Rj3kXXJk'];
+const MULTISIG_CONTRACT_HASHES = process.env.MULTISIG_CONTRACT_HASHES || ['7GQStUCd8bmCK43bzD8PRh7sD2uyyeMJU5h8Rj3kXXJk', 'AEE3vt6S3pS2s7K6HXnZc46VyMyJcjygSMsaafFh67DF'];
```
vgrichina, line 41: may you link relevant contract versions on GitHub as a comment for hashes?

src/utils/twoFactor.js
```diff
@@ -4,6 +4,8 @@ import { WalletError } from './walletError'
 import { promptTwoFactor } from '../actions/account'
 import { ACCESS_KEY_FUNDING_AMOUNT, convertPKForContract, toPK } from './wallet'
 
+window.nearApiJs = nearApiJs
```
vgrichina, line 7: @mattlockyer why these `window.` assignments are needed? Is this debug leftover?

mattlockyer: Yes sorry, will remove and merge. Thanks!

Issue #786, opened 2020-08-06
If a functional call with token transfer attached fails, wallet UI erroneously reports success


janedegtiareva:
**Problem**
If a functional call with token transfer attached fails, wallet UI erroneously reports success

**Expected Behavior**
The UI should wait for the status of function call and report error

**Steps to reproduce**
1.  make a function call with an attached deposit 
2.  set up the call so that it fails (e.g. insufficient balance in a FT transfer)
3.  wallet UI reports success
4.  Observe that explorer shows that transaction failed.

thisisjoshford:
[screen cap](https://drive.google.com/file/d/1CuKfewfb1is1cfkHk-zfWZRL2VBeypHO/view?usp=sharing) of an example

[failed transaction](https://explorer.testnet.near.org/transactions/8UoZxTeR4Ngz9H6zckEoKozMaapfMrKtv2ZYB1eS7zL1) details in explorer

kcole16:
@marcinbodnar Can you take a look at this?

marcinbodnar:
@janedegtiareva @thisisjoshford I'm not sure how to reproduce this bug - if I choose too big donation Wallet will display `Insufficient Funds` and not allow to proceed.

vgrichina:
@marcinbodnar I've been able to repro by making smart contract fail on purpose. Check out https://gitpod.io/#snapshot/5bcbfec4-8edb-4908-8cc8-41b1d276a081

When you try posting message with word "FAIL" – contract execution would fail, but we will still show success in the wallet.

PR #802, opened 2020-08-11
Handle status failure for sign tx (#786)

marcinbodnar:


marcinbodnar:
@vgrichina It's not yet fixed for 2fa, but sign with 2fa doesn't work for now.

kcole16:
@marcinbodnar Looks good. Makes sense that 2FA doesn't work yet (until we update near-api-js) I believe cc @mattlockyer

src/utils/wallet.js
```diff
@@ -708,7 +708,10 @@ class Wallet {
         }
         for (let { receiverId, nonce, blockHash, actions } of transactions) {
             const [, signedTransaction] = await nearApiJs.transactions.signTransaction(receiverId, nonce, actions, blockHash, this.connection.signer, accountId, NETWORK_ID)
-            await this.connection.provider.sendTransaction(signedTransaction)
+            let { status } = await this.connection.provider.sendTransaction(signedTransaction)
+            if (status.Failure !== undefined) {
+                throw new Error('Transaction failure')
```
vgrichina, line 713: let's include transaction hash and receiver in the message (to allow easy lookup for what went wrong)

marcinbodnar: done

PR #812, opened 2020-08-13
Wallet MainNet Release 2020-08-12

vgrichina:


src/actions/account.js
```diff
@@ -187,17 +249,17 @@ export const { addAccessKey, addAccessKeySeedPhrase, clearAlert } = createAction
...
-            ]);
-
-            return walletReturnData;
+        async (accountId, contractName, publicKey, isNew, fundingContract, fundingKey) => {
+            if (isNew) {
+                await wallet.createNewAccount(accountId, fundingContract, fundingKey)
+            }
+            const fullAccess = true;
+            const res = await wallet.addAccessKey(accountId, contractName, publicKey, fullAccess)
+            wallet.postSignedJson('/account/seedPhraseAdded', {
```
vgrichina, line 258: @mattlockyer missing `await`

vgrichina: also I wonder if we should change the order of operation to first update contract helper. So that we don't have silent errors leading to inconsistency (e.g.  here we can add seed phrase key to account, but fail call to contract helper and so not be able to display it in Profile)

PR #818, opened 2020-08-13
2FA Refactor

mattlockyer:
This PR primarily refactors twoFactor.js module to extend Account. In the process several components have has their calls to wallet.js or twoFactor.js updated to conform to the refactor.

**Note:** please ignore the delegate code in the initial commits, it's been removed.

Main changes:
- Account methods called in wallet.js are overridden in twoFactor.js
- wallet.js calls more streamlined and compact, excess methods removed
- Some special casing still occurs, but could be removed in the future

End to end tested locally:
- create account (email, sms, seed)
- recovery account (link, seed)
- send money
- app sign tx (signAndSendTransactions)


src/utils/twoFactor.js
```diff
@@ -202,10 +141,80 @@ export class TwoFactor {
         console.log('deploying multisig contract for', accountId)
         return await account.signAndSendTransaction(accountId, actions);
     }
-}
 
+    /********************************
+    Account overrides
+    ********************************/
+
+    async sendMoney(receiver_id, amount) {
```
vgrichina, line 149: why are these needed? Why not use `signAndSendTransaction`?

mattlockyer:
@vgrichina please have a look. Saw your branch late so this is a bit different but similar concept.

Drops all special case methods like addKey, deleteKey, sendMoney and overrides `signAndSendTransaction`.

I've been e2e testing this, and it works on everything inside the wallet. NOTE app sign functionCall actions are giving me "gasExceeded" error w/linkdrop. I've tried several different configs of where to attach gas to the original multisig request (the functionCall to a multisig contract attaches the gas that's used in confirm/execute. I attach 100Tgas which should be enough...

mattlockyer:
@vgrichina will resolve conflicts by tomorrow. Any comments on structure and approach? I was able to remove a lot of custom code :+1:

src/utils/twoFactor.js
```diff
@@ -2,29 +2,40 @@ import * as nearApiJs from 'near-api-js'
...
+export class TwoFactor extends Account {
     constructor(wallet) {
+        super(wallet.connection, wallet.accountId)
+        this.accountId = wallet.accountId
         this.wallet = wallet
     }
 
+    getAccount() {
+        let account = this
+        if (this.wallet.recoveryAccount) {
```
vgrichina, line 26: why is this needed?

src/utils/twoFactor.js
```diff
@@ -188,34 +113,100 @@ export class TwoFactor {
...
-    }
+    const request_ids = await contract.list_request_ids()
     for (const request_id of request_ids) {
-        try {
-            await contract.delete_request({ request_id })
-        } catch(e) {
-            console.warn(e)
-        }
+        // TODO discuss if need to show these errors; request cooldown prevents deletion and contract panics. Error is pretty useless to console.warn
+        await contract.delete_request({ request_id }).catch((e) => {})
```
vgrichina, line 209: let's not use `.catch()` in `async` functions, it can be confusing.

> Error is pretty useless to console.warn

what is the typical reason for error that it's useless from your POV?

src/utils/wallet.js
```diff
@@ -690,37 +678,26 @@ class Wallet {
...
             networkId: NETWORK_ID,
             provider: { type: 'JsonRpcProvider', args: { url: NODE_URL + '/' } },
             signer: new nearApiJs.InMemorySigner(tempKeyStore)
         })
         await Promise.all(accountIds.map(async (accountId, i) => {
-            const account = new nearApiJs.Account(connection, accountId)
             this.accountId = accountId
+            await this.loadAccountAndState()
+            if (fromSeedPhraseRecovery) {
+                this.has2fa = false
```
vgrichina, line 690: hmm, if account isn't recovered from seed phrase it doesn't mean it has 2FA. So I am not sure what is the logic here.

src/utils/twoFactor.js
```diff
@@ -2,29 +2,40 @@ import * as nearApiJs from 'near-api-js'
...
+} = nearApiJs
 export const METHOD_NAMES_LAK = ['add_request', 'add_request_and_confirm', 'delete_request', 'confirm']
 const VIEW_METHODS = ['get_request_nonce', 'list_request_ids']
 const METHOD_NAMES_CONFIRM = ['confirm']
-const LAK_ALLOWANCE = process.env.LAK_ALLOWANCE || '10000000000000'
-const actionTypes = {
-    'functionCall': 'FunctionCall'
-}
+const GAS_2FA = process.env.REACT_APP_GAS_2FA || '100000000000000'
+const ALLOWANCE_2FA = process.env.ALLOWANCE_2FA || nearApiJs.utils.format.parseNearAmount('1')
```
vgrichina, line 15: let's use `WALLET_2FA_ALLOWANCE` and `WALLET_2FA_GAS` environment variables

src/utils/twoFactor.js
```diff
@@ -188,34 +113,100 @@ export class TwoFactor {
...
+    const { gas, publicKey, methodName, args } = a[type]
+    const action = {
+        ...a[type],
+        type: type[0].toUpperCase() + type.substr(1),
+        gas: (gas && gas.toString()) || undefined,
+        public_key: (publicKey && convertPKForContract(publicKey)) || undefined,
+        method_name: methodName,
+        args: (args && Buffer.from(args).toString('base64')) || undefined,
+    }
+    delete action.publicKey
```
vgrichina, line 176: should we just list all desired fields explicitly instead of copying all and then deleting?

src/utils/wallet.js
```diff
@@ -113,56 +114,56 @@ class Wallet {
...
-        const accessKeys = await this.getAccessKeys(accountId)
-        if (accessKeys) {
-            const localKey = await this.getLocalAccessKey(accountId, accessKeys)
-            const ledgerKey = accessKeys.find(accessKey => accessKey.meta.type === 'ledger')
-            if (ledgerKey && (!localKey || localKey.permission !== 'FullAccess')) {
-                return PublicKey.from(ledgerKey.public_key)
-            }
+    getAccount(accountId = this.accountId) {
+        if (this.has2fa && this.twoFactor.accountId === accountId) {
+            return this.twoFactor;
```
vgrichina, line 123: @mattlockyer this looks like potentially pretty big problem when switching accounts. `Account` for sure doesn't expect being reused between different accounts. Doesn't seem like `TwoFactor` code has anything for switching accounts either (it takes `accountId` in constructors).

src/utils/twoFactor.js
mattlockyer: ok

src/utils/twoFactor.js
mattlockyer: super.signAndSendTransaction

src/utils/twoFactor.js
mattlockyer: cleaner...

src/utils/twoFactor.js
mattlockyer: This method is to clean up any requests we CAN delete (those older than 15 minutes and unconfirmed) for the user's account.

If we fail deleting those requests because of panic (under 15 minutes) we are not concerned.

src/utils/wallet.js
mattlockyer: Apologies for moving methods.

It made sense to have everything related to account management close to the wallet constructor.

Attempting to simplify wallet.js codebase by having TwoFactor extend Account.

When switching accounts, `refreshAccount` is called, and the instance of `twoFactor`, property (child) of wallet, is re-initialized.

https://github.com/near/near-wallet/blob/1c088935cad80500eddca4f17e1b39d9009e214f/src/utils/wallet.js#L121-L173

src/utils/wallet.js
mattlockyer: The account we're loading is a temporary account based on the recovery private key material.

Since it is a seed phrase it is going to be a FAK so we can skip 2FA modal confirm when adding the new localStorage key.

Once the new key is added, the actual account will be loaded and if it has 2FA enabled, it will be a 2FA account.

src/utils/wallet.js
```diff
@@ -202,33 +203,50 @@ class Wallet {
...
+            const localKey = await this.getLocalAccessKey(accountId, accessKeys)
+            const ledgerKey = accessKeys.find(accessKey => accessKey.meta.type === 'ledger')
+            if (ledgerKey && (!localKey || localKey.permission !== 'FullAccess')) {
+                return PublicKey.from(ledgerKey.public_key)
             }
         }
+        return null
     }
 
+    save() {
```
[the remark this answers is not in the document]
mattlockyer: I think it's easier to understand account state management, but can move back if it's bothersome.

https://github.com/near/near-wallet/pull/818/files#r476487810

mattlockyer:
Deprecated in favor of: https://github.com/near/near-wallet/pull/914

PR #819, opened 2020-08-13
Ledger improvements

marcinbodnar:


marcinbodnar:
@kcole16 @vgrichina 
List of bigger changes:
1. Now we have an additional `txSigned` state in redux, which indicates if Ledger transaction was already signed.

2. Now for every Ledger transaction modal, we show different header depends if tx was confirmed pm device:



I think that we might show it a bit better regarding this new feature. For example, display some standard header, and additional information below indicate status (cc @corwinharrell )

3. Additionally, I handled txStatus for `Sign in with Ledger` modal - so now we have an additional state (for each account), that shows `Confirm` if the account needs confirmation on Ledger device. 


@corwinharrell I chose a blue color, but maybe you have a better proposition?

4. I also handled showing proper error if `RetriesExceeded` error appears for every ledger tx. I decided to show the same message for every tx, but it can be easily personalized and we can show more accurate error for each tx if we want to - not sure if it's needed though. Also, this error will appear for not ledger tx too.

corwinharrell:
@marcinbodnar thinking about the pills and how they are used in this context:

They are really meant to indicate state, rather than to be instructions. So, rather than saying `confirm`, let's say `awaiting confirmation...`. Also regarding the color, I'd like to stick to using colors that also do the job of conveying state, so if we can,, let's stick to the traffic light use of color (green, yellow, red, etc...). For `awaiting confirmation...`, let's use the orange or yellow from the original designs to indicate that something is pending or in progress.

marcinbodnar:
@corwinharrell 
> They are really meant to indicate state, rather than to be instructions

Agree, but how to fit `awaiting confirmation` into this small field - maybe some similar expression?

> Also regarding the color, I'd like to stick to using colors that also do the job of conveying state, so if we can,, let's stick to the traffic light use of color (green, yellow, red, etc...). For awaiting confirmation..., let's use the orange or yellow from the original designs to indicate that something is pending or in progress.

Sure, we use oragne for `pending` `background: #f4c898;color: #ae6816;` - what yellow use?

corwinharrell:
@marcinbodnar what are the conditions for displaying the `pending` state? I'm just wandering if there is an opportunity to combine `pending` and `awaiting confirmation` into a single state vs having two since they are fairly similar from my perspective.

marcinbodnar:
@corwinharrell 
Pending is shown with animated three dots after you confirm on Ledger device, and before the operation is complete. So first you see Confirm, then Pending, and finally Approved. The thing is though that sometimes it takes several seconds till account comes from Pending to Approved.

We added this feature because we want user to know that he already confirmed, and now he needs to wait till tx is send.

corwinharrell:
@marcinbodnar gotcha. How would you feel then about keeping what you have, but then changing "approved" to a green check mark, just so we can cut down on the number of status types that we have, and create a more clear indication for when the operation is complete?

marcinbodnar:
@corwinharrell Sure, so it will be still `Confirm` text but with yellow background and font?

corwinharrell:
@marcinbodnar lets just keep it how you have it with the blue bg for now

marcinbodnar:
@corwinharrell Will do. 
Also, what do you think about the second point from my main comment?

corwinharrell:
@marcinbodnar i like the change of the header depending on whether or not some operation is processing. I think it's a good way to indicate change or that something is happening. I think we'll end up using it sparingly but for this instance I like it.

corwinharrell:
@marcinbodnar can you add a screenshot for the changes that include the new check mark + labels?

marcinbodnar:
@corwinharrell

marcinbodnar:
@chadoh will do

marcinbodnar:
@corwinharrell

corwinharrell:
@marcinbodnar looks great! 🚀

Issue #679, opened 2020-06-15
Filter out LAK that reference the user's account in Authorized Apps


kcole16:
### Overview
Most users will be using limited access keys scoped to the multi-sig contract deployed on their own account, and these keys are currently showing up in "Authorized Apps". This is quite confusing to the user, as it show's their own account as an "Authorized App". 

We should filter these out (e.g. `user.testnet` should not show up in "Authorized Apps" if the selected account is also `user.testnet`

### Acceptance Criteria
- [ ] `/authorized-apps` does not show any keys that are scoped to the user's own account 
- [ ]  The dashboard does not show these keys

kcole16:
@marcinbodnar I thought this was completed (my fault), but seems to still be happening?

Can we prioritize this for this week?

marcinbodnar:
@kcole16 sure thing

PR #821, opened 2020-08-17
Filter Authorized Apps (#679)

marcinbodnar:


src/utils/wallet.js
```diff
@@ -211,8 +211,15 @@ class Wallet {
                 balance: await this.getBalance(),
                 accountId: this.accountId,
                 accounts: this.accounts,
-                authorizedApps: accessKeys.filter(it => it.access_key && it.access_key.permission.FunctionCall),
-                fullAccessKeys: accessKeys.filter(it => it.access_key && it.access_key.permission === 'FullAccess'),
+                authorizedApps: accessKeys.filter(it => (
+                    it.access_key 
+                    && it.access_key.permission.FunctionCall 
+                    && it.access_key.permission.FunctionCall.receiver_id !== this.accountId
```
vgrichina, line 217: @marcinbodnar wonder if we should start using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining given we use Babel anyway

marcinbodnar: Yeah, sounds good.

Issue #828, opened 2020-08-21
Managed to Brick New Account Using Ledger via Linkdrop


mattlockyer:
**Problem**

I followed a linkdrop from Kendall with 40N, chose matt.near, connected ledger, installed NEAR app, opened NEAR app on ledger. When I pressed continue on wallet, it showed:
![image](https://user-images.githubusercontent.com/321340/90918837-3c0a5400-e39a-11ea-8581-8aa03b1479db.png)

Retries using seed phrase or otherwise gives me:
![image](https://user-images.githubusercontent.com/321340/90918790-2301a300-e39a-11ea-8e2b-d11bf500bcec.png)

**Expected Behavior**

I can set up my account. My account in not unrecoverable.

**Steps to reproduce**

1. Follow linkdrop
2. Read instructions
3. Connect Ledger
4. Unlock
5. Enable experimental features
6. Install NEAR app
7. Open NEAR app on device
8. Press continue on wallet
9. Fail

kcole16:
@vgrichina Should we validate each account in the list on initial load? If an account in the list isn't valid, we can just drop it.

PR #829, opened 2020-08-21
refactor create account w/ ledger

Patrick1904:
Create account with Ledger key instead of randomly generated key. Mainly to fix this https://github.com/near/near-wallet/issues/828

### DIFF
--- a/src/translations/en.global.json
+++ b/src/translations/en.global.json
@@ -66,6 +66,9 @@
         },
         "deleteRecoveryMethod": {
             "modal": "You will need to confirm delete key details on your Ledger."
+        },
+        "LedgerPublicKey": {
+            "modal": "You will need to confirm public key access on your Ledger"
         }
     },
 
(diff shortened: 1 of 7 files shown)


kcole16:
@Patrick1904 @vgrichina @mattlockyer Do we think this is ready to merge as a hotfix?

vgrichina:
@kcole16 not yet. I'll make a pass

src/translations/en.global.json
```diff
@@ -66,6 +66,9 @@
         },
         "deleteRecoveryMethod": {
             "modal": "You will need to confirm delete key details on your Ledger."
+        },
+        "LedgerPublicKey": {
```
vgrichina, line 70: @Patrick1904 where is this string used. I wonder if I'm confused or some file is missing from commit.

Patrick1904: @vgrichina this what [`modal.textId`](https://github.com/near/near-wallet/blob/ledger-create-account/src/components/accounts/ledger/LedgerConfirmActionModal.js#L24) returns for the `Translate` `id` in the `LedgerConfirmActionModal` while requesting the Ledger key

PR #832, opened 2020-08-22
Show default translations on missing ones

picturepan2:
- [ ] Show default translations (English) on missing ones.
- [ ] Still keep active when refreshing pages.

image

@Patrick1904 Please review.

### DIFF
--- a/src/components/Routing.js
+++ b/src/components/Routing.js
@@ -52,9 +52,7 @@ const theme = {}
 
 const PATH_PREFIX = process.env.PUBLIC_URL
 
-const onMissingTranslation = ({ translationId }) => {
-    return `${translationId}`
-  };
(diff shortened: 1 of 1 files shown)


Patrick1904:
@picturepan2 - on the guest landing page, trying to switch the language to Russian (since that's missing some translations), but getting this error (breaks the wallet). Please take a look.
Screen Shot 2020-08-24 at 11 25 27 AM

vgrichina:
@picturepan2 @Patrick1904 I think it needs a fallback for the case when default translation isn't available either

picturepan2:
@vgrichina I don't think so. English default translations should be always available and aligned to feature dev progress. If there is no English translations, just remove the `<Translate>` tags.

PR #844, opened 2020-08-26
Multisig Contract Update

mattlockyer:
Latest multisig contract

MUST update ENV on CH wherever this is deployed to include new hash:
`45gayUD7tUKFc3vvGwzoPEeFS6RjdQWu7SHGpxAJe13F`


vgrichina:
@mattlockyer how is the binary? How can I reproduce the build to make sure it matches?

mattlockyer:
> @mattlockyer how is the binary? How can I reproduce the build to make sure it matches?

Just added PR here with the script you will need to produce a matching binary.
https://github.com/near/core-contracts/pull/95

Run ./build.sh then in ./res/minify.sh multisig.wasm

mattlockyer:
@vgrichina just spoke to @evgenykuzyakov and looks like minification is not safe for production yet.

Adding regular build of multisig and update hash.

src/utils/wallet.js
```diff
@@ -43,7 +43,9 @@ const MULTISIG_CONTRACT_HASHES = process.env.MULTISIG_CONTRACT_HASHES || [
     // https://github.com/near/core-contracts/blob/fa3e2c6819ef790fdb1ec9eed6b4104cd13eb4b7/multisig/src/lib.rs
     '7GQStUCd8bmCK43bzD8PRh7sD2uyyeMJU5h8Rj3kXXJk',
     // https://github.com/near/core-contracts/blob/fb595e6ec09014d392e9874c2c5d6bbc910362c7/multisig/src/lib.rs
-    'AEE3vt6S3pS2s7K6HXnZc46VyMyJcjygSMsaafFh67DF'
+    'AEE3vt6S3pS2s7K6HXnZc46VyMyJcjygSMsaafFh67DF',
+    // https://github.com/near/core-contracts/blob/636e7e43f1205f4d81431fad0be39c5cb65455f1/multisig/src/lib.rs
```
vgrichina, line 47: why building this commit instead of latest one?

vgrichina:
@nearmax @evgenykuzyakov is there a way to have reproducible builds with Rust smart contracts? Seems like I'm getting different hash than @mattlockyer

src/utils/wallet.js
```diff
@@ -43,7 +43,9 @@ const MULTISIG_CONTRACT_HASHES = process.env.MULTISIG_CONTRACT_HASHES || [
...
+    '8DKTSceSbxVgh4ANXwqmRqGyPWCuZAR1fCqGPXUjD5nZ'
```
mattlockyer, line 48: here is the latest hash

mattlockyer:
@vgrichina here is sample account with latest multisig deployed from this branch:
https://explorer.testnet.near.org/accounts/za5.testnet

src/utils/wallet.js
mattlockyer: @vgrichina I'm wondering if we have to have exactly the same rustc and do clean build?

I just ran `./build.sh` on https://github.com/near/core-contracts/tree/master/multisig

vgrichina: When I built/deployed the same commit I'm getting following hash: `2BmyXMfHVJH3W82wY9d8tSYMhrHqpKViFeKBoofdxfff`

@nearmax @evgenykuzyakov how should we approach building a contract so that we both get the same binary?

PR #845, opened 2020-08-27
fix missing russian translations w/ html in string

Patrick1904:


Patrick1904:
@picturepan2 I've added English as translation for any missing Russian translations that contain code. Can you please add any missing translations for Chinese to this PR?

PR #847, opened 2020-08-27
Update both Chinese translations

picturepan2:
- [ ] Update both Chinese translations


picturepan2:
@Patrick1904 This is a PR which can be merged into `fix-missing-translations` branch. Later we can merge `fix missing russian translations` PR to the`master`.

PR #858, opened 2020-08-31
Fix displaying modal if a new account is being created with email, phone or seed phrase recovery

marcinbodnar:
This fixes the issue when `ledger confirm action modal` was displayed while creating a new account if current account had Ledger recovery enabled.

### DIFF
--- a/src/reducers/ledger/index.js
+++ b/src/reducers/ledger/index.js
@@ -27,7 +27,7 @@ const ledgerModalReducer = handleActions({
     [combineActions(sendMoney, addAccessKey, signAndSendTransactions, removeAccessKey, addAccessKeySeedPhrase, deleteRecoveryMethod, setupRecoveryMessage, addLedgerAccessKey, getLedgerPublicKey)]: (state, { ready, meta, type }) => ({
         ...state,
         modal: {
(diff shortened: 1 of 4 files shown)


src/reducers/ledger/index.js
```diff
@@ -26,7 +26,7 @@ const ledgerModalReducer = handleActions({
     [combineActions(sendMoney, addAccessKey, signAndSendTransactions, removeAccessKey, addAccessKeySeedPhrase, deleteRecoveryMethod, setupRecoveryMessage, addLedgerAccessKey, getLedgerPublicKey)]: (state, { ready, meta, type }) => ({
```
vgrichina, line 26: @marcinbodnar is there any reason why we have to list all these actions explicitly? Seems like we can do following:
1) dispatch action from `Signer` when there is actual signing / public key request sent to Ledger
2) handle that action specifically instead of trying to piggyback on existing actions.

marcinbodnar: @vgrichina This is a good idea. The only thing is that at this moment we want to show a specific message for each tx (`textId` state) - I'm using prefix from `defaultCodesFor` for that, which is handy. But I guess I can check `actionsPending` state to know what type of tx will be signed. Or maybe you have a better idea?

marcinbodnar: @vgrichina I resolved it in #862 (few last commits)

PR #859, opened 2020-08-31
Update lockup contract support for new naming convention

kcole16:
Add support for the lockup contract naming convention <account_name_without_suffix>.lockup.<network_suffix>


src/utils/wallet.js
```diff
@@ -527,7 +527,8 @@ class Wallet {
         const balance = await account.getAccountBalance()
 
         // TODO: Should lockup contract balance be retrieved separately only when needed?
-        const lockupAccountId = accountId + '.' + LOCKUP_ACCOUNT_ID_SUFFIX
+        const lockupAccountId = accountId.replace('.' + ACCOUNT_ID_SUFFIX, '') + '.' + LOCKUP_ACCOUNT_ID_SUFFIX
```
vgrichina, line 530: this will fail for e.g. `testnet.is.testnet` account

kcole16: I suppose I'd call that an edge case, though it's a fair point.

kcole16: Seems like there could be a few edge cases here, which is the issue with not using the entire account name (including the account suffix) in the lockup schema.

vgrichina: I think it's relatively straighforward if we actually replace only suffix, i.e. smth like `accountId.replace(/\.testnet$/, '.' + LOCKUP_ACCOUNT_ID_SUFFIX)` would work

kcole16:
@vgrichina Thoughts on the current state of this? It's a blocker if we want to deploy lockup contracts this week

PR #861, opened 2020-09-01
refactor: streamline running on localnet

chadoh:
The main goal: if you're already running a local NEAR network, make it as easy as possible to run a local contract-helper & Wallet on that network.

Someone in such a situation will very likely be running a different app using Parcel on port 1234, and will need to sign into Wallet using a different port. This switches the local Wallet port to 4000.

There is no need to gitignore `.env.local` since it stores no sensitive info. Including it makes it much easier to run on a local network with one command:

    NODE_ENV=local yarn start

Finally, note that when running a local network with nearup or rainbow-bridge-cli, the only validator account is named `node0`. Using this as the account creator in contract-helper is the simplest way to get started (see near/near-contract-helper#221), which means the account suffix here needs to match.

I think this is a confusing suffix and would much prefer `localnet`, but I don't know if that's worth doing now. Ideally such a change would flow out from `nearup`, rather than being part of the setup in `near-contract-helper` or here.


chadoh:
The `node0` suffix comes from the change explained in https://github.com/near/near-contract-helper/pull/221

package.json
```diff
@@ -38,7 +38,7 @@
     "predeploy": "yarn build",
     "deploy": "gh-pages -d dist/",
     "update:static": "cp src/wasm/multisig.wasm dist/multisig.wasm",
-    "start": "yarn update:static; parcel src/index.html",
+    "start": "yarn update:static; parcel -p 4000 src/index.html",
```
mattlockyer, line 41: Is the port really necessary here? Might screw up others who assume 1234 is the wallet

mattlockyer (dismissed):
Let's leave port default?

package.json
chadoh: See PR description:

> Someone in such a situation will very likely be running a different app using Parcel on port 1234, and will need to sign into Wallet using a different port. This switches the local Wallet port to 4000

If we want to avoid a tiny speed bump for the existing core Wallet devs, we should keep this unchanged. If we want to optimize for "anyone in the future trying to run everything locally", I think it makes sense to assume they might be running their main app using Parcel's default port, and to use something less common here.

chadoh: I tried adding `PORT=4000` to `.env.local`, but Parcel ignores the setting there

vgrichina: if the goal is to support local devs using unmodified wallet, it makes sense to bundle whole thing and serve it without Parcel

chadoh: I completely agree but I don't have time for that right now

chadoh:
@mattlockyer good to merge?

PR #863, opened 2020-09-01
Fixes setup seed phrase - start over turns into new account creation

mattlockyer:
PR fixes adding seed phrase as a "second" recovery option.

Issue: "start over" button flips bit on isNew to true so wallet thinks it's setting up a new account.


vgrichina:
lgtm, but long-term fix should be just removing `isNew` altogether. The source of truth for this is already available on chain.

Passing it around can be error prone as we can see from this issue.

PR #865, opened 2020-09-02
clean up checking multisig min amount

Patrick1904:


corwinharrell:
Let's use the inline error that we currently use for the `Recovery Methods` module. For copy, what do we think about:

`To enable 2FA, your account requires a minimum balance of x`

where the `Enable` button is disabled until the minimum balance requirement is met.

vgrichina:
@Patrick1904 @kcole16 @corwinharrell  we probably need to have 3 separate balance thresholds:
1) absolute minimum required to deploy multisig, you cannot enable 2FA at all without it (e.g. 35 NEAR)
2) balance at which we start to nag users to enable 2FA to protect account (e.g. 100 NEAR)
3) balance at which we basically block user from using anything before they setup 2FA or Ledger (e.g. 10k NEAR)

vgrichina:
Basically we've been promising a lot around progressive security and we should progressively start implementing it :)

Patrick1904:
> @Patrick1904 @kcole16 @corwinharrell we probably need to have 3 separate balance thresholds:
> 
> 1. absolute minimum required to deploy multisig, you cannot enable 2FA at all without it (e.g. 35 NEAR)
> 2. balance at which we start to nag users to enable 2FA to protect account (e.g. 100 NEAR)
> 3. balance at which we basically block user from using anything before they setup 2FA or Ledger (e.g. 10k NEAR)

I agree w/ all above. Number 1 has been added to this PR, I suggest we handle 2 and 3 in a separate PR.

Screen Shot 2020-09-02 at 3 42 40 PM

src/utils/wallet.js
```diff
@@ -25,7 +25,7 @@ export const IS_MAINNET = process.env.REACT_APP_IS_MAINNET === 'true' || process
 export const DISABLE_CREATE_ACCOUNT = process.env.DISABLE_CREATE_ACCOUNT === 'true' || process.env.DISABLE_CREATE_ACCOUNT === 'yes'
 export const DISABLE_SEND_MONEY = process.env.DISABLE_SEND_MONEY === 'true' || process.env.DISABLE_SEND_MONEY === 'yes'
 export const ACCOUNT_ID_SUFFIX = process.env.REACT_APP_ACCOUNT_ID_SUFFIX || 'testnet'
-export const MULTISIG_MIN_AMOUNT = process.env.REACT_APP_MULTISIG_MIN_AMOUNT || '100'
+export const MULTISIG_MIN_AMOUNT = process.env.REACT_APP_MULTISIG_MIN_AMOUNT || '35'
```
vgrichina, line 28: @Patrick1904 seems like `40` is safer default

PR #867, opened 2020-09-02
use CH for accountIds

mattlockyer:
Temp for testing

### DIFF
--- a/src/utils/wallet.js
+++ b/src/utils/wallet.js
@@ -445,7 +445,10 @@ class Wallet {
         // TODO: getXXX methods shouldn't be modifying the state
         await setKeyMeta(publicKey, { type: 'ledger' })
 
-        const accountIds = await getAccountIds(publicKey.toString())
+        // TODO: switch to indexer when working properly for keys added / deleted by contracts (e.g. linkdrop)
+        // const accountIds = await getAccountIds(publicKey.toString())
+        const accountIds = await fetch(`https://near-contract-helper-2fa.onrender.com/publicKey/${publicKey}/accounts`).then((res) => res.json())
+
         const checkedAccountIds = (await Promise.all(
             accountIds
                 .map(async (accountId) => {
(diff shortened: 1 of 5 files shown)


src/utils/wallet.js
```diff
@@ -445,7 +445,10 @@ class Wallet {
         // TODO: getXXX methods shouldn't be modifying the state
         await setKeyMeta(publicKey, { type: 'ledger' })
 
-        const accountIds = await getAccountIds(publicKey.toString())
+        // TODO: switch to indexer when working properly for keys added / deleted by contracts (e.g. linkdrop)
+        // const accountIds = await getAccountIds(publicKey.toString())
+        const accountIds = await fetch(`https://near-contract-helper-2fa.onrender.com/publicKey/${publicKey}/accounts`).then((res) => res.json())
```
vgrichina, line 450: why not use ACCOUNT_HELPER_URL?

vgrichina: note that you can override it in Render for specific PR

src/utils/wallet.js
```diff
@@ -445,7 +445,10 @@ class Wallet {
         // TODO: getXXX methods shouldn't be modifying the state
         await setKeyMeta(publicKey, { type: 'ledger' })
 
-        const accountIds = await getAccountIds(publicKey.toString())
+        // TODO: switch to indexer when working properly for keys added / deleted by contracts (e.g. linkdrop)
```
vgrichina, line 448: no need to switch actually, in both cases you need to call contract helper, switch can happen inside

src/utils/wallet.js
```diff
@@ -444,8 +443,10 @@ class Wallet {
         const publicKey = await this.getLedgerPublicKey()
         // TODO: getXXX methods shouldn't be modifying the state
         await setKeyMeta(publicKey, { type: 'ledger' })
+        const accountIds = sendJson('GET', `${ACCOUNT_HELPER_URL}/publicKey/${publicKey}/accounts`)
+
+        console.log(accountIds)
```
vgrichina, line 448: debug leftover?

src/utils/wallet.js
```diff
@@ -665,7 +666,7 @@ class Wallet {
 
     async recoverAccountSeedPhrase(seedPhrase, accountId, fromSeedPhraseRecovery = true) {
         const { publicKey, secretKey } = parseSeedPhrase(seedPhrase)
-        const accountIds = await getAccountIds(publicKey)
```
vgrichina, line 668: seems to me that it still makes sense to wrap logic into `getAccountIds` vs using `sendJson` directly

PR #871, opened 2020-09-03
Ask user for account ID during "Sign in with Ledger"

marcinbodnar:


Patrick1904:
@marcinbodnar what would be the best/easiest way to test this?

marcinbodnar:
@Patrick1904 Best would be to test it locally, you can set `conts accountId = []` here so it will mimic a situation where Ledger is not able to find any account:
https://github.com/near/near-wallet/blob/master/src/utils/wallet.js#L469

mattlockyer (approved):
Tested e2e and works.
Future improvement: reduce complexity in `account.js` and/or `wallet.js`. Maybe move more ledger func to ledger actions?

src/utils/wallet.js
```diff
@@ -753,6 +763,11 @@ class Wallet {
             }
         }
     }
+
+    dispatchShowLedgerModal(show) {
+        const [ action ] = store.getState().account.actionsPending.slice(-1)
+        store.dispatch(showLedgerModal({show, action}))
+    }
```
mattlockyer, line 770: Any way to move this to ledger actions and import to wallet.js instead?

Patrick1904 (approved):
Tested locally and works as intended.

PR #872, opened 2020-09-04
Prompt user to re-connect Ledger

Patrick1904:


src/utils/wallet.js
```diff
@@ -415,7 +420,18 @@ class Wallet {
 
     async addLedgerAccessKey(accountId, ledgerPublicKey) {
         await setKeyMeta(ledgerPublicKey, { type: 'ledger' })
-        return await this.getAccount(accountId).addKey(ledgerPublicKey)
+        await this.getAccount(accountId).addKey(ledgerPublicKey)
+        await this.postSignedJson('/account/ledgerKeyAdded', { accountId, publicKey: ledgerPublicKey.toString() })
+    }
+
+    async connectLedger(ledgerPublicKey) {
+        const accessKeys =  await this.getAccessKeys()
```
vgrichina, line 428: @Patrick1904 seems like we can just always call `setKeyMeta` in this case

vgrichina: I also think this should potentially merged with `addLedgerAccessKey`, I'll take a look what it takes to consolidate

Patrick1904: @vgrichina thanks let me know if you need me to make any changes

vgrichina: @Patrick1904 I'll merge as is, let's try to rework in follow-up

I think probably makes sense to rework more globally: get rid of `setKeyMeta` as it is now, etc

PR #873, opened 2020-09-04
Disable ability to enable other recovery methods while using ledger

Patrick1904:


vgrichina:
@Patrick1904 let's guard this with environment variable. `DISABLE_RECOVERY_METHODS_WITH_LEDGER`

vgrichina:
> @Patrick1904 let's guard this with environment variable. `DISABLE_RECOVERY_METHODS_WITH_LEDGER`

on a second thought going to merge as is, as looks like we would just have alternate UX for disabling Ledger

PR #891, opened 2020-09-11
Show proper error when signing in with Ledger on Safari

marcinbodnar:


src/utils/wallet.js
```diff
@@ -475,7 +475,15 @@ class Wallet {
     }
 
     async getLedgerAccountIds() {
-        const publicKey = await this.getLedgerPublicKey()
+        let publicKey
+        try {
+            publicKey = await this.getLedgerPublicKey()
+        } catch (error) {
+            if (error.id === 'U2FNotSupported') {
+                throw new WalletError(error.message, 'signInLedger.getLedgerAccountIds.U2FNotSupported')
```
vgrichina, line 483: @marcinbodnar I'll merge as is, but it makes sense to convert into WalletError deeper in Ledger-related code, i.e. at least in `getLedgerPublicKey` or even more deeper

marcinbodnar: @vgrichina Good point.

PR #898, opened 2020-09-14
Fix 2fa deploy multisig keys

mattlockyer:
Fixes trying to delete keys that are NOT part of account when deploying multisig.

### DIFF
--- a/src/utils/twoFactor.js
+++ b/src/utils/twoFactor.js
@@ -189,7 +189,6 @@ export class TwoFactor {
         const account = this.wallet.getAccount(accountId)
         // replace account keys & recovery keys with limited access keys; DO NOT replace seed phrase keys
         const accountKeys = (await account.getAccessKeys()).map((ak) => ak.public_key)
-            .filter((k) => k.indexOf('9t9') === -1)
         const recoveryMethods = await this.wallet.getRecoveryMethods()
         const seedPhraseKeys = recoveryMethods.data
             .filter(({ kind, publicKey }) => kind === 'phrase' && publicKey !== null)
--- a/src/utils/twoFactor.js
+++ b/src/utils/twoFactor.js
@@ -203,9 +203,6 @@ export class TwoFactor {
                 .filter((k) => !!k && accountKeys.includes(k))
         ])]
 
-        console.log(fak2lak)
-        return
-
         const getPublicKey = await this.wallet.postSignedJson('/2fa/getAccessKey', { accountId })
         const confirmOnlyKey = toPK(getPublicKey.publicKey)
         const newArgs = new Uint8Array(new TextEncoder().encode(JSON.stringify({ 'num_confirmations': 2 })));
--- a/src/utils/twoFactor.js
+++ b/src/utils/twoFactor.js
@@ -190,9 +190,8 @@ export class TwoFactor {
         const accountKeys = (await account.getAccessKeys()).map((ak) => ak.public_key)
         const recoveryMethods = await this.wallet.getRecoveryMethods()
         const seedPhraseKeys = recoveryMethods.data
-            .filter(({ kind, publicKey }) => kind === 'phrase' && publicKey !== null)
+            .filter(({ kind, publicKey }) => kind === 'phrase' && publicKey !== null && accountKeys.includes(publicKey))
             .map((rm) => rm.publicKey)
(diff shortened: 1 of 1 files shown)


src/utils/twoFactor.js
```diff
@@ -186,11 +187,22 @@ export class TwoFactor {
         const contractBytes = new Uint8Array(await (await fetch('/multisig.wasm')).arrayBuffer())
         const { accountId } = accountData
         const account = this.wallet.getAccount(accountId)
-        const accountKeys = await account.getAccessKeys();
+        // replace account keys & recovery keys with limited access keys; DO NOT replace seed phrase keys
+        const accountKeys = (await account.getAccessKeys()).map((ak) => ak.public_key)
+            .filter((k) => k.indexOf('9t9') === -1)
```
vgrichina, line 192: @mattlockyer what does this check mean? What is 9t9?

src/utils/twoFactor.js
```diff
@@ -20,6 +20,7 @@ const actionTypes = {
 export class TwoFactor {
     constructor(wallet) {
         this.wallet = wallet
+        window.twoFactor = this
```
vgrichina, line 23: Debug leftover?

mattlockyer: removed

src/utils/twoFactor.js
```diff
@@ -186,11 +187,22 @@ export class TwoFactor {
...
-        const recoveryKeysED = recoveryMethods.data.map((rm) => rm.publicKey)
-        const fak2lak = recoveryMethods.data.filter(({ kind, publicKey }) => kind !== 'phrase' && publicKey !== null).map((rm) => toPK(rm.publicKey))
-        fak2lak.push(...accountKeys.filter((ak) => !recoveryKeysED.includes(ak.public_key)).map((ak) => toPK(ak.public_key)))
+        const seedPhraseKeys = recoveryMethods.data
+            .filter(({ kind, publicKey }) => kind === 'phrase' && publicKey !== null)
+            .map((rm) => rm.publicKey)
+            .filter((k) => !!k && accountKeys.includes(k))
+        const fak2lak = [...new Set([
+            ...accountKeys.filter((k) => !seedPhraseKeys.includes(k)),
+            ...recoveryMethods.data
```
vgrichina, line 198: I don’t understand this, why we need to include anything not in the accountKeys?

src/utils/twoFactor.js
mattlockyer: testing code, removed

src/utils/twoFactor.js
```diff
@@ -186,11 +187,22 @@ export class TwoFactor {
...
+        const accountKeys = (await account.getAccessKeys()).map((ak) => ak.public_key)
+            .filter((k) => k.indexOf('9t9') === -1)
         const recoveryMethods = await this.wallet.getRecoveryMethods()
-        const recoveryKeysED = recoveryMethods.data.map((rm) => rm.publicKey)
-        const fak2lak = recoveryMethods.data.filter(({ kind, publicKey }) => kind !== 'phrase' && publicKey !== null).map((rm) => toPK(rm.publicKey))
-        fak2lak.push(...accountKeys.filter((ak) => !recoveryKeysED.includes(ak.public_key)).map((ak) => toPK(ak.public_key)))
+        const seedPhraseKeys = recoveryMethods.data
+            .filter(({ kind, publicKey }) => kind === 'phrase' && publicKey !== null)
+            .map((rm) => rm.publicKey)
+            .filter((k) => !!k && accountKeys.includes(k))
```
vgrichina, line 195: Why is this not a part of first .filter call in chain?

src/utils/twoFactor.js
mattlockyer: include all account keys that are not seed phrase keys, we leave those as FAKs

src/utils/twoFactor.js
mattlockyer: will try that

src/utils/twoFactor.js
vgrichina: @mattlockyer i understand that we need to include accountKeys. I don’t understand why ‘recoveryMethods’ part is needed. Like what extra keys it will add and why?

mattlockyer: ahhh I see. Makes sense. I will remove and retest with this.

PR #899, opened 2020-09-15
change account id label, placeholder, and sub label

Patrick1904:
Username -> Account ID
Account ID input sub label -> _Your account ID must include a Top Level Account such as **.near**_

### DIFF
--- a/src/components/GlobalStyle.js
+++ b/src/components/GlobalStyle.js
@@ -206,6 +206,13 @@ export default createGlobalStyle`
         -moz-appearance:textfield;
     }
 
+    .input-sub-label {
+        font-size: 12px;
+        font-style: italic;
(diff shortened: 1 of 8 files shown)


Patrick1904:
Screen Shot 2020-09-14 at 5 21 15 PM

@kcole16 @vgrichina @corwinharrell should we reference `ACCOUNT_ID_SUFFIX` here or always say **.near** for placeholder and sub label?

vgrichina:
> Screen Shot 2020-09-14 at 5 21 15 PM
> 
> @kcole16 @vgrichina @corwinharrell should we reference `ACCOUNT_ID_SUFFIX` here or always say **.near** for placeholder and sub label?

as it is an example, I think it makes sense to always say **.near**

PR #905, opened 2020-09-15
seed phrase verify: remove start over

Patrick1904:


Patrick1904:
@vgrichina seems we should try to get this minimal fix out first? For improvement, AFAIU, it could 'fail' after the account was created with the initial seed phrase, but user doesn't know that so they go back and write down a 'new' seed phrase, potentially deleting the original one, ending up with a bricked account. Could we just check if the account was created with the seed phrase, and if there's an error, we simply recover the account for them in the background and then send to profile as usual?

vgrichina:
> Could we just check if the account was created with the seed phrase, and if there's an error, we simply recover the account for them in the background and then send to profile as usual?

that is probably the best. Agree 100% that we should ship hotfix now though.

PR #906, opened 2020-09-15
Fix recovery methods

### LINKED ISSUE #903: Adding e-mail recovery method with 2FA doesn't seem to work
**Problem**

Check test-seed-phrase1.testnet account.

I've tried enabling e-mail recovery method, but it doesn't show up as enabled in Profile (even though I confirmed 2FA). 

Adding it seems like succeeded: https://explorer.testnet.near.org/transactions/ELgY9S2NJVCS61SeFvxzGRTXKqcQ9YEfJbnESbM8dv6U

**Expected Behavior**

E-mail recovery method should show up in Profile

**Steps to reproduce**

Note sure if this is reproducible but I did this:

1. Create account from linkdrop with seed phrase as recovery.
2. Had to configure 2FA (I chose e-mail) as linkdrop had enough tokens.
3. Tried to add e-mail recovery with the same e-mail

mattlockyer:
fix where recovery methods filters by any access key not only full access key

fixes #903

### DIFF
--- a/src/utils/wallet.js
+++ b/src/utils/wallet.js
@@ -591,7 +591,7 @@ class Wallet {
         const balance = await account.getAccountBalance()
 
         // TODO: Should lockup contract balance be retrieved separately only when needed?
-        const re = new RegExp(`\.${ACCOUNT_ID_SUFFIX}$`);
+        const re = new RegExp(`.${ACCOUNT_ID_SUFFIX}$`);
         const lockupAccountId = accountId.replace(re, '.' + LOCKUP_ACCOUNT_ID_SUFFIX)
         try {
             // TODO: Makes sense for a lockup contract to return whole state as JSON instead of method per property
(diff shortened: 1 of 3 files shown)


Patrick1904 (approved):

src/utils/wallet.js
```diff
@@ -590,7 +591,7 @@ class Wallet {
         const balance = await account.getAccountBalance()
 
         // TODO: Should lockup contract balance be retrieved separately only when needed?
-        const re = new RegExp(`\.${ACCOUNT_ID_SUFFIX}$`);
+        const re = new RegExp(`.${ACCOUNT_ID_SUFFIX}$`);
```
vgrichina, line 594: @mattlockyer why this change? It's not what intended in original regexp (matching a `.`)

mattlockyer: yarn recommended the escape character as unnecessary

mattlockyer: image

vgrichina: got catch, looks like has to be `new RegExp(`\\.${ACCOUNT_ID_SUFFIX}$`);

PR #912, opened 2020-09-17
Use chain to determine if it's a new account 

Patrick1904:


src/actions/account.js
```diff
@@ -273,44 +281,49 @@ export const { getAccessKeys, removeAccessKey, addLedgerAccessKey, connectLedger
...
+    dispatch(redirectTo('/profile'))
+}
+
+export const handleCreateAccountWithSeedPhrase = (accountId, recoveryKeyPair, fundingContract, fundingKey) => async (dispatch) => {
+    await dispatch(createAccountWithSeedPhrase(accountId, recoveryKeyPair, fundingContract, fundingKey))
+    const account = await wallet.refreshAccount()
+    const promptTwoFactor = await wallet.twoFactor.checkCanEnableTwoFactor(account)
+
+    if (fundingContract && promptTwoFactor) {
+        dispatch(redirectTo('/enable-two-factor'))
```
Patrick1904, line 295: @marcinbodnar @vgrichina  whenever I'm creating a **new** account w/ seed phrase or link and then dispatch a redirect, a REPLACE action gets dispatched right after redirect, in this case, replacing `/enable-two-factor` with `/`. I can't seem to figure out why this happens. Do you have any idea?

Note that it only happens while creating a new account and then redirecting.

Screen Shot 2020-09-17 at 6 12 18 PM

marcinbodnar: @Patrick1904 
The problem is in `handleCreateAccountWithSeedPhrase`, instead of `await wallet.refreshAccount()` you need to dispatch `refreshAccount` action because now when you redirect two lines later the account is not visible in Redux.

PR #915, opened 2020-09-19
Fix duplicated modal while removing all existing keys in recovery Ledger setup

marcinbodnar:


Patrick1904:
LGTM. Seems we're just missing translated string here:
Screen Shot 2020-09-18 at 5 52 51 PM

marcinbodnar:
@Patrick1904 
Added translation. 
Also, added one additional change, because there are at least three actions needed on Ledger while removing keys, so now it will properly show actions needed (instead of showing `processing...` all the time)

Patrick1904 (approved):

PR #916, opened 2020-09-19
guard against deleting last recovery method

Patrick1904:


Patrick1904:
@corwinharrell let me know if you want to make any changes
Screen Shot 2020-09-18 at 7 31 06 PM

PR #928, opened 2020-09-24
Footer: update support copy

Patrick1904:


Patrick1904:
@picturepan2 can you please update this copy for Chinese too?

ilblackdragon:
Can we push this through?

PR #943, opened 2020-09-29
Profile visible for not login users

marcinbodnar:


Patrick1904:
@marcinbodnar seems that right-side section doesn't show up after being redirected to profile from seed phrase setup.
Screen Shot 2020-09-29 at 8 20 32 AM

PR #944, opened 2020-09-29
remove 'browse apps' from authorized apps page

Patrick1904:


vgrichina:
@Patrick1904 do we still have to remove it given URL isn't dead anymore?

Patrick1904:
> @Patrick1904 do we still have to remove it given URL isn't dead anymore?

@vgrichina good point. Seems worth keeping it then IMO. Thoughts @kcole16 ?

kcole16:
@Patrick1904 @vgrichina It points to non-live apps though (examples.near.org, which is geared towards developers).

We should add an app directory once we have a few we want to point to, but I think it is confusing to point to examples.near.org

Patrick1904:
> @Patrick1904 @vgrichina It points to non-live apps though (examples.near.org, which is geared towards developers).
> 
> We should add an app directory once we have a few we want to point to, but I think it is confusing to point to examples.near.org

That's true, little value for non-devs. I really like the idea of a dedicated page for apps built on NEAR, and think it'd be worth adding sooner than later. Even apps that haven't fully launched yet. Great way to discover the eco system.

vgrichina:
Should we add some page to docs with list of live apps and redirect `builtwithnear.com` to it instead?

vgrichina:
@evgenykuzyakov when mainnet for NEAR Place?

vgrichina:
Closing as we know have awesomenear.com https://github.com/near/near-wallet/commit/a6bbb2526a1b1a49f635e9af730bd1cb3f35cef5

Issue #911, opened 2020-09-16
Handle seed phrase error when account is created


kcole16:
### Overview
Users sometimes encounter an error when creating an account with a seed phrase via linkdrop, even though the transaction is succeeding. We should only show the user an error at this step if it is the actual account creation transaction that is failing, otherwise they will be mislead into thinking their account wasn't successfully created.

A simple solution is to retry the TX after the linkdrop claim TX, as this should eventually succeed. If it doesn't after ~5 retries, we should check if the account has been created, and if it has, log the user in.

https://github.com/near/near-wallet/pull/905#issuecomment-692793925

### Acceptance Criteria
When creating an account using linkdrop
- [ ] At the `/verify` step, users only encounter an error if the linkdrop claim step is failing
- [ ] If the second "addKey" TX fails, the TX is retried
- [ ] If the account is created successfully, the user is redirected to the `/profile` page

![Screen Shot 2020-09-15 at 1 16 51 PM](https://user-images.githubusercontent.com/5624527/93386370-755ca500-f81c-11ea-9eb8-58f32cbbd313.png)

vgrichina:
@kcole16 @marcinbodnar  retries are already happening in near-api-js for about a minute or even more. 

We need to focus on recovery from failure when user lands on /profile, refreshes page, etc

vgrichina:
> At the /verify step, users only encounter an error if the linkdrop claim step is failing

that's not true AFAIK. I think what is happening in many cases is that account creation succeeds but other steps fail (or are interrupted), e.g. marking given key as seed phrase on contract helper

kcole16:
@vgrichina @Patrick1904 @mattlockyer Have you been able to reproduce these errors? I haven't, but the two common states I'm seeing are:
- Users encounter an error on the `/verify` step and never make it to `/profile`, and then think the linkdrop is toast
- Users make it to `/profile`, but do not have the correct seed phrase written down

And seems like the second case is essentially the first, but they selected "Start Over", and then were redirected from `/verify` to `/profile`?

marcinbodnar:
@kcole16 
> Users encounter an error on the /verify step and never make it to /profile, and then think the linkdrop is toast

It might be because error occurs while creating an account or adding key

> Users make it to /profile, but do not have the correct seed phrase written down

In this case, the account was created, the key was added but the contract helper fails to add recovery method

vgrichina:
[...]

PR #948, opened 2020-09-29
Handle errors for Creating an account with Seed phrase and enabling Seed phrase recovery (#911)

marcinbodnar:


marcinbodnar:
@kcole16 @vgrichina 
Implemented flow:
1. Creating account:
 - CH is run before `addAccessKey`, so we will be able to recover the account even if `addAccessKey` fails
 - if an error appears while adding a key for Seed phrase
 - current account state is cleared
 - user is redirected to `/recover-seed-phrase` and he is able to recover the account with the same seed phrase
 - additionally if user was creating a second account (he is already logged in), after clear state for created account, we are selecting the previous account so the user will be still logged in

2. Enabling Seed phrase as second recovery option:
 - if an error appears while adding a key for Seed phrase
 - user is redirected to `/profile` where he is able to try setup recovery once again
 - additional global alert appears with information what did actually happens.

@kcole16 please check error messages text.

src/actions/account.js
```diff
@@ -282,12 +286,27 @@ export const { getAccessKeys, removeAccessKey, addLedgerAccessKey, connectLedger
 })
 
 export const handleAddAccessKeySeedPhrase = (accountId, recoveryKeyPair) => async (dispatch) => {
-    await dispatch(addAccessKeySeedPhrase(accountId, recoveryKeyPair))
-    dispatch(redirectTo('/profile'))
+    try {
+        await dispatch(addAccessKeySeedPhrase(accountId, recoveryKeyPair))
+    } catch (error) {
+
```
vgrichina, line 292: If you are ignoring an error it's best to have explanation in comment, so that it's clearly not accidental.

src/actions/account.js
```diff
@@ -323,8 +354,13 @@ export const { addAccessKey, createAccountWithSeedPhrase, addAccessKeySeedPhrase
             const publicKey = recoveryKeyPair.publicKey.toString()
             const contractName = null;
             const fullAccess = true;
-            await wallet.addAccessKey(accountId, contractName, publicKey, fullAccess)
-            await wallet.postSignedJson('/account/seedPhraseAdded', { accountId, publicKey })
+            
+            try {
+                await wallet.addAccessKey(accountId, contractName, publicKey, fullAccess)
+                await wallet.postSignedJson('/account/seedPhraseAdded', { accountId, publicKey })
```
vgrichina, line 360: Shoudln't `/account/seedPhraseAdded` happen before `addAccessKey` here (same as above)?

src/actions/account.js
```diff
@@ -305,16 +324,28 @@ export const { addAccessKey, createAccountWithSeedPhrase, addAccessKeySeedPhrase
...
             const newPublicKey = newKeyPair.publicKey
-            await wallet.addAccessKey(accountId, contractName, newPublicKey, fullAccess)
-            await wallet.saveAccount(accountId, newKeyPair)
+            
             await wallet.postSignedJson('/account/seedPhraseAdded', { accountId, publicKey })
+            try {
+                await wallet.addAccessKey(accountId, contractName, newPublicKey, fullAccess)
+                await wallet.saveAccount(accountId, newKeyPair)
+            } catch (error) {
+                wallet.clearAccountState()
```
vgrichina, line 343: Let's not ever call `.clearAccountState()`. Definitely not on _any_ error. Means we are removing local keys if we get timeout error or smth like that.

vgrichina: @marcinbodnar I think it's best to remove this method altogether. We can instead have some UI to indicate when account doesn't exist on blockchain / or there are no local keys.

vgrichina (changes requested):
Please do not submit `.clearAccountState`. This can increase the risk of user losing their account.

src/actions/account.js
marcinbodnar: It's intentional. Because there is a scenario where the user might end up with two seed phrase recovery:
 - CH added a recovery option
 - `addAccessKey` throw error but the key will be added in few minutes
 - in the meantime, user tries to add seed phrase recovery the second time and succeded
 - user ends with two seed phrase recovery options.

I have an idea to make it work a bit better. In this case:
 - CH added a recovery option
 - `addAccessKey` throw error but the key will be added in few minutes
 -  user is redirected to `/profile` but we are showing that the recovery option is `pending` and two options:
   - `refresh` button so he can check if the key was added by `addAccessKey`
   - `cancel` button - so he can clear the key from the chain (if exists) and CH and try again.

What do you think about that? It's a bit more complicated than the implemented solution, but in this case, the user might try to rescue the seed phrase that he already keeps somewhere.

src/actions/account.js
marcinbodnar: Sure

src/actions/account.js
marcinbodnar: OK, but in this case, we want to clear the account that was added few lines higher - so the user is able to recover this account once again. We can probably skip `delete this.accounts[this.accountId]` from `clearAccountState` but we need to run `this.accountId = ''` anyway. Is that ok?

src/actions/account.js
vgrichina: > Because there is a scenario where the user might end up with two seed phrase recovery

not sure how current code solves this. With current code:
- `addAcessKey` can succeed, but `/seedPhraseAdded` fail
- in this case user in reality has 2 seed phrases, but only one shows up in UI

it is far better to end up with 2 seed phrases which show up in UI, than with broken state where there is some key for which we don't know where it belongs

@abacabadabacaba wonder whether we can have any good way for user to identify seed phrases without compromising security. E.g. can storing either first or last word compromise security?  Should we generate one extra word just for identification?

src/actions/account.js
vgrichina: > we want to clear the account that was added few lines higher - so the user is able to recover this account once again.

Not sure I understand the problem. Why should recovery fail if we don't clear account state? 

> but we need to run `this.accountId = ''` anyway

Why is this needed unless we know for sure that account creation failed (e.g. account with such name was created before)?

src/actions/account.js
marcinbodnar: When I was discussing changes on Slack, Matt pointed out that "I think we don't want to focus on the CH case though since that will likely never be happening on mainnet.". 
Is this accurate?

src/actions/account.js
marcinbodnar: Recovery will work but it might be a bit weird for the user. He would be asked to recover the account but at the same time, he will be logged in on this account.

vgrichina: Something being a little weird is definitely better alternative than user losing keys :)

marcinbodnar: of course :) but there is an additional issue - for example:
 - if a user is creating an account
 - the key was not added
 - user is redirected to `/recover-seed-phrase`
 - user skips it - he's logged in so it's possible
 - user add email recovery
 - user delete seed phrase recovery
 - both recovery options are deleted.

Not sure why this is happening, but I guess it's because email recovery is created with the main access key (not recovery seed phrase access key, as it does not exist). And this might cause serious problems because the user might not notice it and ends up with no recovery at all.

My idea was to force a recovering account, but if that's forbidden, I think we need to figure out something else to avoid this scenario.

vgrichina: The problem here is that when user is creating account on mainnet – they are claiming linkdrop vs just creating account for free. So if anything goes wrong – there is potential of real money being lost.
If we "force recovery" as you suggest, we can easily end up in situation where all keys to account are lost. 

We need to avoid this dire situation at all cost. This includes recovering from half-broken states, even if it is harder than wiping everything.

src/actions/account.js
vgrichina: > Matt pointed out that "I think we don't want to focus on the CH case though since that will likely never be happening on mainnet.".

not sure what do you mean here? If the assumption that `/seedPhraseAdded` won't every be happening on mainnet – I don't think this is good assumption at all. It is happening I'm pretty sure. Every HTTP request can fail intermittently even if everything is running ok (e.g. network connection issue).

src/actions/account.js
vgrichina: Many users already lost their mainnet accounts because not writing down seed phrase when they got error – so this is not hypothetical situation.

marcinbodnar: yes, I understand it completely.
The only thing is this issue:

> * if a user is creating an account
> * the key was not added
> * user is redirected to `/recover-seed-phrase`
> * user skips it - he's logged in so it's possible
> * user add email recovery
> * user delete seed phrase recovery
> * both recovery options are deleted.

Maybe it's unlikely, but I think we should still handle it. My proposition:
 - if this situation appears, instead of showing that seed phrase is added (on `/profile`) we will show info that some error appears and a button to fix it, that will redirect the user to `/recover-seed-phrase`

vgrichina:
@marcinbodnar is this ok to be merged? Sorry somehow missed before that you already updated it.

marcinbodnar:
@vgrichina np let me resolve conflicts, and it should be ready to go

marcinbodnar:
@vgrichina ready

PR #961, opened 2020-10-04
Update Chinese translations

picturepan2:
- [ ] Update both Chinese translations

### DIFF
--- a/src/translations/zh-hans.global.json
+++ b/src/translations/zh-hans.global.json
@@ -17,8 +17,7 @@
                 "create": "正在检查可用性",
                 "check": "正在检查"
             },
-            "errorAccountNotExist": "创建账户出错，请重试！",
-            "modal": "请在 Ledger 设备上确认添加新账户密钥。"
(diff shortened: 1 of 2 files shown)


picturepan2:
@Patrick1904 https://github.com/near/near-wallet/blob/4d47a6e7871cb01db3a8a9b9046362a54d0e24ee/src/translations/en.global.json#L221 Please look at this line. if the English string is "Delete Recovery Method"?

Patrick1904:
> @Patrick1904
> 
> https://github.com/near/near-wallet/blob/4d47a6e7871cb01db3a8a9b9046362a54d0e24ee/src/translations/en.global.json#L221
> 
> Please look at this line. If the English string is "Delete Recovery Method"?

@picturepan2 we specifically want to communicate to the user they they need to confirm **deleting a key** on their Ledger device.

picturepan2:
@Patrick1904 The PR looks good now.

Patrick1904 (approved):

Issue #931, opened 2020-09-25
Allow sending to implicit accounts


kcole16:
### Overview
Now that we have implicit accounts, we should allow users to send to accounts with exactly 64 characters, even if they don't exist on-chain. We should also explain the risk of doing so, and prompt the user to double check the address.

### Acceptance Criteria
- [ ] Users can send to accounts with exactly 64 characters that do not exist on-chain
- [ ] When sending to non-existent account, the user sees the below message

**Warning Message**
```
Warning: If this is an exchange account, or an account you control, please double check the account ID. If the account name is entered incorrectly, your funds will be lost forever.
```

kcole16:
**Steps if Implicit Accounts Don't Exist**
1. Transfer 0.5 NEAR to the implicit account (to instantiate)
2. Transfer rest of desired amount to the implicit account

cc @vgrichina

vgrichina:
@bowenwang1996 why is this transaction failing https://explorer.testnet.near.org/transactions/BuvCUudDgzJVPY5Pqh6cZRpUy3kMuPfZAAwYGGJgvWM5?

Are implicit accounts enabled on testnet?

bowenwang1996:
@vgrichina testnet had some issue yesterday. It should be fine now

PR #969, opened 2020-10-05
Allow sending to implicit accounts (#931)

marcinbodnar:


kcole16 (approved):

src/components/accounts/AccountFormAccountId.js
```diff
@@ -133,16 +133,21 @@ class AccountFormAccountId extends Component {
 
     handleCheckAvailability = (accountId, type) => (
         accountId
-            && !(
```
vgrichina, line 136: nit: I find it pretty hard to read very nested conditional expressions.

Should we convert stuff like this to use multiple statements (e.g. early return pattern)? Something like:

```
if (!accountId) {
   return false;
}
if (this.isImplictAccount()) {
   return type != 'create';
}
if (type == 'create') {
 ....   
}
```

src/components/accounts/AccountFormAccountId.js
```diff
@@ -133,16 +133,21 @@ class AccountFormAccountId extends Component {
...
-                type === 'create' 
-                && (!this.handleAccountIdLengthState(accountId) 
-                && !this.checkAccountIdLength(accountId))
+            && this.isImplicitAccount()
```
vgrichina, line 136: feels very weird that method takes `accountId` and then we only check `this.state.accountId` here

src/components/accounts/AccountFormAccountId.js
```diff
@@ -163,15 +168,24 @@ class AccountFormAccountId extends Component {
...
+    get implicitAccountRequestStatus() {
+        return {
+            success: false,
+            messageCode: 'account.available.implicitAccount'
+        }
+    }
+
     get requestStatusWithFormValidation() {
         return this.state.accountId
-            ? this.props.formLoader
```
vgrichina, line 168: see above about early return pattern

src/components/accounts/AccountFormAccountId.js
```diff
@@ -163,15 +168,24 @@ class AccountFormAccountId extends Component {
         }
     }
 
+    get implicitAccountRequestStatus() {
+        return {
+            success: false,
```
vgrichina, line 173: I don't think we should represent implicit account warning as failure

vgrichina:
Screen Shot 2020-10-06 at 5 44 56 PM

@corwinharrell feel like we need to adjust this dialog for sending money, doesn't seem to scale well for implicit accounts

vgrichina:
@marcinbodnar can you fix the
"Go Bac
k" and "Transfe
r" right away?

src/components/accounts/AccountFormAccountId.js
marcinbodnar: Sure thing

marcinbodnar:
@vgrichina 
> @marcinbodnar can you fix the

I'm not seeing this but, also looks very weird like the text doesn't fit inside the button, but there is plenty of space. But anyway, I made small changes it should look fine now.

marcinbodnar:
@vgrichina @corwinharrell 
> @corwinharrell feel like we need to adjust this dialog for sending money, doesn't seem to scale well for implicit accounts

I removed the second accountId (small gray font) as it seems unnecessary at all.

kcole16 (approved):

Issue #973, opened 2020-10-07
Amount field allows to type non-number characters


vgrichina:
**Problem**

Amount field allows to type non-number characters.

<img width="608" alt="Screen Shot 2020-10-06 at 5 23 33 PM" src="https://user-images.githubusercontent.com/126710/95274038-9ac84780-07f9-11eb-9f8b-2af16ec6a6a8.png">

**Expected Behavior**

Only digits and dot are allowed.

**Steps to reproduce**

1. Try to send money to somebody
2. Type some letters

PR #977, opened 2020-10-07
Add pattern to amount input field (#973)

marcinbodnar:


marcinbodnar:
@vgrichina wasn't able to reproduce this bug. Added patter, it should help, also it will block the possibility to add to dots in number (now it's possible).

vgrichina:
@marcinbodnar try in Safari?

PR #979, opened 2020-10-07
Revert "Add pattern to amount input field (#973)"

vgrichina:
Reverts near/near-wallet#977


vgrichina:
@marcinbodnar  I've reverted this as otherwise everything crashes when you input character not in the pattern. How did you test the change?

PR #985, opened 2020-10-09
Use tokens from lockup when needed

vgrichina:


kcole16:
@vgrichina Not sure of the best way to test this one

kcole16:
@vgrichina Seems to work, but `Available Balance` on `/send-money` is a bit wonky.

1. Deployed lockup on testnet (`testme123.near`) with `460 N` in the lockup using [these instructions](https://docs.google.com/document/d/1tPB-AUwQjwm45b21-HLgKGWhAU6081QNIdshJj1hbys/edit#) and changing the schema in Wallet locally
2. Went to `/send-money` and `Available Balance` displayed as `16 N` (on `/profile` the `Unlocked Balance` was `460 N`)
3. Sent `13 N` successfully, and went back to `/send-money`
4. `Available Balance` showed as `435 N`
5. Sent `200 N` successfully and went back to `/send-money`
6. Tried to send `100 N`, made it to "Confirm and Send", but received an error "Not enough tokens"
7. Sent `35 N` successfully

Seems like the balance isn't quite loading/reloading correctly. Will test further, and check with 2FA and Ledger as well

src/actions/account.js
```diff
@@ -164,16 +187,12 @@ export const { initializeRecoveryMethod, validateSecurityCode, initTwoFactor, re
...
-        wallet.twoFactor.sendRequest.bind(wallet.twoFactor),
-        () => defaultCodesFor('account.sendTwoFactor')
-    ],
     RESEND_TWO_FACTOR: [
         wallet.twoFactor.resend.bind(wallet.twoFactor),
         () => defaultCodesFor('account.resendTwoFactor')
     ],
     VERIFY_TWO_FACTOR: [
-        wallet.twoFactor.verifyTwoFactor.bind(wallet.twoFactor),
+        (...args) => wallet.twoFactor.verifyRequestCode(...args),
```
mattlockyer, line 195: why these changes?

mattlockyer: @vgrichina disregard this comment. AFAIK the default way to call helper instances based on wallet should be done this way:
```
(...args) => wallet.twoFactor.verifyRequestCode(...args)
```

I forgot the convention and led to lots of confusion re: 2fa refactor.

vgrichina: these are your changes, not mine I think :)

PR #986, opened 2020-10-09
Fixing balance to include locked tokens (hot fix)

mattlockyer:


src/utils/staking.js
```diff
@@ -13,6 +13,7 @@ const {
     }
 } =  nearApiJs
 
+const NEAR_PER_BYTE = new BN(process.env.REACT_APP_NEAR_PER_BYTE || '100000000000000000000')
```
vgrichina, line 16: Should this be available from RPC?  How is it calculated on Profile?

mattlockyer: Was following @evgenykuzyakov examples from near-sdk-rs

kcole16 (approved):

PR #987, opened 2020-10-09
Show proper Login page for Lockup accounts

marcinbodnar:


src/components/login/Login.js
```diff
@@ -55,7 +55,8 @@ class Login extends Component {
     }
 
     render() {
-        const { account: { url }, match } = this.props
+        const { account: { url, balance }, match } = this.props
+        const fullAccess = !url?.contract_id || !!parseFloat(balance?.lockedAmount)
```
vgrichina, line 59: Not sure I correctly understand the logic here. Is it basically "treat login with every contractId as full access if account has locked balance"? That doesn't seem what we need, we need to warn about access key _specific for lockup account id_, not just any account id.

I think the easiest check right now would be just to check for `ACCOUNT_ID_SUFFIX`.

PR #991, opened 2020-10-12
hot fix: twoFactorVerifyModal not returning success with empty return value from rpc

mattlockyer:
This fixes the issue with the 2FA modal erroring on close and subsequently the staking modal not closing as a result.

### DIFF
--- a/src/components/accounts/two_factor/TwoFactorVerifyModal.js
+++ b/src/components/accounts/two_factor/TwoFactorVerifyModal.js
@@ -42,11 +42,10 @@ const TwoFactorVerifyModal = ({ open, onClose }) => {
 
     const handleVerifyCode = async () => {
         const { success, res } = await dispatch(verifyTwoFactor(code))
-        if (success && typeof res !== 'boolean' && res !== false) {
-            onClose(true)
-        } else {
-            onClose(false, new WalletError('Transaction failed with return value: ' + res))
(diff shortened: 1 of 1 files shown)


src/components/accounts/two_factor/TwoFactorVerifyModal.js
```diff
@@ -42,11 +42,10 @@ const TwoFactorVerifyModal = ({ open, onClose }) => {
 
     const handleVerifyCode = async () => {
         const { success, res } = await dispatch(verifyTwoFactor(code))
-        if (success && typeof res !== 'boolean' && res !== false) {
-            onClose(true)
-        } else {
-            onClose(false, new WalletError('Transaction failed with return value: ' + res))
+        if (success === false || res === false) {
```
vgrichina, line 45: @mattlockyer why is result returned in such a weird way that 2 values needs to be checked? Why not throw error on failure from `verifyTwoFactor`?

src/components/accounts/two_factor/TwoFactorVerifyModal.js
```diff
@@ -42,11 +42,10 @@ const TwoFactorVerifyModal = ({ open, onClose }) => {
...
+            return onClose(false, new WalletError('Transaction failed with return value: ' + res))
```
vgrichina, line 46: missing message code

Issue #1006, opened 2020-10-13
Looks like not enough balance error is not handled properly when staking


vgrichina:
**Problem**

> the reason for my transaction failure is an actual scenario when the balance is not enough. So the transaction failure response is throwing it properly ("The balance that can be deposited to the staking pool is lower than the extra amount."). But its not throwing to the staking UI, which might need a fix, and user keeps trying to send the 2fa in loop without knowing the actual error. JS console error: "Error: Transaction failed with return value: undefined"

**Expected Behavior**

UI shows error and doesn't let user just do the same thing in a loop without understanding that it is failing.

**Steps to reproduce**

TODO: Not reproduced yet, this is user report

kcole16:
**Steps to reproduce**
1. Select a validator
2. Attempt to stake your entire available balance
3. Encounter error
4. Attempt to stake 1 NEAR less than available balance

PR #1007, opened 2020-10-13
Fix staking max

mattlockyer:
1. Cap the min displayed "Available lockup balance" at 0.001
2. Allow user to type in exactly what "Available Balance" says and still have staking succeed
3. Included 35N strict reserve vs. storage calculation

Fix for #1006


mattlockyer:
Have to bump min displayed "Available lockup balance" at 0.002

Just some after effects of staking max balance. After 1-2 epochs user will see some amount in "Available lockup balance" again.

PR #1014, opened 2020-10-14
Mainnet Wallet Hotfix: Fix seed phrase recovery for 2FA accounts

vgrichina:


src/utils/wallet.js
```diff
@@ -208,6 +208,7 @@ class Wallet {
             const state = await this.getAccount(this.accountId).state()
             this.twoFactor = new TwoFactor(this)
             const has2fa = this.has2fa = await this.twoFactor.isEnabled()
+            console.log('loadAccount has2fa', this.has2fa);
```
frol, line 211: Do you want to keep the logs?

frol (approved):

src/utils/wallet.js
vgrichina: yes, in case I need to debug further in prod

PR #1017, opened 2020-10-14
hotfix: fixes 2fa recovery links and also allows seed phrase to recover without 2fa

mattlockyer:
solution to https://github.com/near/near-wallet/pull/1014 while preserving 2FA link recovery which was broken.

### DIFF
--- a/src/utils/wallet.js
+++ b/src/utils/wallet.js
@@ -422,11 +422,13 @@ class Wallet {
     // TODO: Why is fullAccess needed? Everything without contractId should be full access.
     async addAccessKey(accountId, contractId, publicKey, fullAccess = false, methodNames) {
         const account = this.getAccount(accountId)
+        console.log('account instance used in recovery add localStorage key', account)
         // update has2fa now after we have the right Account instance for temp recovery
         this.has2fa = await this.twoFactor.isEnabled()
+        console.log('key being added to 2fa account (this.has2fa)', this.has2fa)
         try {
             if (fullAccess || (!this.has2fa && accountId === contractId)) {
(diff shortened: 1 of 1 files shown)


src/utils/wallet.js
```diff
@@ -740,12 +755,23 @@ class Wallet {
...
             if (!accountId || !accountId.length) return
 
-            // set up temp wallet instance
+            // temp account
             this.connection = connection
             this.accountId = accountId
             this.twoFactor = new TwoFactor(this)
             this.twoFactor.accountId = accountId
-            const account = this.getAccount(accountId)
+            const has2fa = this.has2fa = await this.twoFactor.isEnabled()
```
frol, line 763: I fail to understand how this can even work in the first place since you seem to end up modifying a single Wallet instance asyncronously and this should introduce races.

frol (approved):
I certainly fail to understand the whole picture, but this code looks innocent, so I approve to unblock merging if you understand what you are doing better than me :smile:

src/utils/wallet.js
mattlockyer: removed, unused

src/utils/wallet.js
```diff
@@ -208,7 +208,6 @@ class Wallet {
             const state = await this.getAccount(this.accountId).state()
             this.twoFactor = new TwoFactor(this)
             const has2fa = this.has2fa = await this.twoFactor.isEnabled()
-            console.log('loadAccount has2fa', this.has2fa);
```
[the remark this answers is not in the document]
mattlockyer: ok added more comments

Issue #1028, opened 2020-10-14
Show RPC error text in error notification


kcole16:
### Overview
We haven't sufficiently covered the potential errors with human readable ones. In the meantime, we should show the actual error message in the banner to better help us debug user issues.

### Acceptance Criteria
- [ ] When encountering an RPC-delivered error, the user sees both our title and error message, and the RPC-returned error message in the notification

marcinbodnar:
@corwinharrell It looks like this with #1029 changes
![image](https://user-images.githubusercontent.com/36210360/96154887-942f8380-0f0f-11eb-953a-ab4883e07141.png)

corwinharrell:
Here are the designs for the changes to the alert component:

![alert](https://user-images.githubusercontent.com/5934188/96176888-fcc03580-0efa-11eb-9e0c-a276aeaf5080.jpg)

And here are the svg icons @marcinbodnar (they should be constrained to 24px):

[alert-icons.zip](https://github.com/near/near-wallet/files/5387319/alert-icons.zip)

Issue #1019, opened 2020-10-14
Send money errors are wrong


ilblackdragon:
**Problem**
If there is an error with sending money, current error is "Please check that your device is connected and the NEAR app is open."

**Expected Behavior**
Show the actual error

**Steps to reproduce**

One example of errors can be lack of storage balance to cover adding another 2FA confirmation

ilblackdragon:
Duplicate of #1018

bowenwang1996:
why did you reopen it

corwinharrell:
@marcinbodnar can this be closed if it is a addressed in #1029?

kcole16:
Just linked the issue in the PR, so it will be closed once it's merged

marcinbodnar:
@kcole16 thank you

PR #1029, opened 2020-10-14
Global Alert - show the actual error message in the banner (#1019 #1028)

marcinbodnar:


marcinbodnar:
@kcole16 @ilblackdragon 
Wallet already has a functionality to show proper error in Global Alert banner - but the error from RPC need's to have a `type` property. For example, error, when there is a lack of storage balance to cover adding another 2FA confirmation, seems not to have `type`, so it's showing default error. 

With this change, all Global Alerts will also show the actual error message.

Ideally, all RPC errors should have a `type` property, and in Wallet, we should have all these types listed in translations files.

marcinbodnar:
@kcole16 @corwinharrell 
A new Global Alert is ready. I prepare it from scratch, so it's not using semantic ui anymore.

corwinharrell:
> it's not using semantic ui anymore

Praise god

corwinharrell:
@marcinbodnar this looks fantastic. Great job. 🚀

corwinharrell (approved):
@marcinbodnar this looks fantastic. Great job. 🚀

Patrick1904:
Love this! Big improvement. Just a couple of thoughts:
1. After being redirected back to profile post recovery setup, the global alert only shows for ~0.5 
sec before disappearing (might be an existing issue). Can we fix that too?
2. Can we vertically align the global alert icon, it's just slightly off (please see attached)
3. Is there a reason for introducing a new font? (cc @corwinharrell). I am working on replacing our current font with one global font (Inter) [here](https://github.com/near/near-wallet/pull/1011)

Screen Shot 2020-10-16 at 11 39 47 AM

corwinharrell:
@Patrick1904 The only new font is the one being used for the raw RPC output that I'd like to be monospace. Are we using a monospace typeface elsewhere? If so we can default to what we're already using.

marcinbodnar:
@Patrick1904 
> After being redirected back to profile post recovery setup, the global alert only shows for ~0.5
sec before disappearing (might be an existing issue). Can we fix that too?

What kind of recovery did you use?

> Can we vertically align the global alert icon, it's just slightly off (please see attached)

In this situation, when there is no text it looks off, but it looks good that way if there is a text. I wonder, what is the situation when only 'Success` header is displayed?

src/components/responsive/GlobalAlert.js
```diff
@@ -1,150 +1,177 @@
...
-                padding: 0px;
-                cursor: pointer;
-
-                &.white {
-                    filter: brightness(0) invert(1);
-                }
-            }
+const Content = styled.div`
+    min-height: 74px;
+    max-width: 500px;
```
corwinharrell, line 58: Let's go ahead and add `min-width: 275px;`

corwinharrell (changes requested):
Just thought it would be nice for alerts lacking content to not have to be so small, so suggested the addition of a `min-width`.

Patrick1904:
> @Patrick1904
> 
> > After being redirected back to profile post recovery setup, the global alert only shows for ~0.5
> > sec before disappearing (might be an existing issue). Can we fix that too?
> 
> What kind of recovery did you use?

Whenever I am logged in and start the recovery setup flow from the profile, and then get redirected back to the profile


> > Can we vertically align the global alert icon, it's just slightly off (please see attached)
> 
> In this situation, when there is no text it looks off, but it looks good that way if there is a text. I wonder, what is the situation when only 'Success` header is displayed?

I believe this was when I disabled seed phrase from the profile. Might want to align the icon vertically with the main heading/title (cc @corwinharrell)

marcinbodnar:
@Patrick1904 @corwinharrell 
Bug with disappearing alert after setting up recovery was not related to these changes. I fixed it for seed phrase, phone, and email.

Also adjusted Alert with only header visible.

corwinharrell (approved):

Patrick1904 (approved):

PR #1030, opened 2020-10-15
Include small amount for gas in min. balance

Patrick1904:


Patrick1904:
@vgrichina I added 2N as reserved for gas to the Minimum Balance, and subtracted 2N from Available Balance for non-lockup account. Can you please help me do the same for lockup account or let me know how to do it?

src/utils/wallet.js
```diff
@@ -600,9 +601,11 @@ class Wallet {
 
     async getBalance(accountId) {
         accountId = accountId || this.accountId
-
         const account = this.getAccount(accountId)
-        return await account.getAccountBalance()
+        let balance = await account.getAccountBalance()
+        balance.stateStaked = new BN(balance.stateStaked).add(new BN(MIN_BALANCE_FOR_GAS)).toString()
+        balance.available = new BN(balance.available).sub(new BN(MIN_BALANCE_FOR_GAS)).toString()
```
vgrichina, line 607: @Patrick1904 looks good to me, but need to make sure that doesn't add twice with `account-with-lockup`  (when account has lockup).

Patrick1904: @vgrichina I just tested on staging and it works as intended for `account-with-lockup` too. Can you please test and confirm with a mainnet account too? Thank you

src/utils/wallet.js
```diff
@@ -29,6 +29,7 @@ export const DISABLE_SEND_MONEY = process.env.DISABLE_SEND_MONEY === 'true' || p
 export const ACCOUNT_ID_SUFFIX = process.env.REACT_APP_ACCOUNT_ID_SUFFIX || 'testnet'
 export const MULTISIG_MIN_AMOUNT = process.env.REACT_APP_MULTISIG_MIN_AMOUNT || '40'
 export const LOCKUP_ACCOUNT_ID_SUFFIX = process.env.LOCKUP_ACCOUNT_ID_SUFFIX || 'lockup.near'
+export const MIN_BALANCE_FOR_GAS = process.env.REACT_APP_MIN_BALANCE_FOR_GAS || nearApiJs.utils.format.parseNearAmount('2')
```
vgrichina, line 32: @Patrick1904 nit: I think we don't have to use `REACT_APP_` prefix given project not using `react-scripts`. Maybe we should just use `NEAR_WALLET_` as prefix?

PR #1034, opened 2020-10-15
Fix 2FA 'Active since'

Patrick1904:
Fixes:
1. 2FA date 
2. 2FA not showing as active after setup (page refresh was required)


src/components/profile/Profile.js
```diff
@@ -8,24 +8,29 @@ import RecoveryContainer from './Recovery/RecoveryContainer'
...
-            dispatch(getAccessKeys(accountId))
-            dispatch(getLedgerKey())
-            dispatch(get2faMethod())
-            dispatch(checkCanEnableTwoFactor(account))
+            getInitialData()
         }
     }, []);
 
+    const getInitialData = async () => {
+        await dispatch(getAccessKeys(accountId))
```
vgrichina, line 29: @Patrick1904 is it intended that we are waiting for each of these to finish before making another request?

PR #1036, opened 2020-10-15
check if account was created after rpc fails during createNewAccount

mattlockyer:


src/actions/account.js
```diff
@@ -325,7 +325,16 @@ export const { addAccessKey, createAccountWithSeedPhrase, addAccessKeySeedPhrase
     CREATE_ACCOUNT_WITH_SEED_PHRASE: [
         async (accountId, recoveryKeyPair, fundingContract, fundingKey) => {
             await wallet.saveAccount(accountId, recoveryKeyPair);
-            await wallet.createNewAccount(accountId, fundingContract, fundingKey, recoveryKeyPair.publicKey)
+
+            try {
+                await wallet.createNewAccount(accountId, fundingContract, fundingKey, recoveryKeyPair.publicKey)
+            } catch (e) {
+                console.warn('RPC error: createNewAccount\n', e)
+                await new Promise((r) => setTimeout(r, 4000)) // wait 4s for chain
```
vgrichina, line 354: If account creation failed with RetriesExceeded – it generally shouldn't help to wait 4 more seconds (as `RetriesExceeded` should take far longer that that). If default timeout values in near-api-js are too small – we can just increase them.

src/actions/account.js
```diff
@@ -325,7 +325,16 @@ export const { addAccessKey, createAccountWithSeedPhrase, addAccessKeySeedPhrase
...
+                await wallet.getAccount(accountId).state()
+            }
+
+            // if not thrown getting new account state, account exists
```
vgrichina, line 358: This means very different error will be thrown than actual problem with account creation, which isn't optimal.

I think ideally failure during account creation should go 2 directions depending on error:
- known fatal error (e.g. account with such name already exists) – allow user to change params and retry creation
- known non necessarily fatal error (e.g. retries) – keep user on the same page and just allow to hit Retry. Retry first checks if account already exists with correct keys. Also make sure to save the account info regardless, so that if user refreshes page and account was created it gets loaded ok.

andy-haynes:
@marcinbodnar do you know if this is still an issue and whether the changes here still relevant?

If it's not relevant or worth salvaging then I'd vote to close it out, this is a pretty old PR at this point.

marcinbodnar:
@andy-haynes Agree, closing it.

PR #1037, opened 2020-10-15
hotfix: fix error thrown when 2fa tries to check account state and no…

mattlockyer:
… account exists


mattlockyer:
The error from RPC is super cryptic... it basically says "Account ' ' does not exist" this doesn't help.

RPC throwing error on read of account state for account that doesn't exist...

Patrick1904:
Tested this on testnet and it seems to fix the issue

Issue #1042, opened 2020-10-15
Error unselecting staking pool during staking pool withdrawal


kcole16:
**Problem**
Calling `unselect_staking_pool` will fail during withdrawal if the user staked again after triggering Unstake. This will be a more common issue when we launch partial unstaking.

**Expected Behavior**
We should not trigger `unselect_staking_pool` during Withdrawal, only `withdraw_all_from_staking_pool`.

**Steps to reproduce**

1. Stake a small amount (not your entire lockup balance)
2. Unstake 
3. Stake again 
4. When your first stake is ready to withdraw, initiate the withdrawal
5. Attempt to approve `unselect_staking_pool`

**Details**

Failed TX: https://explorer.near.org/transactions/7ujCVsCHSBWfzRSkhhG79N9uj2MNLCCeh4ny7SDR4S8F

Error text: 
```panicked at 'assertion failed: `(left == right)`\n  left: `499999980183345121319654`,\n right: `0`: There is still a deposit on the staking pool', src/owner.rs:56:9```

mattlockyer:
@kcole16 the issue seems to be specifically with calling "unselect" while there is still something staked? Despite calling "withdrawal all" and successfully withdrawing all of the available balance, correct?

I will add a check to lockup withdraw function that will not unselect pool until all balances are "0" with validator.

mattlockyer:
@kcole16 try this on:
https://github.com/near/near-wallet/pull/1043

Might have to wait for betanet...

vgrichina:
@mattlockyer @kcole16 why do we ever need to call `unselect_staking_pool` at all except when selecting new pool?

kcole16:
I don't think we should call it, except when selecting a new pool

mattlockyer:
This was the functionality. @Patrick1904 requested the unselect on withdrawal. Do you still need it?

Patrick1904:
I'm using it right now to determine if you currently have a balance with a validator. Seems unusual to me that we'd keep the validator selected even if I don't have any balance with the validator, but can adjust if needed.

PR #1043, opened 2020-10-15
pull state and check before unselect pool

mattlockyer:
fix for #1042


src/utils/staking.js
```diff
@@ -52,6 +52,8 @@ export class Staking {
     constructor(wallet) {
         this.wallet = wallet
         this.provider = wallet.connection.provider
+
+        console.log(this)
```
vgrichina, line 56: @mattlockyer debug leftover?

PR #1051, opened 2020-10-16
Remove 'create account' link from tool tip

Patrick1904:
I will add it back once we enable simple account creation.


vgrichina:
@Patrick1904 do I understand correctly that we don't need to merge this PR anymore?

Patrick1904:
> @Patrick1904 do I understand correctly that we don't need to merge this PR anymore?

@vgrichina I changed so that it points to implicit account creation on mainnet so we can merge this.

PR #1052, opened 2020-10-16
Add /create-implicit route and page to create implicit accounts with seed phrase

mattlockyer:
PR creates a new route `/create-implicit` for creating implicit accounts
1. copy seed phrase
2. verify
3. copy accountId
4. fund from somewhere
5. click check balance and redirected to `/recover-with-link/:accountId/:seedPhrase`

Some minor changes to other components were necessary to make UI work
- SeedPhraseForm
- Navigation

TODO
Uses localStorage to store seedPhrase and accountId, which should be cleared after user is recovered into wallet successfully.

### DIFF
--- a/src/components/accounts/SetupImplicit.js
+++ b/src/components/accounts/SetupImplicit.js
@@ -1,10 +1,7 @@
 import React, { Component, Fragment } from 'react'
 
 import * as nearApiJs from 'near-api-js' 
-import { parseSeedPhrase } from 'near-seed-phrase'
 import BN from 'bn.js'
-const NETWORK_ID = process.env.REACT_APP_NETWORK_ID || 'default'
-const NODE_URL = process.env.REACT_APP_NODE_URL || 'https://rpc.nearprotocol.com'
 
 import { withRouter, Route } from 'react-router-dom'
 import { connect } from 'react-redux'
@@ -21,6 +18,8 @@ import Container from '../common/styled/Container.css'
 import { KeyPair } from 'near-api-js'
 import FormButton from '../common/FormButton'
 import IconMCopy from '../../images/IconMCopy'
+const NETWORK_ID = process.env.REACT_APP_NETWORK_ID || 'default'
+const NODE_URL = process.env.REACT_APP_NODE_URL || 'https://rpc.nearprotocol.com'
 
 const IMPLICIT_ACCOUNT_KEY = '__IMPLICIT_ACCOUNT_KEY'
 
@@ -58,7 +57,8 @@ class SetupImplicit extends Component {
     componentDidUpdate = (props, state) => {
         const recoveryKeyPair = JSON.parse(localStorage.getItem(IMPLICIT_ACCOUNT_KEY) || '{}')
         if (this.props.location.pathname !== '/create-implicit/fund' && recoveryKeyPair.publicKey) {
-            if (confirm('WARNING! If you have funded your account, press "Cancel".\n\nIf you would like to start over with a new account, press "OK".')) {
+            const confirm = confirm('WARNING! If you have funded your account, press "Cancel".\n\nIf you would like to start over with a new account, press "OK".')
+            if (confirm) {
                 localStorage.removeItem(IMPLICIT_ACCOUNT_KEY)
                 this.refreshData()
                 this.props.history.push('/create-implicit')
--- a/src/components/accounts/SetupImplicit.js
+++ b/src/components/accounts/SetupImplicit.js
@@ -278,7 +278,7 @@ class SetupImplicit extends Component {
                                         !!balance &&
                                         <>
                                             <p style={{marginTop: 16}}>
-                                                Once funded, click "Continue". You will be redirected to a page titled "Restore Account". Click "Continue" again.
+                                                Click "Continue". You will be redirected to a page titled "Restore Account". Click "Continue" again.
                                             </p>
                                             <FormButton
                                                 onClick={() => this.handleContinue()}
--- a/src/components/accounts/SetupImplicit.js
+++ b/src/components/accounts/SetupImplicit.js
@@ -1,0 +1,289 @@
+import React, { Component, Fragment } from 'react'
+
+import * as nearApiJs from 'near-api-js' 
+import { parseSeedPhrase } from 'near-seed-phrase'
+import BN from 'bn.js'
+const NETWORK_ID = process.env.REACT_APP_NETWORK_ID || 'default'
+const NODE_URL = process.env.REACT_APP_NODE_URL || 'https://rpc.nearprotocol.com'
+
+import { withRouter, Route } from 'react-router-dom'
+import { connect } from 'react-redux'
+import { Translate } from 'react-localize-redux'
+
+import { redirectToApp, handleAddAccessKeySeedPhrase, clearAlert, refreshAccount, checkCanEnableTwoFactor, checkIsNew, handleCreateAccountWithSeedPhrase } from '../../actions/account'
+import { generateSeedPhrase } from 'near-seed-phrase'
+import SetupSeedPhraseVerify from './SetupSeedPhraseVerify'
+import SetupSeedPhraseForm from './SetupSeedPhraseForm'
+import copyText from '../../utils/copyText'
+import isMobile from '../../utils/isMobile'
+import { Snackbar, snackbarDuration } from '../common/Snackbar'
+import Container from '../common/styled/Container.css'
+import { KeyPair } from 'near-api-js'
+import FormButton from '../common/FormButton'
+import IconMCopy from '../../images/IconMCopy'
+
+const IMPLICIT_ACCOUNT_KEY = '__IMPLICIT_ACCOUNT_KEY'
+
+class SetupImplicit extends Component {
+    state = {
+        seedPhrase: '',
+        enterWord: '',
+        wordId: null,
+        requestStatus: null,
+        successSnackbar: false,
+        sending: false,
+        snackBarMessage: 'setupSeedPhrase.snackbarCopySuccess'
+    }
+
+    async checkBalance() {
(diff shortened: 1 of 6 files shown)


src/utils/staking.js
```diff
@@ -218,7 +218,13 @@ export class Staking {
         } else {
             lockupId = getLockupAccountId(accountId)
         }
-        await (await new nearApiJs.Account(this.wallet.connection, lockupId)).state()
+        try {
```
vgrichina, line 221: there is `wallet.accountExists` helper BTW, we probably just need to extract it into helper function, this pattern now is used in way too many places :)

src/utils/staking.js
```diff
@@ -228,7 +234,8 @@ export class Staking {
             await (await new nearApiJs.Account(this.wallet.connection, contractId)).state()
             return await new nearApiJs.Contract(this.wallet.getAccount(), contractId, { ...methods })
         } catch(e) {
-            throw new WalletError('No lockup contract for account', 'staking.errors.noLockup')
+            console.warn('No lockup contract for account')
+            // throw new WalletError('No lockup contract for account', 'staking.errors.noLockup')
```
vgrichina, line 238: is this change desired? Best not to commit commented out code without explanation

src/components/navigation/Navigation.js
```diff
@@ -58,23 +58,27 @@ class Navigation extends Component {
     }
 
     render() {
-
         const { menuOpen } = this.state;
 
+        let showNavLinks = this.showNavLinks
```
vgrichina, line 63: seems like this logic should be in `showNavLinks` getter? Also why do we need different logic specifically for `/create-implicit`?

src/components/accounts/SetupSeedPhraseForm.js
```diff
@@ -67,7 +68,7 @@ const SetupSeedPhraseForm = ({
                     <IconMCopy color='#6ad1e3' />
                 </FormButton>
                 <FormButton
-                    linkTo={`/setup-seed-phrase/${accountId}/verify/${fundingContract ? `${fundingContract}/${fundingKey}/` : ``}`}
+                    linkTo={linkTo ? linkTo : `/setup-seed-phrase/${accountId}/verify/${fundingContract ? `${fundingContract}/${fundingKey}/` : ``}`}
```
vgrichina, line 71: nit: should we always pass `linkTo` as parameter (maybe as `continueLinkTo`)? Looks like this component itself doesn't really use `accountId` / `fundingContract` / `fundingKey` otherwise.

src/components/accounts/SetupImplicit.js
```diff
@@ -0,0 +1,289 @@
...
+        const recoveryKeyPair = JSON.parse(localStorage.getItem(IMPLICIT_ACCOUNT_KEY) || '{}')
+        this.setState({ sending: true })
+        const account = new nearApiJs.Account(this.connection, recoveryKeyPair.implicitAccountId)
+        try {
+            const state = await account.state()
+            if (new BN(state.amount).gte(new BN('1000000000000000000000000', 10))) {
+                window.removeEventListener('beforeunload')
+                window.location.href = `${window.location.origin}/recover-with-link/${recoveryKeyPair.implicitAccountId}/${recoveryKeyPair.seedPhrase}`
+            }
+        } catch (e) {
```
vgrichina, line 47: see another comment – I think we really should just make it easier to check whether account exists (have a helper function)

src/components/accounts/SetupImplicit.js
```diff
@@ -0,0 +1,289 @@
...
+
+    async checkBalance() {
+        const recoveryKeyPair = JSON.parse(localStorage.getItem(IMPLICIT_ACCOUNT_KEY) || '{}')
+        this.setState({ sending: true })
+        const account = new nearApiJs.Account(this.connection, recoveryKeyPair.implicitAccountId)
+        try {
+            const state = await account.state()
+            if (new BN(state.amount).gte(new BN('1000000000000000000000000', 10))) {
+                window.removeEventListener('beforeunload')
+                window.location.href = `${window.location.origin}/recover-with-link/${recoveryKeyPair.implicitAccountId}/${recoveryKeyPair.seedPhrase}`
```
vgrichina, line 45: is it intentional to replace URL without pushing to router?

src/components/accounts/SetupImplicit.js
```diff
@@ -0,0 +1,289 @@
...
+        enterWord: '',
+        wordId: null,
+        requestStatus: null,
+        successSnackbar: false,
+        sending: false,
+        snackBarMessage: 'setupSeedPhrase.snackbarCopySuccess'
+    }
+
+    async checkBalance() {
+        const recoveryKeyPair = JSON.parse(localStorage.getItem(IMPLICIT_ACCOUNT_KEY) || '{}')
```
vgrichina, line 38: why do we have special storage for implicit account key? Is there any reason not to put it into common key store?

src/components/accounts/SetupImplicit.js
```diff
@@ -0,0 +1,289 @@
...
+                requestStatus: {
+                    success: false,
+                    messageCode: 'account.verifySeedPhrase.error'
+                }
+            }))
+            return false
+        }
+        console.log('verified', recoveryKeyPair)
+        recoveryKeyPair.seedPhrase = seedPhrase
+        recoveryKeyPair.implicitAccountId = Buffer.from(recoveryKeyPair.publicKey.data).toString('hex')
```
vgrichina, line 146: best not to save copies of values which can be derived from master value (or you risk weird bugs when they get out of sync). Like in this case saving private key is enough. Both public key and implicit account ID can be derived from private key. 

same with seed phrase – if you are saving seed phrase, saving anything else makes no sense – it can be computed from seed phrase.

src/components/accounts/SetupImplicit.js
```diff
@@ -0,0 +1,289 @@
...
+            }).catch(err => {
+                console.log(err.message);
+            });
+        } else {
+            this.handleCopyDesktop(textToCopy);
+        }
+    }
+
+    handleCopyDesktop = (textToCopy) => {
+        copyText(textToCopy ? document.getElementById('implicit-account-id') : document.getElementById('seed-phrase'));
```
vgrichina, line 167: not sure I understand the logic with `textToCopy`. May you explain?

src/components/accounts/SetupImplicit.js
```diff
@@ -0,0 +1,289 @@
...
+                throw(e)
+            }
+        }
+        setTimeout(() => this.setState({ sending: false }), 1000)
+    }
+
+    componentDidUpdate = (props, state) => {
+        const recoveryKeyPair = JSON.parse(localStorage.getItem(IMPLICIT_ACCOUNT_KEY) || '{}')
+        if (this.props.location.pathname !== '/create-implicit/fund' && recoveryKeyPair.publicKey) {
+            const confirm = confirm('WARNING! If you have funded your account, press "Cancel".\n\nIf you would like to start over with a new account, press "OK".')
```
vgrichina, line 60: Shouldn't we check ourselves whether user has funded account indeed?

src/utils/staking.js
mattlockyer: removed change

src/utils/staking.js
mattlockyer: not sure I want to handle this as part of this PR

src/components/navigation/Navigation.js
mattlockyer: if the user already has an account we don't want to show it, unless we do?

src/components/accounts/SetupSeedPhraseForm.js
mattlockyer: Yes, for the implicit account some components are overkill, but this was the easiest working model of seed phrase.

vgrichina: @mattlockyer not sure what do you mean. I'm suggesting to have this component always take `linkTo` option (renamed as `continueLinkTo`). Then this whole thing can be always passed from outside:

```
`/setup-seed-phrase/${accountId}/verify/${fundingContract ? `${fundingContract}/${fundingKey}/` : ``
```

Feels kinda weird to pass it from outside sometimes and not pass in other times.

src/components/accounts/SetupImplicit.js
mattlockyer: the functions were purpose built to copy seed phrase, I'm just trying to generalize so I can copy seed phrase as well.

src/components/accounts/SetupImplicit.js
```diff
@@ -0,0 +1,289 @@
...
+            const state = await account.state()
+            if (new BN(state.amount).gte(new BN('1000000000000000000000000', 10))) {
+                window.removeEventListener('beforeunload')
+                window.location.href = `${window.location.origin}/recover-with-link/${recoveryKeyPair.implicitAccountId}/${recoveryKeyPair.seedPhrase}`
+            }
+        } catch (e) {
+            if (e.message.indexOf('exist while viewing') > -1) {
+                console.warn('account does not exist yet')
+            } else {
+                throw(e)
```
vgrichina, line 51: nit: just `throw e` is enough

src/components/accounts/SetupSeedPhraseForm.js
mattlockyer: It's used by a completely different flow, setupSeedPhrase.

I'm just borrowing the component, I'm not fixing these other components as part of this PR.

src/components/accounts/SetupImplicit.js
mattlockyer: added comment

src/components/accounts/SetupImplicit.js
mattlockyer: Not sure what we should do if they want to start over though?

kcole16 (approved):

kcole16 (approved):

mattlockyer:
ezgif-4-e2b32aba4b6a

Issue #1056, opened 2020-10-19
User can create, fund and sign in to implicit account


mattlockyer:
## Story

As a USER who wants to get off exchange / other wallet and into NEAR Wallet, I want to create an account in the NEAR wallet, fund it and then transfer my funds to that account, so I can move my assets into the NEAR Wallet.

## Acceptance Criteria

- [ ] User can create implicit account via seed phrase
- [ ] Seed phrase is verified
- [ ] User sees funding accountId and can fund account
- [ ] Once funded user can continue with recovery link to sign in

## Unhappy Path

User creates account, funds it, clears browser cache and loses seed phrase.
User clears browser cache and loses option to continue, they need to manually enter seed phrase in recovery options of Wallet.

## Testing

https://wallet.testnet.near.org/create-implicit
Then fund it from another account on testnet (You just need 1N).
Then click continue and recover into account.

## Assets

- N/A

## NOTES

- Sprinted last friday: https://github.com/near/near-wallet/pull/1052#issuecomment-712191127

corwinharrell:
Alright so I walked through this in its current state to provide some feedback through the UX lens. Here is my [video recording](https://streamable.com/z5qyiy) with moment-to-moment commentary if you'd rather watch than read (or both).


## UX Audit
1. `/create-implicit`
    - Starting at seed phrase setup is jarring, there is no introduction to the user to indicate that they are creating a new account.
    - What if they want to use a method other than seed phrase?

2. `/create-implicit/fund` 
    - Instructing the user to perform these actions is odd
    - It assumes that the user has another account which they can use to fund the new account.
        - What if they don't have another account?
    - User: _"Why do I have to send 1 NEAR?"_
    - User: _"Why is it such a small amount?"_
    - No way to guarantee that the user would return to this window/tab without closing it first.

3. `/send-money`
    - What's the likelihood that the user would have recorded their implicit account ID?
    - UI: `Always double-check your accountID`
        - No reason to put this in the UI. We can't just give disclaimers for every important action in our UI. That doesn't solve anything and doesn't **_truly_** communicate the level of severity.
    - When the user inputs an implicit account ID, the "valid" state of the input just says `Always double check the account ID`. It doesn't let the user know that they've entered a valid Account ID.

4. Account Dashboard
    - User: _"OK so I've now funded the new account, but I'm on the dashboard of a different account. How do I get to the account I've created?"_
[...]

PR #1059, opened 2020-10-19
fix continue screen

mattlockyer:


mattlockyer:
Implements: https://github.com/near/near-wallet/issues/1056#issuecomment-712358673

kcole16 (approved):

PR #1066, opened 2020-10-21
Allow to fund account creation from another account

vgrichina:
This is a bit work in progress (you need to know URL http://wallet.testnet.near.org/create-from/your-account.testnet), but generally should be releasable.


kcole16:
One strange quirk (though somewhat expected) when creating an account with seed phrase and recovering:
- Two accounts are recovered, since they both use the same seed phrase (implicit and user-defined accounts)

Not sure we can easily get around this, but it's a bit confusing for the user to see "2 accounts were successfully recovered" if they expect to have just one. Same happens with Ledger.

vgrichina:
@kcole16 I think the fix is to use up whole balance and remove implicit account, going to add it

PR #1071, opened 2020-10-21
hotfix implicit accounts LAK/FAK and dup accountIds

mattlockyer:


kcole16:
@mattlockyer Testing this locally, I still get the 2FA prompt when recovering with seed (for implicit account)

mattlockyer:
@kcole16 you will once, can't fix that unless we push to near-api-js

kcole16 (approved):

vgrichina (changes requested):
Merged smaller hotfix, so let's figure out what are most desired long term changes

PR #1075, opened 2020-10-21
Revert "Fix implicit 2"

vgrichina:
Reverts near/near-wallet#1070

We should merge smaller fix for now instead:
https://github.com/near/near-wallet/pull/1074


mattlockyer:
I'm not sure there's a different fix for this. There are still users with LAK seed phrases from enabling 2fa on implicit accounts that are not being rotated back to FAKs.

mattlockyer (approved):

vgrichina:
> There are still users with LAK seed phrases from enabling 2fa on implicit accounts that are not being rotated back to FAKs.

I don't think this is a big issue. As I understand they can still require anyway with 2FA approval, add different recovery method, etc. Correct?

PR #1076, opened 2020-10-22
Turn "Receive" arrow icon into a "down arrow" instead of a "left arrow"

evgenyboxer:
Hey team!

I started using the wallet and noticed that the "receive" icon points "outside", and this is kinda counter-intuitive. I suggest rotating it by `90deg` instead of `180deg` (so the arrow would point "down").

Before:
Screen Shot 2020-10-22 at 3 14 10 pm

After:
Screen Shot 2020-10-22 at 3 14 37 pm

Let me know what you think!

### DIFF
--- a/src/components/navigation/NavLinks.js
+++ b/src/components/navigation/NavLinks.js
@@ -41,7 +41,13 @@ const NavLink = styled(Link)`
         background-size: contain;
     }
 
-    &.rotate {
+    &.rotate-up {
+        &:before {
+            transform: rotate(-90deg);
+        }
+    }
+
+    &.rotate-down {
         &:before {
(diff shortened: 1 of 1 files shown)


corwinharrell:
Thanks for sharing @evgenyboxer! I can understand why this might communicate better. I have no issue with the change, but I might recommend that we do the same to "Send" for the sake of consistency (so "Send" would point up). @jakestutzman any opinions?

evgenyboxer:
@corwinharrell 

Here's how it looks with the arrow up: (have pushed the changes in the PR)

Screen Shot 2020-10-23 at 1 05 45 pm

PR #1077, opened 2020-10-22
Add Storybook for isolated component development

corwinharrell:
I'd like to add [Storybook](https://storybook.js.org/docs/react/get-started/introduction) to our Wallet repo. This enables and benefits us in the following ways:

1. **Isolated component development:** Storybook provides an environment where individual components can be isolated and refined without having to make changes to components in the context of a view. This allows for focused development and for us to really start improving the quality, versatility, and accessibility of our components. This environment also allows me, a "designer who codes", to more easily contribute to the design of our UI. I plan to spend more time in code and less time in Figma in the coming months as we continue to improve our UI.

2. **Testing for Components:** As we scale, it is important that our UI meets certain standards in accessibility. Storybook offers [tooling for testing the accessibility](https://github.com/storybookjs/storybook/tree/master/addons/a11y) or our components, flagging things like contrast issues or missing states, so that we can keep inline with [web accessibility standards](https://www.w3.org/WAI/standards-guidelines/wcag/).

3. **Design System documentation:** There's a lot of overhead in maintaining a design system. With little configuration, Storybook helps to generate automatic documentation of your components which can be customized and added to over time. Component state and props are clearly displayed and are made interactive, making it easy for anyone using our components to ensure they understand what the components enable, and how best to use them. I would like to own the creation of our design system documentation, and will be adding to and improving it over time, with the goal of eventually packaging our component library so that developers across our team and other products may also benefit from a well constructed and documented component library.

Storybook

4. **Add as you go:** Starting with a single component is all it requires to get going with Storybook. In our case, I've started with `Button`. Each component will be added individually, and can be accompanied with pieces of documentation as we have time.

**USING STORYBOOK IS ENTIRELY OPTIONAL.** It stays out of the way, and doesn't require any changes, additional configuration or rearchitecting of existing code (With one exception detailed below). This means that it is there for you if you'd like to use it, but if not, project devs can simply act like it doesn't exist. The caveat to this, which is actually a benefit, is that Storybook is most useful when using `PropTypes` or TypeScript, where the properties of components are typed. In doing this, Storybook generates documentation automatically based on these `PropTypes`. We should be doing this anyways, and IMO this is a good motivation to start type-ing our current component properties. @Patrick1904 to start, I've just added a few `PropTypes` to our current `Button` component.


vgrichina:
@corwinharrell do we have to install it into project vs run vs `npx`?

corwinharrell:
@vgrichina I believe so since it installs a few dependencies and sets up a default configuration based on your project. You then need to write the "story" files which import your project's components, and act as your component documentation. Not sure if that answers your question. 

I'd take a look at their docs to see if there are any alternative config options that you might favor, I've just always used it by initiating it in the root directory, which is what they recommend.

corwinharrell:
@vgrichina I know this is low priority, but any reason we can't merge this this week? Would love to start in on some component work.

Patrick1904 (approved):

Patrick1904:
@corwinharrell this looks good to me. Please just fix failing tests.

PR #1079, opened 2020-10-22
Deduct 2N from available staking balance to align with profile page

Patrick1904:
For non-lockup accounts, we were only deducting 1N from available balance to stake. This deducts 2N from available balance, so that users see the same 'Available Balance' on the staking dashboard and /profile

### DIFF
--- a/src/utils/staking.js
+++ b/src/utils/staking.js
@@ -3,6 +3,7 @@ import BN from 'bn.js'
 import { WalletError } from './walletError'
 import { getLockupAccountId } from './account-with-lockup'
 import { queryExplorer } from './explorer-api'
+import { MIN_BALANCE_FOR_GAS } from './wallet'
 
 const {
     transactions: {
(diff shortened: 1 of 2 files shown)


src/utils/staking.js
```diff
@@ -203,7 +203,7 @@ export class Staking {
             validators = await this.getValidators()
         }
 
-        let totalUnstaked = new BN(balance.available, 10).sub(new BN(parseNearAmount('1'), 10))
+        let totalUnstaked = new BN(balance.available, 10).sub(new BN(parseNearAmount('2'), 10))
```
vgrichina, line 206: @Patrick1904 let's use `MIN_BALANCE_FOR_GAS` from `wallet.js`

vgrichina: nit: note that you generally don't need to specify `10` as a base for `new BN()`

PR #1080, opened 2020-10-23
Lockup staking: fix stakeAllowed flag

Patrick1904:
Since we do not call `unselect_staking_pool` until user stakes with a new validator, we need to check if the user is staking. The `currentValidators` array should be empty if there's no balance.


kcole16 (approved):

mattlockyer:
tested with fresh lockup

was able to select validator, browse all validators without warnings
then unselect
then select another validator and stake

mattlockyer (approved):

PR #1106, opened 2020-10-27
Staging disclaimer banner + modal

Patrick1904:


src/components/common/NetworkBanner.js
```diff
@@ -41,16 +43,36 @@ const Container = styled.div`
...
+    const setBannerHeight = () => {
+        const bannerHeight = document.getElementById('top-banner') && document.getElementById('top-banner').offsetHeight
+        const app = document.getElementById('app-container')
+        const navContainer = document.getElementById('nav-container')
+        navContainer.style.top = bannerHeight ? `${bannerHeight}px` : 0
+        app.style.paddingTop = bannerHeight ? `${bannerHeight + 85}px` : '75px'
+    }
+
+    const isMainnetGuestFlow = IS_MAINNET && !account.accountId && !window.location.href.includes('staging')
+    const isMainnetStaging = IS_MAINNET && window.location.href.includes('staging')
```
Patrick1904, line 71: @vgrichina do you know if there's a better way to determine if it's staging? Not sure if it would make sense to change the `IS_MAINNET` variable to `CURRENT_NETWORK` or similar.

vgrichina: @Patrick1904 we always can just set specific env variable in given environment. E.g. `SHOW_BETA_WARNING` or something like this.

vgrichina: @Patrick1904 you can specify it for staging here https://github.com/near/near-wallet/blob/2529bab9ca6a95efd01e8d4a0a0a00c5be6024be/netlify.toml#L5

vgrichina (approved):
Looks good overall, but let's change env variable used to control this (see inline comment) @Patrick1904

PR #1107, opened 2020-10-27
twoFactor enable/disable

mattlockyer:
Pre-requisite: https://github.com/near/near-api-js/pull/445


vgrichina:
> To set up and use dependency branch of near-api-js: near-api-js/ms-disable

I think best to reder to specific branch  in `package.json` in such case (so that all deploys can still be tested easily), but then remove the change after near-api-js release

mattlockyer:
> > To set up and use dependency branch of near-api-js: near-api-js/ms-disable
> 
> I think best to reder to specific branch in `package.json` in such case (so that all deploys can still be tested easily), but then remove the change after near-api-js release

https://github.com/near/near-wallet/blob/2523dd3d157f03293c8f393e93d579b6c32ebbc3/package.json#L13

kcole16 (approved):

src/utils/wallet.js
```diff
@@ -819,7 +818,7 @@ class Wallet {
             const keyPair = KeyPair.fromString(secretKey)
             await tempKeyStore.setKey(NETWORK_ID, accountId, keyPair)
             account.keyStore = tempKeyStore
-            account.inMemorySigner = new nearApiJs.InMemorySigner(tempKeyStore)
+            account.inMemorySigner = account.connection.signer = new nearApiJs.InMemorySigner(tempKeyStore)
```
vgrichina, line 821: `account.connection.signer = `

why is this needed? This doesn't look good to me. Why not create temporary `Account` instance  with necessary config instead of patching existing one?

mattlockyer: was always confused by this:
https://github.com/near/near-wallet/blob/2fea47b429f844af95a4b21c60a0e081369575e1/src/utils/wallet.js#L629

vgrichina: looks like it should have been `this.inMemorySigner` (field in `Wallet` object)

vgrichina:
@mattlockyer when I try to disable 2FA this crashes here:

```
    async disableMultisig() {
        const contractBytes = new Uint8Array(await (await fetch('/main.wasm')).arrayBuffer())
        const result = await this.disable(contractBytes)   // <<<<<<<<<<<<<<<<<<<<
        await store.dispatch(refreshAccount())
        return result
    }
```

(there is no `.disable` method)

mattlockyer:
> (there is no `.disable` method)

Can you check which version of near-api-js you're running? Are you linked to one locally?

I just disabled/enabled using https://near-wallet-pr-1107.onrender.com/ worked fine. Cache?

vgrichina (changes requested):
has dangerous bugs blocking release

vgrichina:
```
[I] ➜ NODE_ENV=testnet near keys berries.testnet
Keys for account berries.testnet
[
  {
    public_key: 'ed25519:5z2QWPyCmw69ic5vgPHLGoGCpHzMJ8PLS2NadEvKnWzV',
    access_key: { nonce: 0, permission: 'FullAccess' }
  },
  {
    public_key: 'ed25519:7Ywn6QxxYLhbSRTtiJupssWZ42bbxQvQ7Pa9YVWpnFzm',
    access_key: { nonce: 0, permission: 'FullAccess' }
  },
  {
    public_key: 'ed25519:GLwGYLqFq2wEnakLj3gwoH2jGwKqKARBuCYtqbZcfVLs',
    access_key: { nonce: 1, permission: 'FullAccess' }
  },
  [length]: 3
]
```

looks like after disabling 2FA my limited access key got replaced by FullAccess key

vgrichina:
Screen Shot 2020-11-11 at 3 26 23 PM

re-enabling 2FA after disable gave error but seemingly succeeded

vgrichina:
Needs more careful testing (including regression tests in near-api-js) to proceed. Need to make sure works fine with app logins as well (i.e. doesn't touch app keys).

vgrichina:
@mattlockyer not a blocker for this fix, but "2fa code expired" error from server doesn't seem to be handled in UI yet, so kinda hard to understand what is going on in that case.

vgrichina:
Also "Resend Code" should be clearing code, as you'll need to enter new one

PR #1108, opened 2020-10-27
Refactor alerts and alert redux state

marcinbodnar:


marcinbodnar:
@vgrichina @kcole16 
This is a completely refactored alerts, it will be much easier to use and will be handier to debug.

Main features:
1. Whole alert stuff and `formLoader` (now `mainLoader`) are now in the separate reducer. This is a part of refactoring Redux state.
2. Instead of `requestStatus` state that was stored action status for the last action, now there is `actionStatus` that stores status for all actions that ware dispatched. It has a form of the object instead of an object array, so it will be easier to handle. It also stores all error info, so we can easily clear `globalAlerts` without worrying about clearing error info.
3. GlobalAlert now handles multiple alerts, so if there is a need we can display several alerts.
4. I simplified default translations for alerts. Now it will be based on action `type`, so we don't have to add (but it's available as an option) alert translation ID in meta. Also, all alerts translations will be in one place.
5. There is a `mainLoader` that checks if any action is pending.
6. There is also a separate `pending` state for every action in `actionStatus` state.
7. `requestStatus` state is very similar, but now it's called `localAlert`.
8. Now we can decide to show `globalAlert` in three ways:
 - `showAlert` added to action meta
 - `walletError`
 - `dispatchWithAlert` - for a situation where we want to show an alert only for specific dispatch.
9. Now `globalAlert` handles a few additional options like:
 - `onError` and `onSuccess`, to show an alert for error or success only
 - `console` to show or hide the console error message
 - `messageCodeHeader` for custom alert header
10. There is also `actionsPending` function, that will check if action or actions are in pending state - so it will be easier to use than previously `actionsPending` state.

@vgrichina For now, new alerts are only implemented for a few actions. Please let me know what do you think - if you would like to add any additional features. I implement new alerts for the whole project then.

src/utils/wallet.js
```diff
@@ -781,7 +781,7 @@ class Wallet {
         accountIds = [...accountsSet]
 
         if (!accountIds.length) {
-            throw new WalletError('Cannot find matching public key', 'account.recoverAccount.errorInvalidSeedPhrase', { publicKey })
+            throw new WalletError('Cannot find matching public key', 'allert.RECOVER_ACCOUNT_SEED_PHRASE.errorInvalidSeedPhrase', { publicKey })
```
vgrichina, line 784: `allert` seems like a typo

src/utils/alerts.js
```diff
@@ -0,0 +1,41 @@
...
+
+export const dispatchWithAlert = (action, data) => store.dispatch({
+    ...action,
+    meta: {
+        ...action.meta,
+        ...showAlert(data)
+    }
+})
+
+export const actionsPending = (types) => (typeof types === 'string' ? [types] : types).reduce((x, type) => {
```
vgrichina, line 24: do I understand correctly that `.reduce` is used here as https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some?

src/components/profile/Recovery/RecoveryContainer.js
```diff
@@ -91,7 +92,7 @@ const RecoveryContainer = () => {
         missingKinds.forEach(kind => activeMethods.push({kind: kind}));
     }
 
-    const loading = account.actionsPending.includes('LOAD_RECOVERY_METHODS') || account.actionsPending.includes('REFRESH_ACCOUNT');
+    const loading = actionsPending(['LOAD_RECOVERY_METHODS', 'REFRESH_ACCOUNT'])
```
vgrichina, line 95: I think generally makes sense to refactor this in similar way to alerts – have flag in `meta` whether action should be part of `loading`. By default makes sense to count everything in global loading status.

marcinbodnar:
@vgrichina @ihaid 
It's ready.
Big request, if we could handle it quickly, and don't merge any other changes before it (except for urgent stuff) - the problem is that it's touching many files and functionalities, and on every rebase I need to check again many things and it takes a lot of time.

src/utils/wallet.js
marcinbodnar: done

src/utils/alerts.js
marcinbodnar: done

src/components/profile/Recovery/RecoveryContainer.js
marcinbodnar: I'm not sure if it would be handy - the problem is that we might want to use some action as a loader in one place and not use it in another. I will take a closer look and check if a situation like this appears.

ihaid:
Found only one issue that alert shows on background
image

ihaid (approved):

marcinbodnar:
Thanks @ihaid fixed it.

src/actions/account.js
```diff
@@ -192,32 +191,31 @@ export const {
     checkNewAccount,
     createNewAccount,
     checkAccountAvailable,
-    clear,
     clearCode
 } = createActions({
     INITIALIZE_RECOVERY_METHOD: [
         wallet.initializeRecoveryMethod.bind(wallet),
-        () => defaultCodesFor('account.initializeRecoveryMethod')
+        () => ({})
```
vgrichina, line 198: @marcinbodnar  is empty function needed here for some reason or we can just omit `meta` for such cases?

vgrichina: also what is the implied meaning – does this mean that we expect whoever dispatched this action to handle errors locally?

marcinbodnar: >is empty function needed here for some reason or we can just omit meta for such cases?

We can omit it, but let me take care of that while refactoring redux state

> also what is the implied meaning – does this mean that we expect whoever dispatched this action to handle errors locally without error alert?

Information about action success or failure will be stored anyway automatically in `actionStatus` object - by adding or not adding `showAlert` we decide do we want to show an alert about this error/success or not.

src/reducers/status/index.js
```diff
@@ -0,0 +1,117 @@
...
+                        || (error
+                            ? payload.type
+                                ? `reduxActions.${payload.type}`
+                                : `reduxActions.${type}.error`
+                            : `reduxActions.${type}.success`),
+                    console: error && (meta.alert?.console || payload.data?.console)
+                }
+                : undefined
+        },
+        localAlert: typeof ready === 'undefined'
```
vgrichina, line 65: why there are differences in logic between `localAlert` and `globalAlert`? Like different code for `show` and `messageCode`, etc

seems like stuff would be simpler if data structure is always the same

src/reducers/status/index.js
```diff
@@ -0,0 +1,117 @@
...
+    localAlert: {}
+}
+
+const alertReducer = (state, { error, ready, payload, meta, type }) => {
+    const actionStatus = {
+        ...state.actionStatus,
+        [type]: typeof ready === 'undefined'
+            ? undefined
+            : {
+                success: typeof ready === 'undefined'
```
vgrichina, line 23: this also seems to be almost the same but slightly different than `localAlert` and `globalAlert`. Is there any reason for this?

vgrichina:
@marcinbodnar looks good overall and should be ok to merge into current release after resolving conflicts. See questions on alerts structure however.

src/reducers/status/index.js
marcinbodnar: - `show` - for `globalAlert` it depends also on `onlyError/onlySuccess` param - at this moment we don't use these params for `localAlert` in the project
 - `messageCode` - for `globalAlert` we use `messageCode` from `WalletError` and we also use `type` from RPC error types (for example `RetriesExceeded`) - we don't use it for `localAlert`.

But I guess in both cases we can implement it, maybe in the future we can make use of these. Let me add it.

src/reducers/status/index.js
marcinbodnar: I plan to use `actionStatus` to store basic data for all actions that were dispatched - sometimes we dispatch many actions one after another and it's difficult to track their results. So `actionStatus` stores data for all actions, and `globalAlert/localAlert` stores only data that we want to display in UI. I also use `actionStatus` for checking if action is pending, and in `globalAlert` component for animations.

vgrichina (approved):
@marcinbodnar please resolve conflicts and merge

Issue #1082, opened 2020-10-23
"Unknown App" when log in with near-cli


frol:
**Problem**

![image](https://user-images.githubusercontent.com/304265/96975746-49a29e00-14f1-11eb-9f1b-3474a90a6f82.png)

**Expected Behavior**

It should display "NEAR CLI" as passed in the URL (the URL has `?title=NEAR+CLI`

**Steps to reproduce**

1. `env NEAR_ENV=mainnet near login`

kcole16:
@frol We intentionally do not display the title, as that can easily mislead users.

@marcinbodnar We should display something like `NEAR CLI` when the `referrer` is `localhost`, as this is a special case

Issue #726, opened 2020-07-08
Require /login transactions to be executed in callback URL


kcole16:
### Overview
The fix in #719 provides the users with more info to prevent the below vulnerability. To further reduce the likelihood of phishing, we can move the execution of the `addKey` TX on `/login` out of the wallet, and instead return the signed TX to the URL in the `callback_url`. 

Steps to expose vulnerability:
1. I wish to gain access to Kendall's Flux account.
2. I initiate a signin to Flux. At this point a new keypair K is generated in my session, and the private key is stored in my local storage. I am directed to a URL that will ask me for permission, and that URL contains the public key of that keypair.
3. I send that URL to Kendall. When he opens it, the screen legitimately asks for access to Flux, and is from near.org domain. Kendall has no reason not to trust it and not to Allow access.
4. Once Kendall allows access, the public key is added to his account, while the private key is in my local storage. I have full access to his Flux account.

### Acceptance Criteria
- [ ] If `callback_url` isn't valid, display error "This app cannot be authorized. This may be due to a problem with the app, or can be an attempt at phishing. We do not recommend using the app at this time."
- [ ] After user clicks "Allow", user is redirected to `callback_url` with `signed_tx` as an argument

kcole16:
Note that this likely adds a bit of work for apps when it comes to handling and communicating TX errors to the user. They should be doing this already anyway, but can effectively offload some of this to the wallet in the current flow.

vgrichina:
@kcole16 yeah, biggest problem I can see is that technically it's possible to use up same nonce already before app gets chance to do it. However thankfully it's pretty unlikely in normal usage. Like you'll have to login into other app (through same wallet) or sign transaction from app (through same wallet) while app is submitting transaction from existing login flow. 

So it's likely just ok to have generic "something went wrong, please login again" flow on the receiving side.

kcole16:
From @vgrichina 
**Problem**

Referrer not always reliable as suggested by @abacabadabacaba:

Many websites allow users to post arbitrary link, and these links usually don't block the referrer. This is generally not considered a security issue, unless the page's URL is secret.
Likewise, many websites have redirectors that can redirect to an arbitrary URL. This is also not considered to be problematic by itself, even by Google.
Referrer may be blocked in some browsers.

**Proposed Solution**

Callback URL should be used as primary source of truth (especially after #726 lands). However if Referrer is set and isn't matching callback URL – we should report it as might be a sign of something sketchy going on.

[...]

PR #1112, opened 2020-10-28
Show proper title if login with near-cli (#1082)

marcinbodnar:


marcinbodnar:
@kcole16 @vgrichina 
We are not able to detect that it was in fact CLI - referrer will be empty, like when entering URL directly into the browser tab. So in this case I'm checking the title, and if the title is `NEAR CLI` it is shown. It should be secure because the referrer is checked first, so this title will appear only if the referrer is empty.

Also, we are showing `title` from URL on the details page and on the confirm account id page. So I also changed it to show the same title like on the main login page.

vgrichina:
@marcinbodnar we cannot trust title passed through URL unfortunately, so we cannot show it to user. For NEAR CLI we should show something like "Unknown App".

We also will need to switch to using callback URL domain instead https://github.com/near/near-wallet/issues/726

marcinbodnar:
@marcinbodnar even for this particular situation (when no referref)?

"Unknown app" is shown now, #1082

kcole16:
@marcinbodnar My fault here. I thought we could determine whether the referrer was localhost. If this is not the case, we should keep it as-is. Otherwise, it's abusable (anyone could put "NEAR CLI" as the title, and we can't easily verify).

marcinbodnar:
@kcole16 
> anyone could put "NEAR CLI" as the title, and we can't easily verify

It's true but only in case you copy and paste this URL into a new browser tab, or the page will be launched from the console.

But if that's also dangerous then let me remove the title and we can merge fixes for details and confirm account id page.

vgrichina:
> It's true but only in case you copy and paste this URL into a new browser tab, or the page will be launched from the console.

@marcinbodnar not sure what do you mean. Literally everyone can construct such URL and e.g. post it in some public forum.

marcinbodnar:
> not sure what do you mean. Literally everyone can construct such URL and e.g. post it in some public forum.

@vgrichina Yes, but if you click on this link the `referrer` will not be empty, and the title will not appear.

PR #1123, opened 2020-10-30
fix identify and setting for user profile

icerove:
1. add function `identity` for user profile setting pre-requisite
2. fix some setting for user profile setting.

### DIFF
--- a/src/middleware/index.js
+++ b/src/middleware/index.js
@@ -46,6 +46,7 @@ const readyStatePromise = store => next => action => {
 
 if (process.env.MIXPANEL_TOKEN) {
     mixpanel.init(process.env.MIXPANEL_TOKEN, {'property_blacklist': ['$current_url']})
+    window.mixpanel = mixpanel
 }
 
 const analyticsMiddleware = store => next => action => {
@@ -69,6 +70,8 @@ const analyticsMiddleware = store => next => action => {
             timestamp: new Date().toString(),
             account_id: wallet.accountId
         })
+        let id = mixpanel.get_distinct_id();
+        mixpanel.identify(id);
         mixpanel.people.set({
             network_id: networkId,
             stake: 0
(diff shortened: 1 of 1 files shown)


src/middleware/index.js
```diff
@@ -46,6 +46,7 @@ const readyStatePromise = store => next => action => {
 
 if (process.env.MIXPANEL_TOKEN) {
     mixpanel.init(process.env.MIXPANEL_TOKEN, {'property_blacklist': ['$current_url']})
+    window.mixpanel = mixpanel
```
vgrichina, line 49: is this debug leftover?

src/middleware/index.js
```diff
@@ -69,31 +70,31 @@ const analyticsMiddleware = store => next => action => {
             timestamp: new Date().toString(),
             account_id: wallet.accountId
         })
+        let id = mixpanel.get_distinct_id();
+        mixpanel.identify(id);
```
vgrichina, line 74:
```
mixpanel.get_distinct_id

Returns the current distinct id of the user. This is either the id automatically generated by the library or the id that has been passed by a call to identify().
```

@icerove not sure I understand the idea here. Docs say that `.get_distinct_id` returns ID set by `.identify`

Can you explain big picture of what are you trying to achieve?

src/middleware/index.js
```diff
@@ -69,31 +70,31 @@ const analyticsMiddleware = store => next => action => {
...
         mixpanel.people.set({
             network_id: networkId,
             stake: 0
         })
         if (trackingEvents.includes(action.type)){
             mixpanel.track(action.type, details);
         }
-        if (action.type === 'CREATE_NEW_ACCOUNT'){
```
vgrichina, line 79: Following changes don't seem to match PR topic of __fix identify__.

So either PR description should be updated or these changes should be in the separate PR with separate explanation.

src/middleware/index.js
icerove: if I want to set up user profile, I have to call function `identify` and then use function `people.set`. see this: https://developer.mixpanel.com/docs/javascript#setting-profile-properties

src/middleware/index.js
icerove: actually, not. I am thinking leaving this. But I will delete this for now since this is not for this PR.

src/middleware/index.js
icerove: it is part of setting user profile and I will change the description.

src/middleware/index.js
vgrichina: @icerove why is it necessary to keep `window.mixpanel`?

icerove: I used to think of easy checking for us.

Issue #1124, opened 2020-10-30
Sign in with Ledger UI layout issue


vgrichina:
**Problem**

<img width="845" alt="Screen Shot 2020-10-30 at 3 05 35 PM" src="https://user-images.githubusercontent.com/126710/97760871-75a9ba80-1ac1-11eb-8ac4-6c2eead6b131.png">

PR #1125, opened 2020-10-30
Allow importing remaining Ledger accounts if one fails

vgrichina:
Context: whoever added 2FA after Ledger (e.g. by importing account using other recovery method) has troubles recovering any accounts connected to the same Ledger (so this is relatively urgent fix for people in such situation).

As a drive-by it also allows to press "Reject" for some accounts and skip them (seems like more desired behavior than restarting whole process).

Known issues:
- missing translations
- need design for Error/Rejected state @corwinharrell 
- need to distinguish Error/Rejected (just shows Error now)
- account list doesn't scroll after rejection (not sure whether regression here)
- there seem to be Ledger UI issues which weren't introduced in this PR https://github.com/near/near-wallet/issues/1124

### DIFF
--- a/src/actions/account.js
+++ b/src/actions/account.js
@@ -140,9 +140,14 @@ export const signInWithLedger = () => async (dispatch, getState) => {
 }
 
 export const signInWithLedgerAddAndSaveAccounts = (accountIds) => async (dispatch, getState) => {
-    for (let i = 0; i < accountIds.length; i++) {
-        await dispatch(addLedgerAccountId(accountIds[i]))
-        await dispatch(setLedgerTxSigned(false, accountIds[i]))
+    for (let accountId of accountIds) {
+        try {
+            await dispatch(addLedgerAccountId(accountId))
+            await dispatch(setLedgerTxSigned(false, accountId))
+        } catch (e) {
+            console.warn('Error importing Ledger-based account', accountId, e)
+            // NOTE: We still continue importing other accounts
+        }
     }
 
     return dispatch(saveAndSelectLedgerAccounts(getState().ledger.signInWithLedger))
--- a/src/components/accounts/ledger/LedgerSignInModal.js
+++ b/src/components/accounts/ledger/LedgerSignInModal.js
@@ -125,7 +125,8 @@ const LedgerSignInModal = ({
     open, 
     onClose, 
     ledgerAccounts, 
-    accountsApproved, 
+    accountsApproved,
+    accountsError,
     totalAccounts, 
     txSigned, 
     handleAdditionalAccountId, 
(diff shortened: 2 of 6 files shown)


Patrick1904 (approved):

Patrick1904:
Probably rare use case but would be nice with a recover/continue button that you can click in case you have many accounts and only want to select some of the first. Closing the modal doesn't continue with sign in. Should we add ability to select specific accounts?

cc @marcinbodnar @kcole16 @corwinharrell

marcinbodnar:
@vgrichina 
 - fixed list animation so it will handle `error` state
 - added style for `error` state
 - fix `line-height` for indicator
 - handle error/rejected state - now it will only consider `RequiresFullAccess` error as `error` but we can add any other type if needed

@Patrick1904 There is an issue for it, I will handle it after finish with Redux state refactoring

vgrichina:
@marcinbodnar thanks for help!

https://github.com/near/near-wallet/pull/1125/commits/5b4388410edb3fb420b399b70b8fe7fb167366ea I think we need to reverse this check. Makes sense to check for specific Ledger error which corresponds to rejection, instead of assuming that everything that's not `RequiresFullAccess` is rejection (quite likely it's just different error).

corwinharrell:
@vgrichina @marcinbodnar sorry I lost track of this one in the mix. Re: Errors, do we feel like all of the design challenges were adequately addressed or do you feel like there is still some room for enhancement?

marcinbodnar:
@corwinharrell this is how it looks now.

corwinharrell:
@marcinbodnar do we show anything to give context about why the error occurred? Is that possible?

marcinbodnar:
> Makes sense to check for specific Ledger error which corresponds to rejection, instead of assuming that everything that's not RequiresFullAccess is rejection (quite likely it's just different error).

@vgrichina 
Error thrown by Ledger is not much specific - it throws `Ledger device: UNKNOWN_ERROR (0x9100)` and the standard name `TransportStatusError`. So I thought a better idea would be to specify all error situations by listing RPC types that might appear (`RequiresFullAccess` is the error situation that you describe with 2fa and Ledger).

I changed it to check `TransportStatusError` error for `rejected`, and `error` for every other situation. #1135 

Also added error if no accounts were accepted. #1135

corwinharrell:
Gotcha @marcinbodnar. Might I also suggest a small visual change to help better distinguish between `Error` and `Rejected`

Screen Shot 2020-11-03 at 1 28 31 PM

marcinbodnar:
@corwinharrell 
Added Global Alert if error appears - this is the error message for now, please suggest something better: `Error appears while recovering the account.`. 

Also changed colors for `error` and `rejected` status.

All available here: #1135

PR #1127, opened 2020-10-30
Receive page: copy account ID only

Patrick1904:


src/components/receive-money/ReceiveMoney.js
```diff
@@ -137,7 +137,7 @@ class ReceiveMoney extends Component {
     }
 
     get receiveUrl() {
```
vgrichina, line 139: @Patrick1904 should rename this from `receiveUrl`

src/components/receive-money/ReceiveMoney.js
```diff
@@ -155,7 +151,7 @@ class ReceiveMoney extends Component {
                             <Address onClick={this.handleCopyAddress}>
                                 {this.props.account.accountId}
                                 <UrlAddress ref={this.urlRef}>
```
vgrichina, line 153: @Patrick1904 nit: `UrlAddress` also seems a bit misleading?

PR #1128, opened 2020-10-31
Allow enabling other recovery methods while using Ledger

Patrick1904:


kcole16:
Can we remove the "You have no method to recover your account. please add a method below." when Ledger is enabled?

Screen Shot 2020-11-02 at 5 11 31 PM

kcole16 (approved):

vgrichina:
@Patrick1904 @kcole16 I've merged master here so that we can re-test together with other changes (as there was some Ledger related stuff). Other than that looks good so far.

vgrichina:
Screen Shot 2020-11-04 at 3 52 26 PM

@Patrick1904 not a blocker, but looks like we still have issue with page layout. Not sure if a regression as you've previously had a fix.

vgrichina:
@Patrick1904 @mattlockyer seems like there is a problem if you mess up anything about signing transaction after entering the code from e-mail:

Screen Shot 2020-11-04 at 3 55 58 PM

Patrick1904:
> seems like there is a problem if you mess up anything about signing transaction after entering the code from e-mail:

@vgrichina can you please elaborate on this? Did you enter the wrong code and that caused verification to fail completely even when you entered the correct code?

vgrichina:
> > seems like there is a problem if you mess up anything about signing transaction after entering the code from e-mail:
> 
> @vgrichina can you please elaborate on this? Did you enter the wrong code and that caused verification to fail completely even when you entered the correct code?

no, didn't accept transaction in time on Ledger. It all worked out ok when I sent new code. Don't think it's a blocker, but we need to improve the flow.

Patrick1904:
@vgrichina not sure if this is the same bug that you encountered but I just found one by:

1. Create account with Ledger
2. Initiate flow to setup email recovery (`/setup-recovery`)
3. Enter the email verification code
4. Reject the TX on Ledger
5. Go back to `/setup-recovery`
6. Notice that email shows as active because the key was successfully saved in CH, but not on chain.

Patrick1904:
Two issues:
1. Key is added to CH _before_ it's added to chain.
We validate the security code in CH and then add the key to chain. Only adding the key to chain is confirmed on Ledger
2. We're not double checking on `/setup-recovery` that the recovery key exists on chain before, only CH. Will fix that.

vgrichina:
> Notice that email shows as active because the key was successfully saved in CH, but not on chain.

I see, this is indeed a blocker, let's postpone this PR for next release cut

src/utils/wallet.js
```diff
@@ -718,11 +718,19 @@ class Wallet {
...
+        try {
+            await this.addNewAccessKeyToAccount(accountId, publicKey)
+        } catch(e) {
+            const account = this.getAccount(accountId)
+            const accountKeys = await account.getAccessKeys()
+            if (!accountKeys.some(it => it.public_key.endsWith(publicKey))) {
+                method.publicKey = publicKey
+                await this.deleteRecoveryMethod(method)
+                throw new WalletError('Recovery setup was not successful! Please try again.', 'errors.recoveryMethods.setupMethod')
+            }
```
mattlockyer, line 732:
```
else {
    console.warn(e)
}
```

mattlockyer: otherwise if the error is something else, we'll never know

Patrick1904:
@vgrichina now filtering out recovery methods that do not have key on chain

Patrick1904:
> Screen Shot 2020-11-04 at 3 52 26 PM
> 
> @Patrick1904 not a blocker, but looks like we still have issue with page layout. Not sure if a regression as you've previously had a fix.

@marcinbodnar is this something you can include a fix for in your other Ledger PR? Otherwise I can. Also seems it easily overflows on mobile.

marcinbodnar:
@Patrick1904 all Ledger PR are closed - I can add a fix here?

Patrick1904:
> @Patrick1904 all Ledger PR are closed - I can add a fix here?

@marcinbodnar that'd be awesome! Thanks

marcinbodnar:
@Patrick1904 done

vgrichina:
@Patrick1904 when I recover using e-mail link Ledger doesn't show up in the list of devices. When I "Connect" it again – it gets enabled there, but I still have a local key for signing.  Not sure if really a blocker @kcole16

src/utils/wallet.js
```diff
@@ -718,12 +722,15 @@ class Wallet {
...
-        await this.validateSecurityCode(accountId, method, securityCode);
-        const newPublicKey = recoveryKeyPair.publicKey
-        await this.addNewAccessKeyToAccount(accountId, newPublicKey)
-        await store.dispatch(redirectTo('/profile', { globalAlertPreventClear: true }))
+        const { publicKey } = parseSeedPhrase(recoverySeedPhrase)
+        await this.validateSecurityCode(accountId, method, securityCode)
+        try {
+            await this.addNewAccessKeyToAccount(accountId, publicKey)
+        } catch(e) {
+            throw new WalletError(e, 'errors.recoveryMethods.setupMethod')
```
vgrichina, line 730: @Patrick1904 I think expected first argument to `WalletError` is error message, not another error. Probably best to do `console.error` with caught error.

kcole16:
@vgrichina @Patrick1904 Not having the Ledger show is borderline a blocker. If this is a quick fix, we should implement it. If not, we should aim to have a followup fix in shortly thereafter.

As for adding a localstorage key during recovery, I'd say that is intended behavior, but we do not communicate the ramifications well. There isn't much reason to enable email recovery if the user intends to enforce Ledger signing on every TX.

We should add more warning to users when they attempt to add an additional recovery method with only Ledger installed, as some may not realize that means Ledger signing will not always be enforced. Will create a separate issue for this.

Patrick1904:
> @Patrick1904 when I recover using e-mail link Ledger doesn't show up in the list of devices. When I "Connect" it again – it gets enabled there, but I still have a local key for signing. Not sure if really a blocker @kcole16

@vgrichina @kcole16 - currently when user recovers their account with link (email etc) we encourage the user to 'reconnect' their Ledger when they land on `/profile`. I believe this is the intended behavior?

However, as @vgrichina mentioned, we currently **do not** delete the localStorage key if the user 'reconnects' the Ledger. I can fix that if we consider this a blocker?

kcole16:
@Patrick1904 I don't thin

src/utils/wallet.js
Patrick1904: @vgrichina fixed it here: https://github.com/near/near-wallet/pull/1128/commits/f11ee9270bd751a6ecfa8f856474cfd0ead588ff

Issue #1129, opened 2020-11-02
Improve "Available Balance" clarity


kcole16:
### Overview
We've had some confusion about the `available balance` since we've subtracted the 2N "tolerance". 

While we work on a more complete solution, we should remove the 2N "tolerance", and display the `available balance` like before.

### Acceptance Criteria
- [ ] `available balance` = `actual balance` - `minimum balance`
- [ ] Remove "Use Max" from staking flow
- [ ] If user enters an amount > `available balance` - 0.1 N, they receive an error
   - [ ] `/staking`
   - [ ] `/send-money`

vgrichina:
> If user enters an amount > available balance - 0.1 N, they receive an error

hmm, doesn't it just mean changing minimum balance from 2N to 0.1 N? Or am I missing something?

I think part of the problem is that there is also actual minimum balance (i.e. locked for storage), for this reason some math was off. + generally we are pretty off with info on staked tokens

corwinharrell:
@kcole16 in the scenario where a user inputs an amount equal to `Available balance` - `0.1N`, the transaction is guaranteed to fail is it not?

kcole16:
@corwinharrell Why would it fail? The idea is that the 0.1N should be sufficient to cover gas costs (at least in most cases). If they input their actual full account balance, then it would fail.

corwinharrell:
Sorry @kcole16 I meant to say greater than, not equal to:

My reason in asking is that we can simply disable the Stake/Send submission buttons if the amount exceeds `Available balance - 0.1N`, and display a simple message inline, rather than showing a more obtrusive warning once they've submitted.

corwinharrell:
@Patrick1904 @vgrichina @kcole16 @mattlockyer Curious to know what you guys think of this approach:

![Network Fees](https://user-images.githubusercontent.com/5934188/98023350-c9bfe200-1dd4-11eb-9a84-cb255365a906.gif)

Patrick1904:
> @Patrick1904 @vgrichina @kcole16 @mattlockyer Curious to know what you guys think of this approach:
> 
> ![Network Fees](https://user-images.githubusercontent.com/5934188/98023350-c9bfe200-1dd4-11eb-9a84-cb255365a906.gif)

[...]

PR #1132, opened 2020-11-03
Improve staking and send flows

Patrick1904:
Changes:
1. Show the same 'Available balance' on profile
2. Show contextual 'Avilable to {action}' - stake, unstake, send
3. `WALLET_APP_MIN_AMOUNT` is equal to = `MIN_BALANCE_FOR_GAS` + `ACCESS_KEY_FUNDING_AMOUNT`
3. Change default value for `MIN_BALANCE_FOR_GAS` from 2N to 0.1N

https://github.com/near/near-wallet/issues/1129


Patrick1904:
@kcole16 @vgrichina 

I didn't make the `0.1 N` a variable yet, because I'm wondering if we need to create a new env variable, or if we can change `MIN_BALANCE_FOR_GAS` from `2 N` to `0.1 N`? I can see that `MIN_BALANCE_FOR_GAS` is being used for some lockup and implicit account stuff.

corwinharrell (approved):

Patrick1904:
@vgrichina @kcole16 just want to make sure we don't forget to about this PR. Ideally this should be running on master/testnet for a few days before we ship to production. I think it greatly improves clarity around the users available balance and would love to get it in before we cut for this weeks release.

kcole16:
@Patrick1904 @vgrichina I vote we go ahead and merge it, if there haven't been any issues reported

kcole16 (approved):

vgrichina:
@Patrick1904 I cannot send my "Avaliable Balance"  Screen Shot 2020-11-09 at 5 36 58 PM

kcole16:
@vgrichina That is intended. We are no longer subtracting the 2 NEAR buffer from "Available Balance" in the background, and now telling the user they must leave some NEAR in their account.

@Patrick1904 Why did we change the amount from 0.1 to 0.35?

vgrichina:
> @vgrichina That is intended. We are no longer subtracting the 2 NEAR buffer from "Available Balance" in the background, and now telling the user they must leave some NEAR in their account.
> 

I think we will get a clear problem if we ship as is. People will definitely try to send all "Available Balance" (they did it before according to community chats. I did it immediately myself) and get confused/disappointed when it's not possible. Also mental math would be needed to send all tokens.

> @Patrick1904 Why did we change the amount from 0.1 to 0.35?
'"
@kcole16 I think it makes sense to keep this limit at "access key gas" + "min balance". So that you can still login to few apps if needed.

Maybe the safest change is not to change the logic of available balance – but just reduce the amount for now?

Patrick1904:
I agree with Vlads comment that current UX is suboptimal when attempting to ‘send all’. This will be fixed by adding a [‘Use max’](https://github.com/near/near-wallet/issues/1154) button to `/send` page.

Would it be better if we showed a contextual ‘available balance’ on send and stake? ‘Available to send’ and ‘Available to stake’? That way, the number you see, is the number you can stake/send. As a consequence, those numbers won’t match with the ‘available balance’ on /profile, which will most likely cause some confusion, though we can provide info bubbles to solve for that.

@vgrichina @kcole16 @corwinharrell what do you think?

kcole16:
Not being able to send your full balance is annoying, but I think it's better than complicating the balances further than they already are (our balance are way more confusing for various reasons than other chains).

I think we need to avoid adding more balances, especially ones that differ slightly. By adding "Use Max", we make it pretty easy to send the feasible max amount, and we're only requiring users to leave ~35 cents behind. 

Users are upset when they have to leave 35 NEAR behind since it's a substantial amount. 0.35 IMO is far from that, and while mildly annoying, isn't exactly a deal breaker.

corwinharrell:
> I think we need to avoid adding more balances, especially ones that differ slightly. By adding "Use Max", we make it pretty easy to send the feasible max amount, and we're only requiring users to leave ~35 cents behind.

+1 to the above

> ...though we can provide info bubbles to solve for that.

We should always operate as if info bubbles solve nothing for the user 😆 We just can't depend on people looking to these to answer any questions they might have, and so they should never be the lone source of critical information.

Patrick1904:
@kcole16 @corwinharrell @vgrichina I added the same 'Use max' UX to the `/send` page in case we decide to move forward with this iteration of available balance.

@corwinharrell we really should re-design the send flow:) Especially mobile UX.

ihaid:
image

1. Use an account with a balance lower than 0.35
2. Click on the "Use max" button

ihaid:
I`m able to send an amount more than the "use max" button is saying. You need to input any special character in front of the amount and the button "Send" becomes active.

This amount after "use max":
telegram-cloud-photo-size-8-5289519539949777253-y
This is when you put special character in front:
telegram-cloud-photo-size-2-5289519539949777252-y
telegram-cloud-photo-size-2-5289519539949777254-y

Patrick1904:
@kcole16 @vgrichina, I've implemented the new balance treatment for stake/unstake and send flows. Please check it out and let me know if you find anything weird.

kcole16:
Managed to get a negative amount in the input on both "Send" and "Staking" by selecting "Use Max". 

My total balance was 0.12: https://near-wallet-pr-1132.onrender.com/profile/testbal.testnet

Screen Shot 2020-12-02 at 5 20 43 PM

Not a huge deal as the "Submit" is disabled, but could be confusing for users if the input can be negative.

Patrick1904:
> Managed to get a negative amount in the input on both "Send" and "Staking" by selecting "Use Max".

@kcole16 thanks for catching. Should be fixed now.

Patrick1904:
> Managed to get a negative amount in the input on both "Send" and "Staking" by selecting "Use Max".

@kcole16 thanks for catching! Fixed this on staking and send now

kcole16 (approved):

Patrick1904:
@vgrichina any objections on this one? Think it'd be great to get it into the next release

PR #1133, opened 2020-11-03
Mainnet Wallet Release 2020-11-03

vgrichina:
### Release Notes 
- Add ability to delete lockup contract and withdraw 35 NEAR storage requirement
- Highlight account switching in staking UI
- Highlight that the implicit account is temporary during account creation
- Allow to import accounts connected to Ledger selectively (i.e. single Reject just skips account instead of cancelling process).
- Bug and UI fixes


marcinbodnar:
@vgrichina we may want to add this one #1135

marcinbodnar (approved):

mattlockyer (approved):
Can you please try to include:

https://github.com/near/near-wallet/pull/1137

PR #1137, opened 2020-11-04
use wrapped account vs. account-with-lockup when staking from account

mattlockyer:


src/utils/staking.js
```diff
@@ -191,13 +191,10 @@ export class Staking {
...
 
     async updateStakingAccount(recentlyStakedValidators = []) {
-
         const account_id = this.wallet.accountId
         await this.wallet.refreshAccount()
         const account = this.wallet.getAccount(this.wallet.accountId)
-        const balance = await account.getAccountBalance()
-
-
+        const balance = await account.wrappedAccount.getAccountBalance()
```
vgrichina, line 197: @mattlockyer what if account doesn't have lockup?

mattlockyer: It's fine. The original Account instance is in `accountWithLockup.wrappedAccount`.

mattlockyer:
Should fix #1141

src/utils/staking.js
vgrichina: ah, I forgot that we are wrapping Account instance regardless of whether account has lockup

I think it still helps to check whether `wrappedAccount` exists first, as this is likely to change later. I think it's only done this way because `getAccount` isn't async

mattlockyer: good point! updated

PR #1144, opened 2020-11-05
Fund using MoonPay prototype

vgrichina:


vgrichina:
If you try to create account outside of US – "Fund with Moonpay" button appears

PR #1149, opened 2020-11-07
Staking account selector

Patrick1904:
Changes:
1. Add new account selector on `/staking`
2. Add heavier font weight to balances and headings
3. Clean up Balance component and add missing tests
4. Delete unused components: `BalanceBanner` and `AccountSwitcher`
5. Some minor cleanups
6. Change main CTA on `/staking` from 'Select Validator' --> 'Stake my tokens'


corwinharrell:
The goal of the move away from the contract dropdown is ultimately to provide more clarity about:
1. which contract the user is staking from
2. the balances of each contract
3. the ability for the user to switch between staking from their regular account balance, and their lockup contract balance.

With this in mind, I've been experimenting with two different directions for handling the UX for this iteration:

### Option 1: Inline contract switch
**Pros**:
- Contract switching happens inline.
- It's easier for the user to see right off the bat which contracts they can stake from.
- Smaller scope

**Cons**:
- Takes up a lot of vertical real estate. Not super friendly for smaller devices.
- Not scalable. In the event that we need to support > 2 sources, this option doesn't scale and will naturally need to evolve into Option 2.
-  Doesn't give us any space to articulate the "why" behind including this functionality in the UI

Contract Switch

### Option 2: Separate Modal/View
**Pros**:
- Scalable
- Gives us an opportunity to give a little context into reasoning behind and the outcome of this action. Helps set expectations for the user.
- Frees up vertical real estate for smaller devices.
- Allows us to show more info about the contracts if we wish to in the future.

**Cons**:
- Extra click
- Increased scope

Contract Switch Alt

Would love to here any feedback from the wallet team on weighing the pros/cons of each direction cc @kcole16 @vgrichina @Patrick1904 @mattlockyer @marcinbodnar

Patrick1904:
I'm all for option 1. I think it's the most clear, easy and fast (less clicks) UX. AFAIU, the number of users staking from lockup contracts will decrease with time. I also don't see the added vertical real estate as an issue, users will scroll down. And important CTAs are above the fold.

Patrick1904:
@corwinharrell @kcole16 this is the current state of the prototype. Should we update the design to your option (1) attached above? Essentially adding an additional line in each box.

Screen Shot 2020-11-09 at 9 42 44 AM

corwinharrell:
@Patrick1904 yeah I've thought about it over the weekend and imo if we decide to go with Option 1 we should use labels for each of the balances, simply because I think it benefits a lot from the additional clarity. Better to lose 20px of height than have users guess at what a certain figure in our UI represents.

corwinharrell:
Also a question, why did we switch CTA's from "Stake..." to "Select Validator"? Imo the primary CTA of the view should reflect the user's ultimate goal (ie "Stake my tokens"). Once they reach the point of selecting a validator from the validator profile, I think it makes sense to use "Select validator", but not as much as an CTA that functions as an entry into the staking flow. Thoughts?

Patrick1904:
> Also a question, why did we switch CTA's from "Stake..." to "Select Validator"? Imo the primary CTA of the view should reflect the user's ultimate goal (ie "Stake my tokens"). Once they reach the point of selecting a validator from the validator profile, I think it makes sense to use "Select validator", but not as much as an CTA that functions as an entry into the staking flow. Thoughts?

Yes we changed it because we received feedback from the community that 'Select Validator' is more accurate. I agree with your point and would prefer 'Stake Near', but also see an argument for keeping it as 'Select Validator.  @kcole16 curious about your thoughts.

Patrick1904:
> > Also a question, why did we switch CTA's from "Stake..." to "Select Validator"? Imo the primary CTA of the view should reflect the user's ultimate goal (ie "Stake my tokens"). Once they reach the point of selecting a validator from the validator profile, I think it makes sense to use "Select validator", but not as much as an CTA that functions as an entry into the staking flow. Thoughts?
> 
> Yes we changed it because we received feedback from the community that 'Select Validator' is more accurate. I agree with your point and would prefer 'Stake Near', but also see an argument for keeping it as 'Select Validator. @kcole16 curious about your thoughts.

@corwinharrell @kcole16 @vgrichina final design is implemented and ready for review. I also changed the main CTA on `/staking` from 'Select Validator' --> 'Stake NEAR'

kcole16:
@corwinharrell @Patrick1904 I'm open to the copy change, as I do think it is clearer overall. The user feedback was that the initiation of the staking flow seems to begin after selecting a validator, since we have the user select "Stake with validator" on the validator page. 

It's out of scope for this issue, but I think it's worth making this change as well to clear things up even further.
- After pressing "Select", user is moved directly to `/staking/<validator_name>/stake`, rather than the validator detail page

corwinharrell:
@Patrick1904 Thoughts on something slightly different like "Stake my tokens" or "Stake your tokens"?

> After pressing "Select", user is moved directly to /staking/<validator_name>/stake, rather than the validator detail page

This is bit prescriptive since it basically removes the validator profile/details page from the flow entirely. Some users are going to value seeing that information before they make the decision to stake, especially as the validator profile becomes more robust (I know we've also heard things like "How do I decide who to delegate to? In these scenarios, it doesn't benefit the user to circumvent the validator profile if the profile contains information that might be helpful in making a decision with whom to stake. I understand that right now the profiles are sparse unless you've already staked with one, but just wanted to point this out to remind us of the long-term vision and intent of the validator profile). I do think it's worth discussing an option for users to skip ahead in the process though if they know ahead of time who they'd like to stake with. I think this is something that we can accommodate, but let's frame it as a specific problem we are trying to solve and lay out what the user's motivations are for doing so, and then work to decide what the best solution is.

On that note, I want to make sure that when we receive feedback about things like this, that we're not rushing to make changes based on one or two comments that we receive (I'm not accusing anyone of this, I'm simply saying this because right now, there's not a lot of aggregation of feedback in a way that makes it easy to see how often we're hearing things and from whom. It's entirely anecdotal and handed down from whoever heard it first).

I'm in favor of reacting quickly if it's obvious that something is causing problems for a majority, but I think we need to take feedback with a grain of salt and understand that different users have different motivations and contexts coming into a given scenario, and that this is going to affect what they want or expect to see at any given moment. We should be paying attention to who the users are who request certain changes, and try to understand what the motivation is behind a particular request. We can then hold this up against the user-type that we're actively designing for to determine if it's something that's worth acting in swift response to.

Patrick1904:
> @Patrick1904 Thoughts on something slightly different like "Stake my tokens" or "Stake your tokens"?

@kcole16 @corwinharrell updated copy to 'Stake my tokens'

corwinharrell:
@Patrick1904 looking good. I say we kill the half black / half grey text treatment in the account name. I think it adds visual noise more than anything else. Still keen to hear what the rest of the team has to say regarding the two directions I presented, but if this is the direction we head in I think it looks good.

kcole16:
@corwinharrellv @Patrick1904  I'm in favor of Option 1. We've had enough confusion from users with lockups that it's worth making the account choice explicit. Additionally, we should see the % of users with a lockup contract steadily declining over time, which makes it less crucial to attempt to save space. 

@corwinharrell Great points re: not being too reactive (which I have definitely been guilty of). I've opened a discussion re: how we empower users to compare and choose validators, which I think is worth aligning on before we consider other changes (like the one I proposed): https://github.com/near/near-wallet/discussions/1153#discussion-31188

PR #1157, opened 2020-11-12
Clean up reconnect Ledger

Patrick1904:
Changes:
1. Delete `connectLedger` and use `addLedgerAccessKey` for **enabling** and **reconecting** the Ledger.

### DIFF
--- a/src/actions/account.js
+++ b/src/actions/account.js
@@ -277,14 +277,13 @@ export const {
     CLEAR_CODE: null
 })
 
-export const { getAccessKeys, removeAccessKey, addLedgerAccessKey, connectLedger, disableLedger, removeNonLedgerAccessKeys, getLedgerAccountIds, addLedgerAccountId, saveAndSelectLedgerAccounts, setLedgerTxSigned, clearSignInWithLedgerModalState, showLedgerModal } = createActions({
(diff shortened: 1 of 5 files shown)


Patrick1904:
@vgrichina any concerns with this or can we merge?

vgrichina:
@Patrick1904 I've reconnected Ledger to account recovered with seed phrase and the machine still had local signing key.  Is this expected behavior?

Patrick1904:
> @Patrick1904 I've reconnected Ledger to account recovered with seed phrase and the machine still had local signing key. Is this expected behavior?

@vgrichina yes I didn't change any logic between enabling and reconnecting the Ledger. Currently we leave the local signing key after first enabling Ledger too. However I can remove the local key if we want to do that?

vgrichina:
> Currently we leave the local signing key after first enabling Ledger too. 

@Patrick1904 I think we used to have a flow where we asked user whether they want to keep / remove other keys. Is this not happening anymore?

Patrick1904:
> > Currently we leave the local signing key after first enabling Ledger too.
> 
> @Patrick1904 I think we used to have a flow where we asked user whether they want to keep / remove other keys. Is this not happening anymore?

Yep that's correct. That still applies if the user chooses 'Delete all other keys'. But we don't do that when the user chooses to keep the keys, or reconnects the Ledger post initial setup.

vgrichina:
> or reconnects the Ledger post initial setup.

I see, then it shouldn't block this PR, but I feel like we need to reconsider this. Maybe we shouldn't break the flow and do something like proposed here:
https://github.com/near/near-wallet/issues/1182

PR #1161, opened 2020-11-13
Fix 2fa not showing as active on profile

Patrick1904:
For recovery methods that DO NOT have a `publicKey`, check that the methods `confirm` flag is `true`


kcole16 (approved):

mattlockyer (approved):
tested, works

Patrick1904:
Made some changes based on feedback from @vgrichina

PR #1164, opened 2020-11-16
Fix connection after seed phrase recovery

marcinbodnar:


Patrick1904:
Tested and works great.

@vgrichina, this fix solves the issue with Ledger not always being enforced (before page refresh) when enabling/reconnecting Ledger.

Patrick1904 (approved):

src/utils/wallet.js
```diff
@@ -806,6 +806,8 @@ class Wallet {
             signer: new nearApiJs.InMemorySigner(tempKeyStore)
         })
         
+        const connectionConstructor = this.connection
```
vgrichina, line 809: @marcinbodnar @mattlockyer I think a proper fix would be avoiding modifying `this.connection` in the first place

marcinbodnar: @vgrichina agree, this is just a hotfix from my side, that I think should be merged because it might affect users that will be working with wallet straight after account recovery. @mattlockyer can you handle it? Refactor `getAccount` to use custom connection is easy, but there is also `this` passed to `TwoFactor` which you know the most about.

vgrichina: @marcinbodnar yeah, discussing longer term fix with @mattlockyer

PR #1167, opened 2020-11-16
TEST: Create Account Flow on Testnet

mattlockyer:
This instance is to open the create account flow on testnet so @corwinharrell and others can test it.

It disables the developer account creation flow and instead you MUST fund the implicit account and claim your named account as you would do on mainnet with NO linkdrop.

### DIFF
--- a/src/utils/wallet.js
+++ b/src/utils/wallet.js
@@ -695,6 +695,7 @@ class Wallet {
         await this.validateSecurityCode(accountId, method, securityCode);
         await this.saveAccount(accountId, recoveryKeyPair);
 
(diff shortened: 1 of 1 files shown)


vgrichina:
@corwinharrell @mattlockyer I can send you some tokens for mainnet testing if you want

corwinharrell:
no worries @vgrichina I just needed to capture the current account funding flow and wanted to avoid creating unneeded accounts on mainnet

MaximusHaximus:
Closing this as it seems to be quite out-of-date by now :)

PR #1180, opened 2020-11-18
adding gh validators to validators list

mattlockyer:
pulls more validators (historic) from GH list:
`https://raw.githubusercontent.com/frol/near-validators-scoreboard/scoreboard-${networkId}/validators_scoreboard.json`

streamlines validator list to currentValidators only when user is staking

### DIFF
--- a/src/utils/staking.js
+++ b/src/utils/staking.js
@@ -280,6 +280,7 @@ export class Staking {
     async getValidators(accountIds) {
         const { current_validators, next_validators, current_proposals } = await this.provider.validators()
         const currentValidators = current_validators.map(({ account_id }) => account_id)
+        const nextValidators = current_proposals.map(({ account_id }) => account_id)
         
         if (!accountIds) {
             const rpcValidators = [...current_validators, ...next_validators, ...current_proposals].map(({ account_id }) => account_id)
@@ -301,6 +302,7 @@ export class Staking {
                     const validator = {
                         accountId: account_id,
                         current: currentValidators.includes(account_id),
+                        next: nextValidators.includes(account_id),
                         contract: await this.getContractInstance(account_id, stakingMethods)
                     }
                     const fee = validator.fee = await validator.contract.get_reward_fee_fraction()
--- a/src/utils/staking.js
+++ b/src/utils/staking.js
@@ -290,7 +290,7 @@ export class Staking {
                 ghValidators = (await fetch(`https://raw.githubusercontent.com/frol/near-validators-scoreboard/scoreboard-${networkId}/validators_scoreboard.json`).then((r) => r.json()))
                 .map(({ account_id }) => account_id)
             }
-            
+
             accountIds = [...new Set([...rpcValidators, ...ghValidators])]
                 .filter((v) => v.indexOf('nfvalidator') === -1 && v.indexOf(networkId === 'mainnet' ? '.near' : '.m0') > -1)
         }
@@ -308,9 +308,8 @@ export class Staking {
                     fee.percentage = fee.numerator / fee.denominator * 100
                     return validator
                 } catch (e) {
-                    console.log('error with', account_id)
                     if (!/No contract for account|cannot find contract code|wasm execution failed/.test(e.message)) {
-                        throw(e)
+                        throw e
                     }
                 }
             })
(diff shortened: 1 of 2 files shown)


vgrichina:
@mattlockyer I've tried testing this, but validator list seems to stay completely empty (using https://near-mainnet-staging-wallet-pr-1180.onrender.com/ in Safari):

Screen Shot 2020-11-22 at 12 31 33 AM

mattlockyer:
> @mattlockyer I've tried testing this, but validator list seems to stay completely empty (using https://near-mainnet-staging-wallet-pr-1180.onrender.com/ in Safari):

@vgrichina it's been fixed. It was networkId issue. Now I rely on RPC provider connection url.

I forgot the "gold standard" for network id introspection in the wallet. Is there a better method?

src/utils/staking.js
```diff
@@ -275,23 +277,39 @@ export class Staking {
...
+
+            // TODO use indexer - getting all historic validators from raw GH script .json
+            const networkId = this.provider.connection.url.indexOf('mainnet') > -1 ? 'mainnet' : 'testnet'
+            if (!ghValidators) {
+                ghValidators = (await fetch(`https://raw.githubusercontent.com/frol/near-validators-scoreboard/scoreboard-${networkId}/validators_scoreboard.json`).then((r) => r.json()))
+                .map(({ account_id }) => account_id)
+            }
+            
+            accountIds = [...new Set([...rpcValidators, ...ghValidators])]
+                .filter((v) => v.indexOf('nfvalidator') === -1 && v.indexOf(networkId === 'mainnet' ? '.near' : '.m0') > -1)
```
vgrichina, line 295: @mattlockyer why do we have to filter out here? Isn't scoreboard already filtered by network ID?

src/utils/staking.js
```diff
@@ -275,23 +277,39 @@ export class Staking {
...
+        const nextValidators = current_proposals.map(({ account_id }) => account_id)
+        
         if (!accountIds) {
-            const { current_validators, next_validators, current_proposals } = await this.provider.validators()
-            const allValidators = [...current_validators, ...next_validators, ...current_proposals]
-            accountIds = [...new Set(allValidators.map(({ account_id }) => account_id))]
+            const rpcValidators = [...current_validators, ...next_validators, ...current_proposals].map(({ account_id }) => account_id)
+
+            // TODO use indexer - getting all historic validators from raw GH script .json
+            const networkId = this.provider.connection.url.indexOf('mainnet') > -1 ? 'mainnet' : 'testnet'
```
vgrichina, line 288: `this.provider.connection.networkId`

src/utils/staking.js
```diff
@@ -275,23 +277,39 @@ export class Staking {
...
+                        current: currentValidators.includes(account_id),
+                        next: nextValidators.includes(account_id),
+                        contract: await this.getContractInstance(account_id, stakingMethods)
+                    }
                     const fee = validator.fee = await validator.contract.get_reward_fee_fraction()
                     fee.percentage = fee.numerator / fee.denominator * 100
                     return validator
                 } catch (e) {
-                    if (!/cannot find contract code|wasm execution failed/.test(e.message)) {
+                    console.log('error with', account_id)
```
vgrichina, line 311: makes sense to log error itself? Also probably should use `console.warn`. Should we log in case of expected errors however? Seem like only unexpected would be useful.

src/utils/staking.js
```diff
@@ -275,23 +277,39 @@ export class Staking {
...
+                    if (!/No contract for account|cannot find contract code|wasm execution failed/.test(e.message)) {
                         throw(e)
```
vgrichina, line 313: nit: `throw e`

src/utils/staking.js
mattlockyer: No. It includes a lot of legacy validators that we don't want to process. It would just mean another check against their account and additional RPC call.

src/utils/staking.js
mattlockyer: showing as undefined...?

vgrichina: ah, looks like `this.provider.connection` is a scam, not an actual `Connection` instance. I guess given it's in wallet, reasonable to use `NETWORK_ID` as well.

PR #1181, opened 2020-11-18
Redirect users with 0 balance to Profile page

marcinbodnar:


marcinbodnar:
@kcole16 @vgrichina 
It's working as described in the issue, so if an account was found but adding key fails due to not enough balance, the user will be redirected to `/profile` page. 

If there will be multiple accounts for seed phrase and at least one account will successfully add a key, the user will be logged in to that account. If all accounts fail, the user will be redirected to the profile.

I think that if a few multiple accounts are found we should show some kind of summary page, where all accounts will be listed with success or error information, and the user should be able to:
 - see what went wrong with which account
 - choose on which account he wants to be logged in. 

At this moment it will be confusing to the user if anything but success happens, and even on the success he still might be logged into a different account than would like to use.

I think it would be a nice upgrade but for sure nothing urgent.

marcinbodnar:
@kcole16 @vgrichina 
Accounts and phrases for tests:

```
light local edit glory basic chuckle like beauty science guard force system 
 - marcinh20.testnet - `LackBalanceForState` error
 - marcinh21.testnet - success
 - marcinh22.testnet - success


fade brother canvas sugar tone key sight enjoy soccer foot soda betray 
 - marcinh30.testnet - `LackBalanceForState` error
 - marcinh31.testnet - `LackBalanceForState` error
```

kcole16 (approved):

src/utils/wallet.js
```diff
@@ -841,20 +843,43 @@ class Wallet {
...
         }))
 
         this.connection = connectionConstructor
 
-        return {
-            numberOfAccounts: accountIds.length,
-            accountList: accountIds.flatMap((accountId) => accountId.account_id).join(', '),
+        if (!!accountIdsSuccess.length) {
+            await Promise.all(accountIdsSuccess.map(async ({ accountId, newKeyPair }, i) => {
+                if (i === accountIdsSuccess.length - 1) {
```
vgrichina, line 866: nit: this is not a blocker, but seems like logic can be somewhat simpler if we call `saveAccount` for every item in list and then call `selectAccount` for the last one after loop

marcinbodnar: Sure. Additionaly refactored the same thing in `saveAndSelectLedgerAccounts` and small refactor to `saveAndSelectAccount`. 

I'm not sure about `setAccountConfirmed`, which is part of `saveAndSelectAccount` that was previously needed for confirming if account was ever loaded (createAccount error handling) - but I'm not sure if it's needed with the latest changes for this flow. I will add an issue for it.

PR #1194, opened 2020-11-22
enforce semi

mattlockyer:
We should enforce this style. Partly introduced by my bad habits but can be resolved using eslint.

yarn fix now enforces and fixes missing semi-colons

**A LOT of files will be updated in this PR and should be merged with outstanding PRs.**


MaximusHaximus:
Closing as this is too stale to merge and we can re-regenerate it using `eslint --fix`.
However, we do want this enabled (Vlad wasted quite a bit of time on a hard to debug bug that was caused by no semicolon behaviour in a corner case a couple of months ago, this would have saved him the hassle if we had merged it.)

I'll make a new PR that adds this and `import/order` rule that also has organized imports codebase-wise in a single PR.

PR #1203, opened 2020-11-23
add refresh_staking_pool_balance call

mattlockyer:
call refresh_staking_pool_balance and get liquid before throwing not enough tokens error

for users unstaking their rewards:
https://github.com/near/core-contracts/tree/master/lockup#refresh-the-current-total-balance-on-the-staking-pool


vgrichina (approved):
got anecdotal report from user that this helps + otherwise doesn't look risky

PR #1228, opened 2020-11-25
small update to filter out 2fa methods from recoveryMethods

mattlockyer:
fixes when user doesn't confirm 2fa their 2fa method will disappear from profile

now it won't

### DIFF
--- a/src/actions/account.js
+++ b/src/actions/account.js
@@ -9,7 +9,6 @@ import {
     setKeyMeta, MULTISIG_MIN_PROMPT_AMOUNT
 } from '../utils/wallet'
 import { PublicKey, KeyType } from 'near-api-js/lib/utils/key_pair'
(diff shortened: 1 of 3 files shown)


Patrick1904 (approved):

vgrichina:
> fixes when user doesn't confirm 2fa their 2fa method will disappear from profile
> now it won't

@mattlockyer  hmm, does it work if not confirmed however? Or is it basically happening for every request?

mattlockyer:
> > fixes when user doesn't confirm 2fa their 2fa method will disappear from profile
> > now it won't
> 
> @mattlockyer hmm, does it work if not confirmed however? Or is it basically happening for every request?

It was happening if user makes 2FA request and then cancels.

2FA recovery methods don't use confirmed field anymore to be shown on profile.

PR #1237, opened 2020-11-30
Throw all error while recovering account with a seed phrase

marcinbodnar:


marcinbodnar:
@Patrick1904 so the problem with the user you mention and seed phrase recovery has to be that `addAccessKey` throws an error but it wasn't displayed, success ware displayed instead - now it should work well, the error will be displayed.

src/utils/wallet.js
```diff
@@ -870,6 +870,8 @@ class Wallet {
             if (lastAccount) {
                 store.dispatch(redirectTo(`/profile/${lastAccount.accountId}`, { globalAlertPreventClear: true }))
                 throw lastAccount.error
+            } else {
+                throw accountIdsError.reverse()[accountIdsError.length - 1].error
```
vgrichina, line 874: @marcinbodnar why call `.reverse()` and then take last element? Seems like this can be just `accountIdsError[0].error`?

marcinbodnar: @vgrichina I made it to be consistent with throwing `LackBalanceForState` error (first account) - but it's not necessary. Removed it.

vgrichina: @marcinbodnar BTW I just realized that before fix it would also use wrong order, as `.reverse()` is happening in-place

marcinbodnar: @vgrichina yeah, you're right

Issue #1248, opened 2020-12-01
Resend 2FA code button is not working


ihaid:
**Problem**
Nothing happens when you click the "resend" button on 2fa verify modal window
![image](https://user-images.githubusercontent.com/38665104/100765754-858a2680-3400-11eb-86bb-7fc99a161068.png)

**Expected Behavior**
Code must be sent again 

**Steps to reproduce**

1. Create an account
2. Enable 2FA
3. Go to /send-money
4. Send some Near
5. Wait for the first confirmation
6. Click on the "resend" button

vgrichina:
@ihaid please make sure the fix works ok

ihaid:
> @ihaid please make sure the fix works ok

@vgrichina I still see this issue on testnet

PR #1249, opened 2020-12-01
hotfix for resend 2fa codes

mattlockyer:
There was a mismatch between twoFactor.js and Account2FA from near-api-js.

Discovered via Sentry this morning.

Reduced code. e2e tested branch for resending 2fa code:
- deploying 2fa account
- sending 2fa tx

### DIFF
--- a/src/actions/account.js
+++ b/src/actions/account.js
@@ -199,11 +199,11 @@ export const {
         () => defaultCodesFor('account.initTwoFactor')
     ],
     REINIT_TWO_FACTOR: [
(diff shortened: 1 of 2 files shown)


vgrichina:
@mattlockyer do I understand correctly that this should fix https://github.com/near/near-wallet/issues/1248?

mattlockyer:
> @mattlockyer do I understand correctly that this should fix #1248?

I was unaware of this issue but this will fix it!

PR #1250, opened 2020-12-01
Fix 2fa enable disable

mattlockyer:
This rotates the localStorage key when user disables 2FA.

Requires: https://github.com/near/near-api-js/pull/462


vgrichina:
@mattlockyer needs merge with `master`

mattlockyer:
> @mattlockyer needs merge with `master`

Roger, done. Keep in mind it's still referencing https://github.com/near/near-api-js/pull/462 so package and lock might be different.

ihaid:
Sometimes I see this error in console 
image

vgrichina:
> Sometimes I see this error in console
> image

this one is hiding actual error. Unfortunately I messed up error polyfill setup for Firefox/Safari.

Fix is here: https://github.com/near/near-api-js/commit/56e45fb649d8d794d4a126726b353ef1e882993f

mattlockyer:
@ihaid can you try testing enable/disable today please? This has the hotfix #1280 so there should be no issues with 2FA accounts :crossed_fingers:

vgrichina:
@mattlockyer build on Netlify is broken

vgrichina:
@mattlockyer I've reverted as was breaking build (https://github.com/near/near-wallet/pull/1250/commits/0f91817df577af826c4a1555a9eface67cd37205). Also as I said before I don't think this approach is going to help.

mattlockyer:
> @mattlockyer I've reverted as was breaking build ([0f91817](https://github.com/near/near-wallet/commit/0f91817df577af826c4a1555a9eface67cd37205)). Also as I said before I don't think this approach is going to help.

What's the alternative?

vgrichina:
> What's the alternative?

I think there are such viable options:
- don't release until we don't have to do workaround
- release without workaround (accept risks)
- release with key rotation (accept risks, account is still recoverable)
- release with key rotation + per account backups

I'm not sure when `checkForPublicKeyWithRetries` as implemented could help:
1) it was called after successful `deployMultsig`. Why would it be needed then?
2) if the idea is to increase number of retries – we can increase it for sending transaction from near-api-js in general, no need for specific handing here
3) if you use high amount of retries (like 100 in your case) – user will get bored waiting for result and is likely e.g. to refresh the page breaking the implemented mitigation

Generally for everything we do in browser we shouldn't ever expect that any async function is guaranteed to finish. User can close page / refresh at any time (or hit network error, etc). We should be able to recover from half finished states.

mattlockyer:
I don't think this was quite ready. Still WIP I'll pull it into draft.

While we're here:

> it was called after successful deployMultsig. Why would it be needed then?
In case TX came back as timeout or failed buy actually succeeded
> if you use high amount of retries (like 100 in your case) – user will get bored waiting for result and is likely e.g. to refresh the page breaking the implemented mitigation
I am storing key in localStorage and then retrying after browser refresh
> release with key rotation + per account backups
Also added this

mattlockyer:
@vgrichina fairly substantial update. I know you're not fond of the approach but it checks some boxes.

1. user will deploy/disable multisig which includes a rotation to a new key
2. we don't know if the new key was deployed (even if the TX fails / timeout) since it could succeed on backend
3. so we put the new key in localStorage mapped to the accountId
4. we start to check for the new key using a timeout loop
5. if the user reloads the page, it will check that accountId and start the loop again
6. if the key is found it will update the wallet's localStorage key to the new key

Best case: this happens instantly after deploy/disable multisig
Bad case: the key update doesn't happen instantly, perhaps there's some failed UX since the user has an old key, but given the wide timeout window, they might not be able to sendSignedJson to CH etc...
Worst case: the retries end and no new key is added, the user is confused, reloads and the check for a new key starts over

???

vgrichina:
> I don't think this was quite ready. Still WIP I'll pull it into draft.

@mattlockyer the thing is we want to be able to send this PR to people who really want to disable 2FA right now

vgrichina:
> I know you're not fond of the approach but it checks some boxes.

@mattlockyer 2FA is already a dumpster fire as it is. Stuff barely works in normal scenarios. I really don't want to be responsible for the flaky stuff you push ignoring my recommendations.

mattlockyer:
> > I know you're not fond of the approach but it checks some boxes.
> 
> @mattlockyer 2FA is already a dumpster fire as it is. Stuff barely works in normal scenarios. I really don't want to be responsible for the flaky by design stuff you push ignoring my recommendations.

I am simply acknowledging that you prefer we:
> don't release until we don't have to do workaround

However, if you care to look, I think it takes into account some of your recommendations? 

> release with key rotation + per account backups

It's backing up the new key vs. the old key and then waiting to rotate the key on the frontend.

> We should be able to recover from half finished states.

This is also handled.

I don't think my recent effort on this branch warrants the dismissive commentary. I am trying to communicate an approach that I think has merit. If you disagree that's fine. If you think there are more pressing concerns, that's fine.

However, I am not finding anything constructive (or any specifically instructive recommendations) in your comments from today.

vgrichina:
@mattlockyer 

> I don't think my recent effort on this branch warrants the dismissive commentary. I am trying to communicate an approach that I think has merit. If you disagree that's fine. If you think there are more pressing concerns, that's fine.

2FA has been live for more than a quarter. For all of this time frame we have been repeatedly having issues with basic operations using 2FA:
- account recovery (most recent occurrence just today when I was testing latest release – this is still a production bug)
- app login
- signing app transactions

Throughout all of this period we are getting feedback that 2FA is shit from our users. People were getting stuck in the loop unable to recover their accounts for example. I also experienced this with one of my employee tokens account and it was not fun at all.

How would you call it if not "dumpster fire"? Like do you think doing this to token holders for months and months without much recourse is great success?  This is really not about this PR, this is about current state of 2FA in production. I constantly have to be responsible for it before community and Illia and it's never any positive feedback. 

I gave you constructive comments in https://github.com/near/near-wallet/pull/1250#issuecomment-758362310 as I did many times before. This is not the first time I'm asking you to abandon the loop to check for keys approach for this specific problem.

You are doing a lot of awesome stuff for NEAR. Nearnames is super cool and you've uncovered very important security bugs and I feel bad I end up being negative. I agree I'm not thankful enough about stuff you do well.

But unfortunately there were multiple times when you submit wallet / near-api-js PR where I can see that a large chunk of code just was clearly never executed (as it obviously fails). Or I just try doing basic smoke test and stuff doesn't work. This decreases trust and increases burden on me and other team members on actually making sure your stuff works. So if you also aren't following recommendations it doesn't help as it requires spending even more effort. 

There are quite a few reasons why checkAndAddNewKey would be flaky:
- it's spun up every time `getAccount` is called
- it can run for whole 6 minutes (for a case where key addition actually failed)

Just these 2 together mean that at any given moment we likely can have multiple concurrent long running non-cancellable background async jobs running in a wallet and messing with it's state. This is very unlikely to work in robust way and definitely isn't something I'd want to debug. Concurrency is typically very hard to debug, especially in production when you cannot easily repro locally. Good example is the "Confirm 2FA" message I'm still getting every now and then (instead of transaction confirmation) when trying to recover my account.

mattlockyer:
Some critical issues NOT related to my work:
- Account creation bug leading to **lost funds and names** on mainnet (found by me last summer)
- Modal popping, double send / stake bug leading to **lost funds on mainnet** (found by me)
- Nonce issue with re-used keys and potential attacks **on mainnet** (found by me)
- Key deletion gas bug with potential to **crash mainnet** (found by me)
- Deletion of 2FA method details in production database **on mainnet** (???)
- Account balances showing incorrectly **on mainnet** (???)
- Users unable to transfer lockup funds **on mainnet** (???)

I have appreciated all your reviews and respect the experience you bring to the process. It helped me reduce code complexity and dependencies several times.

However, I do not appreciate the following:

You jumped on this PR yesterday while it was still WIP, when there was **NO review requested** and started calling me out. When I tried to explain what I was working on, you begun a negative personal attack. You scoped your replies to only focus on the fact that a WIP branch wasn't ready for GA testing and declared that everything related to 2FA is a "dumpster fire" (I'm smart enough to read this as a proxy for my work). Your most recent comment, where you speak for the whole team, calls into question whether my work can be trusted. This is not the forum and barely passes for productive discourse.

It has also been particularly challenging the past 7 months to submit features or fixes, without a discussion, architecture guidance, whiteboard session or any specifications **prior to implementation** beyond some curt problem definitions and half-baked solutions littered around GH issues and chat channels. If not trial by review, then what? What is the strategy? I am tired of being forced into quagmires and then personally attacked for speaking up, proposing solutions, taking initiative and writing code.

Have I committed code with imperfections and missed test cases in the past? Of course! We all make mistakes. **See above!** Please feel free to point this out and I am happy to fix my mistakes.

However, implying that we should be infallible; that PRs should be perfect, fully tested and GA the moment code is committed is quite a ludicrous expectation. That's why we are a **team**, have QA, have CI and a have community.

Since Dec, review comments have been less helpful and not particularly healthy. I find them to be reactionary, conflating, personal and borderline malicious.

I do not think this communication is befitting of a teammate and colleague. 

As a result, I am thankful that my time on the wallet team is ending.

Outside of this PR:

You helped identify the race condition between nonces **yesterday**, thank you.

I attempted a fix here: https://github.com/near/near-api-js/pull/482 but closed the issue after reading your comments. I'm not interested in participating in this style of communication as mentioned above.

PS

I would be surprised if anyone can explain **exactly** what your recommendation is, to get disable 2FA out to GA, based on this comment:
https://github.com/near/near-wallet/pull/1250#issuecomment-758362310

PR #1252, opened 2020-12-02
Redirect back to login screen after importing account

mesquka:


vgrichina:
@marcinbodnar I think you were doing some similar changes before, don't remember what's the latest on them

marcinbodnar:
@vgrichina yeah, I assume it's not about `importing` but `creating` account @mesquka ?

mesquka:
This change is specifically for `importing` accounts. The usecase is to allow users that are signing in on a new device to import accounts without abandoning the sign-in flow.

Adding the same flow to `createing` wallets probably makes sense as well, but the funding wallet step takes the user out of the sign-in flow so wasn't as high priority to me.

marcinbodnar:
@mesquka ok, but `EnableTwoFactor.js` is used only for creating account flow. Why changes to this component?

marcinbodnar:
@mesquka 
So it's already implemented for both creating and importing account and it should work well.
It's happening starting from here:
https://github.com/near/near-wallet/blob/master/src/components/Routing.js#L113
with `handleRefreshUrl`, `handleRedirectUrl ` and `handleClearUrl` actions.

The only thing that I can see is not working is skipping 2fa while creating an account.

mesquka:
> @mesquka ok, but `EnableTwoFactor.js` is used only for creating account flow. Why changes to this component?

Sorry, yep that's correct, was replying with outdated information (haven't had my morning coffee yet).

src/components/accounts/two_factor/EnableTwoFactor.js
```diff
@@ -166,7 +176,7 @@ export function EnableTwoFactor(props) {
                     >
                         <Translate id={`button.continue`} />
                     </FormButton>
-                    <FormButton className='link' linkTo='/profile'><Translate id='button.skip' /></FormButton>
+                    <FormButton className='link' onClick={() => skipUrl()}><Translate id='button.skip' /></FormButton>
```
marcinbodnar, line 179: In this case you just need use `dispatch(redirectToApp('/profile'))` it will redirect to `/login`

mesquka:
> @mesquka
> So it's already implemented for both creating and importing account and it should work well.
> It's happening starting from here:
> https://github.com/near/near-wallet/blob/master/src/components/Routing.js#L113
> with `handleRefreshUrl`, `handleRedirectUrl ` and `handleClearUrl` actions.
> 
> The only thing that I can see is not working is skipping 2fa while creating an account.

Redirecting after creating an account wasn't working, hence the changes.

src/components/login/Login.js
```diff
@@ -52,7 +52,9 @@ class Login extends Component {
     }
 
     redirectCreateAccount = () => {
-        this.props.history.push('/create')
+        const { success_url } =  this.props.account.url
+        sessionStorage.setItem("success_url", success_url)
+        this.props.history.push(`/create`)
```
marcinbodnar, line 57: these changes are unnecessary, it's been already done here:
https://github.com/near/near-wallet/blob/master/src/components/Routing.js#L113

marcinbodnar (changes requested):

marcinbodnar:
@mesquka 
> Redirecting after creating an account wasn't working, hence the changes.

Just tested with recovery seed phrase - seems to work well on testnet.

src/components/accounts/two_factor/EnableTwoFactor.js
```diff
@@ -98,9 +98,19 @@ export function EnableTwoFactor(props) {
         handleDeployMultisig()
     }
 
+    const skipUrl = () => {
```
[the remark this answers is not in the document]
marcinbodnar: if you wanted to redirect back to App, the better idea would be to redirect to `/login` page - then the user is able to cancel the operation and will be redirected back to App.

mesquka:
Closing PR, will reopen later if needed.

Issue #1175, opened 2020-11-17
Add functionCall args to /sign detail page


kcole16:
### Overview
If a user is signing a transaction with a `functionCall`, and doesn't have 2FA or Ledger enabled, they have no way of viewing the arguments for the `functionCall`. Since we have access to these from the transaction, we should display them on the `/sign` detail page.

### Acceptance Criteria
On `/sign`
- [ ] User can view the arguments for any `functionCall` actions after selecting "More Information"

<img width="418" alt="Screen Shot 2020-11-17 at 12 20 08 PM" src="https://user-images.githubusercontent.com/5624527/99443396-9e0e2300-28cf-11eb-8fe0-339f7bbef327.png">

PR #1257, opened 2020-12-03
Add functionCall args to /sign detail page (#1175)

marcinbodnar:


kcole16:
@marcinbodnar Noticed this when testing with berryclub.io on the staging link:
Screen Shot 2020-12-07 at 12 49 40 PM

marcinbodnar:
@kcole16 check now please.

kcole16 (approved):

kcole16:
@marcinbodnar Looks good!

vgrichina (changes requested):
@marcinbodnar looks good overall, but need to be prepared to handle non-JSON args. Can fail JSON parsing – makes sense to assume it's arbitrary binary then. Can use something like https://github.com/bma73/hexdump-js for binary, or just show a warning about binary args for now.

marcinbodnar:
@vgrichina got it, do we have any FunctionCall with binary args to test?

marcinbodnar:
@vgrichina this will work?

vgrichina:
> @vgrichina this will work?

I don't think this will work, I think `args` is always binary before you parse it from JSON, so basically you have to handle error when parsing

I think @chadoh should have an example sending binary args with bridge. Is there anything live we can use for testing wallet?

vgrichina:
@marcinbodnar try https://gitpod.io/#snapshot/8f156a4b-9376-41f7-995e-27a79e23b9a5 – it should do non-JSON method call. You need to attach some tokens as donation to get redirect to wallet.

marcinbodnar:
@vgrichina now it should work fine

vgrichina (approved):

Issue #1214, opened 2020-11-24
Strange records in Authorized apps section


ihaid:
**Problem**
I have several accounts in the wallet, for account **kilo.testnet** I add an authorized app, for account **qw121.testnet** I don`t have any authorized app and when I switch between them I see wrong records in the authorized apps section on the summary tab.
After refreshing the page it is empty, as it should be.
![image](https://user-images.githubusercontent.com/38665104/100124401-4732bb80-2e84-11eb-985d-eb9326f91d77.png)
Same behavior on _/authorized-apps_
![image](https://user-images.githubusercontent.com/38665104/100125271-43ebff80-2e85-11eb-9f76-0f306bf9ad6c.png)

**Steps to reproduce**

You can try to reproduce it with these 2 accounts by adding them and switching between:

- `practice gap seminar rib fatigue raw siege vivid equal camp slide bench `

- `suffer mandate oxygen alcohol vacant hat prefer tent peace ticket vacuum normal `

Patrick1904:
@vgrichina @marcinbodnar @kcole16, do you have any idea what's happening here? I've seen this myself and not quite sure why/when this happens.

marcinbodnar:
@Patrick1904 apparently, this shows a list of transactions - in this case `kilo.testnet` is a tx receiver. Also, it is shown only after switching account (it's not visible right after page refresh). Let me handle that.

PR #1258, opened 2020-12-03
Fix getting access keys while account load (#1214)

marcinbodnar:


marcinbodnar:
@mattlockyer @vgrichina 
The problem was, that while switching account's from 2fa account to non-2fa account, new account was retrieving accessKeys from previous account, due to this condition:
https://github.com/near/near-wallet/blob/master/src/utils/wallet.js#L595
At this point `this.twoFactor.isEnabled()` was returning true, because twoFactor still had data from previous account.

kcole16 (approved):

PR #1280, opened 2020-12-10
hotfix: refactor twoFactor with getAccount and account actions

mattlockyer:
`this.twoFactor.isEnabled` is returning true during new account creation flow
This means a new account is thrown into a 2FA confirmation based on the previous 2FA account

Fix addresses this by passing accountId to `this.twoFactor.isEnabled`


mattlockyer:
@kcole16 still getting double confirmation or was that a cache issue?

mattlockyer:
@ihaid find anything else?

ihaid:
> @kcole16 still getting double confirmation or was that a cache issue?

I got it just now after send-money

mattlockyer:
@kcole16 can you try in incognito, Igor has not been able to repro ~10 times in incognito, which leads me to think it's a caching issue

kcole16:
@mattlockyer Still happening in incognito. I get multiple messages both when initially setting up 2FA, and while transferring.

src/actions/account.js
```diff
@@ -161,6 +162,14 @@ export const signInWithLedgerAddAndSaveAccounts = (accountIds) => async (dispatc
 
 const defaultCodesFor = (prefix, data) => ({ successCode: `${prefix}.success`, errorCode: `${prefix}.error`, prefix, data})
 
+const twoFactorMethod = async (method, wallet, args) => {
```
vgrichina, line 165: nit: passing method name as a string breaks code navigation, so better to avoid. Like if I wanted to generalize smth similar I'd do:

```js
withAccount(account => account.sendCode())
```

```js
async function withAccount(fn) {
   const account = await wallet.getAccount(wallet.accountId)
   await fn(account)
}
```

src/utils/wallet.js
```diff
@@ -810,15 +780,17 @@ class Wallet {
...
+            let recoveryKeyIsFAK = false
             // check if recover access key is FAK and if so add key without 2FA
-            if (has2fa) {
+            if (await TwoFactor.has2faEnabled(account)) {
                 const accessKeys = await account.getAccessKeys()
-                const recoveryAccessKey = accessKeys.find(({ public_key }) => public_key === publicKey)
-                if (recoveryAccessKey.access_key.permission && recoveryAccessKey.access_key.permission === 'FullAccess') {
+                recoveryKeyIsFAK = accessKeys.find(({ public_key, access_key }) => 
+                    public_key === publicKey &&
+                    access_key.permission &&
```
vgrichina, line 790: nit: this check isn't needed, `access_key` always has `permission`

src/utils/wallet.js
```diff
@@ -810,15 +780,17 @@ class Wallet {
...
-            const has2fa = await this.twoFactor.isEnabled()
             let account = await this.getAccount(accountId)
+            let recoveryKeyIsFAK = false
             // check if recover access key is FAK and if so add key without 2FA
-            if (has2fa) {
+            if (await TwoFactor.has2faEnabled(account)) {
                 const accessKeys = await account.getAccessKeys()
-                const recoveryAccessKey = accessKeys.find(({ public_key }) => public_key === publicKey)
-                if (recoveryAccessKey.access_key.permission && recoveryAccessKey.access_key.permission === 'FullAccess') {
+                recoveryKeyIsFAK = accessKeys.find(({ public_key, access_key }) =>
```
vgrichina, line 788: nit: best to be explicit about boolean conversion with `!!`

src/utils/wallet.js
```diff
@@ -831,7 +803,7 @@ class Wallet {
             const newKeyPair = KeyPair.fromRandom('ed25519')
             
             try {
-                await this.addAccessKey(accountId, accountId, newKeyPair.publicKey, fromSeedPhraseRecovery)
+                await this.addAccessKey(accountId, accountId, newKeyPair.publicKey, fromSeedPhraseRecovery, '', recoveryKeyIsFAK)
```
vgrichina, line 806: @mattlockyer not sure I understand the logic. Do we basically add full access key if recovering through seed phrase? Or do I misunderstand something?

mattlockyer: This means that when you recover with seed phrase we use the seed phrase full access key and not a 2FA tx to add the key

Means seed phrase recovery method can add key without going through 2FA flow

PR #1283, opened 2020-12-14
Fix sending to implicit account ID

Patrick1904:


Patrick1904:
@frol I've tested using the account ID you provided on Slack. Please let me know if you see anything weird:)

PR #1289, opened 2020-12-14
Fix crash when connecting Ledger

vgrichina:
Introduced by missing await here https://github.com/near/near-wallet/pull/1171/files


mattlockyer:
Do we have render instances for this PR?

vgrichina:
> Do we have render instances for this PR?

we have Netlify instances #

PR #1292, opened 2020-12-17
Remove setFormLoader from everywhere

vgrichina:
It is something that makes it hard to have consistent state for `formLoader` and so breaking all logic to disable forms, etc

### DIFF
--- a/src/components/accounts/AccountFormAccountId.js
+++ b/src/components/accounts/AccountFormAccountId.js
@@ -109,9 +109,6 @@ class AccountFormAccountId extends Component {
         
         handleChange(e, { name, value })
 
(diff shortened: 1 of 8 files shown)


src/components/accounts/AccountFormAccountId.js
```diff
@@ -109,9 +109,6 @@ class AccountFormAccountId extends Component {
         
         handleChange(e, { name, value })
 
-        !this.props.formLoader && this.checkAccountIdLength(value) && this.props.setFormLoader(true)
-        this.props.formLoader && !this.checkAccountIdLength(value) && this.props.setFormLoader(false)
```
marcinbodnar, line 113: @vgrichina this is necessary - without that `checking...` does not appear correctly, if we don't want to use it we need to figure out some replacement. I can handle it after #1108  will be merged.

PR #1294, opened 2020-12-17
Keyboard related fixes to Send, Stake, Recovery, 2FA

Patrick1904:
- Guard against attempts to 're-submit' action related to: send, stake, recovery, 2FA. Any click or key press (enter).
- Guard against submitting incomplete forms or while things are loading
- Clean up and use dispatch vs history.push
- Add missing await(s)
- Show globalAlert for incorrect code during 2FA setup

Note: most fixes above are for **Safari** browser.

### DIFF
--- a/src/components/accounts/SetupSeedPhrase.js
+++ b/src/components/accounts/SetupSeedPhrase.js
@@ -27,7 +27,8 @@ class SetupSeedPhrase extends Component {
         enterWord: '',
         wordId: null,
         requestStatus: null,
-        successSnackbar: false
+        successSnackbar: false,
+        submitting: false
     }
 
     componentDidMount = () => {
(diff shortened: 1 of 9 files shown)


src/components/accounts/SetupSeedPhrase.js
```diff
@@ -166,6 +176,7 @@ class SetupSeedPhrase extends Component {
                                             formLoader={this.props.formLoader}
                                             requestStatus={this.state.requestStatus}
                                             globalAlert={this.props.globalAlert}
+                                            submitting={this.state.submitting}
```
vgrichina, line 179: @Patrick1904 should few just pass `formLoader={this.props.formLoader || this.state.submitting}` here? Basically seems to me that this only needs to be handled once upstream vs passing downstream everwhere.

src/components/send/SendContainer.js
```diff
@@ -50,7 +50,7 @@ export function SendContainer({ match }) {
     const [success, setSuccess] = useState(null)
     const amountAvailableToSend = new BN(balance.available).sub(new BN(parseNearAmount(WALLET_APP_MIN_AMOUNT)))
     const sufficientBalance = !new BN(parseNearAmount(amount)).isZero() && (new BN(parseNearAmount(amount)).lte(amountAvailableToSend) || useMax) && isDecimalString(amount)
-    const sendAllowed = ((requestStatus && requestStatus.success !== false) || id.length === 64) && sufficientBalance && amount && !formLoader
+    const sendAllowed = ((requestStatus && requestStatus.success !== false) || id.length === 64) && sufficientBalance && amount && !formLoader && !success
 
     onKeyDown(e => {
```
vgrichina, line 55: @Patrick1904 I wonder if handling key down specifically can be giving us more problems than onsubmit. Or maybe just different kind of problems.

Patrick1904: It should be solid after I added on `!formLoader && !success` to `sendAllowed`. I've tested it thoroughly and haven't been able to get any weird behavior

Patrick1904: `onSubmit` was actually the cause of most of the issues in Safari, even though I disabled the submit button. Also, we need the `onKeyDown` since the user isn't focused on any input once the confirmation modal shows

PR #1300, opened 2020-12-21
fix check validators pass accountId to wallet, cache account instance

mattlockyer:


mattlockyer:
Should be able to see validator list on mainnet now

src/utils/staking.js
```diff
@@ -295,21 +295,24 @@ export class Staking {
                 .filter((v) => v.indexOf('nfvalidator') === -1 && v.indexOf(networkId === 'mainnet' ? '.near' : '.m0') > -1)
         }
 
+        const account = await this.wallet.getAccount(this.wallet.accountId)
+
         return (await Promise.all(
             accountIds.map(async (account_id) => {
+                if (!account_id) return
```
vgrichina, line 302: why would we have empty account_id in the list?

src/utils/staking.js
```diff
@@ -295,21 +295,24 @@ export class Staking {
...
-                        contract: await this.getContractInstance(account_id, stakingMethods)
+                        contract: await this.getContractInstance(account_id, stakingMethods, account)
                     }
                     const fee = validator.fee = await validator.contract.get_reward_fee_fraction()
                     fee.percentage = fee.numerator / fee.denominator * 100
                     return validator
                 } catch (e) {
                     if (!/No contract for account|cannot find contract code|wasm execution failed/.test(e.message)) {
-                        throw e
+                        console.error(e)
```
vgrichina, line 315: `.warn` is a better fit as it's not a breaking error most likely

src/utils/staking.js
```diff
@@ -486,10 +489,13 @@ export class Staking {
     Helpers
     ********************************/
 
-    async getContractInstance(contractId, methods) {
+    async getContractInstance(contractId, methods, account) {
```
vgrichina, line 492: @mattlockyer seems like best for us to remove this method altogether, as most use cases should use `.getAccount()` from wallet (otherwise would fail with 2FA / lockup)

vgrichina: what I mean is `await (await new nearApiJs.Account(this.wallet.connection, contractId)).state()` is completely redundant

vgrichina: it would be basically: 

```
const contract = new Contract(await this.wallet.getAccount(accountId), contractId, contractMethods)
```

src/utils/staking.js
mattlockyer: I'll revert was from before, I thought validator_id was undefined

src/utils/staking.js
mattlockyer: I don't know how we would get the contract instance of validators then?
we need to calc fee percentage from contract
```
const fee = validator.fee = await validator.contract.get_reward_fee_fraction()
```

kcole16 (approved):

src/utils/staking.js
```diff
@@ -421,6 +425,9 @@ export class Staking {
         } else {
             lockupId = getLockupAccountId(accountId)
         }
+        if (this.wallet.accountExists(lockupId)) {
```
vgrichina, line 428: @mattlockyer this looks like never tested. Missing `await`. Also missing `!`. Basically seems like it would always throw as `accountExists` would always return a not null promise.

src/utils/staking.js
vgrichina: @mattlockyer I meant we might just inline it when making a call, as you anyway are passing different `account` instance.

but also you generally can do `await account.viewFunction(validator.accountId,  'get_reward_fee_fraction')`  in cases when you only need a one off

vgrichina:
Made minimal fix here https://github.com/near/near-wallet/pull/1310/files

PR #1303, opened 2020-12-21
Update UI for implicit account creation

Patrick1904:
**The goal of this UI update is to more clearly convey that the temporary funding address is indeed temporary and that the user should use their Account ID for any future NEAR operations.**

Follow up TODO's (separate PR's):
1. Add progress bar to create account flow
2. Update [design](https://www.figma.com/file/y1kyUHVr11lUmDkDSzMKU8/Create-New-Account?node-id=91%3A423) of recovery options on `/set-recovery`.


Patrick1904:
@vgrichina, once the account is created, the modal closes and there's a couple of seconds delay before the router redirects to the success page. I've tried to add different state flags etc to keep the modal open until redirect happens, but doesn't seem to help.

I set [this](https://github.com/near/near-wallet/pull/1303/files#diff-f48dfb66ac301888e066f197e5967502aa841139af9ab0528b98cb2f27183d9bR82) flag `creatingAccount` to `true` before, and never set it to `false`, expecting it to stay `true` until the router redirects to the next page.

I'm trying to avoid adding on the `formLoader` flag for showing the modal in case there's a pending action in progress once the user lands on `/fund-create-account`, which would trigger the 'account funded modal' right away.

Any idea what might be the issue?

Patrick1904:
@ihaid, can you please do a round of testing on this updated account creation flow? Thank you

ihaid (approved):

Patrick1904:
@corwinharrell I excluded the 'progress bar' because depending on the route the user takes at the recovery setup step, the current and total number of steps can change. Probably not a good idea to change the step/total mid flow.

I do think that a progress bar improves the UX in general, but I'm not sure how we can solve for the step/total issue.

kcole16 (approved):

Issue #1308, opened 2020-12-22
400+ RPC requests until staking page is ready


ilblackdragon:
When opening the staking page for specific pool - https://wallet.near.org/staking/buildlinks.poolv1.near/stake, it takes 400 RPC requests to get to the state when "Amount" is editable.

<img width="581" alt="Screen Shot 2020-12-21 at 7 21 36 PM" src="https://user-images.githubusercontent.com/175486/102845176-cf43ac80-43c1-11eb-8979-5873594e9005.png">

mattlockyer:
Working on a PR here to improve performance of /staking while we only load everything once when the user navigates to the route and again when the user performs a staking action, it's a lot of calls.
https://github.com/near/near-wallet/pull/1314
Couple of notes:
1314 removes an extra call to getValidators that results in all validators being pulled (reduces RPC calls by 53 on testnet and probably > 60 on mainnet)
We can bring down RPC calls even more when we use the new indexer. Here's why:

Currently, we have to make at least 1 RPC call across all validators to see if the user even has a balance with the validator. Only if they do, we make subsequent calls.

We can replace another > 60 RPC calls on mainnet because we'll have all this info from indexer. **Already done here:**
https://github.com/near/near-wallet/pull/1273

Was waiting on reviews and time for all this to be merged.

https://github.com/near/near-wallet/pull/1273 relies on https://github.com/near/near-contract-helper/pull/288

mattlockyer:
Can confirm RPC requests cut down significantly by https://github.com/near/near-wallet/pull/1273

PR #1314, opened 2020-12-22
staking perf improvement, check allValidators once per account

mattlockyer:


mattlockyer:
Fixes: https://github.com/near/near-wallet/issues/1308

PR #1317, opened 2020-12-23
Hotfix: Fix liquid owners balance calculation

vgrichina:


src/utils/account-with-lockup.js
```diff
@@ -128,7 +128,7 @@ async function getAccountBalance() {
             : totalBalance.sub(BN.max(unreleasedAmount, LOCKUP_MIN_BALANCE))
 
         const lockedAmount = totalBalance.sub(ownersBalance)
-        const liquidOwnersBalance = BN.max(new BN(0), ownersBalance.sub(stakedBalance))
+        const liquidOwnersBalance = BN.min(ownersBalance, new BN(lockupBalance.total))
```
mattlockyer, line 131: I think this fixes something from before, but what about staking rewards earned by the lockup?

mattlockyer (approved):
Should probably see if there's a proper way to tease out the liquid staking rewards portion.

src/utils/account-with-lockup.js
vgrichina: @mattlockyer staking rewards are part of `stakedBalance` above fetched from staking pool directly (and so are counted in `ownersBalance`).

PR #1318, opened 2020-12-23
Hotfix: Take hardcoded lockup storage amount into account

vgrichina:


vgrichina:
@evgenykuzyakov may you take a look if I caught all edge cases?

Issue #1330, opened 2020-12-30
Newly created accounts with Ledger won't recover sometimes


kcole16:
**Note**
This might be due to me using the same Ledger to create and fund multiple accounts (using the same funding address), at least one of which I kept as just the implicit account (never created the .near account)

**Problem**
After creating and funding a new account with Ledger as the only key, account recovery doesn't work.

**Expected Behavior**
Account recovery should work.

**Steps to reproduce**

1. Go to /create and choose Ledger as the recovery method
2. Fund the account with 1.1 NEAR
3. Finish creating the account (deleting the funding account)
4. Try to login in incognito: https://wallet.near.org/sign-in-ledger
5. Observe error

Error message is showing that the implicit account doesn't exist (which is true).

In below message, `c006...05824` was the funding account.

```
Uncaught (in promise) Error: [-32000] Server error: account c00604c6988da8767e178347a613f063e4f2d3eb4b72ce3aa285d213ccd05824 does not exist while viewing
    at d.sendJsonRpc (json-rpc-provider.js:163)
    at async d.query (json-rpc-provider.js:70)
    at async p.fetchState (account.js:50)
    at async p.state (account.js:58)
```

mattlockyer:
@kcole16 looks like it's erroring on the implicit account that once had the ledger key.

It finds the implicit and the named account (I'm assuming) during the indexer call, then throws and never gets to try the named account.

kcole16:
@marcinbodnar Can you take a look at this?

PR #1334, opened 2021-01-04
Handle non existing account for Sign in with Ledger (#1330)

marcinbodnar:


kcole16 (approved):

vgrichina:
@marcinbodnar looks like similar issue exists for seed phrase recovery as well, reproduced when working on:
https://github.com/near/near-wallet/pull/1340

I think the solution to both is to update near-contract-helper indexer query to exclude removed keys, though I guess I need to double check what happens with keys which are added back.

marcinbodnar:
@vgrichina 
> I think the solution to both is to update near-contract-helper indexer query to exclude removed keys

For sure, the best idea if it's technically possible.

I'm not able to reproduce it for Seed Phrase.

vgrichina:
> I'm not able to reproduce it for Seed Phrase.

do you mean that when you delete accounts they aren't picked up for seed phrase recovery anymore?

marcinbodnar:
@vgrichina I tried scenario like for ledger, when you claim account (implicit is deleted) and implicit account still appears in CH

vgrichina:
> @vgrichina I tried scenario like for ledger, when you claim account (implicit is deleted) and implicit account still appears in CH

I see, in my test I was deleting original account explicitly and then reusing same seed phrase for different accounts

Looks like we have to merge this for mitigation right now anyway and when I'm 100% sure about new indexer for explorer
we have a fix here:
https://github.com/near/near-contract-helper/pull/304

PR #1340, opened 2021-01-06
Add very basic e2e test for seed phrase recovery

vgrichina:
This also includes automatic alerting on PagerDuty

Currently limited to vlad@near.org so that it is automatically Acknowledged.
If runs stable for few days we can update to use oncall.


vgrichina:
@chefsale PTAL in terms of creating PagerDuty incidents

chefsale (approved):
LGTM, can we discuss only other e2e tests and alerting for wallet, happy to help on implementing these once we know what would they be.

e2e-tests/index.js
```diff
@@ -0,0 +1,109 @@
...
+    const pd = api({ token: process.env.PAGERDUTY_API_KEY });
+
+    console.log('Creating incident on PagerDuty')
+    await pd.post('/incidents', {
+        data: {
+            incident: {
+                type: "incident",
+                title: "wallet e2e-tests failure",
+                service: {
+                    id: "PJ9TV6C", // wallet
```
mhalambek, line 39: Could you put in check if there is an active incident for this, so you don't trigger new incident if there is already an active one on pagerduty.

Also on other note we have similar use case for this e2e tests with bridge. So we could potentially reuse this, but too early to tell for now.

vgrichina: @mhalambek I think we just need to solve flakiness (so that there aren't false incidents) vs check for whether there is active incident. I'm working to fix the underlying issues.

chefsale: The problem still is that everytime it runs it opens a new incident for the same issue, not flakiness. If there is the same issue running for 30 minutes and the cron job runs every 5 minutes, 6 incidents will be created which is spamming too much. Pagerduty doesn't know that it is the same incident or not. We need to implement a check which lists all the current open incidents and check if this one is present if not trigger.

vgrichina: Gonna change this in follow-up, for now just added flag to enable alerts explicitly

PR #1338, opened 2021-01-05
Release 2021-01-05

vgrichina:
### Release Notes
- Activity feed displays all transactions for 2FA accounts
- The role of the one-time use funding account is explained during account creation
- Fix Ledger issue with deleted and recreated implicit accounts

PR #1343, opened 2021-01-11
Create w/ implicit: fix delay

Patrick1904:


Patrick1904:
@kcole16 @ihaid, this fixes the delay that happened after creating with implicit account. Can you please test it out and confirm that it behaves as expected.

Patrick1904:
@vgrichina would be great if we can include this in https://github.com/near/near-wallet/pull/1338 to improve the implicit account UX.

vgrichina:
@Patrick1904  @kcole16 @ihaid I've merged, let's make sure to test on staging wallet

PR #1344, opened 2021-01-12
Disable 2FA (pre-release version)

vgrichina:
Should be safe enough to use to disable 2FA to remove remaining funds from accounts.

### DIFF
--- a/src/components/profile/two_factor/TwoFactorAuth.js
+++ b/src/components/profile/two_factor/TwoFactorAuth.js
@@ -74,9 +74,7 @@ const TwoFactorAuth = ({ twoFactor, history }) => {
                             </div>
                             <div className='detail'>{twoFactor.detail}</div>
                         </div>
-                        {
-                            false && <FormButton onClick={() => setConfirmDisable(true)} className='gray-red'><Translate id='button.disable'/></FormButton>
-                        }
+                        <FormButton onClick={() => setConfirmDisable(true)} className='gray-red'><Translate id='button.disable'/></FormButton>
                     </div>
                     <div className='bottom'>
                         <span className='color-green'>
--- a/src/utils/twoFactor.js
+++ b/src/utils/twoFactor.js
@@ -16,6 +16,8 @@ export const MULTISIG_CONTRACT_HASHES = process.env.MULTISIG_CONTRACT_HASHES || 
     '55E7imniT2uuYrECn17qJAk9fLcwQW4ftNSwmCJL5Di',
 ];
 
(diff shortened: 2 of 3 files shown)


Patrick1904:
@vgrichina @mattlockyer I noticed while testing this that the "Active since XX" on `/profile` doesn't update in CH when I re-enable 2FA.
Screen Shot 2021-01-29 at 5 38 27 PM

mattlockyer:
I think contract helper is using some default `createdAt` / `updatedAt` timestamps (not sure of the exact name) in the model.

2FA methods are stored as recoveryMethod models in CH.
https://github.com/near/near-contract-helper/blob/master/models/recovery-method.js

@vgrichina if the user is trying to re-enable, `requestId` will be -1. I think we'd have to manually update `createdAt`?

Patrick1904:
> I think contract helper is using some default `createdAt` / `updatedAt` timestamps (not sure of the exact name) in the model.
> 
> 2FA methods are stored as recoveryMethod models in CH.
> https://github.com/near/near-contract-helper/blob/master/models/recovery-method.js
> 
> @vgrichina if the user is trying to re-enable, `requestId` will be -1. I think we'd have to manually update `createdAt`?

@mattlockyer yes I think any time user enables 2FA, we can just override current `createdAt`

vgrichina:
I've seen the requestId = -1 thing when I try to recover 2FA account. Sometimes it won't create multisig request but send e-mail (with wrong content).

mattlockyer:
> I've seen the `requestId = -1` thing when I try to recover 2FA account. 

Yes that is another issue unrelated to the date.

Attempted fix here: https://github.com/near/near-api-js/pull/482

If you'd like to fix the date issue in CH you would update the necessary date fields when `requestId == -1` in CH.

Patrick1904:
@vgrichina @MaximusHaximus I think this PR may have gotten too out of sync with contract helper because I'm not able to recover any 2FA account anymore using the staging link: https://deploy-preview-1344--near-wallet-staging.netlify.app/

I just keep seeing in console: `Invalid security code. Please try again.`

vgrichina:
> @vgrichina @MaximusHaximus I think this PR may have gotten too out of sync with contract helper because I'm not able to recover any 2FA account anymore using the staging link: https://deploy-preview-1344--near-wallet-staging.netlify.app/
> 
> I just keep seeing in console: `Invalid security code. Please try again.`

yeah, this PR needs to be rebased + use latest version of near-api-js

MaximusHaximus:
Key rotation no longer necessary due to changes in nearcore

PR #1377, opened 2021-01-21
Allow to pass metadata with transaction to sign

vgrichina:
Needs this 2 PRs to work:
https://github.com/near/near-contract-helper/pull/347
https://github.com/near/near-api-js/pull/529

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "fetch-send-json": "0.0.2",
     "js-sha256": "^0.9.0",
     "mixpanel-browser": "^2.41.0",
-    "near-api-js": "^0.38.0",
+    "near-api-js": "near/near-api-js#2fa-update",
     "near-ledger-js": "^0.1.2",
     "near-seed-phrase": "^0.1.0",
     "prop-types": "^15.7.2",
--- a/src/actions/account.js
+++ b/src/actions/account.js
@@ -81,9 +81,9 @@ export const handleRefreshUrl = () => (dispatch, getState) => {
             dispatch(refreshUrl(loadState()))
         }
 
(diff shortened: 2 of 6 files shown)


vgrichina:
Looks like this almost works besides:

>         const transactionHashes = await dispatch(signAndSendTransactions(transactions, accountId))

`transactionHashes` need to be propagated upstream from `signAndSendTransactions`

"Close" button on success screen shouldn't call `handleDeny`?

jimmy3dita:
From product meeting: this is a blocker for the bridge launch, needed in the next release.

abacabadabacaba:
A couple of remarks:

1. As I understand, the only purpose of `meta` parameter is to pass some data to the callback URL. In this case, the name is confusing. I would suggest to call it `state`, which is the name used for a parameter with a similar function in widely used [OAuth 2.0](https://tools.ietf.org/html/rfc6749#section-4.1.1) and [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) protocols.
  Also, any app not using this parameter may be vulnerable to CSRF. What if a malicious website redirects the user to the app's callback URL?
2. To improve security, I would pass the parameters (both to and from the wallet) in `location.hash` rather than `location.search`. There are two reasons for this. The first is that it is better to avoid sending potentially sensitive values to the server, where they may be logged. The second reason is to reduce the chance of leaking the values in [`Referer` header](https://developer.mozilla.org/en-US/docs/Web/Security/Referer_header:_privacy_and_security_concerns).

vgrichina:
@abacabadabacaba moved into separate issue:
https://github.com/near/near-wallet/issues/1546

neither of these are security regressions so not going to change for Monday release

PR #1386, opened 2021-01-26
SMS service setup: do not allow specific geos (+86)

Patrick1904:


kcole16:
@Patrick1904 Can we make the disallowed geos (e.g. 'CN') a config variable, and also support a list/multiple geos? Decent chance we will need to add more

vgrichina (approved):
Looks good but I have same concern as @kcole16 – let's make this a list specified in env variable.

Patrick1904:
> @Patrick1904 Can we make the disallowed geos (e.g. 'CN') a config variable, and also support a list/multiple geos? Decent chance we will need to add more

@kcole16 @vgrichina I updated it to check a list of country codes defined in env file. E.g. `SMS_BLACKLIST = "CN, SE, FI"`

ihaid (approved):

PR #1388, opened 2021-01-26
Default translation

marcinbodnar:


src/components/common/GlobalAlert.js
```diff
@@ -181,14 +180,22 @@ const GlobalAlertNew = ({ globalAlert, actionStatus, clearGlobalAlert, closeIcon
...
                                 </Header>
-
-                                <Translate>
-                                    {({ translate }) => 
-                                        translate(alert.messageCode) === ''
-                                            ? <Translate id={`reduxActions.default.${alert.success ? 'success' : 'error'}`} />
-                                            : <Translate id={alert.messageCode} data={alert.data} />
+                                <Translate id={alert.messageCode} data={alert.data} options={{
+                                    onMissingTranslation: ({ languageCode }) => {
+                                        if (languageCode === 'en') {
```
vgrichina, line 185: @marcinbodnar not sure I understand this. Why do we have this check for multiple languages with same message hardcoded here?

marcinbodnar: @vgrichina It didn't work well so I refactored it like this to handle default translations for all languages. But now I found a better solution, will work well with default translations in a translation file. Please check.

PR #1392, opened 2021-01-28
Reimplementation/mixpanel tracking

icerove:
mixpanel browser tracking


src/components/accounts/two_factor/EnableTwoFactor.js
```diff
@@ -79,13 +80,17 @@ export function EnableTwoFactor(props) {
     }, [recoveryMethods]);
 
     const handleNext = async () => {
+        Mixpanel.track("2FA Go to next step for tying in code")
```
Patrick1904, line 83: Can skip this one

src/components/accounts/two_factor/EnableTwoFactor.js
```diff
@@ -187,10 +202,18 @@ export function EnableTwoFactor(props) {
                         type='submit'
                         sending={loading}
                         sendingString='button.enabling'
+                        onClick = {() => Mixpanel.track("2FA Click Continue button")}
```
Patrick1904, line 205: `form` submits `handleNext` already so can remove `onClick = {() => Mixpanel.track("2FA Click Continue button")}`

Patrick1904:
This looks good to me. @vgrichina any concerns?

src/components/Routing.js
```diff
@@ -123,6 +125,13 @@ class Routing extends Component {
 
             handleClearAlert()
         })
+
+        let id = Mixpanel.get_distinct_id()
+        Mixpanel.identify(id)
+        Mixpanel.people.set({enabled_2FA: this.props.account.twoFactor, can_enable_two_factor: this.props.account.canEnableTwoFactor})
+        if(this.props.account.accountId){
```
vgrichina, line 132: nit: please format like the rest of the code, e.g. here should be `if (this.props.account.accountId) {`. Same for other similar instances, not just in this line.

src/components/Routing.js
```diff
@@ -168,6 +177,8 @@ class Routing extends Component {
                                     account.requestPending(verified, error)
                                     // clears requestPending and closes the modal
                                     promptTwoFactor(null)
+                                    //tracking error
```
vgrichina, line 180: nit: `// tracking error` (missing spacw)

src/mixpanel/index.js
```diff
@@ -0,0 +1,25 @@
+import mixpanel from 'mixpanel-browser';
```
vgrichina, line 1: nit: please do 4 spaces indents as in other places for consistency

src/components/accounts/two_factor/EnableTwoFactor.js
```diff
@@ -190,7 +204,14 @@ export function EnableTwoFactor(props) {
                     >
                         <Translate id={`button.continue`} />
                     </FormButton>
-                    <FormButton className='link' type='button' linkTo='/profile'><Translate id='button.skip' /></FormButton>
+                    <FormButton 
+                        className='link' 
+                        type='button' 
+                        linkTo='/profile' 
+                        onClick={() => Mixpanel.track("2FA Click Skip button", {url_link:"/profile"})}
```
vgrichina, line 211: nit: I am a bit confused around event naming. Sometimes words are capitalised, sometimes they don't. Is there some pattern that needs to be followed?

src/mixpanel/index.js
```diff
@@ -0,0 +1,25 @@
...
+
+const BROWSER_MIXPANEL_TOKEN = '9edede4b70de19f399736d5840872910';
+
+mixpanel.init(BROWSER_MIXPANEL_TOKEN);
+mixpanel.register({'timestamp': new Date().toString(), '$referrer': document.referrer})
+export const Mixpanel = {
```
vgrichina, line 7: not sure why this is needed versus just `export const mixpanel`?  Is there plan to override some of these methods later?

vgrichina (commented):
Looks good overall besides few nits

src/mixpanel/index.js
icerove: we need to do so since do not want to init everywhere

src/components/accounts/two_factor/EnableTwoFactor.js
icerove: I will follow the pattern for verb to be capitalized and other not

src/mixpanel/index.js
vgrichina: this doesn't look fixed

src/mixpanel/index.js
vgrichina: you can still init in this module but re-export `mixpanel`

maybe I don't understand something but it's not clear why wrapper object is needed as well if the only goal is to init

PR #1395, opened 2021-01-29
Fix default error for Global Alert

marcinbodnar:


Patrick1904:
Tested and confirmed this fixes the error. Merging as it's a critical fix.

marcinbodnar:
thanks @Patrick1904

Issue #1394, opened 2021-01-29
Improve loading time of Profile page


marcinbodnar:
Improve loading time by adding functionality where `In staking pools` details will be fetched only when the user clicks on the `In staking pools` field - user will see loader and after fetching the data will be displayed.

jimmy3dita:
@marcinbodnar added this card to the current sprint, feel free to drag/drop it in the relative column

marcinbodnar:
@jimmy3dita thanks, done.

PR #1400, opened 2021-02-01
Improve loading time of Profile page (#1394)

marcinbodnar:


Patrick1904:
@marcinbodnar can we please load the 'in staking pools' balances after the initial account balance has loaded vs loading them when the user expands the accordion? That way the initial balance still loads instantly and the user doesn't have to wait for staking pool balances if they expand 'in staking pools'

vgrichina:
Somewhat related https://github.com/near/near-api-js/pull/492 – we can potentially cut some requests at near-api-js level I think.

marcinbodnar:
> @marcinbodnar can we please load the 'in staking pools' balances after the initial account balance has loaded vs loading them when the user expands the accordion? That way the initial balance still loads instantly and the user doesn't have to wait for staking pool balances if they expand 'in staking pools'

@Patrick1904 done, please test, it wasn't so trivial as I thought

src/components/common/Accordion.js
```diff
@@ -29,6 +29,8 @@ const Accordion = ({ className, trigger, children }) => {
     }, [open])
 
     const handleClick = () => {
+        action && action()
```
Patrick1904, line 32: @marcinbodnar can we revert the changes to `Accordion`?

marcinbodnar: @Patrick1904 I left it on purpose, it might be handy in the future to handle action by this component

Patrick1904 (approved):
LGTM

Issue #1309, opened 2020-12-22
Refreshing page resets staking with a lockup


ilblackdragon:
Steps to reproduce
1. Open account with lockup
2. Go to staking and select lockup
3. Select specific pool
4. Go into staking with that pool
5. Refresh the page

Observe that available balance changes from lockup balance to normal account balance.

Patrick1904:
@marcinbodnar can we please persist account selection for page refresh and if the user navigates to let's say profile, and then back to staking again. We can use localStorage.

PR #1408, opened 2021-02-02
Persist account selection for staking (#1309)

marcinbodnar:


Patrick1904 (approved):
Looks and works great!

PR #1390, opened 2021-01-27
Release 2021-01-26

vgrichina:

PR #1413, opened 2021-02-03
translation: fix old data prop

Patrick1904:


src/translations/zh-hans.global.json
```diff
@@ -817,7 +817,7 @@
         "pageTitle": {
             "loading": "正在加载...",
             "notFound": "账户 ${accountId} 未找到",
-            "default": "账户：${accountId}"
```
Patrick1904, line 820: @picturepan2 can you please confirm that this translation still makes sense without an account ID

Patrick1904: @picturepan2 I'm gonna include this temp fix in the current [release](https://github.com/near/near-wallet/pull/1390/) since it's better than showing the undefined variable on the UI.

Screen Shot 2021-02-03 at 1 05 16 PM

picturepan2: Yes, it means "Account". It is good.

picturepan2: @Patrick1904 Yes, it means "Account". It looks still good without account ID.

PR #1417, opened 2021-02-04
Add back fallback for missing translations

picturepan2:
@Patrick1904 This is the fix.

`onMissingTranslation` was removed  in https://github.com/near/near-wallet/commit/8581a373f6db9e0dddac1875b7b81aa92ab4fd3d. Maybe ask @icerove before merge?

### DIFF
--- a/src/components/Routing.js
+++ b/src/components/Routing.js
@@ -57,6 +57,8 @@ const theme = {}
 
 const PATH_PREFIX = process.env.PUBLIC_URL
 
+const onMissingTranslation = ({ defaultTranslation }) => defaultTranslation;
+
 const Container = styled.div`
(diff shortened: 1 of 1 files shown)


Patrick1904:
> @Patrick1904 This is the fix.
> 
> `onMissingTranslation` was removed in [8581a37](https://github.com/near/near-wallet/commit/8581a373f6db9e0dddac1875b7b81aa92ab4fd3d). Maybe ask @icerove before merge?

@icerove we need this piece of code so that whenever there's no translation available for a string, it falls back to English. Did it cause some issue with Mixpanel or something?

Patrick1904 (approved):
Thank you @picturepan2!

PR #1424, opened 2021-02-05
Enable creating account w/ current account

Patrick1904:
Create an account using currently logged in account:
1. First sign in w/ existing account: https://deploy-preview-1424--near-wallet-staging.netlify.app/
2. Navigate to `/create` and start the account creation flow
3. Once you're on the funding page, click the green button that says 'Fund with XXXXXX.near'

Screen Shot 2021-02-05 at 4 11 38 PM

### DIFF
--- a/src/components/accounts/SetupImplicit.js
+++ b/src/components/accounts/SetupImplicit.js
@@ -64,6 +64,12 @@ const StyledContainer = styled(Container)`
         &.green, &.black {
             margin-top: 25px !important;
         }
+
+        &.green {
+            height: auto !important;
+            line-break: anywhere;
+            line-height: normal;
+        }
     }
 `
 
@@ -90,9 +96,9 @@ class SetupImplicit extends Component {
     state = { ...initialState }
 
     handleContinue = async () => {
-        const { dispatch, accountId, implicitAccountId, recoveryMethod } = this.props
+        const { dispatch, newAccountId, implicitAccountId, recoveryMethod } = this.props
         this.setState({ createAccount: true })
-        await dispatch(createAccountFromImplicit(accountId, implicitAccountId, recoveryMethod))
+        await dispatch(createAccountFromImplicit(newAccountId, implicitAccountId, recoveryMethod))
         await dispatch(redirectTo('/fund-create-account/success'))
     }
 
@@ -211,11 +217,11 @@ class SetupImplicit extends Component {
     }
 
     handleFundWithAccount = async () => {
-        const { dispatch, implicitAccountId, accountId, recoveryMethod } = this.props
+        const { dispatch, implicitAccountId, newAccountId, recoveryMethod } = this.props
         this.setState({ createFromAccount: true })
         await dispatch(sendMoney(implicitAccountId, MIN_BALANCE_TO_CREATE.toString()))
-        await dispatch(createAccountFromImplicit(accountId, implicitAccountId, recoveryMethod))
-        await dispatch(redirectTo('/fund-create-account/success'))
+        await dispatch(createAccountFromImplicit(newAccountId, implicitAccountId, recoveryMethod))
+        dispatch(redirectTo('/fund-create-account/success'))
     }
 
     render() {
@@ -229,7 +235,7 @@ class SetupImplicit extends Component {
             createFromAccount
         } = this.state
 
-        const { implicitAccountId, accountId, mainLoader, balance } = this.props
+        const { implicitAccountId, newAccountId, accountId, mainLoader, balance } = this.props
 
         return (
             <Translate>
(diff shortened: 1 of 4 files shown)


vgrichina:
@Patrick1904 looks good, not sure whether this copy still applies however:

```
You must now use this Account ID as your address for all NEAR operations.

Please update your address on any exchanges or other devices.
```

as you didn't have funding address technically? Though I guess you still have seen it on a previous page 🤔

src/components/accounts/SetupImplicit.js
```diff
@@ -203,22 +216,35 @@ class SetupImplicit extends Component {
         });
     }
 
+    handleFundWithAccount = async () => {
+        const { dispatch, implicitAccountId, newAccountId, recoveryMethod } = this.props
+        this.setState({ createFromAccount: true })
+        await dispatch(sendMoney(implicitAccountId, MIN_BALANCE_TO_CREATE.toString()))
+        await dispatch(createAccountFromImplicit(newAccountId, implicitAccountId, recoveryMethod))
```
vgrichina, line 223: I think we shouldn't send money to implicit account first

```
export const createAccountFromImplicit = createAction('CREATE_ACCOUNT_FROM_IMPLICIT', async (accountId, implicitAccountId, recoveryMethod) => {
    const recoveryKeyPair = await wallet.keyStore.getKey(wallet.connection.networkId, implicitAccountId)
    if (recoveryKeyPair) {
        await wallet.saveAccount(accountId, recoveryKeyPair)
    }
    const publicKey = new PublicKey({ keyType: KeyType.ED25519, data: Buffer.from(implicitAccountId, 'hex') })
    await wallet.createNewAccount(accountId, { fundingAccountId: implicitAccountId }, recoveryMethod, publicKey)
})
```

instead it makes sense to create another action similar to this one which would use your current account instead of `implicitAccountId`

downstream in `wallet` there is already handling for implicit vs non-implicit:

```
    async createNewAccountFromAnother(accountId, fundingAccountId, publicKey) {
        const account = await this.getAccount(fundingAccountId)
        await account.functionCall(ACCOUNT_ID_SUFFIX, 'create_account', {
            new_account_id: accountId,
            new_public_key: publicKey.toString().replace(/^ed25519:/, '')
            // TODO: Adjust gas if necessary
            // TODO: Should new account have other than minimum balance? For implicit account?
        }, LINKDROP_GAS, MIN_BALANCE_FOR_GAS)
        if (!this.accounts[fundingAccountId]) {
            // Temporary implicit account used for funding – move whole balance by deleting it
            await account.deleteAccount(accountId)
        }
    }
```

vgrichina (commented):
Looks good overall, but we shouldn't do extra transaction to implicit account (it can be skipped altogether)

src/components/accounts/SetupImplicit.js
Patrick1904: Good point I will update!

Patrick1904:
> @Patrick1904 looks good, not sure whether this copy still applies however:
> 
> ```
> You must now use this Account ID as your address for all NEAR operations.
> 
> Please update your address on any exchanges or other devices.
> ```
> 
> as you didn't have funding address technically? Though I guess you still have seen it on a previous page 🤔

Yea I'm not sure about this one since we've previously introduced the funding address. I'll defer to @corwinharrell on this. Can easily omit this part of the copy if the user funded with current account.

Patrick1904:
> Looks good overall, but we shouldn't do extra transaction to implicit account (it can be skipped altogether)

@vgrichina yep I agree we shouldn't do unnecessary transactions. I'm thinking though that it might be safer and more clear to keep using the implicit account and the creation process the same, as it's already part of the recovery setup process before this step. It's also clearer with 'fund with xxx.near' that your current account will fund the temporary funding address. And if we skip the 'funding', then it should be 'create with xxxx.near' - but that seems weird when you're on this 'fund your account' page.

If you're creating an account using current account, then we don't have to send the user here at all, but can do that directly when the user is finalizing their recovery setup. Though that introduces new questions like asking the user permission to fund the creation of the new account, and if it's only the absolute minimum, 1N, or more.

I guess what I'm trying to say is that we may want to make changes to the flow/design if we change what's happening under the hood. And that the current implementation doesn't require flow/design changes.

vgrichina:
> I guess what I'm trying to say is that we may want to make changes to the flow/design if we change what's happening under the hood. And that the current implementation doesn't require flow/design changes.

I agree we should make changes to the flow in this case. Once you chose to fund with your own account we can redirect user to new account. No extra confirmation steps needed. 

Extra transaction increases risk that user lands in half finished state (as we already see happening with normal funding flow). Best to keep everything atomic when it's possible.

> If you're creating an account using current account, then we don't have to send the user here at all, but can do that directly when the user is finalizing their recovery setup.

I think we still want to give user choice to fund externally, as you don't necessarily want to tie new account to your existing account.

PR #1432, opened 2021-02-08
Point 'HELP' button to Zendesk page

Patrick1904:
Point 'HELP' button to `https://nearhelp.zendesk.com/` instead of Discord.

### DIFF
--- a/src/components/navigation/DesktopContainer.js
+++ b/src/components/navigation/DesktopContainer.js
@@ -1,7 +1,6 @@
 import React, { Component } from 'react';
 import { Translate } from 'react-localize-redux';
 import styled from 'styled-components';
-import helpIcon from '../../images/icon-help.svg';
 import userIcon from '../../images/user-icon-grey.svg';
 import languagesIcon from '../../images/icon-languages.svg';
 import Logo from './Logo';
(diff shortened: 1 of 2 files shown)


kcole16 (approved):

vgrichina (changes requested):
@Patrick1904 we shouldn't embed code loaded from Zendesk as it's extra security risk for wallet.

I think it's best we just link to the page to get support from existing Help button.

Patrick1904:
> @Patrick1904 we shouldn't embed code loaded from Zendesk as it's extra security risk for wallet.
> 
> I think it's best we just link to the page to get support from existing Help button.

@kcole16 are you onboard with just linking to Zendesk page using existing button?

kcole16:
> @kcole16 are you onboard with just linking to Zendesk page using existing button?

I'm onboard. It's not ideal, but security is #1 priority always (I should've thought of that when writing the issue).

Patrick1904:
> @Patrick1904 we shouldn't embed code loaded from Zendesk as it's extra security risk for wallet.
> 
> I think it's best we just link to the page to get support from existing Help button.

The current 'HELP' button only exists on desktop. For mobile users, the 'join community' link in the footer is the only option to get help, and it leads to Discord, not Zendesk.

I suggest we add the 'HELP' button to the mobile nav too then.

@kcole16 @corwinharrell @vgrichina

Issue #1428, opened 2021-02-08
Default Ledger account can be deleted during account creation


kcole16:
**Problem**
Creating a new account with a Ledger leads to the deletion of the default HD path address, which may have been in use. This is usually recoverable, but in a rare (hard to reproduce) case can lead to catastrophic loss of NEAR.

**Expected Behavior**
The default account should not be deleted, especially if it already exists on-chain.

**Steps to reproduce**

1. Create a new account with the default HD path via CLI (`44'/397'/0'/0'/1'`) 
2. Send NEAR to it
3. Create a new account in Wallet and choose Ledger
4. Observe that the first account is deleted

vgrichina:
Looks like underlying reason for this problem is that `create_account` doesn't fail transaction on failure:
https://explorer.near.org/transactions/7MuUQdmto8LJyY5pKMimWUcL5UK57UN2tVD3mPMfaF8N

kcole16:
Closing as @vgrichina fixed in #1435 (right?)

vgrichina:
yeah, the important part has been fixed – account only gets removed if is completely equivalent to the one that's not created yet

PR #1435, opened 2021-02-09
Handle errors in create_account call

vgrichina:


vgrichina:
Addresses big part of https://github.com/near/near-wallet/issues/1428

Patrick1904:
@vgrichina this is in case `create_account` fails during implicit account creation right? Can we use the global alert to surface the error? That way the user will see the error, and the modal should stay open, so that the user can click 'continue' and try again

vgrichina:
> @vgrichina this is in case `create_account` fails during implicit account creation right? Can we use the global alert to surface the error? That way the user will see the error, and the modal should stay open, so that the user can click 'continue' and try again

yeah global alert should be ok

generally this is something that shouldn't normally be happening but we have to handle just in case to avoid fallout from errors (like the one with Ledger setup)

kcole16 (approved):

PR #1438, opened 2021-02-09
Improve Wallet loading time, fix checking if lockup exists, reset Redux state on switch, recover, create account

marcinbodnar:


src/components/profile/Profile.js
```diff
@@ -141,6 +141,7 @@ export function Profile({ match }) {
                 await dispatch(loadRecoveryMethods())
                 dispatch(getAccessKeys(accountId))
                 dispatch(getLedgerKey())
+                await dispatch(getBalance())
```
vgrichina, line 144: @marcinbodnar [not a blocker for this PR]: seems like we can improve even further if parallelise kicking off all the queries vs using `await` to do some of them sequentially? Basically kick off loading balance when we know that we need to do it vs when recovery methods already loaded, etc

src/utils/wallet.js
```diff
@@ -183,7 +184,7 @@ class Wallet {
                         break
                     }   
                 }
-                this.selectAccount(nextAccountId)
+                store.dispatch(selectAccount(nextAccountId))
```
vgrichina, line 187: @marcinbodnar do I understand correctly that larger plan is to break down `wallet.js` into Redux actions?

src/components/profile/Profile.js
marcinbodnar: @vgrichina Right, in this particular case I wanted to load Recovery Methods before fetching for staking balance details, which might take few seconds.

src/utils/wallet.js
marcinbodnar: @vgrichina Yes, that's the larger plan - in this case, we want to clear the account redux state while changing account.

PR #1443, opened 2021-02-10
Random shuffle of Validators list introduced

zavodil:
Alphabetical order of validators on the staking page is wrong and gives motivation to pools to abuse it by naming like "00001_pool" or "aaa_pool". Lets reshuffle the list randomly so every pool will have equal chances to  stay on the top. Existing search bar will help users to find specific pool. 

Clip2net_210210172728


corwinharrell:
I don't think we need to add an additional label. The label shown above is particularly clunky. I think it's fine if we just randomize the order and hopefully add the ability to sort the list in the near future (eg by fee, alphabetically, by number of delegators, etc..)

corwinharrell (changes requested):
I don't think we need to add an additional label. The label shown here is particularly clunky. I think it's fine if we just randomize the order and hopefully add the ability to sort the list in the near future (eg by fee, alphabetically, by number of delegators, etc..)

zavodil:
> I don't think we need to add an additional label. The label shown here is particularly clunky. I think it's fine if we just randomize the order and hopefully add the ability to sort the list in the near future (eg by fee, alphabetically, by number of delegators, etc..)

I added the label only to give a hint for existing users why they see different list now. We can remove it now or in a month after the update.

Patrick1904:
> > I don't think we need to add an additional label. The label shown here is particularly clunky. I think it's fine if we just randomize the order and hopefully add the ability to sort the list in the near future (eg by fee, alphabetically, by number of delegators, etc..)
> 
> I added the label only to give a hint for existing users why they see different list now. We can remove it now or in a month after the update.

@zavodil can you please remove the label and we'll merge this. Thanks!

zavodil:
> > > I don't think we need to add an additional label. The label shown here is particularly clunky. I think it's fine if we just randomize the order and hopefully add the ability to sort the list in the near future (eg by fee, alphabetically, by number of delegators, etc..)
> > 
> > 
> > I added the label only to give a hint for existing users why they see different list now. We can remove it now or in a month after the update.
> 
> @zavodil can you please remove the label and we'll merge this. Thanks!

Done!

Patrick1904 (approved):

PR #1445, opened 2021-02-10
Fix default translation for Global Alert

marcinbodnar:


marcinbodnar:
@Patrick1904 I did check the issue you mention, about staking details not loading sometimes on the Profile page - wasn't able to reproduce it, I checked the implementation and it looks good.

Patrick1904 (approved):

Issue #1399, opened 2021-02-01
Buttons inactive on Ledger enabling page


ihaid:
**Expected Behavior**
Buttons should be clickable when canceling operation 

**Steps to reproduce**

1. Create an account
2. Go to the profile page 
3. Click  enable "Ledger Hardware Wallet"
4. On the pop-up window click "Continue" 
5. Close window without connecting Ledger
6. Click on any button on the page

PR #1447, opened 2021-02-10
Fix closing modal (#1399)

marcinbodnar:


Patrick1904:
@marcinbodnar is there a way to reject/cancel the ledger connect when closing the modal? Otherwise it tries to connect for a while and eventually throws the error.

Screen Shot 2021-02-10 at 4 30 09 PM

Patrick1904:
> @marcinbodnar is there a way to reject/cancel the ledger connect when closing the modal? Otherwise it tries to connect for a while and eventually throws the error.

Actually, canceling should probably be strictly controlled with the Ledger device.

Patrick1904 (approved):

marcinbodnar:
@Patrick1904 
> Actually, canceling should probably be strictly controlled with the Ledger device.
Yes, I don't think it's possible from the Wallet side.

PR #1448, opened 2021-02-11
Add links to transactions in explorer

vgrichina:
Tiny improvement to make it easier to find transactions in explorer


Patrick1904:
Should we make the entire row clickable vs only the time/status section? cc @corwinharrell 

Screen Shot 2021-02-11 at 10 58 45 AM

corwinharrell:
yeah it makes sense to me to make the whole row clickable

Patrick1904:
> yeah it makes sense to me to make the whole row clickable

Looks like semantic-ui is breaking everything when I wrap with `<a>`. I'll merge this for now and will re-write without semantic-ui in the fungible token design update.

PR #1450, opened 2021-02-11
Track/send and stake

icerove:
add track event on:
1. dashboard page links
2. send flow 
3. receive flow
4. stake flow
5. unstake flow


src/components/send/SendContainer.js
```diff
@@ -98,12 +106,19 @@ export function SendContainer({ match, location }) {
...
 
     const handleConfirm = () => {
+        Mixpanel.track("SEND Click submit button")
         setConfirm(true)
     }
 
     const handleSend = async () => {
-        await dispatch(sendMoney(id, parseNearAmount(amount)))
-        await dispatch(getBalance())
+        try {
```
vgrichina, line 114: Looks like this pattern is used so frequently that will benefit from being extracted into function, so that looks more like this:

```js
await withTracking('SEND money`, async () => {
    await dispatch(sendMoney(id, parseNearAmount(amount)))
    await dispatch(getBalance()) 
});
```

basically have something like:

```js
    async function withTracking(name, fn) {
         try {
             Mixpanel.track(`${name} start`)
             await fn();
             Mixpanel.track(`${name} finish`)
         } catch (e) {
             Mixpanel.track(`${name} fail`, {error: e.message})
         }
    }
```

src/components/staking/components/StakingAction.js
```diff
@@ -169,7 +183,13 @@ export default function StakingAction({
                     />
                 }
                 <div className='desc'><Translate id={`staking.${action}Success.descTwo`}/></div>
-                <FormButton linkTo='/staking' className='gray-blue'><Translate id={`staking.${action}Success.button`} /></FormButton>
+                <FormButton
```
vgrichina, line 186: looks like in many cases we are tracking button clicks
I think we should just customise respective components like `FormButton`  so that you don't have to repeat boilerplate every time.


so instead of 
```jsx
<FormButton 
                     linkTo='/staking' 
                     className='gray-blue'
                     onClick={() => Mixpanel.track("STAKE/UNSTAKE Return to dashboard")}
                 >
```

you can have:

```jsx
<FormButton linkTo='/staking' className='gray-blue' trackingId='STAKE/UNSTAKE Return to dashboard'><Translate id={`staking.${action}Success.button`} /></FormButton>
```

src/components/send/SendContainer.js
icerove: will change this

src/components/staking/components/StakingAction.js
icerove: will change this too

vgrichina (changes requested):
Looks good overall, however looks like we keep multiplying the same boilerplate in many places.

I gave some examples of what can be done for such cases:
- async operations
- common UI components

Let me know if you have any questions / other proposals on how to reduce boilerplate needed for analytics.

src/components/profile/Profile.js
```diff
@@ -150,18 +150,20 @@ export function Profile({ match }) {
...
-        if (account.balance?.total) {
+        if (account.balance) {
             let id = Mixpanel.get_distinct_id()
             Mixpanel.identify(id)
             Mixpanel.people.set_once({create_date: new Date().toString(),})
             Mixpanel.people.set({
                 relogin_date: new Date().toString(),
                 enabled_2FA: account.has2fa,
-                [accountId]: formatNEAR(account.balance.total) 
+                [accountId+'total']: formatNEAR(account.balance.total),
```
vgrichina, line 160: how is this processed on mixpanel side? Seems like some delimited between `accountId` and subfields might be needed?

icerove: yes, I add _ instead. Thanks

src/components/accounts/two_factor/EnableTwoFactor.js
```diff
@@ -82,24 +82,19 @@ export function EnableTwoFactor(props) {
...
-                    Mixpanel.track("2FA Deploy multisig start")
-                    try {
-                        await handleDeployMultisig()
-                        Mixpanel.track("2FA Deploy multisig finish")
-                    } catch(e) {
-                        Mixpanel.track("2FA Deploy multisig fail", {error: e.message})
+            await Mixpanel.withTracking("2FA Initialize two factor",
+                async () => response = await dispatch(initTwoFactor(accountId, method)),
+                () => {},
+                async () => {
```
icerove, line 88: @vgrichina Please have a look at this function since I want you to double check whether this one I also code in correct format and do not break the function.

src/components/common/Footer.js
```diff
@@ -155,7 +166,10 @@ class Footer extends Component {
                         className='learn-more'
                         verticalAlign='middle'
                     >
-                        <Translate id='footer.desc' /> <a href='https://nearprotocol.com/'><Translate id='footer.learnMore' /></a>
+                        <Translate id='footer.desc' />{' '}
+                        <a href='https://near.org' onClick={() => Mixpanel.track("Click Learn More")}>
```
Patrick1904, line 170: We might want to include `Footer` key word for these

src/components/navigation/NavLinks.js
```diff
@@ -91,12 +92,12 @@ const NavLink = styled(Link)`
...
-        <NavLink icon={summaryIcon} to='/'><Translate id='link.summary'/></NavLink>
+        <NavLink icon={summaryIcon} to='/' onClick={() => Mixpanel.track("Click summary button on nav") }><Translate id='link.summary'/></NavLink>
         {!DISABLE_SEND_MONEY &&
-            <NavLink icon={arrowIcon} className='rotate-up' to='/send-money'><Translate id='link.send'/></NavLink>
+            <NavLink icon={arrowIcon} className='rotate-up' to='/send-money' onClick={() => Mixpanel.track("SEND Click send button on nav")}><Translate id='link.send'/></NavLink>
         }
-        <NavLink icon={arrowIcon} className='rotate-down' to='/receive-money'><Translate id='link.receive'/></NavLink>
-        <NavLink icon={stakingIcon} to='/staking'><Translate id='link.staking'/></NavLink>
+        <NavLink icon={arrowIcon} className='rotate-down' to='/receive-money' onClick={() => Mixpanel.track("RECEIVE Click receive button on nav")}><Translate id='link.receive'/></NavLink>
+        <NavLink icon={stakingIcon} to='/staking' onClick={() => Mixpanel.track("STAKE Click staking button")}><Translate id='link.staking'/></NavLink>
```
Patrick1904, line 100: I think it's best to align the pattern for all these nav buttons

src/components/staking/components/BalanceBreakdown.js
```diff
@@ -123,7 +124,7 @@ function BalanceBreakdown({ total, onClickAvailable, availableType, error }) {
                         </div>
                     </div>
                     <div className='title'>
-                        <div onClick={() => setOpen(!open)}>
+                        <div onClick={() => {setOpen(!open); Mixpanel.track("Watch available to send")}}>
```
Patrick1904, line 127: This will track on open and close right now

icerove: yes, so I do not track open or close. But track checkout available token

Patrick1904: Right, maybe adjust the language otherwise it seems that it's only tracking when the user expands the dropdown. `STAKE Click toggle available to stake/unstake dropdown`

PR #1460, opened 2021-02-17
add verify success

icerove:
add tracking for success verification 2FA modal one

### DIFF
--- a/src/components/accounts/two_factor/TwoFactorVerifyModal.js
+++ b/src/components/accounts/two_factor/TwoFactorVerifyModal.js
@@ -68,7 +68,7 @@ const TwoFactorVerifyModal = ({ open, onClose }) => {
     }
     
     const handleCancelClose = () => {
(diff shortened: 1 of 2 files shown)


src/components/accounts/two_factor/TwoFactorVerifyModal.js
```diff
@@ -68,7 +68,7 @@ const TwoFactorVerifyModal = ({ open, onClose }) => {
     }
     
     const handleCancelClose = () => {
-        Mixpanel.track("2FA Modal Cancel verification")
```
vgrichina, line 71: not sure why this change, is `handleCancelClose` called in some other case than dismissing 2FA modal?

icerove: reverted

icerove:
I look through 2FA verify modal again and find I misunderstanding `onClose` before, thanks for pointing out and this time it should be correct. @vgrichina

PR #1461, opened 2021-02-17
Don't delete funding account if it has changes to public keys

vgrichina:
Basically implicit account is only deleted in case when it would be re-created intact on repeated transfer.


vgrichina:
Should address https://github.com/near/near-wallet/issues/1428 for cases when user is already using given implicit account id.

vgrichina:
@kcole16 @Patrick1904 please test creating multiple accounts from same Ledger. If at any point you import implicit account directly it shouldn't get deleted on next funding.

Issue #1476, opened 2021-02-22
App login: user is not redirected back to the app if the user switches account before allowing login


Patrick1904:
**Problem**
Error occurs if the user switches account on the `/login` page. It seems that the `redirect_url` is lost after switching account.

**Expected Behavior**
User is redirected back to the app after switching account and approving login

**Steps to reproduce**
1. Go to berryclub.io
2. Click 'Login with NEAR Wallet'
3. Switch account before clicking 'Allow'
4. Click the 'Allow' button
5. Observe that you get stuck on the confirm step

PR #1479, opened 2021-02-23
Fix switching account (#1476)

marcinbodnar:


Patrick1904 (approved):
Able to switch account during app `/login` process and still persist `url` data and successfully redirected back to app. Tested using berryclub.io

PR #1480, opened 2021-02-23
Recover with link: Fix error case

Patrick1904:
The 'link expired' screen was showing while the account was being recovered.


Patrick1904:
Note that this bug is currently not in production - but reproducible on wallet.testnet.near.org

To reproduce the bug:
1. Recover with link on wallet.testnet.near.org
2. Observe that 'link expired' shows while the account is being recovered

Issue #1471, opened 2021-02-21
Make sure saveAccount / saveAndSelectAccount works well with multiple windows open


vgrichina:
**Problem**

Looks like account list is currently saved in `selectAccount` method (i.e. `saveAccount` doesn't save) + to save wallet just serialises current in memory list to local storage. This means that if you open 2 wallet instances and add account in each of them, there is likely conflict and one of these would be excluded from the list.

**Expected Behavior**

When you open 2 wallet windows you can add account from both of them and see it appear in other window when it is refreshed.

Probably the easiest way to ensure it is to read a list from local storage before writing and actually save new accounts from `saveAccount` (vs `selectAccount`), as then you know which account has been added.

**Steps to reproduce**

1. Open 2 wallet windows.
2. Add new account in both of them.
3. Refresh both wallet windows

NOTE: This is P1 as can potentially result in account loss (if user has account without recovery methods somehow still in the wallet). Also generally likely to result in degraded UX (having to import account again).

PR #1489, opened 2021-02-23
Creating an account in multiple windows open

marcinbodnar:


vgrichina:
@marcinbodnar something I wonder is whether we need to maintain this separate list at all.

https://github.com/near/near-api-js/blob/6d4e688c0715b5257e45c4be24d39c98d5631648/src/key_stores/browser_local_storage_key_store.ts#L80

vgrichina:
@marcinbodnar something I wonder about whether we need to maintain this list at all.

We can use `getAccounts` in key store to get full list off accounts with keys https://github.com/near/near-api-js/blob/6d4e688c0715b5257e45c4be24d39c98d5631648/src/key_stores/browser_local_storage_key_store.ts#L80

I guess it might have some false positives we'd need to filter out based on whether account still exists on chain.

MaximusHaximus:
@marcinbodnar Is this PR still relevant, or should we close it?

marcinbodnar:
@MaximusHaximus I'm not sure how this part of creating an account is working now, but creating an account in multiple windows should be handled anyway - I would suggest closing this PR, but create a new issue with a link to this PR - so the old solution will be visible.

MaximusHaximus:
> @MaximusHaximus I'm not sure how this part of creating an account is working now, but creating an account in multiple windows should be handled anyway - I would suggest closing this PR, but create a new issue with a link to this PR - so the old solution will be visible.

Rather than create a new issue for the same problem, can you take a look at this and see if the code is still relevant (there are no conflicts) and still solves the problem?  It looks like this would fix https://github.com/near/near-wallet/issues/1471 which is a P1 issue that also appears to be a well written issue with reproduction case documented; I don't see any benefit in re-creating that issue :).

It looks like what we have here is a minimal impact one-line change.  If this code change still solves the problem that it was supposed to solve and hasn't introduced a negative impact, then I'd like for us to go ahead and get it merged rather than kick this particular can farther down the road with a new issue :). We just need to be sure there isn't a _negative_ consequence from this change, which should be confirmed with basic testing cases that call `saveAccount()` you do, and that the problem is still solved with this one-line change (also easily verifiable via the same reproduction case outlined in #1471).  If we're not confident that this fixes the problem and is safe, then we should close the PR and do whatever we need to to gain that confidence -- but I'm assuming this is ready to go given it's up for review.

Let's get this over the line so that users can benefit from the fix! If you discover that it either (a) doesn't fix the problem or (b) has a negative side-effect, then lets close this PR and triage the issue into a different sprint based on its priority back into _In Development_ -- but right now, this is a ghost in the system, "done", but not _really_ done until it's merged, and not being tracked in our current sprints.

MaximusHaximus:
@marcinbodnar a friendly reminder about this PR

marcinbodnar:
> > @MaximusHaximus I'm not sure how this part of creating an account is working now, but creating an account in multiple windows should be handled anyway - I would suggest closing this PR, but create a new issue with a link to this PR - so the old solution will be visible.
> 
> Rather than create a new issue for the same problem, can you take a look at this and see if the code is still relevant (there are no conflicts) and still solves the problem? It looks like this would fix #1471 which is a P1 issue that also appears to be a well written issue with reproduction case documented; I don't see any benefit in re-creating that issue :).
> 
> It looks like what we have here is a minimal impact one-line change. If this code change still solves the problem that it was supposed to solve and hasn't introduced a negative impact, then I'd like for us to go ahead and get it merged rather than kick this particular can farther down the road with a new issue :). We just need to be sure there isn't a _negative_ consequence from this change, which should be confirmed with basic testing cases that call `saveAccount()` you do, and that the problem is still solved with this one-line change (also easily verifiable via the same reproduction case outlined in #1471). If we're not confident that this fixes the problem and is safe, then we should close the PR and do whatever we need to to gain that confidence -- but I'm assuming this is ready to go given it's up for review.
> 
> Let's get this over the line so that users can benefit from the fix! If you discover that it either (a) doesn't fix the problem or (b) has a negative side-effect, then lets close this PR and triage the issue into a different sprint based on its priority back into _In Development_ -- but right now, this is a ghost in the system, "done", but not _really_ done until it's merged, and not being tracked in our current sprints.

I meant new PR.

marcinbodnar:
@MaximusHaximus Done.

MaximusHaximus:
@ihaid Can you give this a test before we merge it to testnet/staging?

ihaid:
> @ihaid Can you give this a test before we merge it to testnet/staging?

looks go to me!

ihaid (approved):

MaximusHaximus:
<3 rock and roll!

PR #1495, opened 2021-02-25
Track/update without local

icerove:
1. tracking with variable BROWSWER_MIXPANEL_TOKEN only
2. delete unused code

### DIFF
--- a/src/mixpanel/index.js
+++ b/src/mixpanel/index.js
@@ -53,4 +53,4 @@ if (process.env.REACT_APP_NETWORK_ID || process.env.REACT_APP_IS_MAINNET) {
     };
 }
 
-export {Mixpanel};
+export { Mixpanel };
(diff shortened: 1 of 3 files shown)


src/mixpanel/index.js
```diff
@@ -2,40 +2,55 @@ import mixpanel from 'mixpanel-browser';
...
-            errorOpration(e)
-        } finally {
-            await finalOperation()
-        }
-  }
+    withTracking: () => {}
 };
+
+if (process.env.REACT_APP_NETWORK_ID || process.env.REACT_APP_IS_MAINNET) {
+    mixpanel.init(BROWSER_MIXPANEL_TOKEN);
```
vgrichina, line 18: let's pass `BROWSER_MIXPANEL_TOKEN` in environment variable similar to `MIXPANEL_TOKEN`

```
    mixpanel.init(process.env.MIXPANEL_TOKEN, {'property_blacklist': ['$current_url']})
}
```

then if it's not set – tracking can be disabled, no need to tie it to specific networks

icerove: need to add variable `BROWSER_MIXPANEL_TOKEN` to the testnet and betanet, too. But I forget the config file link. @vgrichina

PR #1502, opened 2021-02-27
Fix validate account ID from URL

Patrick1904:
Fixes issue with 'submit' button being disabled when going to `/send-money` with account ID in the url.

To test:
1. Go to `/send-money/aValidAccountId`
2. Verify that the 'submit' button is enabled once you have entered a valid amount

### DIFF
--- a/src/components/send/SendContainer.js
+++ b/src/components/send/SendContainer.js
@@ -45,7 +45,7 @@ const StyledContainer = styled(Container)`
 export function SendContainer({ match, location }) {
     const dispatch = useDispatch()
     const { accountId, balance } = useSelector(({ account }) => account);
-    const { localAlert, mainLoader } = useSelector(({ status }) => status);
+    const { localAlert, mainLoader, actionStatus } = useSelector(({ status }) => status);
     const [useMax, setUseMax] = useState(null)
     const [amount, setAmount] = useState('')
     const [confirm, setConfirm] = useState(null)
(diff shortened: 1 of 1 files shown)


src/components/send/SendContainer.js
```diff
@@ -74,8 +75,10 @@ export function SendContainer({ match, location }) {
     }, [location.key])
 
     useEffect(() => {
-        dispatch(getBalance())
-    }, [])
+        if (id && actionStatus.GET_BALANCE?.success) {
```
vgrichina, line 78: @Patrick1904 fix works alright, but not sure I understand the logic, may you explain?

Patrick1904: Yep there are two actions dispatched when the user lands fresh on the page: `GET_BALANCE` and `CHECK_ACCOUNT_AVAILABLE`. `GET_BALANCE` tends to finish **last**, which overrides the `localAlert` needed to show the validation from `CHECK_ACCOUNT_AVAILABLE`. So I'm just waiting for `GET_BALANCE.success` to be `true` before dispatching the account validation, that way it's the last action to finish.

vgrichina: @Patrick1904 any reason not to await on `dispatch(getBalance())` then? Or maybe just not rely on `localAlert` for form validation and kick off these in parallel to avoid delay?

in any case I think we can merge, just let's make sure to have a tracking issue for cleaner fix

Patrick1904: @vgrichina yea so `GET_BALANCE` gets dispatched as part of account refresh when user lands on the page, that's why I removed it from the component because it was already being dispatched. I agree it's not a super clean solution. @marcinbodnar do you have any idea how to avoid this with localAlert?

PR #1470, opened 2021-02-20
New Navigation, Dashboard, Explore Apps, and UI for Fungible Tokens

Patrick1904:
Changes in preparation for Fungible tokens:
1. New Dashboard/Wallet page
2. New navigation
3. Moved authorized apps from dashboard to profile
4. 'Explore Apps' module on dashboard

Fungible tokens UI is ready (currently hidden in `src/components/Wallet.js`)

PR #1505, opened 2021-03-01
Add balance to every account in Switch account section on the top nav (#1496)

marcinbodnar:
1. Request is starting when the submenu is opened.
2. Balance for accounts is fetched sequentially, to make it a bit lighter.
3. It needed some additional work including `localAlert` mechanism and `mainLoader` - because we rely on `mainLoader` for some local actions, I think that in future works we should avoid using `mainLoader` and instead use `actionsPending` for specific actions. `mainLoader` should be used if we decide to add main loader to Wallet.
4. Also added `Refresh` button for every account balance, it might be handy for users if, for example, you want to check if balance changed without switching account.

Please try to open the submenu, and send money quickly. It works well for 7 accounts or less, but if you have more, you would need to wait longer for the transfer to finish. It might be longer with a slow network.

Not many users have a lot of accounts, but I would suggest adding some improvements in the future, some propositions listed in this comment:
https://github.com/near/near-wallet/pull/1470#issuecomment-785319046

### DIFF
--- a/src/components/common/Balance.js
+++ b/src/components/common/Balance.js
@@ -3,6 +3,7 @@ import styled from 'styled-components'
 import { List } from 'semantic-ui-react'
 import { utils } from 'near-api-js'
 import { BN } from 'bn.js'
+import { Translate } from 'react-localize-redux'
 
 const CustomDiv = styled(List)`
     position: relative;
@@ -64,7 +65,7 @@ const Balance = ({ amount, symbol, className }) => {
             {amount && symbol !== false && symbol !== 'near' && <span className='symbol'>Ⓝ</span>}
             {amount
                 ? amountShow
-                : <div className="dots">Retrieving balance</div>
+                : <div className="dots"><Translate id='loadingNoDots'/></div>
             }
             {amount && symbol === 'near' && <span className='currency'>&nbsp;NEAR</span>}
         </CustomDiv>
--- a/src/components/common/Balance.js
+++ b/src/components/common/Balance.js
@@ -24,7 +24,7 @@ const CustomDiv = styled(List)`
         :after {
             content: '.';
             animation: link 1s steps(5, end) infinite;
-        
+
             @keyframes link {
                 0%, 20% {
                     color: rgba(0,0,0,0);
@@ -64,7 +64,7 @@ const Balance = ({ amount, symbol, className }) => {
             {amount && symbol !== false && symbol !== 'near' && <span className='symbol'>Ⓝ</span>}
             {amount
                 ? amountShow
-                : <div className="dots">Loading</div>
+                : <div className="dots">Retrieving balance</div>
             }
             {amount && symbol === 'near' && <span className='currency'>&nbsp;NEAR</span>}
         </CustomDiv>
--- a/src/components/navigation/Navigation.js
+++ b/src/components/navigation/Navigation.js
@@ -54,6 +54,10 @@ class Navigation extends Component {
         const desktopMenu = document.getElementById('desktop-menu');
         const mobileMenu = document.getElementById('mobile-menu');
 
+        if (e.target.tagName === 'SPAN') {
+            return false
+        }
+
         if (e.target.tagName === 'BUTTON' || e.target.tagName === 'A' || (!desktopMenu.contains(e.target) && !mobileMenu.contains(e.target))) {
             this.setState({ menuOpen: false });
         }
--- a/src/components/navigation/Navigation.js
+++ b/src/components/navigation/Navigation.js
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import { withRouter } from 'react-router-dom';
-import { switchAccount, getAvailableAccountsBalance, getAccountBalance } from '../../actions/account';
+import { switchAccount, getAvailableAccountsBalance, getAccountBalance, getBalance } from '../../actions/account';
 import { connect } from 'react-redux';
 import DesktopContainer from './DesktopContainer';
 import MobileContainer from './MobileContainer';
@@ -95,6 +95,7 @@ class Navigation extends Component {
                     showNavLinks={this.showNavLinks}
                     flowLimitation={flowLimitation}
                     refreshBalance={this.props.getAccountBalance}
+                    getBalance={this.props.getBalance}
                     {...this.props}
                 />
                 <MobileContainer
@@ -121,7 +122,8 @@ const mapStateToProps = ({ account, availableAccounts, router, flowLimitation })
 const mapDispatchToProps = {
     switchAccount,
     getAvailableAccountsBalance,
-    getAccountBalance
+    getAccountBalance,
+    getBalance
 }
 
 export default connect(
--- a/src/components/navigation/Navigation.js
+++ b/src/components/navigation/Navigation.js
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import { withRouter } from 'react-router-dom';
-import { switchAccount } from '../../actions/account';
+import { switchAccount, getAvailableAccountsBalance, getAccountBalance } from '../../actions/account';
 import { connect } from 'react-redux';
 import DesktopContainer from './DesktopContainer';
 import MobileContainer from './MobileContainer';
@@ -114,7 +114,8 @@ const mapStateToProps = ({ account, availableAccounts, router, flowLimitation })
 })
 
 const mapDispatchToProps = {
-    switchAccount
+    switchAccount,
+    getAvailableAccountsBalance,
 }
 
 export default connect(
--- a/src/components/navigation/Navigation.js
+++ b/src/components/navigation/Navigation.js
@@ -65,6 +65,10 @@ class Navigation extends Component {
     }
 
     toggleMenu = () => {
+        if (!this.state.menuOpen) {
+            this.props.getAvailableAccountsBalance()
+        }
+
         this.setState(prevState => ({
             menuOpen: !prevState.menuOpen
         }));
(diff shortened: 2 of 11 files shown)


marcinbodnar:
@corwinharrell Please propose an icon for the refresh, I added a random icon for now
image

src/components/navigation/Navigation.js
```diff
@@ -54,6 +54,10 @@ class Navigation extends Component {
         const desktopMenu = document.getElementById('desktop-menu');
         const mobileMenu = document.getElementById('mobile-menu');
 
+        if (e.target.tagName === 'SPAN') {
+            return false
```
Patrick1904, line 58: I don't think we need this since the 'refresh' icon is within the menu div, maybe I'm missing something? Also a side effect of using 'SPAN' here is that clicking on a span on the page that's outside of the menu won't close the menu now.

src/components/navigation/UserAccounts.js
```diff
@@ -38,24 +48,54 @@ const Account = styled.div`
...
+
     &.additional-account {
         :last-of-type {
             border-bottom: 0;
         }
+
+        .refresh {
+            position: absolute;
+            width: 20px;
+            height: 100%;
```
Patrick1904, line 64: Nice addition to be able to 'refresh' the account balance. Can you please also center the icon vertically (once @corwinharrell has provided a new icon)

Patrick1904: Also, very much a nit, but could skip a lot of the style (and absolute positioning) if you make the icon a component and use flexbox to position the elements within the `<Account>`, similar to how it's done here: https://github.com/near/near-wallet/blob/master/src/components/navigation/NavLinks.js#L79

Patrick1904:
> Please try to open the submenu, and send money quickly. It works well for 7 accounts or less, but if you have more, you would need to wait longer for the transfer to finish. It might be longer with a slow network.

@marcinbodnar yes I tried first with five accounts and the loading is reasonable. >5 accounts it can get slow. Since you added the 'refresh' button, maybe we can add a cap to refresh the first 5 accounts only? Like you suggested before. And the user can still click 'refresh' on an individual account not included in the first 5? It's probably rare with more than five accounts, but could be good with some cap. cc @corwinharrell

src/components/navigation/Navigation.js
marcinbodnar: So the `contains` doesn't seems to work with this element - I guess it's because it's added dynamically - the same is with switching accounts, it closes menu but it's not BUTTON either A. I will test it.

src/components/navigation/UserAccounts.js
marcinbodnar: Sure. I'm actually not sure about this design, just added it there to make it work - but maybe, the icon should be added just after the balance amount? @corwinharrell maybe you can propose something nice here?

marcinbodnar:
> > Please try to open the submenu, and send money quickly. It works well for 7 accounts or less, but if you have more, you would need to wait longer for the transfer to finish. It might be longer with a slow network.
> 
> @marcinbodnar yes I tried first with five accounts and the loading is reasonable. >5 accounts it can get slow. Since you added the 'refresh' button, maybe we can add a cap to refresh the first 5 accounts only? Like you suggested before. And the user can still click 'refresh' on an individual account not included in the first 5? It's probably rare with more than five accounts, but could be good with some cap. cc @corwinharrell

Sure, @corwinharrell are you ok with that?

vgrichina:
@marcinbodnar I think my main concern would be to test whether this in any way impacts sign in / transaction signing flow for apps.

vgrichina:
Also I think we should assume that many users would have more than 5 accounts. Given that all your transactions are public it's the only way to keep stuff somewhat less public.

marcinbodnar:
> @marcinbodnar I think my main concern would be to test whether this in any way impacts sign in / transaction signing flow for apps.

Right, we disabled the submenu for the Sign page, so it's not possible to lunch these actions. But the submenu is still available for the Login page - I guess we should do the same (cc @Patrick1904) ?

corwinharrell:
@marcinbodnar @Patrick1904 made some refinements to the styles for the account dropdown. Only issue I'm encountering now is that the menu closes when you click on `Sync` (It may be because I switched to using an actual button element for the Sync action). If one of you could review my changes then it should be good to go once that issue is addressed.

marcinbodnar:
> @marcinbodnar @Patrick1904 made some refinements to the styles for the account dropdown. Only issue I'm encountering now is that the menu closes when you click on `Sync` (It may be because I switched to using an actual button element for the Sync action). If one of you could review my changes then it should be good to go once that issue is addressed.

@corwinharrell OK, I will check that. There is also an issue with a lot of accounts, please check this comment:
https://github.com/near/near-wallet/pull/1505#issuecomment-790125316

src/components/common/Balance.js
```diff
@@ -64,7 +64,7 @@ const Balance = ({ amount, symbol, className }) => {
             {amount && symbol !== false && symbol !== 'near' && <span className='symbol'>Ⓝ</span>}
             {amount
                 ? amountShow
-                : <div className="dots">Loading</div>
+                : <div className="dots">Retrieving balance</div>
```
Patrick1904, line 67: @marcinbodnar can we please make this a translate string too

marcinbodnar: @corwinharrell It's not a good idea to change this, this component is used to show all amounts in Near in Wallet - if you don't like `loading` we can change it but it should be more universal and also shorter than `Retrieving balance`.

corwinharrell: @marcinbodnar I see. My assumption was that due to the fact that this was called "Balance", that it was scoped to just balances and not all amounts. I thought that saying "retrieving balances..." was more specific and insightful. If this component goes beyond balances and is used for other amounts as well, perhaps we can change its name to reflect its more general usage?

marcinbodnar: @corwinharrell true, it does not reflect its usage anymore. I will add a translation for now, would you like me to change `Loading` for something different?

corwinharrell: @marcinbodnar if this component is more general-purpose than just balances I think we can leave it as "Loading..." Eventually I'd like to take a pass at some of our loading states and introduce some nicer loaders etc...

corwinharrell:
> > > Please try to open the submenu, and send money quickly. It works well for 7 accounts or less, but if you have more, you would need to wait longer for the transfer to finish. It might be longer with a slow network.
> > 
> > 
> > @marcinbodnar yes I tried first with five accounts and the loading is reasonable. >5 accounts it can get slow. Since you added the 'refresh' button, maybe we can add a cap to refresh the first 5 accounts only? Like you suggested before. And the user can still click 'refresh' on an individual account not included in the first 5? It's probably rare with more than five accounts, but could be good with some cap. cc @corwinharrell
> 
> Sure, @corwinharrell are you ok with that?

I think this is fine and we can monitor feedback and make adjustments iteratively

vgrichina:
@marcinbodnar are there any changes needed based on @corwinharrell response or good to go with change as is?

marcinbodnar:
> @marcinbodnar are there any changes needed based on @corwinharrell response or good to go with change as is?

@vgrichina it's ready to go now.

vgrichina:
> > @marcinbodnar are there any changes needed based on @corwinharrell response or good to go with change as is?
> 
> @vgrichina it's ready to go now.

somehow I'm getting stuck on https://github.com/near/near-wallet/issues/1557

vgrichina:
@marcinbodnar seems like better not to merge it before https://github.com/near/near-wallet/issues/1516 lands  because of  https://github.com/near/near-ops/issues/441

marcinbodnar:
> @marcinbodnar seems like better not to merge it before #1516 lands because of [near/near-ops#441](https://github.com/near/near-ops/issues/441)

@vgrichina right, I'm finishing #1516 - should be ready today.

marcinbodnar:
@vgrichina  Added limitation for Login page, so the accounts balances will not be fetched automatically when submenu opens.
Should be ready to go.

Patrick1904:
Looks good! @marcinbodnar it seems that clicking the 'sync' button on the active account doesn't dispatch any `GET_ACCOUNT_BALANCE` action. 

Screen Shot 2021-04-07 at 2 23 08 PM

src/components/navigation/UserAccounts.js
```diff
@@ -13,49 +19,159 @@ const Wrapper = styled.div`
...
+            />
         ))}
     </Wrapper>
 )
 
-export default UserAccounts;
\ No newline at end of file
+const UserAccount = ({ accountId, balance, balanceLoading, refreshBalance, active, onClick }) => (
+    <Account className={active ? 'active-account' : 'additional-account'}>
+        <div className='account-data' onClick={onClick}>
```
corwinharrell, line 149: The tap target here only spans the width of the text. This results in nothing happening if I click the space in-between the account name and the "Sync" button. Ideally, the tap target should span the width of the container so the user can click anywhere other than the "Sync" button to switch to a different account.

corwinharrell (changes requested):

src/components/navigation/UserAccounts.js
marcinbodnar: good catch @corwinharrell , it should be good now.

marcinbodnar:
> Looks good! @marcinbodnar it seems that clicking the 'sync' button on the active account doesn't dispatch any `GET_ACCOUNT_BALANCE` action.
> 
> Screen Shot 2021-04-07 at 2 23 08 PM

Thanks @Patrick1904 , it didn't work on mobile - should be good now.

Patrick1904:
@marcinbodnar, seems I'm still not able to click anywhere inside the box to select the account. I think you may have to move the `onClick` to the box vs just the text wrapper div. And after that, you may need to do `stopPropagation` whenever user clicks 'sync' as it will register as click on the parent div too otherwise. 

Also small thing: There's no pointer cursor whenever I hover on the 'sync' button on the active account.

Patrick1904 (approved):

corwinharrell (approved):

PR #1509, opened 2021-03-02
finish build

icerove:
1. delete some private property for user
2. add network id property

### DIFF
--- a/src/components/profile/Profile.js
+++ b/src/components/profile/Profile.js
@@ -180,7 +180,6 @@ export function Profile({ match }) {
     }, []);
 
     useEffect(() => {
-        console.log(account)
         if (userRecoveryMethods) {
             let id = Mixpanel.get_distinct_id()
             Mixpanel.identify(id)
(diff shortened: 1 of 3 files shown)


src/components/profile/Profile.js
```diff
@@ -181,25 +180,16 @@ export function Profile({ match }) {
...
+                enabled_2FA: account.has2fa
             })
             Mixpanel.alias(accountId)
-        }
-    },[account.balance?.total])
-
-    useEffect(() => {
-        if (userRecoveryMethods) {
-            let id = Mixpanel.get_distinct_id()
-            Mixpanel.identify(id)
```
Patrick1904, line 202: @icerove - we should remove `detail_2FA: twoFactor.detail`

src/components/profile/Profile.js
```diff
@@ -181,25 +180,16 @@ export function Profile({ match }) {
     }, []);
 
     useEffect(() => {
-        if (account.balance?.total) {
+        console.log(account)
```
vgrichina, line 183: nit: need to remove leftover console.log

Patrick1904 (approved):

PR #1518, opened 2021-03-04
Fix activity overflow

Patrick1904:


Patrick1904:
@vgrichina do you think we could hotfix this one? It fixes some overflows for long account IDs and amounts. Should have tested that more thoroughly.

PR #1523, opened 2021-03-04
Do not require adding access key on login

vgrichina:


vgrichina:
@corwinharrell wondering if we need to let user know somehow if app is not requesting to add a key when logging in (basically only asks for account ID)

chadoh (approved):

Issue #114, opened 2019-07-22
Investigate whether some CSS/JS modules can be removed easily


vgrichina:
E.g. looks like we include semantic-ui CSS which is 600K uncompressed and 90K compressed. We likely don't use most of this stuff.

vgrichina:
@marcinbodnar @Patrick1904 can you take a look and make rough estimate on how long it will take to get rid of semantic-ui?

I assume it should be possible to provide our internal shims for `Grid`, `List`, `Form`, `Image` and other commonly used components to avoid changing much code.

marcinbodnar:
@vgrichina My rough estimate would be about 40-50h. We got rid of a lot of it since we last talk about it, but there are still some problematic elements.

vgrichina:
@marcinbodnar so based on current state looks like following components are still used from semantic-ui-react:

- [ ] Image
- [ ] Button
- [ ] List
- [ ] Form
- [ ] Segment
- [ ] Grid
- [ ] Header
- [ ] Container
- [ ] Divider 
- [ ] Popup
- [ ] Modal
- [ ] Responsive
- [ ] Input

I think we need to determine which of these we want to re-implement in wallet (probably makes sense for most commonly used) and which to just remove altogether (makes sense for least commonly used).

I assume we already made a lot of progress since last discussion about this, right?

Patrick1904:
> @marcinbodnar so based on current state looks like following components are still used from semantic-ui-react:
> 
> * [ ]  Image
> * [ ]  Button
> * [ ]  List
> * [ ]  Form
> * [ ]  Segment
> * [ ]  Grid
> * [ ]  Header
> * [ ]  Container
> * [ ]  Divider
> * [ ]  Popup
> * [ ]  Modal
> * [ ]  Responsive
> * [ ]  Input
> 
> I think we need to determine which of these we want to re-implement in wallet (probably makes sense for most commonly used) and which to just remove altogether (makes sense for least commonly used).
> 
> I assume we already made a lot of progress since last discussion about this, right?

@vgrichina thanks for doing the audit. The Popup is the main component still missing. It'll take a few days to replace/remove the remaining semantic-ui-react stuff. I'll make sure to prioritize this better. @marcinbodnar we can coordinate async.

marcinbodnar:
@Patrick1904 sure, at this moment I have a lot of things on my plate, that are a bit urgent. But I will have that in mind, I can handle it whenever there will be nothing more urgent.

[...]

PR #1525, opened 2021-03-05
Replace semantic-ui modal/tooltip components

Patrick1904:
This PR is part of current efforts to remove `semantic-ui-react`. Main changes in this PR include:
1. Create hybrid modal/tooltip info component: `Tooltip.js` - and remove all semantic-ui modals/tooltips

https://github.com/near/near-wallet/issues/114


corwinharrell (changes requested):
LGTM, although we may want to add a bit more padding to the right side of the tooltip on mobile so that the close button doesn't overlap the content:

Screen Shot 2021-03-10 at 10 31 28 AM

corwinharrell (changes requested):
LGTM, although we may want to add a bit more padding to the right side of the tooltip on mobile so that the close button doesn't overlap the content:

Screen Shot 2021-03-10 at 10 31 28 AM

corwinharrell (approved):

PR #1531, opened 2021-03-08
fix the tracking for copy

icerove:
1. fix the code for copy account with link 
2. update token

### DIFF
--- a/src/components/common/ClickToCopy.js
+++ b/src/components/common/ClickToCopy.js
@@ -37,7 +37,7 @@ const ClickToCopy = ({ className, children, copy, translate = 'default' }) => {
     const [show, setShow] = useState(false)
 
     const handleCopy = () => {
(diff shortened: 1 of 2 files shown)


src/components/common/ClickToCopy.js
```diff
@@ -37,7 +37,7 @@ const ClickToCopy = ({ className, children, copy, translate = 'default' }) => {
     const [show, setShow] = useState(false)
 
     const handleCopy = () => {
-        Mixpanel.track(`Copy ${copy} with click`)
+        Mixpanel.track(`Click to copy text`, {text: copy})
```
Patrick1904, line 40: @icerove is there a reason why we want to include `copy`? It's the users account ID

Patrick1904: I recommend that we remove `{name: copy}`

Patrick1904 (approved):
Thank you @icerove!

PR #1532, opened 2021-03-09
Update both Chinese translations

picturepan2:
- [ ] Update both Chinese translations to the latest version of NEAR Wallet.

### DIFF
--- a/src/translations/zh-hant.global.json
+++ b/src/translations/zh-hant.global.json
@@ -1,7 +1,7 @@
 {
     "reduxActions": {
         "RECOVER_ACCOUNT_SEED_PHRASE": {
-            "success": "Recover using Seed Phrase is complete.<br/><b>${numberOfAccounts} account(s)</b> have been successfully recovered.<br/>The last one is set as active.",
+            "success": "成功通過助記詞恢復賬戶。<br/><b>${numberOfAccounts} 個賬戶</b>成功恢復，已切換至最新恢復的賬戶。",
             "error": "恢復賬戶失敗。"
         },
         "SEND_MONEY": {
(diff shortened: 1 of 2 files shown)


src/translations/zh-hant.global.json
```diff
@@ -608,27 +636,46 @@
...
     "footer": {
         "copyrights": "NEAR Inc. 版權所有。",
         "termsOfService": "服務條款",
         "privacyPolicy": "隱私策略",
-        "desc": "NEAR 是開放性網絡 (Open Web) 的基礎架構, 提供了一個去中心化存儲和計算平台。",
+        "desc": "NEAR 是開放性網絡 (Open Web) 的基礎架構, 提供了一個去中心化存儲和計算平臺。",
         "learnMore": "了解更多",
         "needHelp": "需要幫助？",
-        "contactSupport": "聯繫客服"
+        "contactSupport": "聯系客服"
```
think-in-universe, line 652: is "聯系" better than "聯繫"? 聯繫 seems to be correct to me.

src/translations/zh-hant.global.json
```diff
@@ -608,27 +636,46 @@
...
         "activity": "動態"
     },
-    "loading": "Loading...",
+
+    "loading": "加載中...",
+
     "balance": {
-        "balance": "總餘額",
-        "balanceLoading": "餘額加載中"
+        "balance": "總余額",
```
think-in-universe, line 662: "餘額" looks OK to me. is it necessary to replace with "余額"?

think-in-universe (commented):
added some questions about the Traditional Chinese translation

src/translations/zh-hant.global.json
```diff
@@ -887,13 +938,13 @@
             "visibleTo": "是否公開",
             "username": "用戶名",
             "public": "公開",
-            "totalBalance": "餘額",
-            "minBalance": "保留餘額",
+            "totalBalance": "余額",
+            "minBalance": "保留余額",
```
think-in-universe: there are other appearances of "余額" that should be replaced with "餘額"

think-in-universe (changes requested):
more "余額" in traditional Chinese that should be replaced with "餘額"

Patrick1904:
Thank you @picturepan2!

PR #1542, opened 2021-03-12
Fix check for public key

Patrick1904:


mehtaphysical:
sorry to bud in on this. But it may be better to fix the order of the dispatched actions here: https://github.com/near/near-wallet/blob/master/src/actions/account.js#L499

right now its:

```js
dispatch(selectAccount(accountId)) // sets the account back to the initial state which doesn't include url
await dispatch(refreshAccount()) // introduces a race condition that can make it so the account.loader state may become true thus allowing PrivateRoutes to display
dispatch(handleRefreshUrl()) // re-adds the account.url state that was removed by selectAccount
```

May be better to do this:

```js
export const switchAccount = (accountId) => async (dispatch, getState) => {
    dispatch(selectAccount(accountId))
    dispatch(handleRefreshUrl())
    dispatch(refreshAccount())
}
```

1. account to initial state
2. re add acount.url state
3. get account information making it possible to display `PrivateRoutes`

Patrick1904:
> sorry to bud in on this. But it may be better to fix the order of the dispatched actions here: https://github.com/near/near-wallet/blob/master/src/actions/account.js#L499
> 
> right now its:
> 
> ```js
> dispatch(selectAccount(accountId)) // sets the account back to the initial state which doesn't include url
> await dispatch(refreshAccount()) // introduces a race condition that can make it so the account.loader state may become true thus allowing PrivateRoutes to display
> dispatch(handleRefreshUrl()) // re-adds the account.url state that was removed by selectAccount
> ```
> 
> May be better to do this:
> 
> ```js
> export const switchAccount = (accountId) => async (dispatch, getState) => {
>     dispatch(selectAccount(accountId))
>     dispatch(handleRefreshUrl())
>     dispatch(refreshAccount())
> }
> ```
> 
> 1. account to initial state
> 2. re add acount.url state
> 3. get account information making it possible to display `PrivateRoutes`

Really appreciate the input @mehtaphysical!

PR #1555, opened 2021-03-17
Fungible token: always use default icon for now

Patrick1904:


Patrick1904:
@kcole16 @corwinharrell @vgrichina - looked at the [contract](https://github.com/near/near-sdk-rs/blob/44f55b057684dacf79e6195d90b1dd877ec1242b/examples/fungible-token/ft/src/lib.rs) and looks risky to just use whatever is being provided as `icon` right now as it could affect the UI negatively. I'm setting it to always use the default token icon for now and will update once we see what's being provided.

corwinharrell:
> looks risky to just use whatever is being provided as icon right now as it could affect the UI negatively

@Patrick1904 can you elaborate?

Issue #1538, opened 2021-03-11
Improve Usability of "Unstake" UX


corwinharrell:
### Problem Summary
In the current implementation of Staking, users have found that the ability to `Unstake` their tokens is difficult to find, since users must first navigate to their current validator's profile in order to `Unstake`.

### Intended Outcome
Users should be able to initiate the `Unstake` action directly from the staking dashboard. Once they indicate that they would like to `Unstake`, they can then confirm the action without having to navigate to the validator's profile.

### How will it work?
1. From /staking, user clicks `Unstake` next to `Total Amount Staked`.
2. If the user is staking with multiple validators, the user must choose the validator from which they want to Unstake.
3. User inputs the amount they wish to Unstake, and clicks Submit
4. User confirms the action via a confirmation dialogue.
5. User returns to dashboard, where their staking figures are updated.

### Designs
**Figma:** https://www.figma.com/file/UGjVUoDOLrAsRbJw6OosyS/1538-Improve-Unstake-UX?node-id=1%3A2597

corwinharrell:
@jimmy3dita I just want to confirm that currently users can only stake with a single validator? Do we know when this might change?

corwinharrell:
@Patrick1904 As long as our staking UI is limited to a single validator, this should be as simple as adding an "Unstake" button to the dashboard next to "Total Amount Staked". Feel free to go ahead and add this. I don't think a mockup is necessary.

As soon as we enable staking with multiple validators, this experience will need to be expanded to allow the user to select the validator from which they would like to unstake, before indicating an amount.

One small request: Let's knock the 's' off of `Your Current Validators`

Patrick1904:
> @Patrick1904 As long as our staking UI is limited to a single validator, this should be as simple as adding an "Unstake" button to the dashboard next to "Total Amount Staked". Feel free to go ahead and add this. I don't think a mockup is necessary.
> 
> As soon as we enable staking with multiple validators, this experience will need to be expanded to allow the user to select the validator from which they would like to unstake, before indicating an amount.
> 
> One small request: Let's knock the 's' off of `Your Current Validators`

@corwinharrell the single validator restriction only applies when you're staking from a lockup account. You're able to stake with unlimited number of validators when you stake from your main account. We can still add the 'unstake' button for lockup case, but might be confusing with inconsistent UX?

corwinharrell:
[...]

PR #1556, opened 2021-03-17
Improve Staking page loading time (#1446)

marcinbodnar:


Patrick1904:
@marcinbodnar took some notes on potential issues below:

1. Load staking dashboard using an account that has a `lockup`. Then switch account. Notice that main account on staking dashboard is switched, but the lockup account still shows, even though the account you just switched to does not.
2. On staking dashboard: Radio button hides/shows when loading - maybe use skeleton loading to avoid. Also the radio button shows when there’s only one account (should only show if there are multiple)
3. Not able to withdraw available tokens from staking pool. Confirm modal closes without any visible error.
2. Staking dashboard visually loads if you navigate to a different page and back again. Can we avoid that (at least visually) like before?
3. Navigate from staking dashboard to profile and back, and notice that only total amount staked and rewards earned load. It looks a bit off that pending release and available for withdrawal are not loading. Can we make them all load?
3. On profile: refresh the page using account that doesn’t have any lockup. Console: `Error in background action: Error: No contract for account`. Pointing to line 301 in `staking.js`
4. On profile: refresh the page using account that has a lockup. Console: `TypeError: Cannot read property 'accountId' of undefined`. Pointing to line 270 in `staking.js`
5. The globalAlert shows for staking/unstaking - worth setting to onlyError
6. Tried unstaking `0.1N` from a `lockup` account. And got the error below. It actually succeeded, and the `0.1N` was unstaked. However, the validator box says `6.41157 NEAR` but should be `~6.3XXXX NEAR`. I don't think it's getting to balance refresh because of the error.
Screen Shot 2021-03-17 at 5 57 48 PM

src/actions/staking.js
```diff
@@ -1,34 +1,426 @@
...
+            if (process.env.REACT_APP_USE_TESTINGLOCKUP && accountId.length < 64) {
+                lockupId = `testinglockup.${accountId}`
+            } else {
+                lockupId = getLockupAccountId(accountId)
+            }
+
+            let contract
+            try {
+                await (await new Account(wallet.connection, lockupId)).state()
+                contract = await new Contract(await wallet.getAccountBasic(accountId), lockupId, { ...lockupMethods })
```
vgrichina, line 299: @marcinbodnar this looks like it wouldn't work with 2FA (because regular `Account`), did you test with 2FA?

src/actions/staking.js
```diff
@@ -1,34 +1,426 @@
...
+                })
+            )).filter((v) => !!v)
+        }
+    }
 })
+
+const handleGetAccounts = () => async (dispatch, getState) => {
+    await dispatch(handleGetLockup())
+
+    return getState().staking?.lockup?.lockupId
```
vgrichina, line 350: nit: IMO more clear here to just have `if` with 2 separate `return`'s

vgrichina (changes requested):
@marcinbodnar looks good overall but I have concerns regarding 2FA account support

Patrick1904:
@marcinbodnar would it be OK for you if we ship https://github.com/near/near-wallet/issues/1538 before this? Or do you want me to wait?

marcinbodnar:
> @marcinbodnar would it be OK for you if we ship #1538 before this? Or do you want me to wait?

@Patrick1904 Please wait, right now I finished all other issues, and will start making changes to this issue - should be ready tomorrow evening.

vgrichina (approved):
Looks good overall but we need to make sure this is tested well with different edge cases:
- normal account
- 2FA account
- account with lockup

Patrick1904 (approved):
LGTM. I've tested stake, unstake, withdraw using lockup + 2FA.

marcinbodnar:
> Looks good overall but we need to make sure this is tested well with different edge cases:
> 
> * normal account
> * 2FA account
> * account with lockup
> 
> (I didn't test thoroughly this time).

@vgrichina I tested it using testnet account, @Patrick1904 tested with mainnet account. But it would be good if Igor also tested it thoroughly. @ihaid could you please test this PR trying every possible action for Staking?

ihaid:
Found one issue, steps:
Go to stake on account with 2fa
Submit and confirm staking and on the 2fa confirm window click "Cancel"
Then you will see that it is hung in this state and the only refresh will help
image

marcinbodnar:
> Found one issue, steps:
> Go to stake on account with 2fa
> Submit and confirm staking and on the 2fa confirm window click "Cancel"
> Then you will see that it is hung in this state and the only refresh will help
> image

Thanks @ihaid it should be good now, please check.

PR #1558, opened 2021-03-18
Fungible token: link to explorer

Patrick1904:
Changes:
1. Show as much as possible of the contract/account string before applying ellipsis
2. Make the contract/account clickable -> explorer

### DIFF
--- a/src/components/wallet/TokenBox.js
+++ b/src/components/wallet/TokenBox.js
@@ -76,7 +76,7 @@ const TokenBox = ({ token }) => {
             <div className='desc'>
                 <span>{token.symbol}</span>
                 <span title={token.contract}>
-                    <a href={`${EXPLORER_URL}/${token.contract}`} target='_blank' rel='noopener noreferrer'>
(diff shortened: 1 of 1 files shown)


src/components/wallet/TokenBox.js
```diff
@@ -62,7 +74,11 @@ const TokenBox = ({ token }) => {
             </div>
             <div className='desc'>
                 <span>{token.symbol}</span>
-                <span title={token.contract}>{token.contract}</span>
+                <span title={token.contract}>
+                    <a href={`https://explorer.near.org/accounts/${token.contract}`} target='_blank' rel='noopener noreferrer'>
```
Patrick1904, line 78: @vgrichina I hard coded the explorer URL since you said the contracts are currently only on mainnet, otherwise the links will be broken on testnet wallet

vgrichina: @Patrick1904 we have `EXPLORER_URL` in wallet.js, no need to hardcode

> since you said the contracts are currently only on mainnet, otherwise the links will be broken on testnet wallet

that was about specific whitelisted contracts, we can whitelist different ones in config for testnet

PR #1562, opened 2021-03-19
Fix button on Choose Security Method page

marcinbodnar:


src/utils/wallet.js
```diff
@@ -689,7 +689,16 @@ class Wallet {
     async setupRecoveryMessageNewAccount(accountId, method, securityCode, fundingOptions, recoverySeedPhrase) {
         const { secretKey } = parseSeedPhrase(recoverySeedPhrase)
         const recoveryKeyPair = KeyPair.fromString(secretKey)
-        await this.validateSecurityCode(accountId, method, securityCode);
+        
+        try {
+            await this.validateSecurityCode(accountId, method, securityCode);
+        } catch (error) {
+            if (error.toString().includes('Error: Invalid code')) {
+                throw new WalletError(error, 'setupRecoveryMessageNewAccount.invalidCode')
```
vgrichina, line 697: Isn’t first parameter supposed to be a string vs error?

src/utils/wallet.js
```diff
@@ -689,7 +689,16 @@ class Wallet {
     async setupRecoveryMessageNewAccount(accountId, method, securityCode, fundingOptions, recoverySeedPhrase) {
         const { secretKey } = parseSeedPhrase(recoverySeedPhrase)
         const recoveryKeyPair = KeyPair.fromString(secretKey)
-        await this.validateSecurityCode(accountId, method, securityCode);
+        
+        try {
+            await this.validateSecurityCode(accountId, method, securityCode);
+        } catch (error) {
+            if (error.toString().includes('Error: Invalid code')) {
```
vgrichina, line 696: Where is this error coming from? Latest near-api-js should have type field in errors.

src/utils/wallet.js
marcinbodnar: @vgrichina I refactored it, `validateSecurityCode` was already handling this error, just an alert was not showing

src/utils/wallet.js
marcinbodnar: @vgrichina It's from CH, no `type`:
https://github.com/near/near-wallet/blob/master/src/utils/wallet.js#L665

src/utils/wallet.js
```diff
@@ -667,7 +667,7 @@ class Wallet {
                 await this.postSignedJson('/account/validateSecurityCode', body);
             }
         } catch(e) {
-            throw new WalletError('Invalid code', 'setupRecoveryMessage.error')
+            throw new WalletError('Invalid code', 'setupRecoveryMessageNewAccount.invalidCode')
```
vgrichina, line 670: @marcinbodnar let's have a check for specific error from contract helper. I believe it provides 401 code + messages on what exactly happened:
https://github.com/near/near-contract-helper/blob/f16b2f8d5d691c4174e5ee33636005f735775220/middleware/2fa.js#L243

marcinbodnar: @vgrichina this error is for email and phone recovery code, so it's this one in CH:
https://github.com/near/near-contract-helper/blob/f16b2f8d5d691c4174e5ee33636005f735775220/app.js#L291

But it throws `recoveryMethod does not exist` which doesn't seem to be relevant. Want me to check this message string anyway?

vgrichina: @marcinbodnar ah, got it

it is relatively relevant, think of it same as when validating user/password you usually just report that there is no user with such password (vs revealing that user exists)

not sure how important though, maybe we should give more granular info

but I think we need to catch this specific error if we are only expecting it

vgrichina: @marcinbodnar anyway, gonna merge – let's fix in follow-up.

marcinbodnar: @vgrichina 
> it is relatively relevant, think of it same as when validating user/password you usually just report that there is no user with such password (vs revealing that user exists)

Right, got it. I will add this check in my next PR.

Issue #1567, opened 2021-03-22
Redirect to the Login page after account import is not working


marcinbodnar:
**Problem**
Redirect to the Login page after account import is not working

**Expected Behavior**
If the user goes to the import account page from Login, he should be redirected back to the Login page with the same URL params after an account is imported successfully.

**Steps to reproduce**

1. Try to login to app.
2. Go to Import Account page
3. Import account
4. You are redirected to Dashboard instead of Login.

Issue #1516, opened 2021-03-03
Make login / signing page load fast


vgrichina:
**Problem**

Looks like right now when you get redirected from app to login / signing page it takes a while (often few seconds) before UI gets ready and you can press "Approve" button.

I guess it's because we are waiting for some stuff to load which is not necessarily blocking (like user's balances, etc?). 

**Expected Behavior**

Account selection dropdown and "Approve" button are active immediately

**Steps to reproduce**

1. Go to http://berryclub.io
2. Try signing in / buying avocados.

PR #1578, opened 2021-03-24
Make login / signing page load fast and fix redirect to app (#1516 #1567)

marcinbodnar:


marcinbodnar:
@vgrichina I refactored previously implemented main menu hiding, so now it will depend on Redux state (instead of checking URL). 

Sign page will load faster, and the Login page will load a lot faster - for Login flow, I removed balance checking (about 1s faster) and also added functionality to fetch limited data for Account (about 0.3s faster). 

Added new reducer `flowLimitation` that will point what data should be fetched and what part of the menu and submenu are available while on Login and Sign flow.

src/actions/account.js
```diff
@@ -6,14 +6,20 @@ import { TwoFactor } from '../utils/twoFactor'
...
+    WALLET_RECOVER_ACCOUNT_URL, 
+    setKeyMeta,
+    MULTISIG_MIN_PROMPT_AMOUNT
 } from '../utils/wallet'
 import { PublicKey, KeyType } from 'near-api-js/lib/utils/key_pair'
 import { WalletError } from '../utils/walletError'
 import { utils } from 'near-api-js'
 import { BN } from 'bn.js'
 import { showAlert, dispatchWithAlert } from '../utils/alerts'
+import { handleflowLimitation, handleClearflowLimitation } from './flowLimitation'
```
vgrichina, line 22: nit: seems like Flow should be capitalized in these

vgrichina:
@marcinbodnar looks good and already is much faster.

However I think it is necessary to do a follow-up PR to avoid loading staked balances, as this can be slow query depending on the situation.  Generally we need to be careful to not block any other UI on staked balances.

marcinbodnar:
> However I think it is necessary to do a follow-up PR to avoid loading staked balances, as this can be slow query depending on the situation. Generally we need to be careful to not block any other UI on staked balances.

@vgrichina Ok, added issue for it: #1580

PR #1579, opened 2021-03-25
Load tokens list from contract helper

vgrichina:
Depends on https://github.com/near/near-contract-helper/pull/322

- Most likely candidates for token contracts loaded from contract helper
- If required methods are missing tokens are ignored
- `localStorage` caching of the metadata


vgrichina:
Needs contract helper stable release to be tested (or test with separate deployment)

Issue #1552, opened 2021-03-17
Incorrect calculations for unlocked amount in lockup


ilblackdragon:
It seems like wallet doesn't account for non default release date, see difference in unlocked amount:
 - https://near.github.io/account-lookup/#cdot.near
 - https://wallet.near.org/profile/cdot.near

See tx with lockup creation here that it's indeed start date at Feb 10: https://explorer.near.org/transactions/HQECeAYXuUTwhGuvvVy2NmxAhjxKFuwiET77D3NTiTDg

jimmy3dita:
Will look into it with @marcinbodnar early next week

PR #1581, opened 2021-03-25
Fix calculating lockup (#1552)

marcinbodnar:


jimmy3dita (approved):
Did some basic regression testing with Ledger Nano X and checking some other accounts, it looks good to me

PR #1582, opened 2021-03-26
Dashboard: Add new desktop layout

Patrick1904:
Desktop changes:
1. Split dashboard page into two columns to better handle a long list of fungible tokens while also showing recent transactions in the same view (without having to scroll)

Screen Shot 2021-03-25 at 5 50 33 PM


corwinharrell (changes requested):
Looks good! Since we've changed the label for the total balance to read "NEAR Balance", let's drop the "NEAR" from the total balance. Otherwise looks gtg!

corwinharrell (approved):

jimmy3dita:
If I had a magic wand, I would switch the "recent activity" and the "tokens" on the page -> recent activity under the total balance, and tokens on the right, under the (collapsible) "explore apps" banner.

I'm worried that most users will squint to find what happened to their account, while almost ignoring the balance of the tokens that won't change.

corwinharrell:
> If I had a magic wand, I would switch the "recent activity" and the "tokens" on the page -> recent activity under the total balance, and tokens on the right, under the (collapsible) "explore apps" banner.
> 
> I'm worried that most users will squint to find what happened to their account, while almost ignoring the balance of the tokens that won't change.

@jimmy3dita In thinking ahead, I'd like the main column of this layout to be primarily concerned with assets and asset management. The recent activity pane on this dashboard will again only be for the most recent items. With a dedicated activity feed that includes more than just a few txns and that has a dedicated route, it won't be necessary to give the smaller list of recent txns so much real estate on the dashboard.

If it helps to envision how this might evolve in the future, here's some exploration/anticipatory design I've been doing on the side to try and help ensure the decisions made now are acting as steps towards enabling future iterations:

Wallet UI Exploration

jimmy3dita:
Love this design direction!

PR #1607, opened 2021-04-03
Dashboard: Add ability to buy more tokens w/ MoonPay

Patrick1904:
Testing instructions:
1. Navigate to Wallet dashboard
2. Click on 'Buy'
3. Click on 'Buy with MoonPay'

Note: MoonPay module only shows up if you are currently in a supported region. VPN works.

Screen Shot 2021-04-07 at 8 00 25 PM


Patrick1904:
@corwinharrell note that it takes 1-2 seconds to check MoonPay support, which is a bit suboptimal. Might want to consider some type of 'Checking MoonPay availability' state for the UI.

vgrichina:
@kcole16 @Patrick1904 @jimmy3dita should we add gate.io as well as it's the only way to buy in US right now AFAIK?

We also can probably link to CoinMarketCap page with more exchanges as fallback:
https://coinmarketcap.com/currencies/near-protocol/markets/

However options convenient in some places don't even seem to be listed there, e.g. https://coins.black

src/utils/moonpay.js
```diff
@@ -0,0 +1,39 @@
+import sendJson from 'fetch-send-json'
+import { ACCOUNT_HELPER_URL } from '../utils/wallet'
+
+const MOONPAY_API_URL = process.env.MOONPAY_API_URL || 'https://api.moonpay.com'
```
vgrichina, line 4: Looks like SetupImplicit still contains following code instead of using this module

src/utils/moonpay.js
```diff
@@ -0,0 +1,39 @@
...
+const MOONPAY_API_KEY = process.env.MOONPAY_API_KEY || 'pk_test_wQDTsWBsvUm7cPiz9XowdtNeL5xasP9'
+export const MOONPAY_BUY_URL = `${process.env.MOONPAY_BUY_URL || 'https://buy.moonpay.io?apiKey='}${MOONPAY_API_KEY}`
+
+export const isMoonpayAvailable = async () => {
+    const moonpayGet = (path) => sendJson('GET', `${MOONPAY_API_URL}${path}?apiKey=${MOONPAY_API_KEY}`)
+    const isAllowed = ({ isAllowed, isBuyAllowed }) => isAllowed && isBuyAllowed
+
+    const ipAddressInfo = await moonpayGet('/v4/ip_address')
```
vgrichina, line 12: @Patrick1904 given you are raising concern about speed – we can try to make stuff load a bit faster like this:

```js
const [ipAddressInfo, countries, currencies] = await Promise.all([
   moonpayGet('/v4/ip_address'),
   moonpayGet('/v3/countries'),
   moonpayGet('/v3/currencies')
])
```

src/components/accounts/SetupImplicit.js
```diff
@@ -292,7 +274,7 @@ class SetupImplicit extends Component {
                                 open={whereToBuy}
                             />
                         }
-                        {createAccount &&
+                        {false &&
```
vgrichina, line 277: @Patrick1904 is this on purpose or debug change?

vgrichina (approved):
Looks good to me overall, but I didn't test

src/components/accounts/SetupImplicit.js
Patrick1904: Thanks for catching @vgrichina! Will fix

corwinharrell (approved):
LGTM 🚀

jimmy3dita:
> @kcole16 @Patrick1904 @jimmy3dita should we add gate.io as well as it's the only way to buy in US right now AFAIK?
> 
> We also can probably link to CoinMarketCap page with more exchanges as fallback:
> https://coinmarketcap.com/currencies/near-protocol/markets/
> 
> However options convenient in some places don't even seem to be listed there, e.g. https://coins.black

This may change based on the selected language too. 
I agree that we should add at least gate.io, as it's the only way to buy from the US. Looking into the opportunity to generate a bit of revenues from the referrals

vgrichina:
> This may change based on the selected language too.

@jimmy3dita Moonpay even giving us geo IP information already, so we can show based on that

Patrick1904:
@picturepan2 can you please add translations for this when you have a chance.

PR #1644, opened 2021-04-09
Improve unstake & withdraw UX

Patrick1904:
Main changes:
1. Improve unstake/withdraw UX by adding an 'Unstake' and a 'Withdraw' button on main staking dashboard (w/ alternative routes)
2. Minor style changes

Screen Shot 2021-04-13 at 2 45 28 PM


corwinharrell (approved):
I only tested unstake due to the release period required for withdrawal, but it looks good and if everything is consistent then it feels GTG.

PR #1649, opened 2021-04-13
Add ToS modal to /create

Patrick1904:


corwinharrell (approved):

corwinharrell:
LGTM. @jimmy3dita Perhaps you can come up with a better paragraph of copy for modal itself? Not sure if there's a template we can follow or something. I can look as well.

Patrick1904:
@jimmy3dita should we change the ToS and Privacy Policy links in the footer too? If so, should both links go to the same PDF?

Screen Shot 2021-04-13 at 1 14 31 PM

vgrichina:
@corwinharrell let's merge as is and @jimmy3dita can update copy later if needed

Patrick1904:
@picturepan2 can you please add translations for this when you have a chance.

PR #1657, opened 2021-04-14
Add low-level Mixpanel event to facilitate MAU and DAU calculations

icerove:
update wallet balance loading event as Stefano asked

add event loading to calculate how many accounts are open and how many times are open

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "fetch-send-json": "0.0.2",
     "js-sha256": "^0.9.0",
     "lodash.once": "^4.1.1",
+    "lodash.throttle": "^4.1.1",
     "mixpanel-browser": "^2.41.0",
     "near-api-js": "^0.39.0",
     "near-ledger-js": "^0.1.2",
--- a/src/components/Routing.js
+++ b/src/components/Routing.js
@@ -60,6 +60,7 @@ import { Mixpanel } from "../mixpanel/index";
 import classNames from '../utils/classNames';
 import Terms from './terms/Terms';
 import { getAccountIsInactive } from '../utils/localStorage'
+import { reportUiActiveMixpanelThrottled } from '../utils/reportUiActiveMixpanelThrottled';
 
 const theme = {}
 
@@ -182,6 +183,8 @@ class Routing extends Component {
         const { account } = this.props;
 
         const isInactiveAccount = account && (getAccountIsInactive(`${account.localStorage?.accountId || account.accountId}`) || account.accountHelperWalletState?.fundedAccountNeedsDeposit);
+
+        reportUiActiveMixpanelThrottled();
 
         return (
             <Container className={classNames(['App', {'network-banner': (!IS_MAINNET || SHOW_PRERELEASE_WARNING)}])} id='app-container'>
--- a/src/components/profile/Profile.js
+++ b/src/components/profile/Profile.js
@@ -21,6 +21,7 @@ import { Mixpanel } from "../../mixpanel/index"
 import AuthorizedApp from './authorized_apps/AuthorizedApp'
 import FormButton from '../common/FormButton'
 import Tooltip from '../common/Tooltip'
+import { reportUiActiveMixpanelThrottled } from '../../utils/reportUiActiveMixpanelThrottled';
 
 
 const StyledContainer = styled(Container)`
@@ -137,6 +138,8 @@ export function Profile({ match }) {
     const userRecoveryMethods = recoveryMethods[account.accountId]
     const twoFactor = has2fa && userRecoveryMethods && userRecoveryMethods.filter(m => m.kind.includes('2fa'))[0]
     const profileBalance = selectProfileBalance(account.balance)
+
+    reportUiActiveMixpanelThrottled();
 
     useEffect(() => {
         if (!loginAccountId) {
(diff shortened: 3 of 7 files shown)


Patrick1904:
@icerove can you please provide some context on what exactly you're wanting to track here?

icerove:
@Patrick1904 
add event loading to calculate how many accounts are open and how many times are open

vgrichina:
> @Patrick1904
> add event loading to calculate how many accounts are open and how many times are open

what do you mean by "accounts are open", how many accounts are created by user?

@icerove  not sure why tracking this particular event specifically as it's gonna be fired like every time user goes to Wallet page I think

icerove:
@vgrichina 
> not sure why tracking this particular event specifically as it's gonna be fired like every time user goes to Wallet page I think

yes, this is what Stefano want to track

vgrichina:
@jimmy3dita may you clarify what do you want to track, as I still don't understand

jimmy3dita:
> @jimmy3dita may you clarify what do you want to track, as I still don't understand

Measure the number of unique users opening their wallet via https://wallet.near.org, to check the balance or the staking rewards, without signing any transaction or doing any on-chain activity.
This is critical to measure the Active Users (DAU and MAU).

It's unnecessary to measure more than once per user, on a daily basis (CET), and monthly basis (moving, last 30 days).
If the user `x` opens the wallet `x.near`, it will trigger +1 on the DAU, and for 24 hours we can ignore any further event of this type for this user.

icerove:
@VGasparini 
@Patrick1904

Patrick1904:
> > @jimmy3dita may you clarify what do you want to track, as I still don't understand
> 
> Measure the number of unique users opening their wallet via https://wallet.near.org, to check the balance or the staking rewards, without signing any transaction or doing any on-chain activity.
> This is critical to measure the Active Users (DAU and MAU).
> 
> It's unnecessary to measure more than once per user, on a daily basis (CET), and monthly basis (moving, last 30 days).
> If the user `x` opens the wallet `x.near`, it will trigger +1 on the DAU, and for 24 hours we can ignore any further event of this type for this user.

@stefanopepe @icerove it seems to me it may not be sufficient. Currently the `wallet balance loading` event is fired if:

1. User recovers an account and lands on the dashboard
2. User goes to wallet.near.org while logged in 
3. User refreshes the page while on the dashboard
4. User switches account while on the dashboard

A few scenarios that would result in the event not being fired:

1. User doesn't visit the dashboard
2. User refreshes the page on profile, etc., and then navigates to dashboard
3. User uses the Wallet to log into an app or sign an app transaction

vgrichina:
Is there any reason why we cannot calculate DAU/MAU over user sending _any_ event vs one specific?

stefanopepe:
> Is there any reason why we cannot calculate DAU/MAU over user sending _any_ event vs one specific?

Any event is good, we just want to measure how often some user gets back to the wallet and simply look around. 
Here's the catch: if I got right how MixPanel works, we may end up harvesting godzillions of useless events from the same user, while we simply need "the first one for today".

stefanopepe:
@MaximusHaximus adding you to the conversation - I want to understand if MAU and DAU metric is measurable with this work, or we need @icerove to rework this

MaximusHaximus:
> > Is there any reason why we cannot calculate DAU/MAU over user sending _any_ event vs one specific?
> 
> Any event is good, we just want to measure how often some user gets back to the wallet and simply look around.
> Here's the catch: if I got right how MixPanel works, we may end up harvesting godzillions of useless events from the same user, while we simply need "the first one for today".

I think that fetching account balance is probably the lowest-level single action we could rely on for this.  It does make sense that we don't need to report an event for _every single time a users balance is fetched_, as this will indeed create a ton of events in Mixpanel that we don't care about.

The simplest way to avoid spamming but still get some signal would be to put this action report behind a throttled call that can only be sent every X minutes.  I think if we were to make it only report once every 60 minutes it would still be sufficient for our purposes.

While we could try to do something fancy like only trigger it "once per day", even the concept of a "day" is problematic; where do we consider a day to start and where does it end, and do those hours make sense for a user in a different timezone than we are in?

I tend to lean towards getting a few more events than we technically need but be relatively sure we _do get events_ when someone uses the app.  I'll rebase this on to current `master` and make this only report once every 60 minutes at the most, then I think we can merge this and see where we're at from the dashboards side of things :)

MaximusHaximus:
@stefanopepe @icerove This event would have fired _only_ any time someone navigated to the `Wallet` tab of the UI, max once per hour.

- It would **not** fire if someone navigates _from_ the `Wallet` tab to the `Staking` tab
- It would **not** fire if someone navigates from the `Wallet` tab or `Staking` tab to the `Account` tab.
- It would **_only_** fire when someone opens the `Account` tab or switches accounts _while already on the `Account` tab_.

However, from Stefano's message in this thread, it sounds like we want our 'active' users counts to include people who are already on the Staking page using the app, or who are already on the Account page using the app, in addition to people who are on the Wallet page using the app.

As a result, I made two changes to this to facilitate this happening:
1. I added the reporting to the Staking and Account pages (it was already on the Wallet page)
2. I renamed the action string we are reporting to Mixpanel to be identical on all 3 pages for consistency -- it reports now `ui is active`.

I think this is ready for review + landing at this point.

Patrick1904 (commented):
LGTM. Suggestions:
1. Place `reportUiActiveMixpanelThrottled` in `render()` inside `Routing.js` to cover all routes/pages (including `/login` and `/sign`). It will re-render when switching account too.
2. Remove `useRef` in `Wallet.js` since it's not being used

MaximusHaximus:
@Patrick1904 :+1: good ideas -- I made those changes.  If this looks GTG for you, I think we're ready to rock.

Patrick1904 (approved):

PR #1663, opened 2021-04-15
Update Chinese translations

picturepan2:
- [ ] Update both Chinese translations, and fix missing ones.

### DIFF
--- a/src/translations/zh-hans.global.json
+++ b/src/translations/zh-hans.global.json
@@ -482,7 +482,7 @@
         "recoverAccount": "恢复账户",
         "goBack": "返回",
         "createNewAccount": "创建新的账户",
-        "importAccount": "添加其他账户",
+        "importAccount": "导入已有账户",
         "moreInformation": "更多信息",
         "dismiss": "关闭",
         "cancelOperation": "取消操作",
--- a/src/translations/zh-hans.global.json
+++ b/src/translations/zh-hans.global.json
@@ -129,6 +129,9 @@
         },
         "lockup": {
             "transferAllWithStakingPoolBalance": "为将你剩余锁定余额转账至钱包，请从质押池赎回并提现所有质押资产。"
+        },
+        "setupRecoveryMessageNewAccount": {
+            "invalidCode": "无效验证码"
         }
     },
     "account": {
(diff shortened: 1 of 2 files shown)


src/translations/zh-hans.global.json
```diff
@@ -169,7 +172,8 @@
                 },
                 "alertBanner": {
                     "title": "请勿关闭此窗口！"
-                }
+                },
+                "fundWith": "打开"
```
think-in-universe, line 176: `Fund with` should be `支付` ? I don't know exactly where the button is, so may be wrong.

src/translations/zh-hans.global.json
```diff
@@ -478,7 +482,7 @@
         "recoverAccount": "恢复账户",
         "goBack": "返回",
         "createNewAccount": "创建新的账户",
-        "addAccount": "添加其他账户",
+        "importAccount": "添加其他账户",
```
think-in-universe, line 485: `导入其他账户` is preferred if follow the change precisely, but `添加其他账户` is also fine.

src/translations/zh-hant.global.json
```diff
@@ -169,7 +172,8 @@
                 },
                 "alertBanner": {
                     "title": "請勿關閉此窗口！"
-                }
+                },
+                "fundWith": "打開"
```
think-in-universe, line 176: same as above for Simplified Chinese

src/translations/zh-hant.global.json
```diff
@@ -478,7 +482,7 @@
         "recoverAccount": "恢復賬戶",
         "goBack": "返回",
         "createNewAccount": "創建新的賬戶",
-        "addAccount": "添加其他賬戶",
+        "importAccount": "添加其他賬戶",
```
think-in-universe, line 485: same as above for Simplified Chinese

think-in-universe (commented):
added some comments about two strings

src/translations/zh-hans.global.json
picturepan2: 通过 MoonPay 支付的按钮，但是 MoonPay 是图标，「打开 MoonPay」比较合适。

src/translations/zh-hans.global.json
picturepan2: 「导入已有账户」吧

PR #1664, opened 2021-04-15
Redirect for new users onboarded via login button

marcinbodnar:


jimmy3dita:
Looping in Igor, so he can test the whole process.
I tried to open the staging link and works well. My problem now is how to test if it works: all "login with NEAR" buttons send to `https://wallet.near.org` and not `https://deploy-preview-1664--near-wallet-staging.netlify.app/` - so there's no real way to check if the redirect works correctly once the account creation is done.

As far as I understand, `walletUrl:` is a constant that can't be changed via debug console (see [here](https://github.com/evgenykuzyakov/berryclub/blob/ad2b37045b14fa72181ab92831fb741a7c40234b/frontend/src/App.js#L21) for an example), so we need to modify a third-party app to test this one.

@vgrichina any ideas? I think we may use https://examples.near.org/ and deploy a slightly modified application on GitPod (eg. the NFT one). Testing is otherwise blocked

marcinbodnar:
> Looping in Igor, so he can test the whole process.
> I tried to open the staging link and works well. My problem now is how to test if it works: all "login with NEAR" buttons send to `https://wallet.near.org` and not `https://deploy-preview-1664--near-wallet-staging.netlify.app/` - so there's no real way to check if the redirect works correctly once the account creation is done.
> 
> As far as I understand, `walletUrl:` is a constant that can't be changed via debug console (see [here](https://github.com/evgenykuzyakov/berryclub/blob/ad2b37045b14fa72181ab92831fb741a7c40234b/frontend/src/App.js#L21) for an example), so we need to modify a third-party app to test this one.
> 
> @vgrichina any ideas? I think we may use https://examples.near.org/ and deploy a slightly modified application on GitPod (eg. the NFT one). Testing is otherwise blocked

@jimmy3dita You can test it by copying the search part of URL - so for example click `login with NEAR` and copy `/login?...` part of URL and paste it after `https://deploy-preview-1664--near-wallet-staging.netlify.app/`.

jimmy3dita:
> @jimmy3dita You can test it by copying the search part of URL - so for example click `login with NEAR` and copy `/login?...` part of URL and paste it after `https://deploy-preview-1664--near-wallet-staging.netlify.app/`.

I tried, and the redirect is too fast to capture.
I can't "right-click" the login button, it has no link. How do I capture the `/login?...` link via browser console?

jimmy3dita:
I was able to test it, following @marcinbodnar suggestion, and I found slightly inconsistent UI.

To reproduce:
1. open the link `https://berryclub.io`
2. click "login" in the top-left corner
3. store the URL on a text editor for later on
4. open a new incognito window
5. surf the URL https://deploy-preview-1664--near-wallet-staging.netlify.app/
6. replace `https://wallet.near.org/`in the string saved at step 3 with `https://deploy-preview-1664--near-wallet-staging.netlify.app/`

### Create a new account
Once the full funding process is completed, the user is redirected to the login page instead of the wallet dashboard.
The login shows "unknown app"

Schermata 2021-04-16 alle 16 21 44

### Recover from an existing account
If the user recover from a seed passphrase, the page shows also the meta-data of the application asking the redirect:

Schermata 2021-04-16 alle 16 22 20

### Expected Behavior
It may be confusing for a new user to authorize an "unknown application", so the expected behavior is to always display the name of the app. Maybe it's just a small bug?

marcinbodnar:
> I was able to test it, following @marcinbodnar suggestion, and I found slightly inconsistent UI.
> 
> To reproduce:
> 
> 1. open the link `https://berryclub.io`
> 2. click "login" in the top-left corner
> 3. store the URL on a text editor for later on
> 4. open a new incognito window
> 5. surf the URL https://deploy-preview-1664--near-wallet-staging.netlify.app/
> 6. replace `https://wallet.near.org/`in the string saved at step 3 with `https://deploy-preview-1664--near-wallet-staging.netlify.app/`
> 
> ### Create a new account
> Once the full funding process is completed, the user is redirected to the login page instead of the wallet dashboard.
> The login shows "unknown app"
> 
> Schermata 2021-04-16 alle 16 21 44
> 
> ### Recover from an existing account
> If the user recover from a seed passphrase, the page shows also the meta-data of the application asking the redirect:
> 
> Schermata 2021-04-16 alle 16 22 20
> 
> ### Expected Behavior
> It may be confusing for a new user to authorize an "unknown application", so the expected behavior is to always display the name of the app. Maybe it's just a small bug?

@jimmy3dita It's because we use `document.referrer` instead of app title from URL - so if you copy the URL into a new browser tab/window the referrer is undefined - it's not an issue in real-life cases when you click `Login with Near` and proceed in the same tab/window - so the user will always the domain name.

ihaid (approved):

Issue #1086, opened 2020-10-23
Disable 2FA


kcole16:
### Overview
Many users have setup 2FA, only to realize they later want to: 
-  Use Ledger instead
-  Continue without 2FA
-  Create a new account and withdraw their funds

We should allow users to disable (and re-enable) 2FA after setting it up.

### Acceptance Criteria
- [ ] Users can disable 2FA
  - [ ] User can sign TXs with localStorage key
  - [ ] Recovery keys still work
  - [ ] State stake is reduced to < 5 NEAR
- [ ] Users can re-enable 2FA
  - [ ] Able to choose new verification method

mattlockyer:
Requires logic in near-api-js to support:
https://github.com/near/near-api-js/pull/445

Here is PR for wallet including instructions on how to test with above PR of near-api-js:
https://github.com/near/near-wallet/pull/1107

corwinharrell:
@Patrick1904 we're using the same pattern here as we are for disabling other methods right? Just want to make sure nothing additional is needed since it's marked as needing design.

vgrichina:
> Here is PR for wallet including instructions on how to test with above PR of near-api-js:
> #1107

any reason not to point to specific branch temporarily in `package.json` while still in draft?

mattlockyer:
Issues outstanding:

- potential "bad keys" scenario, blocked on indexer (which probably won't work) and a fix
- rust sdk and contract runtime addition for safer key rotations

PR #1667, opened 2021-04-16
Enable Disable 2FA

vgrichina:
Do simplest possible change to enable 2FA (without key rotation as in https://github.com/near/near-wallet/pull/1344)

Key rotation should be unnecessary given nearcore changes around the issue.

Fixes https://github.com/near/near-wallet/issues/1086

### DIFF
--- a/src/components/profile/two_factor/TwoFactorAuth.js
+++ b/src/components/profile/two_factor/TwoFactorAuth.js
@@ -74,9 +74,7 @@ const TwoFactorAuth = ({ twoFactor, history }) => {
                             </div>
                             <div className='detail'>{twoFactor.detail}</div>
                         </div>
-                        {
-                            false && <FormButton onClick={() => setConfirmDisable(true)} className='gray-red'><Translate id='button.disable'/></FormButton>
-                        }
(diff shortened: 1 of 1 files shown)


jimmy3dita:
Following. Is it related to #1086 ? I'm trying to map PRs and Issues

vgrichina:
> Following. Is it related to #1086 ? I'm trying to map PRs and Issues

Thanks, added link to description!

vgrichina:
@jimmy3dita looks like disabling 2FA works now, but looks like an error somehow, gonna debug it

ihaid (approved):

Patrick1904 (approved):

jimmy3dita:
I tested it and I didn't see any errors. 
However, I think we can improve the 2FA message, specifying that we are disabling this feature.

@MaximusHaximus do you have a way to detect this specific action from the contract helper? Currently, the email subject says `Confirm Transaction WARNING - Adding FULL ACCESS KEY to Account: %account%`, while it might specify something in the domain of `Confirm Transaction WARNING - REMOVING 2FA from Account: %account%`.
The email copy should change too (just a few words), so I need an idea of the needed effort.

Issue #1645, opened 2021-04-12
Unlocked and transferrable balance is negative


think-in-universe:
**Problem**

One user with around 80K NEAR reported an issue that the unlocked and transferrable balance is negative (-1,072.9366 NEAR when checked previously, https://wallet.near.org/profile/08d9a90ffd033fdf966c0a04c1feb70627d1a5e7e9328b13ee6a90c322e58bd8). 

So he cannot transfer NEAR as he usually does previously: https://explorer.near.org/accounts/08d9a90ffd033fdf966c0a04c1feb70627d1a5e7e9328b13ee6a90c322e58bd8

**Expected Behavior**

The unlocked balance should be non-negative. 

**Steps to reproduce**

1. Open the wallet profile: https://wallet.near.org/profile/08d9a90ffd033fdf966c0a04c1feb70627d1a5e7e9328b13ee6a90c322e58bd8
2. The unlocked and transferrable balances are negative (-626.9454 NEAR) at the moment. 

![image](https://user-images.githubusercontent.com/46699230/114420836-43ab9100-9be7-11eb-9959-0b09dda3a279.png)

jimmy3dita:
Escalating to Priority 1: prevents the user from spending their balance.
We have an open ticket on Zendesk too

vgrichina:
@jimmy3dita @marcinbodnar according to information from @evgenykuzyakov the source of problem is a bug in earlier version of lockup contract (always using transfers start date instead of lockup start date).

We need to adjust our calculation to account for it.

marcinbodnar:
> @jimmy3dita @marcinbodnar according to information from @evgenykuzyakov the source of problem is a bug in earlier version of lockup contract (always using transfers start date instead of lockup start date).
> 
> We need to adjust our calculation to account for it.

@vgrichina Ok, but if so then the wallet is calculating the locked balance properly. 
 For example for this account `08d9a90ffd033fdf966c0a04c1feb70627d1a5e7e9328b13ee6a90c322e58bd8`, we take `lockup_timestamp` (as it's bigger than `transfers_timestamp`) as a lockup starting date, so it's correct. So the unlocked balance is actually negative. And the only change that we should do is to check if the balance is negative and if so we should show zero Near. Right?

But I wonder how the user ended with this state. We had the same issue on wallet a few weeks ago (always taking `transfers_timestamp` as a starting point), that was fixed. So I guess that before this fix was made, these few users transferred more Near from lockup than they should be able to?

[...]

PR #1668, opened 2021-04-16
Fix calculating unreleased amount (#1645)

marcinbodnar:


vgrichina:
@marcinbodnar we need to check the code hash of contract (return by account.state()) to determine whether to use transfers timestamp or lockup timestamp.

marcinbodnar:
> @marcinbodnar we need to check the code hash of contract (return by account.state()) to determine whether to use transfers timestamp or lockup timestamp.

@vgrichina do you mean that we should check whether code_hash is "11111111111111111111111111111111" or different?

vgrichina:
> @vgrichina do you mean that we should check whether code_hash is "11111111111111111111111111111111" or different?

the value gonna be different in both cases, code_hash "11111111111111111111111111111111" is for when there isn't any contract AFAIK

fixed contract build is here:
https://github.com/near/core-contracts/pull/136#issuecomment-815115064

I think we need to keep a list either of hashes with unfixed code and/or hashes with fixed code.

@frol do you know if there is any way to query indexer for contract hashes?

frol:
@vgrichina we don't save code hashes in indexer database. You can query that with JSON RPC: https://docs.near.org/docs/api/rpc#view-account

frol:
Please, see this fix for getting the real locked/unlocked amounts: https://github.com/near/account-lookup/pull/5

marcinbodnar:
> Please, see this fix for getting the real locked/unlocked amounts: [near/account-lookup#5](https://github.com/near/account-lookup/pull/5)

Thanks @frol

marcinbodnar:
@vgrichina fix ready

src/utils/account-with-lockup.js
```diff
@@ -150,9 +151,13 @@ async function getAccountBalance(limitedAccountData = false) {
 
         const { transfer_poll_account_id, transfers_timestamp } = transferInformation
         const transfersTimestamp = transfer_poll_account_id ? await this.viewFunction(transfer_poll_account_id, 'get_result') : transfers_timestamp
-        const startTimestampBN = lockupTimestamp
-            ? BN.max(new BN(lockupTimestamp), new BN(transfersTimestamp))
-            : new BN(transfersTimestamp)
+
+        const hasBrokenTimestamp = (await lockupAccount.state()).code_hash === '3kVY9qcVRoW3B5498SMX6R3rtSLiCdmBzKs7zcnzDJ7Q' && lockupTimestamp !== null
+        const startTimestampBN = BN.max(
+            new BN(transfersTimestamp),
```
frol, line 157: When the lockup timestamp is correct (in new lockup contract release), lockup start timestamp is:

```js
max(
  transfersTimestamp + lockupDuration,
  lockupTimestamp
)
```

so it should be:

```suggestion
            new BN(transfersTimestamp).addn(lockupDuration),
```

The naming is completely confusing, I know: https://github.com/near/core-contracts/blob/39e029619041b564ec08d33a8d65325b653b121b/lockup/src/types.rs#L37-L39

Ref: https://github.com/near/account-lookup/blob/6980f10bd0905cac9035c9b7011a471eaf1272d2/script.js#L190

`lockupCliff` below should be just `dateNowBN.lt(startTimestampBN)`

frol (changes requested):

src/utils/account-with-lockup.js
```diff
@@ -150,9 +151,13 @@ async function getAccountBalance(limitedAccountData = false) {
...
+            new BN(!hasBrokenTimestamp && (lockupTimestamp || 0))
```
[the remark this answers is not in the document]
marcinbodnar: I know, `new BN(false)` is just '0', so it fine - but I can change it, maybe it would be less confusing

src/utils/account-with-lockup.js
marcinbodnar: @frol Actually, if I understand the lockup cliff correctly, this logic implemented in Wallet is proper and there might be an error in `lockup-account` repo. 

For example, if the account has a 4-year lockup, and a 1-year cliff and releases start date '10/13/2020', then after 1 year from the release start date the account should have 25% unlocked. But in this logic:
https://github.com/near/account-lookup/blob/57ebf8c537452459436cf6733d16109239990528/script.js#L190
the release start point is changed, which propagates to wrong calculation.

That results in the release start date moved one year ahead. Cliff should only result in the number of unlocked Near, not change the lockup schedule.

So if you consider this example (4-year lockup, 1-year cliff) the release will start on 10/13/2021, and after one year from the release start date users will see zero unlocked.

Or am I missing something?

@evgenykuzyakov I think you should double-check it in `core-contracts` repo, as it looks like it has the same logic as `lockup-account`.

src/utils/account-with-lockup.js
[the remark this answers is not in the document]
vgrichina: nit: seems like `hasBrokenTimestamp ? lockupTimestamp : 0` would be more explicit way to write conditional here

src/utils/account-with-lockup.js
vgrichina: > `(await lockupAccount.state()).code_hash === '3kVY9qcVRoW3B5498SMX6R3rtSLiCdmBzKs7zcnzDJ7Q'`

@marcinbodnar let's make it a list of comma separated hashes that can be configured through env variable (as we likely can have newer versions of contract some day as well)

vgrichina: > @frol Actually, if I understand the lockup cliff correctly, this logic implemented in Wallet is proper and there might be an error in `lockup-account` repo.

@marcinbodnar main point here is that we shouldn't implement what is "proper" in wallet implementation. We need to match what lockup contract does (even if it is "incorrect") including compatibility with its bugs.

and lockup contract is functioning as described by @frol:
https://github.com/near/core-contracts/blob/39e029619041b564ec08d33a8d65325b653b121b/lockup/src/getters.rs#L74

frol: > let's make it a list of comma separated hashes that can be configured through env variable (as we likely can have newer versions of contract some day as well)

It is a single known contract version with the bug, all the rest are going to be bug-free.

marcinbodnar: > > @frol Actually, if I understand the lockup cliff correctly, this logic implemented in Wallet is proper and there might be an error in `lockup-account` repo.
> 
> @marcinbodnar main point here is that we shouldn't implement what is "proper" in wallet implementation. We need to match what lockup contract does (even if it is "incorrect") including compatibility with its bugs.
> 
> and lockup contract is functioning as described by @frol:
> https://github.com/near/core-contracts/blob/39e029619041b564ec08d33a8d65325b653b121b/lockup/src/getters.rs#L74

@frol @vgrichina Right, and shouldn't we handle this bug the same, by checking the contract version? Because that's a different bug, resulting in a different problem.

frol: @marcinbodnar Unless we have some update to the lockup contract, there is nothing to workaround here since you won’t be able to withdraw no matter what we want to believe should be “correct”. I am not in charge of the lockup mechanics or implementation

marcinbodnar: @frol @vgrichina 
Yes, I understand that we need to show the same values as a contract no matter what because it's live on the blockchain. 

And that's exactly my point here, with the latest changes in `account-lockup` it will show different values than the lockup contract.

I checked it several times, and I'm pretty sure that I'm right. It's not so obvious, I will try to explain it differently.

If we consider situation:
 - lockup contract: `3kVY9qcVRoW3B5498SMX6R3rtSLiCdmBzKs7zcnzDJ7Q` (previous version)
 - "lockup_duration": "31536000000000000" (1 year)
 - "release_duration": "126144000000000000"  (4 years)
 - "lockup_timestamp": null

So on April 04th 2022 (1 year from now):
1. `account-lockup` will show Liquid amount about 13000 N because it will calculate:
 - lockupReleaseStartTimestamp: 10/13/2021
 - endTimestamp 10/12/2025
2. `core-contracts / get_locked_amount` will return about 38000 N because it will calculate:
 - lockupReleaseStartTimestamp 10/13/2021
 - endTimestamp 10/12/2024

That's because in this `3kVY9qcVRoW3B5498SMX6R3rtSLiCdmBzKs7zcnzDJ7Q` version of the contract the `lockup_timestamp` is calculated wrongly, but it's not taking part in calculating the `end_timestamp`, as you can see here (before the latest change):
https://github.com/near/core-contracts/pull/136/files#diff-2846eb861dc09e92d16b219710c567d81afa363e6d2f5ebc4376385c1d43f168L81

You can simulate this situation locally in `account-lockup` repo by adding one year to `now`. You can use your team-tokens account because I think it's using the same version of the contract as in the example (mine does).

So the correct implementation for this situation should look like this:
https://github.com/near/account-lookup/pull/6
then it will show the same values as the lockup contract.

It seems like there is also a bug in the current version of lockup-contract, I'm just not sure if it was intended. I will add a separate issue in `core-contracts` repo.

marcinbodnar: @frol @vgrichina 
Please check these two repos, you can use your own account from team-tokens, as it will have lockup_duration defined and the same contract:

1. Exact copy of current `account-lookup`, just simulating current date one year from now:
https://marcinbodnar.github.io/account-lookup/

2. Logic from lockup contract with code_hash `3kVY9qcVRoW3B5498SMX6R3rtSLiCdmBzKs7zcnzDJ7Q`, simulating current date one year from now (this is what the old version of the contract will calculate):
https://marcinbodnar.github.io/account-lookup-core/

`Liquid amount` is different.
That's because the old lockup contract don't take the `lockup_timestamp` (`lockupReleaseStartTimestamp`) while calculating `end_timestamp`, and the `account-lookup` do, for the account with `lockup_duretion` defined.

frol: @marcinbodnar @vgrichina you might be interested to join this discussion: https://github.com/near/core-contracts/issues/141

PR #1672, opened 2021-04-20
Fix handling lockups with broken timestamp (#1645)

marcinbodnar:


marcinbodnar:
@vgrichina this is the proper way to handle both versions of lockup contracts - the current and the previous one with the bug.

PR #1692, opened 2021-04-27
Fungible token list: show token icon if available

Patrick1904:


Patrick1904:
@corwinharrell @vgrichina I've changed to use the provided `icon` value as `img` `src`. I'm forcing the size to 32x32px. This is how BANANA and MOON tokens show up:

Screen Shot 2021-04-27 at 2 36 53 PM


cc @evgenykuzyakov

Patrick1904:
@corwinharrell do you think we should force a 50% border-radius or leave it raw? 
Screen Shot 2021-04-27 at 3 00 15 PM
Screen Shot 2021-04-27 at 3 00 27 PM

vgrichina:
Gonna merge this as is for now (so that everyone can get a better feel on staging). Design adjustments can go as follow up.

PR #1703, opened 2021-04-29
Sign error: send back errorType and errorMessage

Patrick1904:


src/components/sign/Sign.js
```diff
@@ -20,10 +20,18 @@ class Sign extends Component {
...
+        const { callbackUrl, meta, actionStatus } = this.props;
         // TODO: Dispatch action for app redirect?
         if (this.props.callbackUrl) {
-            window.location.href = addQueryParams(callbackUrl, { meta, errorCode: 'userRejected' })
+            let errorMessage = 'userRejected';
+            const signTxStatus = actionStatus.SIGN_AND_SEND_TRANSACTIONS;
+            if (signTxStatus?.success === false) {
+                errorMessage = 'Unknown error occured';
+                if (signTxStatus.errorMessage) {
+                    errorMessage = signTxStatus.errorMessage;
```
vgrichina, line 31: is there any way we can pass through `type` from `near-api-js` https://github.com/near/near-api-js/blob/ee975a1edd84dff955302453fbcca1413d362bb7/src/utils/errors.ts#L14? (when it's available)

src/components/sign/Sign.js
```diff
@@ -20,10 +20,18 @@ class Sign extends Component {
...
         Mixpanel.track("SIGN Deny the transaction")
-        const { callbackUrl, meta } = this.props;
+        const { callbackUrl, meta, actionStatus } = this.props;
         // TODO: Dispatch action for app redirect?
         if (this.props.callbackUrl) {
-            window.location.href = addQueryParams(callbackUrl, { meta, errorCode: 'userRejected' })
+            let errorMessage = 'userRejected';
+            const signTxStatus = actionStatus.SIGN_AND_SEND_TRANSACTIONS;
+            if (signTxStatus?.success === false) {
+                errorMessage = 'Unknown error occured';
```
vgrichina, line 29: it's supposed to be code versus a message, so let's do `'unknownError'`

src/components/sign/Sign.js
Patrick1904: @vgrichina good idea! I added `errorType`. Please check again.

Patrick1904:
@vgrichina is there anything blocking this one or do you want me to make any adjustments?

src/components/sign/Sign.js
```diff
@@ -20,10 +20,25 @@ class Sign extends Component {
     handleDeny = e => {
         e.preventDefault();
         Mixpanel.track("SIGN Deny the transaction")
-        const { callbackUrl, meta } = this.props;
+        const { callbackUrl, meta, signTxStatus } = this.props;
```
MaximusHaximus, line 23: 💖  winning

MaximusHaximus (approved):
Lookin' good! error handling++ :)

src/components/sign/Sign.js
```diff
@@ -20,10 +20,25 @@ class Sign extends Component {
...
         // TODO: Dispatch action for app redirect?
         if (this.props.callbackUrl) {
-            window.location.href = addQueryParams(callbackUrl, { meta, errorCode: 'userRejected' })
+
+            if (signTxStatus?.success !== false) {
+                window.location.href = addQueryParams(callbackUrl, {
+                    meta,
+                    errorType: 'userRejected',
+                    errorMessage: 'User rejected transaction'
```
MaximusHaximus, line 31: Sorry about the delayed response here!  Just a re-cap of our discussion in Discord
1. If `errorMessage` is un-bounded and can contain large strings, this could break URL size limits, so we probably don't want to add it unless we are very sure of its content in all cases :)

2. We should keep the code argument as `errorCode`, so that app developers integrations that check for failures don't suddenly stop getting feedback for errors in the `errorCode` query param -- this would effectively be a breaking change for apps that do check for errors in this step, since the _same_ data would be available in a different query param than it used to be.

MaximusHaximus (changes requested):
Added concerns around backwards compat and URL length size in-line, per our off-PR conversation

src/components/sign/Sign.js
MaximusHaximus: Or we could truncate `errorMessage` if it's particularly long :). For backwards compat we want to ensure URLs are < 2048 characters.

Patrick1904: @MaximusHaximus, made the following changes: https://github.com/near/near-wallet/pull/1703/commits/d3a035dde558e39dddd1fd9c6a0b2e9d2c1cc6a9

1. Keep argument as `errorCode`
2. Wrap strings in `encodeURIComponent`
3. Limit `errorMessage` to max 100 characters

MaximusHaximus (approved):
Just one safety thing re: `errorMessage` `substring()` chaining :) Once that's resolved, LGTM!

Issue #1665, opened 2021-04-16
The wallet shows only the Near logo and footer before the account state is loaded


marcinbodnar:
At this moment Near logo and footer are showing immediately after the page loaded, and then we wait for the account state to be loaded before the rest of UI is shown. This is because we want to be sure if the Landing page (no account) or Dashboard should be shown.

The idea is to improve this process, by:
1. First check if there is info about the account in the localStorage, and if so, we can show menu options and some other basic UI, before the account state is loaded.
2. If there is no info about the account in localStorage, we show the Landing page.

jimmy3dita:
What is the issue from the user's point of view?

marcinbodnar:
> What is the issue from the user's point of view?

The user needs to wait 1-2 seconds (or longer if the network is slow) before anything other than the logo and footer is shown.

jimmy3dita:
Thanks, added a priority 2, as the absence of this feature can generate an unnecessarily high number of tickets if the RPC is temporarily slow or unresponsive. 
Also, it's possible that in certain geographic areas this delay is higher anyway.

This issue reminds me of similar sync/async issues, such as the "name is unavailable" in #634 and #1506

vgrichina:
@jimmy3dita I suggest we put it as P1 as the issue is pretty bad if you aren't on excellent network or RPC is slow, e.g. @ilblackdragon is totally frustrated with it in China.

This is something that lies on a critical path to success (like paying for something using NEAR).

PR #1707, opened 2021-04-30
Show UI instantly (#1665)

marcinbodnar:


marcinbodnar:
@ihaid Could you please test it with all routes? The thing is that now we will show components even if the account state is not fetched yet. So I made all changes needed, but want to be sure that all routes work when you refresh the page and don't throw any errors.

marcinbodnar:
@corwinharrell Could you please check one thing? With a slow network, there might be a situation when the UI is shown immediately but we still are loading the account state, so I added animated three dots into the select account box - but maybe you can propose a better solution without many changes needed. Here's how it looks:
image

ihaid (approved):

Issue #1590, opened 2021-03-30
There is no error when access key is already used for another application


ilblackdragon:
**Problem**
When adding access key, if this key is already used - there is an error coming on adding it. But it doesn't get shown to the user, so user is just stuck on the "approve" page.

**Expected Behavior**
Show an error, redirect back to the application.

**Steps to reproduce**

1. Sign in with NEAR in some app
2. Copy the wallet url 
3. Approve
4. Paste the wallet url
5. Click approve
6. Observe that nothing happens

jimmy3dita:
What is the use case that requires to copy&paste the wallet URL twice? 

Anyway, looks to me an easy fix cc @Patrick1904

Patrick1904:
It seems that the key does get added (without any error) even if already used for a different application. Is the idea that we should protect against adding the same key again? cc @vgrichina 

I added the same key a few times for [patrick.lockup.testnet](https://explorer.testnet.near.org/accounts/patrick-lockup.testnet)

<img width="1193" alt="Screen Shot 2021-04-29 at 12 59 59 PM" src="https://user-images.githubusercontent.com/24921205/116611076-d0926200-a8ea-11eb-8fb7-f8cff0c6b952.png">

Steps:
1. Go to https://near-examples.github.io/guest-book and click login
2. Copy the `public_key` from the URL
3. Approve sign in
4. Go to https://near-drop-testnet.onrender.com/ and click login
5. Replace the `public_key` copied in step 2 and refresh the page
6. Approve login
7. Observe in Explorer that the same key was added two times

vgrichina:
@jimmy3dita why is this P1? 

@ilblackdragon when does this situation happen in the wild? Like why do different apps use the same key (this doesn't seem right)?

jimmy3dita:
> @jimmy3dita why is this P1?
> 
> @ilblackdragon when does this situation happen in the wild? Like why do different apps use the same key (this doesn't seem right)?

As mentioned on Discord, I misunderstood the logic behind this error - e.g. a malicious application can re-use previously authorized keys to escalate its credentials and subtract tokens, but this is not the case. Will de-escalate to P3

Issue #429, opened 2020-03-02
Treat ActionError::AddKeyAlreadyExists as a success when adding key


vgrichina:
Maybe should be solved at nearlib level, but generally ActionError::AddKeyAlreadyExists means that key already has been successfully added (need to check if added with same permissions though).

Right now this issue means that if there is any connectivity problem when authorizing an app – wallet gets stuck.

PR #1708, opened 2021-05-03
Add access key: throw error if key already used

Patrick1904:
To test, please see original issue description: https://github.com/near/near-wallet/issues/1590#issue-844059236

### DIFF
--- a/src/components/login/Login.js
+++ b/src/components/login/Login.js
@@ -7,7 +7,7 @@ import LoginForm from './LoginForm'
 import LoginConfirm from './LoginConfirm'
 import LoginDetails from './LoginDetails'
 import LoginIncorrectContractId from './LoginIncorrectContractId'
-import { handleRefreshUrl, switchAccount, allowLogin, redirectToApp } from '../../actions/account'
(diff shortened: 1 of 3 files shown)


vgrichina (changes requested):
Throwing error gonna degrade UX for more common case than described in https://github.com/near/near-wallet/issues/1590

(when one app tries to add the same key second time – e.g. because of user going back in browser)

We need to understand context better: https://github.com/near/near-wallet/issues/1590#issuecomment-831617240 

and provide more specific fix if necessary

vgrichina:
More context on why handling error is needed:
https://github.com/near/near-wallet/issues/429

Patrick1904:
> Throwing error gonna degrade UX for more common case than described in #1590
> 
> (when one app tries to add the same key second time – e.g. because of user going back in browser)
> 
> We need to understand context better: [#1590 (comment)](https://github.com/near/near-wallet/issues/1590#issuecomment-831617240)
> 
> and provide more specific fix if necessary

@vgrichina can you please provide instructions on what I should change? I remember you mentioned something about checking key permissions?

PR #1727, opened 2021-05-07
rollback near-api-js upgrade

think-in-universe:
### Description

We cannot import account into Wallet on testnet or login any DApp on testnet now. Developers from community keep reporting this issue. 

image

error message

```
adding full access key ed25519:DDXd.....
instrument.ts:129 TypeError: Cannot read property 'map' of undefined
    at p.<anonymous> (account-with-lockup.js:29)
    at f (runtime.js:63)
    at Generator._invoke (runtime.js:293)
    at Generator.next (runtime.js:118)
    at l (account-with-lockup.js:234)
    at o (account-with-lockup.js:234)
```

### Cause

This is caused by near-api-js upgraded to **0.41.0** which introduced some compatibility issues.

The account cannot be imported because it failed to **addKey** via `near-api-js`

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "fetch-send-json": "0.0.2",
     "js-sha256": "^0.9.0",
     "mixpanel-browser": "^2.41.0",
-    "near-api-js": "^0.41.0",
+    "near-api-js": "^0.39.0",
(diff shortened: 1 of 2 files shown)


think-in-universe:
Fixes: https://github.com/near/near-wallet/issues/1730

PR #1728, opened 2021-05-07
Update lockup signAndSendTransaction for near-api-js 0.41.0

mehtaphysical:
This fixes the issue caused by updating near-api-js 0.41.0

### DIFF
--- a/src/utils/account-with-lockup.js
+++ b/src/utils/account-with-lockup.js
@@ -23,8 +23,8 @@ export function decorateWithLockup(account) {
 }
 
 async function signAndSendTransaction(...args) {
-    if(args.length === 1) return signAndSendTransactionV2.bind(this,...args);
-    else return signAndSendTransactionV1.bind(this, ...args);
+    if(args.length === 1) return signAndSendTransactionV2.apply(this, args);
+    else return signAndSendTransactionV1.apply(this, args);
 }
 
 async function signAndSendTransactionV1(receiverId, actions) {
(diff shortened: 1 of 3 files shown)


mehtaphysical:
This is also only a quick fix. The better approach may be to create a `LockupAccount` class that extends `Account`. Then instead of: https://github.com/near/near-wallet/blob/b9f26062385d69ba0deac928eb3096a93abef0ef/src/utils/wallet.js#L631-L638

Could do:

```js
async getAccount(accountId) {
    return LockupAccount.create(this.connection, accountId)
}
```

mehtaphysical:
This fixes #1730

src/utils/account-with-lockup.js
```diff
@@ -22,7 +22,16 @@ export function decorateWithLockup(account) {
     return decorated;
 }
 
-async function signAndSendTransaction(receiverId, actions) {
+async function signAndSendTransaction(...args) {
+    if(args.length === 1) return signAndSendTransactionV2.bind(this,...args);
```
vgrichina, line 26: this doesn't look right, should be calling function instead of binding it

MaximusHaximus:
@ihaid Could you give this PR a test and in particular verify that working with a 2FA-enabled account works as expected?

ihaid:
Can't disable 2fa here
image

esaminu:
Continuing on https://github.com/near/near-wallet/pull/2248

PR #1751, opened 2021-05-14
Update translations

picturepan2:
- [ ] Update Chinese translations for reCAPTCHA and drop claiming.
- [ ] Add validatorBox state `active` and `inactive` for translations.
- [ ] Fix a button text wrap bug that all texts are supposed to be in one line (screenshot below):

Before:
Before
Fixed:
Fixed


Patrick1904 (approved):
Thank you @picturepan2!

PR #1756, opened 2021-05-17
Persist redux state

marcinbodnar:


Patrick1904:
Loading is FAST. Seems there's a minor issue with the GlobalAlert when I recover with seed phrase. For some reason I'm not able to dismiss the success alert.

vgrichina:
@marcinbodnar something we should take in mind is that local storage can be pretty limited

Wonder what we can do to fail gracefully if we run out of storage (ideally we should be able to drop cache but keep accounts with keys intact)

PR #1757, opened 2021-05-17
Remove '/profile' URL requirement

Patrick1904:


Patrick1904:
@vgrichina do you see any issues with scalability here? For example, username `near` works because we don't have a route `/near` - otherwise it'd be an issue. I've tried the current routes and there doesn't seem to be any clashes between routes and active account IDs. Routes always get priority.

vgrichina:
> @vgrichina do you see any issues with scalability here? For example, username `near` works because we don't have a route `/near` - otherwise it'd be an issue. I've tried the current routes and there doesn't seem to be any clashes between routes and active account IDs. Routes always get priority.

once top level names get auctioned there's quite likely going to be `profile` or `staking` top level accounts, etc

so we for sure should keep support for explicit `/profile`

MaximusHaximus:
Just to add my 2 cents, I can't think of a clean way to structure our routes that would allow this, and any changes we make that would namespace them differently would almost certainly be breaking changes for any apps integrating with the wallet... unless we've got a driving reason to kill `/profile` that is pretty critical I think it's fine to have a /profile route exist especially given Vlad's mentioning of top level domains which could be almost any string..

Patrick1904:
Thanks for the input on this. I agree, it seems the drawbacks are greater than the benefits. Will close this for now.

PR #1758, opened 2021-05-18
UI fixes

Patrick1904:
Changes:
1. GlobalAlert: check if `messageCode` is a `string` before using `includes` (not always the case)
2. Verify seed phrase _and_ verify recovery link code: fix `sendingString` for new account:
- New account: 'Creating account...'
- Existing account: 'Verifying...'
3. Fix reCAPTCHA layout and spacing

PR #1763, opened 2021-05-18
Release 2021-05-18

vgrichina:
- Return back `errorType` and `errorMessage` for the transaction signing requests that failed
- Various UI fixes


Patrick1904:
@vgrichina would be really great if we can include this one too: https://github.com/near/near-wallet/pull/1758

Patrick1904 (approved):

Issue #783, opened 2020-08-05
Add banner to indicate network/RPC issues


kcole16:
### Overview
Recent network issues have made the wallet slow/unusable for periods of time. Additionally, some of our error messages are quite vague/misleading in these cases, and we do not warn users that the network may be the underlying issue (e.g. "Invalid Seed Phrase"). 

Adding a banner with a warning and link to nearprotocol.statuspal.io will help provide some indication of what's happening.

Perform health check by calling RPC endpoint `/health`

### Acceptance Criteria
- [ ] When RPC is unresponsive, display a red banner at the top of the page
  - [ ] "Our network nodes are experiencing issues. Please check back later, as you will likely encounter high error rates, including during login. To view the latest network status updates, click here"
  - [ ] Link to nearprotocol.statuspal.io

kcole16:
@bowenwang1996 Is there a good way to perform a network health check?

bowenwang1996:
There is a rpc endpoint for health check `/health`

ilblackdragon:
How about statuspal? We ideally should use something third party

Patrick1904:
@vgrichina @kcole16 here's an example response from Statuspal: `https://statuspal.io/api/v1/status_pages/nearprotocol/status`

![Screen Shot 2020-10-12 at 1 37 47 PM](https://user-images.githubusercontent.com/24921205/95788345-37726580-0c90-11eb-9b5f-80b3fc90cd90.png)

Should we base the network error on all (RPC, Explorer frontend, Explorer) or just RPC?

kcole16:
@Patrick1904 Seems like we should use the RPC endpoint `/health` instead (checking https://rpc.mainnet.near.org), as it's more direct.

Patrick1904:
@vgrichina @bowenwang1996 apparently the RPC /health end point returns an error if the node hasn’t responded with a block for 20 seconds, so AFAIU it may not catch the frequently reported issue with recovering accounts, i.e. ‘no accounts recovered…’

In theory, if 'no accounts recovered...' happened and RPC is unresponsive, would /health return an error?

bowenwang1996:
> In theory, if 'no accounts recovered...' happened and RPC is unresponsive, would /health return an error?

Yeah we will work on tweaking the health endpoint to be more informative but right now that may not return an error

marcinbodnar:
@corwinharrell We should get back to this topic because there is an additional issue now.
[...]

PR #1764, opened 2021-05-19
Don't block on indexer when sending transactions

marcinbodnar:


marcinbodnar:
@vgrichina @MaximusHaximus 
It's ready. I used the limited balance feature that I prepared for previous functionalities.

 - Sign - it was already implemented, we fetch only a limited balance to make it work.
 - Login - it's not fetching balance after recent changes for flow limitation that I prepared
 - Send - now it's fetching only limited balance - additionally moved full balance fetch at the end on success flow so it will not block the Success screen.

Also, I disabled several other functionalities and show loaders if CH is not working, I'm using one of the staking balance property to check whether CH is working or not:
 - Recovery methods - before it was showing boxes even if the values are not loaded - now it will show skeleton until the full balance is loaded (CH is working)
 - Staking - the same - before, it was showing page with empty data, now it will skeletons until full balance is loaded
 - Recent Activities - the same.

I disabled CH in the last commit so you can test how it's working.

Two things that I wonder about:
1. Recover account - we use CH only for checking accountId by publicKey, so we can still allow user to recover his account even if CH is down, the only change needed would be to add accountId input so the user can manually type accountId - the only issue is to figure out when to show accountId input (to figure out if the CH is not working, or working slowly), or we can show the input always but marked it as optional which is not ideal.
2. I think that we should also show some kind of banner or other notifications for the users when CH is not responding for several seconds - I'm thinking something about "Sorry, we have some issues, but you can Send Money, Login, Sign (if you are not using 2FA)".

src/utils/wallet.js
```diff
@@ -30,7 +30,8 @@ export const WALLET_CREATE_NEW_ACCOUNT_FLOW_URLS = ['create', 'set-recovery', 's
 export const WALLET_LOGIN_URL = 'login'
 export const WALLET_SIGN_URL = 'sign'
 export const WALLET_RECOVER_ACCOUNT_URL = 'recover-account'
-export const ACCOUNT_HELPER_URL = process.env.REACT_APP_ACCOUNT_HELPER_URL || 'https://near-contract-helper.onrender.com'
+// export const ACCOUNT_HELPER_URL = process.env.REACT_APP_ACCOUNT_HELPER_URL || 'https://near-contract-helper.onrender.com'
```
vgrichina, line 33: @marcinbodnar this seems like accidental debug leftover?

vgrichina:
> 1. Recover account - we use CH only for checking accountId by publicKey, so we can still allow user to recover his account even if CH is down, the only change needed would be to add accountId input so the user can manually type accountId - the only issue is to figure out when to show accountId input (to figure out if the CH is not working, or working slowly), or we can show the input always but marked it as optional which is not ideal.

I think generally this query is pretty cheap, so less likely to be extremely slow, etc.  So shouldn't be a blocker.

> 2. I think that we should also show some kind of banner or other notifications for the users when CH is not responding for several seconds - I'm thinking something about "Sorry, we have some issues, but you can Send Money, Login, Sign (if you are not using 2FA)".

I think it's potentially gonna be many false positives and annoying. Most common failure mode isn't contract helper not responding. More often it is just fetching staking deposits super slow or smth like that. If we can send money without fetching staking deposit, it solves the problem for most of the cases. 

I think we should introduce some offline/online status indicator at some point. I think there is existing tracker:
https://github.com/near/near-wallet/issues/783

vgrichina:
somehow getting stuck here when creating account with seed phrase:

Screenshot 2021-05-19 at 17 14 44

src/utils/wallet.js
marcinbodnar: I disabled CH in the last commit so you can test how it's working.

marcinbodnar:
>I think it's potentially gonna be many false positives and annoying. Most common failure mode isn't contract helper not responding. More often it is just fetching staking deposits super slow or smth like that. If we can send money without fetching staking deposit, it solves the problem for most of the cases.

Right, but the thing is that the user will see that the Wallet is not working, for example, if the user enters the dashboard will see that balance is loading, recent activities are loading - the user will think that the Wallet is not working. I think we should consider informing the user that he can still do some important stuff because it's not obvious.

src/components/profile/Recovery/RecoveryContainer.js
```diff
@@ -43,7 +43,7 @@ const RecoveryContainer = ({ type, recoveryMethods }) => {
     const currentActiveKinds = new Set(activeMethods.map(method => method.kind));
     const missingKinds = allKinds.filter(kind => !currentActiveKinds.has(kind))
     const deleteAllowed = [...currentActiveKinds].length > 1 || account.ledgerKey;
-    const recoveryLoader = (actionsPending('LOAD_RECOVERY_METHODS') && !userRecoveryMethods.length) || !account.accountId
+    const recoveryLoader = (actionsPending('LOAD_RECOVERY_METHODS') && !userRecoveryMethods.length) || !account.accountId || !account.balance.stakedBalanceMainAccount
```
vgrichina, line 46: not clear why recovery loader is tied to `account.balance.stakedBalanceMainAccount`

src/components/wallet/Activities.js
```diff
@@ -84,7 +85,8 @@ const StyledContainer = styled.div`
 
 const Activities = ({ transactions, accountId, getTransactionStatus }) => {
     const [transactionHash, setTransactionHash] = useState()
-    const activityLoader = actionsPending(['GET_TRANSACTIONS', 'REFRESH_ACCOUNT_OWNER'])
+    const { balance } = useSelector(({account}) => account)
+    const activityLoader = actionsPending(['GET_TRANSACTIONS', 'REFRESH_ACCOUNT_OWNER']) || !balance.stakedBalanceMainAccount
```
vgrichina, line 89: not clear why activity loader is tied to `account.balance.stakedBalanceMainAccount`

vgrichina:
Screenshot 2021-05-26 at 01 50 40

@marcinbodnar I've tried testing it locally by running wallet and not starting contract helper and cannot really send money (see screenshot)

src/components/profile/Recovery/RecoveryContainer.js
marcinbodnar: before it was showing boxes even if the values are not loaded - now it will show skeleton until the full balance is loaded (CH is working)

src/components/wallet/Activities.js
marcinbodnar: Same here, now the activity box on the dashboard will show three dots, without it it was showing an empty box, so it was suggesting that you don't have any recent activities.

vgrichina:
> In this case, you can't send because the balance is still loading, but it should load the basic balance from the account state, I checked and it's working fine for me, not able to reproduce - please check what console is showing for you.


I think following steps are giving consistent repro:
1) Run local wallet without contract helper
2) Go to send money page for some account
3) Switch to another account using dropdown

Looks like the issue doesn't happen by itself, but happens when you try switching account.

marcinbodnar:
> @marcinbodnar
> 
> > In this case, you can't send because the balance is still loading, but it should load the basic balance from the account state, I checked and it's working fine for me, not able to reproduce - please check what console is showing for you.
> 
> I think following steps are giving consistent repro:
> 
> 1. Run local wallet without contract helper
> 2. Go to send money page for some account
> 3. Switch to another account using dropdown
> 
> Looks like the issue doesn't happen by itself, but happens when you try switching account.

Thanks @vgrichina , fixed.

marcinbodnar:
> somehow getting stuck here when creating account with seed phrase:
> 
> Screenshot 2021-05-19 at 17 14 44

Yes, it will not work because we are using `CONTRACT_CREATE_ACCOUNT_URL`, which uses `ACCOUNT_HELPER_URL` to compose a POST URL for creating an account.

andy-haynes:
@marcinbodnar is the work done here still actionable? It seems like a lot of the code here has diverged, but also since moving to a separate service/REST API for indexer queries the risk of contract helper downtime is mostly mitigated.

Do you think this is still an issue or can it be closed out?

marcinbodnar:
@andy-haynes I mean it was pretty important PR when created, but there ware no response, I asked about it on slack as well a few times. I mean if the risk of downtime is minor, then we can close it.

PR #1771, opened 2021-05-24
Feature Flag Editor

MaximusHaximus:
## Adds new `feature-flags` package
- Leverages `inquirer` prompts to edit feature flag states and keep some basic metadata about the flag states in sync.  This metadata will allow us to easily keep track of flags that have been enabled for some time but not yet removed ("stale" flags), and nag the owner of the flags to remove them :)
- Pulled in some code in `feature-flags/gitTools` directory that parses git configs and extracts git usernames from those configs to minimize dependencies. Left references to authors in comments.
- Added `ini` dev dependency to feature-flags package as I think this one is reasonable to use directly given its _massive_ adoption (21 million d/l per week)
- Added `fs-extra` dev dependency for simpler file handling and no cross-platform corner cases
- Can be run from any subdirectory in the repository; looks for the closest `NEARFlags.json` file in a parent directory.
- Writes formatted JSON to the file.

## Updates `frontend` package
- Adds a single new flag, `EXAMPLE_FLAG`, so that the shape of the flag state is clear
- Adds a new `ExampleFlag` component that references `EXAMPLE_FLAG`
- Added a route so you can see the flag in action; route is `/example_flag` :)
- Added new environment variable: `NEAR_WALLET_ENV`. **You will need to add this to your .env file for the wallet build to work.**.  Valid options are the values of the object hash exported from `features/environments.json` file (probably `testnet` :))
**After pulling this code, and updating your `.env` file, you will likely need to run `yarn` and `lerna bootstrap`  at the top level of the repo, then `cd packages/frontend; rm -Rf .cache` before you next run `yarn start` there.  Use the `/example_flag` route to verify local dev env is working sanely.**

#### Example route that renders a component that varies depending on `NEAR_WALLET_ENV`:
- https://near-wallet-pr-1771.onrender.com/example_flag (`testnet` flag state - `true`)
- https://deploy-preview-1771--near-wallet-staging.netlify.app/example_flag (`mainnet_STAGING` flag state - `false`)

## Need to add later:
- Add 'create new environment' codepath that prompts the user to ask them which environment to use as default flag states for the new environment

## Nice to add later:
- Add nag for stale flags
- Use `yargs-interactive` to allow `action` and `flag name` to be provided directly from the command line without any interactive prompts.
- Native ESM import support


FlagEditor/FlagEditor.js
```diff
@@ -0,0 +1,125 @@
...
+const path = require('path');
+
+const getGitUsername = require('./gitTools/git-user-name');
+
+const { ACTIONS, ENVIRONMENTS } = require('./constants');
+
+const FLAGS_FILENAME = 'NEARFlags.json';
+const DEBUG_LOGGING = false;
+
+class FlagEditor {
```
vgrichina, line 11: seems like this can be separate open source npm package vs built into wallet

FlagEditor/gitTools/git-config-path.js
```diff
@@ -0,0 +1,35 @@
+/*!
```
vgrichina, line 1: is there a specific reason these are included directly vs used as npm dependencies?

NEARFlags.json
```diff
@@ -0,0 +1,31 @@
+{
+  "EXAMPLE_FLAG": {
+    "createdBy": "Daryl Collins",
+    "createdAt": "2021-05-24T09:33:27.489Z",
+    "development": {
```
vgrichina, line 5: do I understand correctly that this environment gonna get used for PR deploys?

NEARFlags.json
```diff
@@ -0,0 +1,31 @@
+{
+  "EXAMPLE_FLAG": {
```
vgrichina, line 2: given that each flag description is pretty sizeable, seems like it makes sense to do one feature per file to allow easier merging and editing

NEARFlags.json
```diff
@@ -0,0 +1,31 @@
...
+    "createdBy": "Daryl Collins",
+    "createdAt": "2021-05-24T09:33:27.489Z",
+    "development": {
+      "enabled": true,
+      "lastEditedBy": "Daryl Collins",
```
vgrichina, line 7: I personally feel that it is more convenient to have `git blame` as single source of truth for "when this was edited by who" (especially given how good GH UX around it is these days), but don't have hard stance if other people find it convenient cc @Patrick1904 @marcinbodnar @stefanopepe

MaximusHaximus: I agree git blame is awesome <3 :) This is mainly here so that we can easily build some basic tooling around this like 'how long as this flag been on in production env' and add something to prompt us to remove older flags that have not been toggled anywhere in a while :)

NEARFlags.json
MaximusHaximus: Related to your point around git blame -- it's a lot easier to track additions and removals to a file then the addition and removal of entire files inside of a directory, when doing forensics it's nice to be able to continually step back through the feature json file without needing to use a tool that is good at this (Webstorm does it very well, but I'm keenly aware that many people don't use it :))

With the flags organized alphabetically there will almost never be merge conflicts in my experience using this type of pattern previously on teams with 20+ devs committing to the same repo.  What is your concern around ease of editing here?

What do you think about creating 1 file _per environment_ to split things up a bit without creating a new file for every flag?

NEARFlags.json
MaximusHaximus: I was thinking this would be code that should only be on for the developer's local environments (not yet running in any other environment).

The problem I've seen with our PR builds is that they can create a chicken-and-egg problem for required backend code changes that are required for the new functionality in a PR.  With feature flags like this, we could probably eliminate PR builds and make the equivalent to one of our current PR builds that is ready for testing be 'I flagged it on in `testnet-staging` and merged it', so that it's clear from the codebase what code is going to be running against which environment.

We could also retain PR builds/deploy, but make them use `testnet-staging`, which is closest equivalent to what we do now.  However, it's really up to us; we could even do N number of builds from any PR that are configured for whichever environments we'd like to be able to test from the PR (e.g. we could do a build for mainnet staging and testnet staging on every PR if we want), but see my above concerns about crossing environment boundaries and not having a clear idea of what code is running against the backend.

FlagEditor/gitTools/git-config-path.js
MaximusHaximus: I was looking to minimize the overall # of dependencies that are not **extremely** widely used, and it seemed to me that anything that people explicitly install to interface with git at the filesystem level as the expected behavior for the package was at a much higher-than-usual risk for supply chain attacks (attackers could expect that this code will run directly on privileged machines that have developers credentials on the filesystem in many cases).  Maybe my risk->reward judgement here is wrong, I'm happy to bring them in directly :)

FlagEditor/FlagEditor.js
MaximusHaximus: **Absolutely**.  For initial implementation I'd like to get it reviewed and tested here, then move it into its own NPM package so that it could be used for any JS ecosystem packages at NEAR (or elsewhere, with some basic configuration!)

features/features.js
```diff
@@ -0,0 +1,11 @@
+const { initFeatureFlags } = require('@near-wallet/feature-flags');
```
MaximusHaximus, line 1: Note that this resolves to installing the actual NPM package in production builds, but links to the local filesystem version of the package during local dev so you can change `@near-wallet/feature-flags` and see your changes immediately take effect in e.g. `frontend`

MaximusHaximus:
@ihaid This PR ended up involving some refactoring of how we resolve NPM dependencies , which could have unexpected consequences  -- please do a full set of smoke tests on both the testnet preview and mainnet preview for this PR <3

ihaid:
image
the seed phrase is broken @MaximusHaximus

ihaid:
Recovery also 
image

andy-haynes:
@ihaid would you mind taking another look when you have time please? I've resolved the dependency conflicts with `master` that I believe were causing some problems, now I'm able to create and import accounts in the PR preview.

cc @MaximusHaximus

ihaid:
@andy-haynes create, recovery, send, 2fa looking good

ihaid (approved):

PR #1772, opened 2021-05-24
Fix selecting validator on stake action and disabled UI for StakingAction component if staking not allowed

marcinbodnar:


Patrick1904 (approved):
LGTM

PR #1774, opened 2021-05-25
Account creation: new implicit and faucet flows

Patrick1904:
Changes:
1. Account ID input: add TLA suffix
2. Add new 'terms page' before creating account
3. Modify implicit account creation flow
4. Require faucet users to fund their account before using the wallet

For details please see: [Figma design](https://www.figma.com/file/X0TuszmPFSnxOcb4dq0ZtM/Update-Create-Account-w%2F-Faucet?node-id=262%3A15)

Staging link:
https://deploy-preview-1774--near-wallet-staging.netlify.app/


picturepan2:
@Patrick1904 I have updated the Chinese translations to the latest version. 
Btw, who is doing other languages? Is it possible to create a group for these people and commit updates when major update is ready.

src/translations/zh-hant.global.json
```diff
@@ -525,8 +552,9 @@
     },
 
     "copy": {
+        "title": "複製",
         "default": "已復制！"
```
think-in-universe, line 556: 可以统一成`複製`

picturepan2:
@Patrick1904 @think-in-universe Updated better translations of "copy".

src/components/accounts/CreateAccount.js
```diff
@@ -133,18 +167,59 @@ class CreateAccount extends Component {
...
+        if (DISABLE_CREATE_ACCOUNT && !hasFunding && !termsAccepted) {
+            return (
+                <StyledContainer className='small-centered border'>
+                    <FundNearIcon/>
+                    <h1><Translate id='createAccount.termsPage.title'/></h1>
+                    <h2><Translate id='createAccount.termsPage.descOne' data={{ data: utils.format.formatNearAmount(MIN_BALANCE_TO_CREATE) }}/></h2>
+                    <h2><Translate id='createAccount.termsPage.descTwo'/></h2>
+                    <FormButton
+                        onClick={() => this.setState({ whereToBuy: true })}
+                        color='link blue'
```
MaximusHaximus, line 187: Should this be (or at least some of this be) in `className` prop?

src/components/accounts/CreateAccount.js
```diff
@@ -133,18 +167,59 @@ class CreateAccount extends Component {
...
-            showTerms,
-            termsChecked,
-            privacyChecked,
+            termsAccepted,
+            whereToBuy
         } = this.state
 
-        const { localAlert, mainLoader, checkNewAccount, resetAccount, clearLocalAlert } = this.props
+        const { localAlert, mainLoader, checkNewAccount, resetAccount, clearLocalAlert, fundingContract, fundingKey } = this.props;
+        const hasFunding = fundingContract && fundingKey;
```
MaximusHaximus, line 175: Does this represent e.g. `isLinkDrop`? Or could hasFunding be true for reasons other than a linkdrop?

src/components/accounts/CreateAccount.js
```diff
@@ -133,18 +167,59 @@ class CreateAccount extends Component {
...
         const useLocalAlert = accountId.length > 0 ? localAlert : undefined;
 
+        if (DISABLE_CREATE_ACCOUNT && !hasFunding && !termsAccepted) {
```
MaximusHaximus, line 178: What will happen on `testnet`, where `DISABLE_CREATE_ACCOUNT` is false?

src/components/accounts/EnterVerificationCode.js
```diff
@@ -173,16 +188,20 @@ const EnterVerificationCode = ({
...
 
             <div className='resend'>
                 <div><Translate id='setRecoveryConfirm.didNotRecive'/></div>
                 <div>
-                    <span onClick={onResend} className='link'><Translate id='setRecoveryConfirm.resendCode'/></span>
-                    &nbsp;<Translate id='setRecoveryConfirm.or'/>&nbsp;<span onClick={onGoBack}
-                                                                             className='link'><Translate
-                    id='setRecoveryConfirm.sendToDifferent'/> <Translate
-                    id={`setRecoveryConfirm.${useEmail ? 'email' : 'phone'}`}/></span>
+                    <span onClick={!reSending ? onResend : () => {}} className='link'><Translate id={`setRecoveryConfirm.${!reSending ? 'resendCode' : 'resending'}`}/></span>
```
MaximusHaximus, line 202: What is the difference between `loading` and `reSending` props? Feels weird that `sending={loading}` on L171 but we use `reSending` here.  Should L171 also be referencing `reSending` for consistency?

src/components/accounts/SetupImplicit.js
```diff
@@ -134,167 +107,143 @@ class SetupImplicit extends Component {
...
+            moonpaySignedURL,
+            balance,
+            claimMyAccount,
+            creatingAccount
         } = this.state
 
-        const { implicitAccountId, accountId, mainLoader } = this.props
+        const { implicitAccountId, accountId, mainLoader } = this.props;
+
+        if (createAccount) {
```
MaximusHaximus, line 175: Not a blocker, but it feels like we actually have two-components-in-one here -- each with different purposes

src/components/accounts/create/FundWithMoonpay.js
```diff
@@ -0,0 +1,52 @@
...
+
+const FundWithMoonpay = ({ moonpaySignedURL }) => {
+
+    return (
+        <Container>
+            <h3><Translate id='account.createImplicit.pre.moonPay.title'/></h3>
+            <h2><Translate id='account.createImplicit.pre.moonPay.desc'/></h2>
+            <FormButton
+                linkTo='https://support.moonpay.com/'
+                color='link normal underline'
```
MaximusHaximus, line 35: should this be `className`?

src/components/wallet/ActivityDetailModal.js
```diff
@@ -16,6 +16,7 @@ const StyledContainer = styled.div`
     h2 {
         text-align: center;
         margin-top: 20px;
+        color: #24272a !important;
```
MaximusHaximus, line 19: !important 😱  Is this a smell that we've got styles that are too broad in their scope?

src/utils/wallet.js
```diff
@@ -671,6 +671,21 @@ class Wallet {
         });
     }
 
+    async getAccountHelperWalletState(accountId) {
+        const state = await sendJson('GET', ACCOUNT_HELPER_URL + `/account/walletState/${accountId}`);
+        const localStorageInactive = `wallet:account:${accountId}:inactive`
```
MaximusHaximus, line 676: Magic string and setItem should be encapsulated in `localStorage.js` :)

MaximusHaximus (changes requested):
:+1: 😮 This is a beast of a PR! Nice work :D. Added just a few questions/comments in-line

src/components/accounts/AccountFormAccountId.js
```diff
@@ -66,14 +56,42 @@ class AccountFormAccountId extends Component {
...
+        const isSafari = /Safari/.test(navigator.userAgent) && /Apple Computer/.test(navigator.vendor);
+        const width = this.getTextWidth(this.input.current.inputRef.current.value, '16px Inter');
+        const extraSpace = isSafari ? 21.5 : 22
+        this.suffix.current.style.left = width + extraSpace + 'px';
+        this.suffix.current.style.visibility = 'visible';
+        if (this.input.current.inputRef.current.value.length === 0) this.suffix.current.style.visibility = 'hidden';
+    }
+
+    getTextWidth = (text, font) => {
+        let canvas = this.getTextWidth.canvas || (this.getTextWidth.canvas = document.createElement('canvas'));
```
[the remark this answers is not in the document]
Patrick1904: I found this way of measuring the width of the text here: https://stackoverflow.com/questions/118241/calculate-text-width-with-javascript/21015393#21015393

Apparelty

src/components/accounts/SetupImplicit.js
```diff
@@ -134,167 +107,143 @@ class SetupImplicit extends Component {
...
-            },
-            (e) => { 
-                if (e.message.indexOf('exist while viewing') === -1) {
-                    throw e
-                }
-                this.setState({ balance: false })
+        const { implicitAccountId, dispatch } = this.props
+        const { createAccount } = this.state
+
+        if (!document.hidden) {
```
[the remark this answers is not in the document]
Patrick1904: I added a check for `!document.hidden` to make sure the current window tab is active to avoid checking the balance every 2 seconds unless the user is likely to be interacting with the page. 

For example, if the user gets to this page and delays sending funds to the address for a few hours. While at the same time, using a different tab for other things. `document.hidden` will always be `false` as long as the tab is visible/active.

corwinharrell (approved):
Looking and feeling good. Funding via the Account ID feels much better than funding via the implicit address. @Patrick1904 the instances of underlined text are meant to link out to documentation that explains the account storage costs btw

src/components/accounts/EnterVerificationCode.js
Patrick1904: @MaximusHaximus yep good point. We only want the `loading` prop on that button to be true if we're either setting up a recovery message for a new account or creating an account + setting up a recovery method. The `reSending` flag can therefore not be used there since we don't want the button to show the 'Verifying...' string if the user is only re-sending the code. I made some changes here to clarify what the flag is doing and to make sure the button is only loading when it should be: https://github.com/near/near-wallet/pull/1774/commits/965f76cedc0ef83c720ed12205aa3611f473ba14

MaximusHaximus (approved):
:+1: Nice work :D This is awesome.

PR #1780, opened 2021-05-26
Added missing requirements for account length and updated note keys

### LINKED ISSUE #1442: Requirements for the account name is not fully described
**Problem**
The creation of an account with the Account ID that has a length of more than 64 characters is not possible and it is not clarified in the hint. The current hint is:

Your account ID can contain any of the following:
Lowercase characters (a-z)
Digits (0-9)
Characters (_-) can be used as separators

Your account ID CANNOT contain:
Characters "@" or "."

**Expected Behavior**
Users should know the maximum length of the Account ID (rules can be found here: https://docs.near.org/docs/concepts/account#account-id-rules). Also, it should be clarified, that acc length consists of acc name and network suffix. 

**Steps to reproduce**
1. Go to https://wallet.testnet.near.org/create
2.  Enter Account ID with 65 or more characters

brookeperkins:
Fixes #1442

- Added `minCharacters` and `maxCharacters` keys to notes in English; need to be added in all other included languages.

Screen Shot 2021-05-26 at 1 36 50 PM


src/components/common/AccountNote.js
```diff
@@ -20,17 +21,19 @@ const Container = styled.div`
...
+            <li><Translate id='createAccount.note.digits'/></li>
+            <li><Translate id='createAccount.note.separators'/></li>
         </ul>
-        <p><Translate id='createAccount.note.five'/></p>
+        <p><Translate id='createAccount.note.cannotContain'/></p>
         <ul>
-            <li><Translate id='createAccount.note.six'/></li>
+            <li><Translate id='createAccount.note.characters'/></li>
+            <li><Translate id='createAccount.note.minCharacters'/></li>
+            <li><Translate id='createAccount.note.maxCharacters' data={{ accountSuffix: ACCOUNT_ID_SUFFIX }}/></li>
```
brookeperkins, line 34: New keys! 🔑 Needs to be translated to other languages.

brookeperkins:
I noticed it might be easier to distinguish the ID requirements list if the lines related to "canContain" and "cannotContain" were in bold. Would it be alright if I created a ticket to do that?

corwinharrell (approved):

src/components/common/AccountNote.js
Patrick1904: @picturepan2 when you have a chance can you please add translation for min and max chars?

Patrick1904:
> I noticed it might be easier to distinguish the ID requirements list if the lines related to "canContain" and "cannotContain" were in bold. Would also be nice if each account ID requirement were crossed off or removed from the list as the user types and the criteria are met. Would it be alright if I created a ticket to do that?

Thank you @brookeperkins, this is really great. Please feel free to create a ticket for account ID requirements.

@corwinharrell do you have any thoughts on UX when account ID requirements are met?

Patrick1904 (approved):

corwinharrell:
@brookeperkins Thanks for your contribution!

Re ID requirements: Rather than crossing each item off the list as they are met, I'd suggest that we treat the user's input as correct by default, and use the list of requirements to highlight the errors/unmet requirements. This way it acts like inline input validation and tells the user exactly what they might need to change to get to a valid ID.

Might look something like this (don't worry about the copy changes, this is from a newer design):

Screen Shot 2021-05-28 at 1 07 46 PM

Patrick1904:
@brookeperkins - as @corwinharrell already mentioned, please ignore changes outside outside of the account ID requirements, those are being addressed here already (suffix, etc): https://github.com/near/near-wallet/pull/1774

PR #1781, opened 2021-05-27
Add Mixpanel events for funded account flow

### LINKED ISSUE #1779: Add Mixpanel tracking for funded accounts flow
**Problem**
Currently, we don't have visibility into the funded account flow in Mixpanel.  We should add events that will allow us to know:

[X] Did the screen think there was a funded account when the user landed on it? 
[X] Do we think the recaptcha LOADED or FAILED to load 
[X] Did the user ever solve the recaptcha? 
[X] Did the user use the 'escape hatch' button to bypass the reCaptcha due to it failing to load? 
[X] When the user attempted to create the account, did it fail due to NotEnoughBalance?
[X] Did the user attempt to create a funded account, but get blocked by our backend validation of their reCaptcha response?

MaximusHaximus:
Closes #1779

Adds new Mixpanel events throughout the funded account flow
See #1779 for the things we can now track via Mixpanel events.

Note: I moved `setupRecoveryMessageNewAccount()` out of `wallet.js` because I wanted to add Mixpanel tracking to this logic, but didn't want to leak Mixpanel awareness into the code in `wallet.js`.  It also was a bit of a mis-fit for `wallet.js` because it was one of the few places there that used our `store.dispatch()` method directly to dispatch a bunch of sub-actions.  Incidentally, this had caused me to add handling of `NotEnoughBalance` errors during account creation inside `wallet.js` _only for this one case_ where the other account creation flows had the logic in the UI components -- so by moving this code out of the actions, we avoided adding Mixpanel to wallet.js and eliminated some coupling to `dispatch()` that made it awkward to handle error cases consistently across the UI components. 🎉  :)

### DIFF
--- a/src/actions/account.js
+++ b/src/actions/account.js
@@ -221,7 +221,6 @@ export const {
     getLedgerKey,
     getLedgerPublicKey,
     setupRecoveryMessage,
-    setupRecoveryMessageNewAccount,
     deleteRecoveryMethod,
     checkNearDropBalance,
     claimLinkdropToAccount,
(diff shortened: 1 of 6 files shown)


icerove (approved):

Patrick1904 (approved):
LGTM

PR #1785, opened 2021-05-27
Refactor loading fungible token balances using Redux

marcinbodnar:


src/utils/tokens.js
```diff
@@ -0,0 +1,47 @@
...
+        return {
+            contract,
+            balance
+        }
+    }
+}
+
+const logError = (error) => {
+    console.warn(error)
+    Sentry.captureException()
```
vgrichina, line 46: missing parameter to `captureException`

src/components/wallet/TokenAmount.js
```diff
@@ -3,9 +3,12 @@ import { formatTokenAmount } from '../../utils/amounts'
 
 const FRAC_DIGITS = 5
 
-const TokenAmount = ({ token: { balance, decimals, symbol}, className }) => (
+const TokenAmount = ({ token: { balance, decimals, symbol }, className }) => (
     <div className={className} title={showFullAmount(balance, decimals, symbol)}>
-        {balance && formatToken(balance, decimals)}
+        {balance
+            ? formatToken(balance, decimals)
+            : <span className='dots' />
```
vgrichina, line 10: it is probably best not to spin dots for every balance being loaded (screen gonna look too busy, especially given that balances load pretty fast)

I think if we are to improve something in loading is actually try to make not loaded state closer to loaded (e.g. right now one item is smaller in height before loaded, so it kinda looks glitchy)

src/components/wallet/Wallet.js
```diff
@@ -130,6 +163,9 @@ export function Wallet() {
     const hideExploreApps = localStorage.getItem('hideExploreApps')
     const linkdropAmount = localStorage.getItem('linkdropAmount')
     const linkdropModal = linkdropAmount && showLinkdropModal !== false;
+    const { tokens } = useSelector(({ tokens }) => tokens)
+    const { actionStatus } = useSelector(({ status }) => status)
+    const tokensLoader = actionsPending(['TOKENS/LIKELY_CONTRACTS/GET', 'TOKENS/TOKENS_DETAILS/GET_METADATA', 'TOKENS/TOKENS_DETAILS/GET_BALANCE_OF']) || !balance?.total
```
vgrichina, line 168: feel like it would be nice to just be able to pass `TOKENS/` prefix here

src/utils/tokens.js
```diff
@@ -0,0 +1,47 @@
...
+        logError(e);
+    } finally {
+        return {
+            contract,
+            metadata
+        }
+    }
+}
+
+export const getBalanceOf = async (contract, account, accountId) => {
```
vgrichina, line 29: nit: seems like you can avoid some boilerplate:

```js
const logAndIgnoreError = (promise) => promise.catch(error => console.warn(error));

export const getBalanceOf = async (contract, account, accountId) => {
  let balance = await logAndIgnoreError(account.viewFunction(contract, 'ft_balance_of', { account_id: accountId }))
  return { contract, balance }
}
```

src/actions/tokens.js
```diff
@@ -0,0 +1,42 @@
...
+    const { accountId } = getState().account
+
+    await dispatch(tokens.likelyContracts.get(accountId))
+
+    let contracts = [...new Set([...getState().tokens.likelyContracts, ...WHITELISTED_CONTRACTS])].reduce((x, contract) => ({
+        ...x,
+        [contract]: { contract }
+    }), {})
+
+    dispatch(tokens.tokensDetails.set(contracts))
```
vgrichina, line 17: not sure why this action is needed vs doing same work in the reducer for `tokens.likelyContracts.get`

src/actions/tokens.js
```diff
@@ -0,0 +1,42 @@
...
+
+    const account = await wallet.getAccount(accountId)
```
vgrichina, line 19: we'd better should just create `Account` object directly to avoid extra RPC calls, as there is no special handling needed for 2FA / lockup

vgrichina: this also should allow to create it in `getMetadata` and `getBalanceOf` actions directly vs passing it in (which feels a bit weird, I don't think action params are typically used to pass dependencies)

src/reducers/tokens/index.js
```diff
@@ -0,0 +1,68 @@
...
+        (!ready || error)
+            ? state
+            : ({
+                ...state,
+                likelyContracts: payload
+            }),
+}, initialState)
+
+const tokensReducer = handleActions({
+    [tokens.tokensDetails.set]: (state, { payload }) => ({
```
vgrichina, line 22: feel like we should have some better handling here than current one (replacing whole list)

like if we start from cached list – we probably only want to invalidate balances for existing contracts, but keep cached metadata as is until latest version is loaded

src/actions/account.js
```diff
@@ -545,6 +547,7 @@ export const switchAccount = (accountId) => async (dispatch, getState) => {
     dispatch(handleRefreshUrl())
     dispatch(staking.clearState())
     dispatch(refreshAccount())
+    dispatch(tokens.clearState())
```
vgrichina, line 550: wonder whether we want to have `clearState` explicit vs handling this in reducer for `switchAccount`, etc

feel like if only goal is to update state, then it makes sense to update state in whatever actions reducer vs spin off more actions to do this

unless I'm missing smth important

src/utils/tokens.js
marcinbodnar: done

src/components/wallet/TokenAmount.js
marcinbodnar: > I think if we are to improve something in loading is actually try to make not loaded state closer to loaded (e.g. right now one item is smaller in height before loaded, so it kinda looks glitchy)

This is fixed, by blocking showing token until metadata are available

I'm also not sure about the dots that appear in the balance place until the balance is loaded. But I wanted to handle situations when the network will be slow. I was thinking about another solution when we don't display token at all until the balance is loaded. In this case, dots will be visible only on section title (Tokens...)
What do you think?

src/components/wallet/Wallet.js
marcinbodnar: added `actionsPendingByPrefix` function

src/utils/tokens.js
marcinbodnar: Good idea, done

src/actions/tokens.js
marcinbodnar: Right, done

src/actions/tokens.js
marcinbodnar: done

src/reducers/tokens/index.js
marcinbodnar: done

src/actions/account.js
marcinbodnar: You're right, but I would leave it for now, because I would like to change it completely while refactoring the redux state - then the account state will not be cleared at all

src/actions/tokens.js
```diff
@@ -0,0 +1,36 @@
...
+            GET: [
+                getLikelyContracts,
+                () => WHITELISTED_CONTRACTS
+            ],
+        },
+        TOKENS_DETAILS: {
+            GET_METADATA: getMetadata,
+            GET_BALANCE_OF: getBalanceOf
+        },
+        CLEAR_STATE: null
```
MaximusHaximus, line 34: Given there are so few keys and the intention of the actions can easily be described without nesting -- why not define these in a flat fashion like e.g. `getLikelyTokenNames`, `getTokenMetadata` and `getTokenBalances`?

src/components/wallet/TokenAmount.js
```diff
@@ -3,9 +3,12 @@ import { formatTokenAmount } from '../../utils/amounts'
 
 const FRAC_DIGITS = 5
 
-const TokenAmount = ({ token: { balance, decimals, symbol}, className }) => (
+const TokenAmount = ({ token: { balance, decimals, symbol }, className }) => (
     <div className={className} title={showFullAmount(balance, decimals, symbol)}>
-        {balance && formatToken(balance, decimals)}
+        {balance
+            ? formatToken(balance, decimals)
```
MaximusHaximus, line 9: :nit: It would be easier to follow this file if helper methods were defined before they are used (especially when they are very similarly named to things we import, and/or wrap things that we import)

src/components/wallet/TokenAmount.js
```diff
@@ -22,9 +25,9 @@ const formatToken = (amount, decimals) => {
     return removeTrailingZeros(formattedAmount)
```
MaximusHaximus, line 25: Just happened to notice this logic on L22/L23.  It looks like we are changing the value we're displaying in such a way that it is not actually an accurate number -- is this essentially performing a `round down` or a `.toFixed()` kind of operation via string manipulation?  If so, could we use BN.js for this instead so it is declarative rather than imperatively written?

If we need to leave it the way it is rather than use BN.js, It would be nice to have a comment on this, or (even better) move this logic into a named function so it is readily apparent that we are rounding the value or converting it to a fixed precision, without having to mentally parse the entire expression to understand what it is for, and have an explanation co-located with the code as to _why_ we need to transform the last digit a `1` here.

src/utils/tokens.js
```diff
@@ -0,0 +1,31 @@
...
+    const account = await wallet.getAccountBasic(accountId)
+
+    let balance = await logAndIgnoreError(account.viewFunction(contract, 'ft_balance_of', { account_id: accountId }))
+    return {
+        contract,
+        balance
+    }
+}
+
+const logAndIgnoreError = (promise) => promise.catch(error => (console.warn(error), Sentry.captureException(error)));
```
MaximusHaximus, line 31: I'm curious about this indirection and lazy attachment of a .catch() handler -- why not `.catch(logError)` on each async call that we want to add a catch handler to? `logAndIgnoreError(...)` is more characters to type than `.catch(logError)`, it requires wrapping a function call rather than chaining on it, and the code doesn't read sequentially because of it being handled via a wrapper function rather than being chained on the call. 
 I don't think it's really any DRYer (we repeat `logAndIgnoreError` the same # of times we would repeat `.catch(logError)`), and it adds an extra layer of indirection that doesn't seem to add any value over .catch(logError)` unless I am missing something here.

src/utils/tokens.js
```diff
@@ -0,0 +1,31 @@
...
+    await logAndIgnoreError(sendJson('GET', `${ACCOUNT_HELPER_URL}/account/${accountId}/likelyTokens`))
+)
+
+export const getMetadata = async (contract, accountId) => {
+    const account = await wallet.getAccountBasic(accountId)
+
+    let metadata = await logAndIgnoreError(account.viewFunction(contract, 'ft_metadata'))
+    return {
+        contract,
+        metadata
```
MaximusHaximus, line 17: Curious -- does `payload.metadata` include `contract` key? It would be nice to add a comment here referencing the spec that we expect this object shape to follow, if it's documented somewhere, or just describe the shape of the expected response if there is no spec that we can reference.

src/utils/tokens.js
```diff
@@ -0,0 +1,31 @@
...
+import { wallet } from './wallet'
+
+export const getLikelyContracts = async (accountId) => (
+    await logAndIgnoreError(sendJson('GET', `${ACCOUNT_HELPER_URL}/account/${accountId}/likelyTokens`))
+)
+
+export const getMetadata = async (contract, accountId) => {
+    const account = await wallet.getAccountBasic(accountId)
+
+    let metadata = await logAndIgnoreError(account.viewFunction(contract, 'ft_metadata'))
```
MaximusHaximus, line 14: :nit: For consistency, if we don't need to re-assign the value of a variable, we should use `const`.  No big deal here since the entire variable is optional so its lifecycle is tiny, but we should enforce this with `eslint` so that it's auto-fixed for us on save if we forget and doesn't come up in code reviews as a nit/lint fix.

src/utils/tokens.js
```diff
@@ -0,0 +1,31 @@
+
+import sendJson from '../tmp_fetch_send_json'
+import { ACCOUNT_HELPER_URL } from './wallet'
+import * as Sentry from '@sentry/browser'
+import { wallet } from './wallet'
+
+export const getLikelyContracts = async (accountId) => (
+    await logAndIgnoreError(sendJson('GET', `${ACCOUNT_HELPER_URL}/account/${accountId}/likelyTokens`))
```
MaximusHaximus, line 8: :nit: superfluous `await`; no need to have this be an expression at all and the `await` doesn't do anything in this context.

src/utils/alerts.js
```diff
@@ -24,6 +24,16 @@ export const dispatchWithAlert = (action, data) => store.dispatch({
 
 export const actionsPending = (types) => (typeof types === 'string' ? [types] : types).some((type) => store.getState().status?.actionStatus[type]?.pending)
 
+export const actionsPendingByPrefix = (typePrefix) => {
+    const { actionStatus = {} } = store.getState().status
+
+    return Object.keys(actionStatus).some((type) => {
+        if (type.includes(typePrefix)) {
```
MaximusHaximus, line 31: I think this should be `startsWith` -- or for backwards compatibility, maybe `.slice(typePrefix.length) ===`, to avoid false positive matching other actions that have the same string somewhere else in their action name.

MaximusHaximus (changes requested):
This is looking :fire: awesome, 💖

I added a few ignorant questions and a few requests in-line, and added one comment that isn't a blocker for this PR, but mainly a discussion point about the repetitive nature of our action handlers repeating the !ready || error boilerplate for _every async action everywhere_, and how that impacts our reducer content, context where we know about errors, and limits our ability to define error handlers vs. success handlers as explicit actions.

marcinbodnar:
Thanks @MaximusHaximus .

src/actions/tokens.js
```diff
@@ -0,0 +1,36 @@
...
+export const handleGetTokens = () => async (dispatch, getState) => {
+    const { accountId } = getState().account
+
+    await dispatch(tokens.likelyContracts.get(accountId))
+
+    const { tokens: contracts } = getState().tokens
+
+    await Promise.all(Object.keys(contracts).map(async contract => {
+        await dispatch(tokens.tokensDetails.getMetadata(contract, accountId))
+    }))
```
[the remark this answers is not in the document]
marcinbodnar: My intention was to load them parallelly but wait with fetching balance until the details are loaded, so the list of all tokens will show as soon as possible, without waiting on balances.

src/actions/tokens.js
```diff
@@ -0,0 +1,36 @@
...
+
+    Object.keys(contracts).map(async contract => {
+        await dispatch(tokens.tokensDetails.getBalanceOf(contract, accountId))
+    })
```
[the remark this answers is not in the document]
marcinbodnar: Right, removed async/await - if something goes wrong it will be caught by `tokens.tokensDetails.getBalanceOf` action.

src/actions/tokens.js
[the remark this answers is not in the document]
marcinbodnar: One thing that I forget, I also wanted to wait for token details, to know if fetching balance is needed - so fetching balance only for tokens with metadata. Added.

src/components/wallet/Tokens.js
```diff
@@ -26,7 +26,7 @@ const StyledContainer = styled.div`
 const Tokens = ({ tokens }) => {
     return (
         <StyledContainer>
-            {tokens.map((token, i) => (
+            {tokens.map((token, i) => token.spec && (
```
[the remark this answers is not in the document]
marcinbodnar: If `token.spec` is available then we know that the details are loaded.

The idea is not to show token until the metadata is loaded - before, we were showing the list of all likelyContracts (but only with contract name, no token name), and after that, we ware starting fetching details - the result was that each token was shown incomplete (no name), and full data ware shown after a while, additionally some tokens ware removed after they ware shown (no metadata) - so not the best UX.

src/components/wallet/Wallet.js
```diff
@@ -130,6 +163,9 @@ export function Wallet() {
     const hideExploreApps = localStorage.getItem('hideExploreApps')
     const linkdropAmount = localStorage.getItem('linkdropAmount')
     const linkdropModal = linkdropAmount && showLinkdropModal !== false;
+    const { tokens } = useSelector(({ tokens }) => tokens)
+    const { actionStatus } = useSelector(({ status }) => status)
```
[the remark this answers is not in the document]
marcinbodnar: Done

src/reducers/tokens/index.js
```diff
@@ -0,0 +1,73 @@
...
+            }),
+}, initialState)
+
+const tokensReducer = handleActions({
+    [tokens.likelyContracts.get]: (state, { ready, error, payload, meta }) => 
+        (!ready || error)
+            ? state
+            : ({
+                ...state,
+                tokens: [...payload, ...meta].reduce((x, contract) => ({
```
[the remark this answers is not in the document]
marcinbodnar: Done

src/reducers/tokens/index.js
```diff
@@ -0,0 +1,73 @@
...
+                    ...x,
+                    [contract]: { 
+                        contract
```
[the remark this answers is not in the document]
marcinbodnar: >We can assume that contract is the string representation of the contract name; can we call it contractName so it's apparent that it's not a rich object without looking at how we use it and/or where it comes from?

Agree, renamed everywhere.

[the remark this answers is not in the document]
marcinbodnar: >Also, curious about why we are creating an object keyed with the name of contract, but with only one key named contract that is the same as the containing object's key; what is the purpose of the nested contract string?

Right, refactored it a bit - now we will save likelyTokens as a separate object, but at the same time we will keep details for tokens with metadata only. It will be much simpler.

src/reducers/tokens/index.js
```diff
@@ -0,0 +1,73 @@
+import { handleActions } from 'redux-actions'
+import reduceReducers from 'reduce-reducers'
+
+import { tokens } from '../../actions/tokens'
+
+const initialState = {
+    likelyContracts: [],
```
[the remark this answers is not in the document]
marcinbodnar: Right, but I'm not sure about it, because when I start refactoring it, it was like that:
https://github.com/near/near-wallet/blob/master/src/components/wallet/Wallet.js#L169

So `likelyTokens` endpoint becomes `likelyContracts`, I thought that's weird (like you), and started using `likelyTokens` everywhere, you can see it at my first commits - but after that, it was much more confusing because next we are using `contract` (instead of `token`) everywhere - so I changed it back to `likelyContracts`.

So I'm not sure how to handle it, we can use `likelyTokens`, but then we should use `tokens` instead of `contract` but that also confusing since it's not a token :) Maybe we should change the endpoint name, at some point?

src/utils/tokens.js
marcinbodnar: Done

src/utils/alerts.js
marcinbodnar: Right

src/utils/tokens.js
marcinbodnar: Done

src/reducers/tokens/index.js
```diff
@@ -0,0 +1,73 @@
...
+        (!ready || error)
+            ? state
+            : ({
+                ...state,
+                likelyContracts: payload
+            }),
+}, initialState)
+
+const tokensReducer = handleActions({
+    [tokens.likelyContracts.get]: (state, { ready, error, payload, meta }) =>
```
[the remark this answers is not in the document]
marcinbodnar: It's removed now, based on changes for the previous comment.

src/utils/tokens.js
marcinbodnar: >Curious -- does payload.metadata include contract key?

No. I'm not sure what is the possible object shape - @vgrichina where we can check that?

vgrichina: There is fungible token standard defining all these details:
https://github.com/near/NEPs/tree/master/specs/Standards/FungibleToken

Probably makes sense to link to it from the implementation given there is clearly confusion over where to look this up.

src/components/wallet/Wallet.js
```diff
@@ -15,10 +14,16 @@ import { Mixpanel } from "../../mixpanel/index"
 import Activities from './Activities'
 import ExploreApps from './ExploreApps'
 import Tokens from './Tokens'
-import { ACCOUNT_HELPER_URL, wallet } from '../../utils/wallet'
 import LinkDropSuccessModal from './LinkDropSuccessModal'
+import { selectTokensDetails } from '../../reducers/tokens'
+import { selectActionStatus } from '../../reducers/status'
+import { selectTransactions } from '../../reducers/transactions'
+import { selectAccountId } from '../../reducers/account'
```
MaximusHaximus, line 21: :nit: we could combine multiple imports from `account` 🤷

src/reducers/tokens/index.js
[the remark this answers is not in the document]
MaximusHaximus: It sounds like the sanest thing to do is re-name `likelyTokens` everywhere, including the `near-contract-helper` route, to be `/likelyTokenContracts`? :).

If that sounds correct, could you create an issue for that change? I'm happy to add a  route for`/likelyTokenContracts` to `near-contract-helper` that is bound to the same route handler as `/likelyTokens`, then once that is landed we can rename the call to the route as a separate PR to `near-wallet`.

For this PR, let's rename everything to use the naming convention we want the contract-helper route to be (`likelyTokenContracts`?) :). That way the follow-up PR after we've added the new route to the contract-helper will only be a change to that one line that performs the API call

MaximusHaximus (changes requested):
:+1: I'd say we're just about there :D. Just a couple things in-line and then we can get this sucker merged to `master` :)

marcinbodnar:
@MaximusHaximus Done.

src/components/wallet/TokenAmount.js
marcinbodnar: If I remember correctly, that was the logic copied from `near-api-js` - I think we were planning to export this fn from `near-api-js` and use it here. But not sure :)

MaximusHaximus (approved):
:+1: Nice work! Let's get this sucker merged to testnet & staging :) :D

PR #1786, opened 2021-05-28
Don't use --https by default

vgrichina:
It gets in a way every time server is restarted and only required for limited set of features.
It is possible to use `yarn start --https` to enable HTTPS on demand when needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
   "scripts": {
     "predeploy": "yarn build",
     "update:static": "cp src/wasm/multisig.wasm dist/multisig.wasm && cp src/wasm/main.wasm dist/main.wasm",
-    "start": "yarn update:static; parcel src/index.html --https",
+    "start": "yarn update:static; parcel src/index.html",
     "prebuild": "rm -rf dist/",
     "build": "parcel build src/index.html && node scripts/sentry-send-release.js && yarn update:static",
     "test": "jest",


vgrichina:
@Patrick1904 @marcinbodnar @MaximusHaximus wonder what are your thoughts on this?

I'm basically thinking we should instead just have info on README https://github.com/near/near-wallet/pull/1787

Patrick1904 (approved):

MaximusHaximus:
Given that Ledger is a core functionality of the wallet that everyone should be able to test locally, I'd like to optimize towards it being the default configuration of the start command rather than making it an exceptional case. I actually didn't even realize that using non-https would break `reCaptcha` testing locally until I read the README.md change here.  If that is true, then in my mind this is even more of a reason to provide a first-class resolution to the nag screen issue, rather than to revert to a configuration that provides partial functionality during local development.  

The error generated when you attempt to use the ledger without HTTPS isn't enabled is cryptic and takes some intuition to resolve, whereas a screen about a local HTTPS certificate is explicit and immediately understandable.  I'd prefer if we eliminated the nag screen but left HTTPS on, rather than remove the nag screen by disabling HTTPS, only to have developers presented with it _anyway_, as soon as they need to test ledger or reCaptcha functionality, and to have them need to debug the issue or remember a note from the README about why the feature doesn't work. Bypassing the nag screen is a single click from the user that is well understood. Debugging a broken feature requires time and context, then would require them to bypass the nag screen _anyway_ once they have figured that out.

Also, just to be clear about how much this impacts local development, could you confirm you get this prompt _every time you restart parcel?_  In my testing, restarting my parcel server doesn't re-prompt me with the nag screen, and Chrome remembers my selection for days and days across a boot, and even in inPrivate sessions without a new presentation of the nag screen.  How often are you seeing the nag screen and under what circumstances?  I have seen it twice since this was merged a few weeks ago (hard reboots, I think).

IMO, if we want to add any 'special behavior' for our `start` command, the _exception_ should be to _disable_ local usage of ledger and reCaptcha functionality rather than to _enable_ it. e.g. "if you don't intend to use ledger or test reCaptcha locally, and you don't want to trust a self-signed cert for localhost development, and seeing a screen about the self-signed certificate once-per-boot is unacceptable to you, then you can instead run `yarn start --disableHTTPSRequiredFeatures`, and use plain `http` for your convenience.  _However_, if you do want to use features that require HTTPS, you will see a warning screen if you have not trusted the local dev certificate <link to trusting cert instructions>.  Note that the warning displayed by chrome _is only a warning_ and you do not _have_ to trust the local certificate in order to test ledger functionality, trusting the local certificate just eliminates the chrome warning for convenience."  Then we could stop rendering the portions of the UI that aren't supported, with a user friendly reason why they aren't being rendered, if someone opts out of the feature, rather than having a stack trace bubble up when they try to do so.  However, even reading all of that explanation makes me think that if someone can easily eliminate the nag screen, they would likely never use the `--disableHTTPSRequiredFeatures` option in the first place, so their development flow is _consistent_ no matter what subset of features they are using in development.  It seems like a solution searching for a problem given the nag screen can be eliminated anyway, without disabling HTTPS, and rarely appears.

Long-to-short: Given that 99% of the time during development, reloads are handled by HMR, and Chrome remembers the selection so you don't get repeatedly prompted, I'm not sure why this would present a big problem, but if it _does_ present a serious impediment to local development flow, I'd like for us to make the elimination of that nag screen when using HTTPS a first-class part of the codebase rather than disabling HTTPS, but then still providing a less-than-ideal experience for those who enable it, and leaving it up to them to debug out why parts of the system are broken when it happens.

vgrichina:
> The error generated when you attempt to use the ledger without HTTPS enabled is cryptic and takes some intuition to resolve, whereas a screen about a local HTTPS certificate is explicit and immediately understandable.

not sure what are you talking about, this is the error generated:

```
Error: U2F browser support is needed for Ledger. Please use Chrome, Opera or Firefox with a U2F extension. Also make sure you're on an HTTPS connection
```

it literally says **make sure you're on an HTTPS connection**

I'd like to hear opinion from @marcinbodnar and @Patrick1904 what is more convenient default. 
It's not too big deal for me, just inconvenient as the very first thing I have to do when I switch to another branch I have to apply this edit to `package.json`.

> Also, just to be clear about how much this impacts local development, could you confirm you get this prompt every time you restart parcel? 

looks like it doesn't indeed happen on every restart, but it happens when I nuke `.cache` folder – but I do it pretty frequently as you have to do it often when either env variables or dependencies have changed

MaximusHaximus:
@vgrichina Could you pull down https://github.com/near/near-wallet/pull/1790 and verify that it resolves the issue with the nag screen re-appearing for you when you destroy the .cache directory?  I understand your frustration with the nag screen, and I'm keen to find a solution to that issue that doesn't mean that our default `yarn start` configuration doesn't allowing us to use ledger and reCaptcha... and any other things that we add in the future that require HTTPS to work.

marcinbodnar:
Yes, It's a bit inconvenient when I have to edit .json every time I start working with a new branch or after rebase.

But I also understand Daryl's point of view, it makes sense.

The best idea would be to find a solution without any inconveniences for us.

MaximusHaximus:
Closing in favor of https://github.com/near/near-wallet/pull/1790 which resolves the recurring nags by providing stable dev certificates that won't change even when you destroy the `.cache` directory

PR #1787, opened 2021-05-28
Update README.md to include HTTPS info

vgrichina:
Follow up to https://github.com/near/near-wallet/pull/1786


MaximusHaximus:
Closing in favor of https://github.com/near/near-wallet/pull/1790 which resolves the recurring nags by providing stable dev certificates that won't change even when you destroy the `.cache` directory

PR #1790, opened 2021-05-29
Stable SSL certificate for localhost https development

MaximusHaximus:
Adds stable certificates, so that users are not nagged when using https on localhost and the .cache folder has been regenerated.  Works with both 127.0.0.1 and localhost https URLs.

This allows us to use HTTPS in development, with no nag screens re-appearing when .cache is erased.


MaximusHaximus:
Merging this  -- it should resolve the nagging certificate prompt for everyone without us needing to disable https for our default `start` command; LMK if you have any problems with it.

marcinbodnar:
> Merging this -- it should resolve the nagging certificate prompt for everyone without us needing to disable https for our default `start` command; LMK if you have any problems with it.

Great, thanks @MaximusHaximus

Issue #1791, opened 2021-05-29
Login screen flashes full access key request


ilblackdragon:
When wallet login screen opens, first few seconds it shows the full access key request (with warning and everything) and only after that time switches to the request to share account id.

# Steps to reproduce

1. Go to https://app.ref.finance
2. Click "Login"
3. Observe first few seconds the screen like this:

<img width="654" alt="Screen Shot 2021-05-29 at 5 55 04 AM" src="https://user-images.githubusercontent.com/175486/120071081-6e8c6000-c042-11eb-8bfe-745d5cbbfc02.png">

stefanopepe:
Tagging @marcinbodnar, as probably was introduced by the recent changes to redux state

marcinbodnar:
Thanks @stefanopepe

PR #1798, opened 2021-05-31
Fix checking account confirmation form (#1791)

marcinbodnar:


MaximusHaximus (changes requested):
Since this expression is pretty central to the wallet login UI, and it is already pretty complex and only likely to grow more complex as time goes on, it'd be nice to move the logic into a more readable format with some human readable descriptions about what the cases represent from a product perspective -- comment in-line :)

src/components/login/Login.js
```diff
@@ -61,7 +61,7 @@ class Login extends Component {
     render() {
         const { account: { url, accountId }, match, appTitle } = this.props
         const accountConfirmationForm = url
-            ? (url.public_key && (!url.contract_id || url.contract_id?.endsWith(`.${LOCKUP_ACCOUNT_ID_SUFFIX}`))) || url.contract_id === accountId
+            ? (url.public_key && (!url.contract_id || url.contract_id?.endsWith(`.${LOCKUP_ACCOUNT_ID_SUFFIX}`))) || (accountId && url.contract_id === accountId)
```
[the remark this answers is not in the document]
marcinbodnar: Done.

src/components/login/Login.js
```diff
@@ -58,11 +58,28 @@ class Login extends Component {
         this.props.history.push('/create')
     }
 
+    get shouldRenderAccountConfirmationForm() {
+        const { account: { url, accountId } } = this.props
+
+        if (!url) {
+            return undefined
```
MaximusHaximus, line 65: Just a :nit:, since we're always doing a coercive check on this value currently, but in general we should try to return the same data type for all return branches of a method like this for consistency, or when we do an explicit === check (which is best practice, generally) on the result of this function somewhere else, it would be incorrect just for this one case, or we'd have to check === false || === undefined.

MaximusHaximus (approved):
<3  🚢 :shipit:

src/components/login/Login.js
marcinbodnar: Right, I didn't dig into it since I just jumped into this issue to fix it fast - so I refactored the previous exact conditions. I think the best implementation would be to leave `undefined` as it's telling us that we don't know yet if we should show confirmation form or not, and then use it instead of `if (!url)`. Made that change, let me know what do you think?

MaximusHaximus: :+1: I can get behind that, the duplication of the !url check in both the render method and the `shouldRenderAccountConfirmationForm` method bothered me a little, actually :D :+1:

MaximusHaximus (approved):
:+1: legit

PR #1812, opened 2021-06-09
TMP: Always recreate transaction when signing

vgrichina:


ilblackdragon:
What's the timeline on this landing?

MaximusHaximus:
@vgrichina Keen to land this in this week's release unless there is an issue with this approach as a temporary fix?

ilblackdragon:
People still have this issue: https://discord.com/channels/831453845833383957/831453846286106636/854319474827853855
```
Error: Transaction nonce 19 must be larger than nonce of the used access key 40070466000000
```

Also the nonce is really weird.

PR #1813, opened 2021-06-09
Add note about near.bet when creating account name

vgrichina:
Should increase awareness of the fact that you can trade .near names


Kouprin:
I don't feel confident this change is needed for wallet for several reasons:

stefanopepe:
I don't support this change either.
Near.bet is not an "official" NEAR Collective product, and I'm worried that other platforms (let's say ENS) might need the same later on

MaximusHaximus:
Closing as `won't fix` for now based on the objections of Alex and Stefano above. Please feel free to re-open it to make a case about their concerns if desired.

Issue #1818, opened 2021-06-11
redirect url during linkdrop doesn't work


think-in-universe:
**Problem**

When user opens linkdrop with a `redirectUrl` param, the page is not redirected to the url after account is created. 

**Expected Behavior**

User should be redirected to the url as described in https://github.com/near/near-wallet/issues/1641 and implemented in https://github.com/near/near-wallet/pull/1652

**Steps to reproduce**

1. Open the create page with link: https://wallet.testnet.near.org/create/testnet/4JHcJkmy8cQ92b1BLZUUUJZKw6ej2URu6rwoiYN3AL8gba626LspzLjQ97FSPs1aPUVb7EjwwAvHGCkQVQe7EMVW?redirectUrl=https://berryclub.io/  or https://wallet.testnet.near.org/create/testnet/5waZiox2Q11VKAeWpDjTxpkDNaVaXKGSbodBHBqhy2nYt1uLXqH74ctPJPnxfACp9HL3uEspnRGADB9YCzfVZJFy?redirectUrl=https://berryclub.io/
2. Create a new account with linkdrop
3. The page is not redirected to https://berryclub.io/

stefanopepe:
cc @marcinbodnar and @Patrick1904 , we were discussing about this the other for the Account Creation v1

think-in-universe:
I tested again just now for several times, and cannot reproduce the issue any more. the url redirect is working now. 

we indeed could reproduce this issue yesterday as reported by several developers. I have asked them to try again and will share the result here.

marcinbodnar:
> cc @marcinbodnar and @Patrick1904 , we were discussing about this the other for the Account Creation v1

@Patrick1904 Want to check it? If I remember correctly you were working on this feature lately, I think it would be good to double-check the code if everything looks fine.

think-in-universe:
> we indeed could reproduce this issue yesterday as reported by several developers. I have asked them to try again and will share the result here.

The developers could still reproduce this issue. I'm doing more investigation and will share more findings here.

PR #1819, opened 2021-06-14
Fix URL handling for /linkdrop page (#1818)

marcinbodnar:


marcinbodnar:
@think-in-universe Regarding #1818, checked the code, found two issues with linkdrop URL, and fixed it. I couldn't reproduce the exact same bug but found two situations where the URL might be cleared. Please check if it fixed the issue.

think-in-universe:
@marcinbodnar thanks Marcin. I also noticed that `redirectUrl` was lost when redirected to `linkdrop`

I asked the developers to test and they confirm this PR fixed #1818. 

thanks @marcinbodnar @Patrick1904 @stefanopepe

think-in-universe (approved):
LGTM. 

per feedback from developers, this change does fix the issue.

PR #1829, opened 2021-06-17
Refactor Redux state to be keyed by accountId and new folder/file structure

marcinbodnar:


marcinbodnar:
@Igor could you please test it - the problem is that these changes touch every single functionality that we have. I did test it after every single change, and again when it's done. But it would be great if you can test it again in every possible way.

ihaid:
image
Here you can click "Disable 2FA" many times what causes sms spam

ihaid:
The rest looks good for me

ihaid (approved):

Issue #1827, opened 2021-06-17
redirect urls during linkdrop should also work for existing account (when user doesn't create new account)


think-in-universe:
### Overview

As a linkdrop sender, I want to redirect the receiver to a URL after he/she received the dropped NEAR even if the user claimed NEAR with an existing account (didn't create a new account), so I can the linkdrop redirect workflow will be the same for all the users.

As described in #1641, implemented in #1652, and fixed recently in #1818, new users will be redirected to the `redirectUrl`. Besides new created accounts, we also want to redirect the existing users to the `redirectUrl`.

### Acceptance Criteria
- [ ] When a user open a linkdrop (e.g. https://wallet.testnet.near.org/create/testnet/5waZiox2Q11VKAeWpDjTxpkDNaVaXKGSbodBHBqhy2nYt1uLXqH74ctPJPnxfACp9HL3uEspnRGADB9YCzfVZJFy?redirectUrl=https://berryclub.io/), and claim with an existing account, he will still be redirected to https://berryclub.io
- [ ] The existing account's name is passed as an argument to the redirect URL (`accountId`): https://berryclub.io?accountId=someone.testnet

think-in-universe:
The enhancement is also important to one dapp that will be launched around June 25th. so hopefully we could add the enhancement recently.

stefanopepe:
Robert, thank you for this suggestion.
I'll leave the technical discussion to @MaximusHaximus , while my main concern is around phishing and spear phishing attacks.

An attacker might leverage this feature to create malicious airdrops to be distributed via chat channels (or similar). A small amount of tokens might convince the user to click and claim the funds, but then get redirected to a fake application that could ask for the seed passprhase or require the authorization of a full-access key that is capable to drains funds (imagine a copycat of berryclub).
Let us wait for a rapid feedback by @vgrichina and @kcole16 as they spent some time on these type of issues too.

think-in-universe:
thanks Stefano @stefanopepe 

the security concern of phishing makes sense. 

IMHO, since we already supported `redirectUrl` for newly created account, adding `redirectUrl` support for existing users is not less secure than the current model.

Issue #1641, opened 2021-04-09
Allow apps to pass redirect urls during linkdrop


kcole16:
### Overview
As an application user, I want to sign up for an account on Flux, receive a linkdrop to create a NEAR account, and be redirected back to Flux immediately after creation, so I can start predicting the NEAR price.

We have several outstanding, high priority requests to re-implement the redirect after account creation (used to work). Specifically, we should allow apps to pass a `redirectUrl` into a linkdrop link, and redirect rather than sending the user to `/profile`.

### Acceptance Criteria
- [ ] The `/create_account` route accepts a `redirectUrl` parameter
- [ ] After account creation, users are redirected to `redirectUrl`
- [ ] The user's account name is passed as an argument to the redirect URL (`accountId`)

jimmy3dita:
@Patrick1904 @marcinbodnar we need this live within April 22nd, with at least 5 days for testing.
I need your feedback on the resources needed to implement this feature

marcinbodnar:
@Patrick1904 you were working on linkdrop lately (i think), want to handle it?

Patrick1904:
@jimmy3dita @marcinbodnar sure I'm finishing up https://github.com/near/near-wallet/issues/1538 right now so will fix https://github.com/near/near-wallet/issues/1641 after.

ETA for https://github.com/near/near-wallet/issues/1641 is Wednesday the 14th. That should give us enough time to test.

PR #1830, opened 2021-06-18
Add linkdrop redirect for existing users

think-in-universe:
### Description

Resolves: #1827

Similar to #1641 which is designed specific for new users, add url redirect support after an existing user claimed the linkdroip.

### DIFF
--- a/src/utils/wallet.js
+++ b/src/utils/wallet.js
@@ -18,7 +18,6 @@ import {
     redirectTo,
     fundCreateAccount,
     finishAccountSetup,
-    finishLinkdropClaim,
     selectAccount
 } from '../actions/account'
 
@@ -439,8 +438,6 @@ class Wallet {
         });
 
         await contract.claim({ account_id: accountId }, LINKDROP_GAS);
-
-        await store.dispatch(finishLinkdropClaim());
     }
 
     async saveAccount(accountId, keyPair) {
--- a/src/utils/wallet.js
+++ b/src/utils/wallet.js
@@ -18,6 +18,7 @@ import {
     redirectTo,
     fundCreateAccount,
     finishAccountSetup,
+    finishLinkdropClaim,
     selectAccount
 } from '../actions/account'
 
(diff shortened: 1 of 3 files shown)


think-in-universe:
Before we merge the PR, we require more security review as mentioned here by @stefanopepe https://github.com/near/near-wallet/issues/1827#issuecomment-863440319

think-in-universe:
Any ideas about the CI error? does it fail because I didn't submit the branch to https://github.com/near/near-wallet repo directly?

MaximusHaximus:
> Any ideas about the CI error? does it fail because I didn't submit the branch to https://github.com/near/near-wallet repo directly?

Yes, I think that is the problem; it looks like we are initializing Sentry during the build process when we don't need to and really shouldn't be for branch PR builds.  I'll take a look into this and figure out the right way to resolve it :+1:, but I don't think it's a blocker.

src/utils/wallet.js
```diff
@@ -438,6 +439,8 @@ class Wallet {
         });
 
         await contract.claim({ account_id: accountId }, LINKDROP_GAS);
+
+        await store.dispatch(finishLinkdropClaim());
```
MaximusHaximus, line 443: I know that we're already dispatching using `store` in this file, so this wasn't obvious, but we're trying to extricate Redux awareness from `wallet.js`.  Instead of dispatching here, can you please enhance the `CLAIM_LINKDROP_TO_ACCOUNT` action in `actions/account.js` to call this wallet method, then dispatch this action instead of having this wallet method dispatch the action?

MaximusHaximus (changes requested):
Keen to get this merged this sprint, but just one change requested around the usage of `store` in wallet.js, which we're trying to move away from :)

src/utils/wallet.js
think-in-universe: sure. I have updated the action `CLAIM_LINKDROP_TO_ACCOUNT `

MaximusHaximus: :+1: thanks so much

MaximusHaximus (approved):
:+1: :shipit:

PR #1832, opened 2021-06-19
1774#2 - Account creation: new implicit and faucet flows

MaximusHaximus:
Re-adds temporarily reverted code from #1774 :)


src/components/accounts/create/status/AccountNeedsFunding.js
```diff
@@ -0,0 +1,49 @@
+import React from 'react'
+import { Translate } from 'react-localize-redux'
+import CopyIcon from '../../../svg/CopyIcon'
+import ClickToCopy from '../../../common/ClickToCopy'
+import Balance from '../../../common/Balance'
+import StyledContainer from './Style.css'
+
+const AccountNeedsFunding = ({
+    accountId,
+    fundingAddress,
```
MaximusHaximus, line 10: :nit: not a blocker, but it feels like `fundingAddress` and `accountId` are really the same thing, but there are two different `type`s of `accountId` -- `singleUse` vs. `nearName`, from a conceptual perspective (especially given `fundingAddress` really is a full-fledged account on-chain). From the engineering side, it's preferable to keep the concepts as close to reality as possible (reduces cognitive overhead and the numbers of ways we name the same logical _thing_), even if the UI tries to make the user feel like they are totally different things :)

src/components/accounts/recovery_setup/SetupRecoveryMethod.js
```diff
@@ -48,14 +49,16 @@ const StyledContainer = styled(Container)`
...
+        align-items: center;
     }
 
 `
 
 class SetupRecoveryMethod extends Component {
 
     state = {
-        option: 'phrase',
+        option: this.props.router.location.method || 'phrase',
```
MaximusHaximus, line 61: :nit: I think that as we move into container/presentational components, we're going to want this to be a prop named e.g. `method` rather than leak the awareness of `router` into this component (so we can render it as a storybook without injecting the router into our presentational component)

src/components/accounts/recovery_setup/SetupRecoveryMethod.js
```diff
@@ -64,15 +67,22 @@ class SetupRecoveryMethod extends Component {
...
+    emailInput = createRef()
+    phoneInput = createRef()
+
     async componentDidMount() {
         const { router, checkIsNew } = this.props;
         const { method } = router.location;
 
         if (method) {
             this.setState({ option: method });
+            if (method === 'email' || method === 'phone') {
```
MaximusHaximus, line 83: :nit:not a blocker, but at some point "soon" we should ensure that these are encapsulated in a constant somewhere we can import them from so it's immediately identifiable which files use any of these magic strings without needing to search for every possible magic string that's part of the `methods` enumeration in order to do so.

src/components/accounts/recovery_setup/SetupRecoveryMethod.js
```diff
@@ -285,36 +300,49 @@ class SetupRecoveryMethod extends Component {
...
-                        <h4><Translate id='setupRecovery.basicSecurity'/></h4>
+                        <h4>
+                            <Translate id='setupRecovery.basicSecurity' />
+                            <Tooltip translate='profile.security.lessSecureDesc' icon='icon-lg'/>
+                        </h4>
                         <RecoveryOption
-                            onClick={() => this.setState({ option: 'email' })}
+                            onClick={() => {
+                                this.setState({ option: 'email' });
+                                if (option !== 'email') {
```
MaximusHaximus, line 340: At first, I didn't understand why if the option is _not_ email, there would be an emailInput for us to focus? Is this using the correct magic string and ref?  But then I realised we're expecting `setState()` to have finished and the input to exist when this line runs, and we're actually checking `option` from the _pre-update state_.  For correctness, this focus code should exist in `componentDidUpdate()` or perhaps scheduled in the second, optional callback argument to the `setState()` call so that we're sure the update has been applied before we try to focus the input.  I'm not _entirely_ sure that even that would work, as I'm not positive that the DOM has been re-rendered or if only the state update has been completed when the callback fires.  It makes me wonder: could we just add an `autofocus` property to the input(s) that we care about focusing here and avoid this logic entirely?

src/components/accounts/recovery_setup/SetupRecoveryMethod.js
```diff
@@ -330,12 +358,20 @@ class SetupRecoveryMethod extends Component {
...
                                         tabIndex='1'
+                                        ref={this.emailInput}
                                     />
                                 )}
                             </Translate>
                         </RecoveryOption>
                         <RecoveryOption
-                            onClick={() => this.setState({ option: 'phone' })}
+                            onClick={() => {
+                                this.setState({ option: 'phone' });
```
MaximusHaximus, line 368: As above, could we leverage `autofocus` instead?

MaximusHaximus (commented):
:+1: This is a _beast_ !  Nice work.  Let's get it on to staging and hammer on it some.
I added non-blocking comments throughout where I spotted things that are sub-optimal or confusing, but none of them are blockers -- I just wanted to make sure I documented them since I spent a fairly large amount of time reviewing the code impacted by these changes and didn't want to lose that context.

Patrick1904 (approved):

PR #1834, opened 2021-06-20
1775 - Add support for multiple ledger transports

### LINKED ISSUE #1775: Implement ledger support for WebHID & WebUSB
**Problem**
Recent updates to Chromium have broken U2F functionality on all Chromium based browsers.  Furthermore, U2F has already been marked as deprecated by Ledger with an intention to remove support for U2F entirely in Q1 2021.

Although Ledger's official migration guide documentation for moving off of the now deprecated U2F transport indicates that we should replace U2F with WebUSB, both U2F and WebUSB have recently been rendered unusable by the latest Chrome v91.xx update.

There is an alternative transport, WebHID, which still works on the latest version of Chromium, however, both WebUSB and WebHID are still in draft stages, and **neither** one is yet supported by Firefox, which _only_ works with the (now deprecated) U2F transport.

Given these facts, and our desire to maintain ledger support for major browsers for as long as we can, we should implement fallback logic that tries each of the available transports and uses whichever one is supported by the current browser.

This pattern has already been introduced in some other open source projects -- an example code pattern that implements fallback for each of the transport types can be found here:
https://github.com/vacuumlabs/adalite/pull/1007/files

We depend on `near-ledger-js` (https://github.com/near/near-ledger-js) for ledger interactions, along with multiple other near ecosystem projects (like `near-cli`, `near-multistaker`), so it would be nice to have this fallback logic in `near-ledger-js` to avoid duplicating it into multiple repositories, especially given how volatile these standards are; we may very well need to implement integration with Ledger Live with the _next_ iteration of changes to Chrome, Firefox, etc.

Reference migration guide **_(obsolete, given Chrome no longer supports WebUSB as of v91.x)_**:
https://github.com/LedgerHQ/ledgerjs/blob/master/docs/migrate_webusb.md
_"We plan to remove the support of U2F channel proxying on our apps in Q1 2021"_

MaximusHaximus:
Fixes #1775

WiP

## Design
https://www.figma.com/file/TJ8gGKjPfOhnanLH34rzOJ/1834:-Multiple-Ledger-Transports?node-id=370%3A2314

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "js-sha256": "^0.9.0",
     "mixpanel-browser": "^2.41.0",
     "near-api-js": "^0.39.0",
-    "near-ledger-js": "^0.1.2",
+    "near-ledger-js": "0.2.0",
     "near-seed-phrase": "^0.2.0",
     "prop-types": "^15.7.2",
     "qrcode.react": "^1.0.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1722,10 +1722,25 @@
     rxjs "^6.6.6"
     semver "^7.3.4"
 
+"@ledgerhq/devices@^5.51.1":
+  version "5.51.1"
+  resolved "https://registry.yarnpkg.com/@ledgerhq/devices/-/devices-5.51.1.tgz#d741a4a5d8f17c2f9d282fd27147e6fe1999edb7"
+  integrity sha512-4w+P0VkbjzEXC7kv8T1GJ/9AVaP9I6uasMZ/JcdwZBS3qwvKo5A5z9uGhP5c7TvItzcmPb44b5Mw2kT+WjUuAA==
+  dependencies:
+    "@ledgerhq/errors" "^5.50.0"
+    "@ledgerhq/logs" "^5.50.0"
+    rxjs "6"
+    semver "^7.3.5"
+
 "@ledgerhq/errors@^5.34.0", "@ledgerhq/errors@^5.46.0":
   version "5.46.0"
   resolved "https://registry.yarnpkg.com/@ledgerhq/errors/-/errors-5.46.0.tgz#573e1c758dce36acf5f414278d8b5970f45fe2b8"
   integrity sha512-1/q/Tqv+aznX/rO3kdpg3Hv7O3xv68dWHkcfai8BZGTdTIwh6vdguFsdUNJ7eiNxEMKNA9gU+p78ZS/PDzoitQ==
+
+"@ledgerhq/errors@^5.50.0":
+  version "5.50.0"
+  resolved "https://registry.yarnpkg.com/@ledgerhq/errors/-/errors-5.50.0.tgz#e3a6834cb8c19346efca214c1af84ed28e69dad9"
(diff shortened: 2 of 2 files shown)


MaximusHaximus:
@corwinharrell We'll need some design work here, because the new ledger logic will need to be triggered _as a result of a user action_, (e.g. an onClick on a button) and can no longer be done in the background when we render the screen that shows e.g. 'please connect your ledger...' or 'please confirm on your ledger' like we currently do

Also, the new transports are _active_ transports, whereas the old U2F transport was _passive_, meaning the entire negotiation and connection (and failure) would happen at the moment the browser sent the request to the ledger.  With active transports, if someone connects to the ledger, but _then_ they launch the NEAR app, or if they launch the NEAR app, connect to the ledger, then _close the NEAR app_, they will be disconnected from the ledger and need to re-connect (by clicking a button somewhere in the UI!).

See the demo in `near/near-ledger-js`for a barebones example of how I wired this up for a contrived example.  We will need the wallet UI to have basically the same functionality as this demo as far as:
- Presenting a button for the user to initialize a connection to the ledger
- Showing the connection state in the UI
- Updating the connection state in the UI if and when the connection to the ledger is closed
- Allowing the user to re-initialize the connection to the ledger in cases of an error being thrown while attempting to communicate with the ledger -- trying to re-use the same transport after some types of errors have been encountered will _always fail_, so re-connecting with an entirely new transport is the best way to recover from a fatal error communicating with the ledger.

vgrichina:
@MaximusHaximus we generally don't initiate transactions without user clicking on something. Is there any problem with initialising the connection when existing UI actions are performed? Will connection only allow one request per click?

MaximusHaximus:
> @MaximusHaximus we generally don't initiate transactions without user clicking on something. Is there any problem with initialising the connection when existing UI actions are performed? Will connection only allow one request per click?

You can do multiple signings per click, but I believe there is a new behavior to consider vs. U2F (passive), which is that if you open the connection without the NEAR app running, then open the NEAR app _after_ opening the connection, it will be closed 'out from under you', and you will need to create a new connection to actually sign the transaction after you've got the NEAR app running... so you could prompt the user to connect to the ledger... they can click "ok, please do", then when you tell them 'please approve the transaction', they run the NEAR app...it'll drop the connection and force them to re-connect (with a new permission prompt from the browser) before the transaction can be submitted to the ledger for signing.

corwinharrell:
## Design Summary

**Figma Link:** https://www.figma.com/file/TJ8gGKjPfOhnanLH34rzOJ/1834-Multiple-Ledger-Transports?node-id=216%3A193

After navigating to `/sign-in-ledger`, a user must first establish a connection with their device via a button pressed while the device is connected:

Screen Shot 2021-08-19 at 10 33 20 PM

Once the connection has been established, the section state reflects this. They may then click the second button that triggers the existing ledger key addition flow:

Screen Shot 2021-08-19 at 10 33 28 PM

Error states are displayed as modal messages with context sensitive messaging depending on whether the initial connection failed, or if an existing connection was disrupted:

Screen Shot 2021-08-19 at 10 37 50 PM

zavodil:
@vgrichina @stefanopepe 
Is it ready to be merged? I have ongoing questions about chrome&ledger issue from users and developers

andy-haynes:
@MaximusHaximus @heycorwin is this still relevant since fixing Chrome support for ledger? Can we close these out?

Issue #1808, opened 2021-06-07
Don't require CAPTCHA for already funded accounts


vgrichina:
**Problem**

Looks like wallet currently always requires CAPTCHA during account creation, even if user is claiming already funded linkdrop.

**Expected Behavior**

User only has to pass CAPTCHA when account is funded for free by coin-op.near. Linkdrop flow should be as easy as possible as it's not abusable, CAPTCHA gets in a way of onboarding on events, etc.

**Steps to reproduce**

1. Create linkdrop, e.g. using http://redpacket.near.org
2. Try to claim it in the wallet, creating new account.

vgrichina:
@Patrick1904 another small issue with CAPTCHA – I think we should hide the UI component once user already passed captcha, as right now some peeps get confused into doing it over and over again before proceeding.

PR #1840, opened 2021-06-22
Create account: Modify cases where reCaptcha prompt is displayed to the user

Patrick1904:
Closes https://github.com/near/near-wallet/issues/1808

Changes:
1. Remove call to contract-helper to see if faucet is available (`fundedAccountAvailable`). Instead, assume that it is, and allow the existing fallback logic to implicit flow if account creation fails due to NotEnoughBalance.  This will make any spammers who are looking to spam funded accounts have to spend some $$$ solving reCaptchas without any guarantee that when they _do_ solve the reCaptcha that they will actually get a funded account.
2. Only require reCaptcha for non-linkdrop account creation (#1808)


vgrichina:
I kinda feel that explaining what change does right here not gonna be very effective against spammers ;)

MaximusHaximus (commented):
> I kinda feel that explaining what change does right here not gonna be very effective against spammers ;)

Open Source :tm:. It sure has some drawbacks doesn't it?

src/components/accounts/SetupSeedPhrase.js
```diff
@@ -233,17 +234,18 @@ class SetupSeedPhrase extends Component {
...
-                                            localAlert={this.state.localAlert}
+                                            mainLoader={this.props.mainLoader || submitting}
+                                            localAlert={localAlert}
                                             globalAlert={this.props.globalAlert}
                                             onRecaptchaChange={this.handleRecaptchaChange}
                                             ref={(ref) => this.recaptchaRef = ref}
-                                            isNewAccount={this.state.isNewAccount}
+                                            isNewAccount={isNewAccount}
                                             onSubmit={this.handleOnSubmit}
+                                            isLinkDrop={parseFundingOptions(this.props.location.search)}
```
MaximusHaximus, line 248: For type consistency, and to avoid the 'don't surprise me' principle, it would be best if we passed in a boolean for `isLinkDrop` here rather than an object that we check in a truesy way internally.  Since you made `parseFundingOptions()` return `null` we can just update this to `parseFundingOptions(this.props.location.search) !== null` safely :+1:

MaximusHaximus (approved):
Just one thing re: providing a boolean input for `isLinkDrop` property in one spot, otherwise LGTM :+1:

PR #1847, opened 2021-06-22
Revert "Fix limit must number by NEP-181"

Patrick1904:
Reverts near/near-wallet#1844

Reverting this as it causes issues with NFTs from Pluminite. Will have to be fixed.
Screen Shot 2021-06-22 at 3 29 29 PM

### DIFF
--- a/src/actions/nft.js
+++ b/src/actions/nft.js
@@ -61,7 +61,7 @@ async function getTokens(contractName, accountId, { base_uri }) {
         }
 
         // TODO: Pagination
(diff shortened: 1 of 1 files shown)


zavodil:
Great catch!
I redeployed the contract https://github.com/zavodil/pluminite-ui/commit/190781d90ba7775a55bd1a0cf2253ad43631380f  but we have to make a global fix since pluminite is a fork of nfr-simple https://github.com/near-apps/nft-market/blob/main/contracts/nft-simple/src/enumerable.rs#L82

Issue #1845, opened 2021-06-22
NEP-181: invalid type: string \\\"50\\\", expected u64


ilyar:
**Problem**

```
errors.js:18 Uncaught (in promise) Error: Querying call/collection.jmw.near/nft_tokens_for_owner,ZMKC1ggrxR5fACQFWMt9eTYvf3AYLdbUWUFQLZufKt7duBJxQJmHTEMNMAEkicp4XpTGKX7c8ZYWX9i26qGZ8p failed: wasm execution failed with error: FunctionCallError(HostError(GuestPanic { panic_msg: "panicked at \'Failed to deserialize input from JSON.: Error(\"invalid type: string \\\"50\\\", expected u64\", line: 1, column: 62)\', contract/arterra/arterra.rs:138:1" })).
{
  "error": "wasm execution failed with error: FunctionCallError(HostError(GuestPanic { panic_msg: \"panicked at \\'Failed to deserialize input from JSON.: Error(\\\"invalid type: string \\\\\\\"50\\\\\\\", expected u64\\\", line: 1, column: 62)\\', contract/arterra/arterra.rs:138:1\" }))",
  "logs": [],
  "block_height": 40699732,
  "block_hash": "8ZWG5GmKvVfPXtEZucbeYNnWg8SvA2zsGTrRXAsGuaoh"
}
    at f.query (json-rpc-provider.js:80)
    at async g.viewFunction (account.js:271)
```

https://nomicon.io/Standards/NonFungibleToken/Enumeration.html

```rust
// Get list of all tokens owned by a given account
//
// Arguments:
// * `account_id`: a valid NEAR account
// * `from_index`: a string representing an unsigned 128-bit integer,
//    representing the starting index of tokens to return
// * `limit`: the maximum number of tokens to return
//
// Returns a paginated list of all tokens owned by this account
function nft_tokens_for_owner(
  account_id: string,
  from_index: string|null, // default: 0
  limit: number|null, // default: unlimited (could fail due to gas limit)
): Token[] {}
```

https://github.com/ilyar/near-sdk-rs/blob/aeceb993eed7e7a9d6baf9e8e31b1034a663caa8/near-contract-standards/src/non_fungible_token/enumeration/enumeration_impl.rs#L60
```rust
    fn nft_tokens_for_owner(
        &self,
        account_id: ValidAccountId,
        from_index: Option<U128>,
        limit: Option<u64>,
    ) -> Vec<Token> {
```

**Expected Behavior**

```
[...]

PR #1855, opened 2021-06-23
Release 2021-06-23

MaximusHaximus:
- Fixes using incorrect data type for NFT queries using `limit` parameter (https://github.com/near/near-wallet/issues/1845)


Patrick1904 (dismissed):

vgrichina (approved):
LGTM as long as we verified change works with latest contract updates from @mattlockyer

Note that I also merged another change which should help to decrease errors when signing.

mattlockyer:
@vgrichina you should still have the same test NFTs for hip hop in your testnet account.

I updated the testnet contracts as well.

Issue #1714, opened 2021-05-05
Transfer Fungible Tokens - Version 0


stefanopepe:
## Feature recap
In an effort to make NEAR Wallet an asset management tool, this feature will enable users to send their fungible tokens directly from the wallet UI.

This is a blocking feature for our thriving DeFi ecosystem.

More info: [Fungible tokens: manage transactions](https://github.com/near/near-wallet-roadmap/issues/15)

## Story

As a wallet user who wants to send part of his `Fungible Tokens` balance, I want to `select the fungible token from the available list` and define the amount, so I can send the balance to other wallets.

## Acceptance Criteria

- [ ] The `/send-money` function allows the selection of one of the available fungible tokens
- [ ] This function allows the user to define the amount to send
- [ ] This function allows to send "max"
- [ ] This function won't send an amount higher than what is available
- [ ] This function won't list fungible tokens with zero balance
- [ ] This function would show the NEAR tokens needed for the gas
- [ ] The recipient dashboard shows an updated balance (upon refresh)
- [ ] Mobile ok
- [x] #1909

## Unhappy Path
- The user won't see an updated fungible token balance
- The user is not able to select a fungible token even if the balance is > 0
- The user runs out of NEAR tokens to pay for the gas, and the transaction fails
- The contract call fails for unexpected error at the RPC level, and the user is not informed

## Resources

- Figma Document: https://www.figma.com/file/unSbZpLnFv0oYBcU1gURWR/Transfer-Fungible-Tokens?node-id=304%3A593

## NOTES
This is related to epic #456 and [Fungible Tokens: display balance](https://github.com/near/near-wallet-roadmap/issues/14)

jimmy3dita:
@corwinharrell please edit/add/remove any of the content above, it's still a draft

Patrick1904:
@stefanopepe @vgrichina @MaximusHaximus: my understanding is that storage deposit is required for the receiver to be able to receive a FT. The sender could pay for storage deposit on behalf of the receiver before sending tokens, however, the tokens won't show up in the receivers wallet unless the storage deposit was paid for by the receiver (current indexer limitation?).

Is this something that we have to take into consideration when designing this? For example, do we need to add ability for users to manually 'add' FT to their wallet by initiating a storage deposit for a specific token?

stefanopepe:
@Patrick1904 I'm not sure how to address your concern: we have to track this information from the indexer anyway, users are expecting to see tokens in their wallet (e.g. received via Bridge).

[...]

PR #1863, opened 2021-06-26
Transfer fungible tokens: Add 'TX Receipt' module

Patrick1904:

Changes:
1. Adds a send TX 'Receipt' module that will be part of the new fungible token transfer flow. 
2. Adds component for each type of TX receipt, e.g. status, amount, receiver, etc, that will be re-used in other parts of the send flow

Screen Shot 2021-06-25 at 7 59 47 PM


src/utils/getDateAndTime.js
```diff
@@ -0,0 +1,3 @@
+export default function getDateAndTime(block_timestamp) {
+    return new Date(block_timestamp).toLocaleString('en-US', {dateStyle: 'short', timeStyle: 'short'})
```
MaximusHaximus, line 2: Rather than assume EN-US, we can use the browser's locale settings for this:
https://github.com/wojtekmaj/get-user-locale/blob/master/src/index.js

src/components/send/components/entry_types/DateAndTime.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import { Translate } from 'react-localize-redux'
+import StyledContainer from './Style.css'
+import getDateAndTime from '../../../../utils/getDateAndTime'
+
+const DateAndTime = ({ timeStamp, translate }) => {
+    return (
+        <StyledContainer>
+            <Translate id={translate} />
+            <div className='time'>{getDateAndTime(timeStamp)}</div>
```
MaximusHaximus, line 10: I propose we rename `getDateAndTime` `formatTimestampForLocale` (including the file name)

src/components/send/components/TXStatus.js
```diff
@@ -0,0 +1,41 @@
...
+    width: 9px;
+    height: 9px;   
+    background-color: ${props => props.color};
+    border-radius: 50%;
+    margin-right: 10px;
+`
+
+const TXStatus = ({ status }) => {
+
+    const getStatusColor = () => {
```
MaximusHaximus, line 20: Would prefer if this was a pure method that lived in module scope above our component definition, since it doesn't need to be re-created every time we render; we just need to make it take `status` as an argument instead of snorkeling it from the parent closure

src/components/send/components/Receipt.js
```diff
@@ -0,0 +1,73 @@
...
+            decimals: 24,
+            icon: ''
+        },
+        network_fees: '200000000000000000000',
+        receiver_id: 'corwin.near',
+        block_timestamp: 1624648200755
+    }
+}) => {
+
+    const translate = (type) => `sendV2.TXEntry.title.${type}`;
```
MaximusHaximus, line 38: Since this is a pure function that doesn't need to be re-defined each render() cycle, would prefer if it lived in module scope rather than being defined inside the closure

src/components/send/components/Receipt.js
```diff
@@ -0,0 +1,73 @@
...
+        status: 'SuccessValue',
+        token: {
+            symbol: 'NEAR',
+            amount: '50000000000000000000000000000',
+            decimals: 24,
+            icon: ''
+        },
+        network_fees: '200000000000000000000',
+        receiver_id: 'corwin.near',
+        block_timestamp: 1624648200755
```
MaximusHaximus, line 34: Would prefer if we didn't even commit these values (they have a way of being overlooked later and ending up in production).  When we get a storybook defined for this, it should provide this prop value for us so we don't have defaults for testing in the definition of the component.

MaximusHaximus (changes requested):
_**This is awesome!**_ Nice work :+1: I'm digging the small/modular components, and reviewing this was pretty fast due to the tight scope of the changeset 💖 🏅 🏆 

Just a few (minor) things commented on in-line, and we can land this sucker

MaximusHaximus (approved):
:+1: Legit! Merge at your convenience good sir!

PR #1869, opened 2021-06-29
Revert "Create account: Modify cases where reCaptcha prompt is displayed to the user"

Patrick1904:
Reverts near/near-wallet#1840


Patrick1904:
@MaximusHaximus I had to revert this as it I noticed issue with account creation post merge. Will investigate.

PR #1873, opened 2021-06-29
Revert "Revert "Create account: Modify cases where reCaptcha prompt is displayed to the user""

Patrick1904:
Reverts near/near-wallet#1869


MaximusHaximus (approved):
:+1:

PR #1874, opened 2021-06-30
New send flow: Add AmountInput and AccountIDInput components

Patrick1904:

Cleanups:
1. Remove all `semanti-ui-react` `Input` component usage and cleanup
4. Remove various unused code in `PaginationBlock.js`
5. Delete `Search.js` component

New components:

1. Send flow: Add `AmountInput.js` component
2. Send flow: Add `AccountIDInput.js` component

Changes to existing components:
1. Some cleanup of `AccountFormAccountId` component (used on `/send-money` and `/create`)


src/components/send/components/AccountIDInput.js
```diff
@@ -0,0 +1,212 @@
+import React, { Component, createRef } from 'react'
```
MaximusHaximus, line 1: <3 I love that this component doesn't have branching to handle corner cases around whether we're trying to create a new account with the input; the assumption that this is used for only accounts that already exist is really nice from that perspective.

MaximusHaximus (changes requested):
This is looking awesome <3 <3 Some comments in-line :)

src/components/send/components/AmountInput.js
```diff
@@ -0,0 +1,47 @@
...
+const getFontSize = (charLength) => {
+    const baseSize = 60
+    if (charLength >= baseSize) {
+        charLength = baseSize - 5
+    }
+    const fontSize = baseSize - charLength
+    return fontSize
+}
+
+const AmountInput = ({ value, onChange, error }) => {
```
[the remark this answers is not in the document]
Patrick1904: Yep the idea was to keep this as pure UI as possible we can either wrap it in a new component or provide it with props. Similar to how the current `SendContainer.js` is handling amount validation.

> Should we have any formatting behavior here for decimals?

Curious what type of decimal formatting behavior you'd prefer? My POV is cleaner/less is more in this scenario of inputting an amount.

> Should we make the default value be a float including a decimal?

It might be a good idea, current is just the [default design](https://www.figma.com/file/unSbZpLnFv0oYBcU1gURWR/Transfer-Fungible-Tokens?node-id=415%3A291) but can tweak it before we ship.

> Assuming we'll be using this to send NFTs:
> 
> - Should there be validation here so that we can force people to only provide integer values in some cases?
> - Or should we perhaps create an entirely separate input for sending NFTs that disallows entering a float value?

I believe we'd want a separate flow for sending NFTs since you'd be sending a 'collectible' and may benefit from seeing the 'picture' vs a number.

Patrick1904:
@MaximusHaximus Thank you for reviewing. I made the changes you requested. I was able to remove the `input` `ref` but kept the `ref` for the prefix/suffix since there's no event on the element.

Please check the changes, and if no major blockers, would be great if we can merge and add improvements as followup.

src/components/send/components/AccountIDInput.js
```diff
@@ -0,0 +1,214 @@
...
+        const { wrongChar } = this.state
+        const success = localAlert?.success
+        const problem = !localAlert?.success && localAlert?.show
+
+        return (
+            <Translate>
+                {({ translate }) => (
+                    <InputWrapper className={classNames([{ 'success': success }, { 'problem': problem }, { 'wrong-char': wrongChar }])}>
+                        <input
+                            ref={this.input}
```
MaximusHaximus, line 191: 🔪  Extra `ref` here

src/components/send/components/AccountIDInput.js
```diff
@@ -0,0 +1,214 @@
...
+                            placeholder={translate('input.accountId.placeholder')}
+                            required
+                            autoComplete='off'
+                            autoCorrect='off'
+                            autoCapitalize='off'
+                            spellCheck='false'
+                            tabIndex='1'
+                            disabled={disabled}
+                        />
+                        <span className='success-prefix' ref={this.prefix}>
```
MaximusHaximus, line 204: Not a blocker, but this also could be prop driven rather than using a `ref`, since all we are doing with it is setting some styles on it based on props/state.

src/components/accounts/AccountFormAccountId.js
```diff
@@ -222,18 +215,20 @@ class AccountFormAccountId extends Component {
...
             <>
                 <Translate>
                     {({ translate }) => (
-                        <InputWrapper type={type}>
-                            <Input
-                                className={classNames([{'success': localAlert && localAlert.success}, {'problem': localAlert && localAlert.success === false}, {'wrong-char': wrongChar}])}
+                        <InputWrapper className={classNames([type, {'success': success}, {'problem': problem}, {'wrong-char': wrongChar}])}>
+                            <input
                                 name='accountId'
                                 ref={this.input}
```
MaximusHaximus, line 228: :hocho: extra `ref`

MaximusHaximus (approved):
Seems like there's an extra `ref` still left in 2 spots; added comments in-line for those two

Also, we could simplify things further and remove the `ref` we have for `suffix` and `prefix`, since all we're doing is setting some computed `style` values on the span we directly control via our `render()` method, but it's not a blocker.

Good to land, once the `ref={this.input}` entries are removed :+1:

src/components/send/components/AmountInput.js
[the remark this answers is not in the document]
MaximusHaximus: :+1: Seems like a new component for the transfer of NFTs is GTG :+1:

PR #1876, opened 2021-06-30
add zh language support for NFTs

think-in-universe:
Add Simplified and Traditional Chinese support for NFT components

image


Patrick1904:
Thanks @think-in-universe for adding this!

Patrick1904 (approved):

Issue #1875, opened 2021-06-30
`redirectUrl` is lost when claiming linkdrop with an account recovered by seed phrase


think-in-universe:
**Problem**

As implemented in https://github.com/near/near-wallet/pull/1830, now users claim linkdrop with existing account (already logged in) will be redirected successfully. However, if the user didn't login in the browser before and want to login with seed phrase, the `redirectUrl` parameter is lost, and the redirect won't happen.

**Expected Behavior**

The linkdrop redirect should work for users who want to recover their account with seed phrase first. 

**Steps to reproduce**

1. Open linkdrop with link such as https://wallet.testnet.near.org/create/testnet/5waZiox2Q11VKAeWpDjTxpkDNaVaXKGSbodBHBqhy2nYt1uLXqH74ctPJPnxfACp9HL3uEspnRGADB9YCzfVZJFy?redirectUrl=https://berryclub.io/  in incognito mode of your browser
2. After opening the page, the user hasn't logged in before, and needs to recover account with seed phrase.
3. After the account is recovered, the `redirectUrl` parameter is lost, and the redirect won't happen after claiming the linkdrop.

PR #1883, opened 2021-06-30
Fix linkdrop redirect after recover account

think-in-universe:
### Description

This PR fixes https://github.com/near/near-wallet/issues/1875  The redirect doesn't work if a user import a new account and then claim the linkdrop. 

To fix the issue, we need to: 
1. ensure the `redirectUrl` parameter (or the state that saves this parameter) is not lost when jumping back from the `recover-account` and `recover-seed-phrase` pages;
2. the `redirectUrl` parameter was saved into state correctly, so the redirection could happen successfully

### DIFF
--- a/src/components/accounts/LinkdropLanding.js
+++ b/src/components/accounts/LinkdropLanding.js
@@ -3,7 +3,7 @@ import { Translate } from 'react-localize-redux';
 import { connect } from 'react-redux';
 import styled from 'styled-components';
 
-import { checkNearDropBalance, claimLinkdropToAccount, redirectTo } from '../../actions/account';
+import { checkNearDropBalance, claimLinkdropToAccount, redirectTo, handleRefreshUrl } from '../../actions/account';
 import { clearLocalAlert } from '../../actions/status';
 import { Mixpanel } from '../../mixpanel/index';
 import { actionsPending } from '../../utils/alerts';
@@ -71,9 +71,10 @@ class LinkdropLanding extends Component {
     }
 
     componentDidMount() {
-        const { fundingContract, fundingKey } = this.props;
+        const { fundingContract, fundingKey, handleRefreshUrl } = this.props;
         if (fundingContract && fundingKey) {
             this.handleCheckNearDropBalance();
+            handleRefreshUrl();
         }
     }
 
@@ -96,12 +97,15 @@ class LinkdropLanding extends Component {
     }
 
     render() {
-        const { fundingContract, fundingKey, accountId, mainLoader } = this.props;
+        const { fundingContract, fundingKey, accountId, mainLoader, history } = this.props;
         const { balance, invalidNearDrop } = this.state;
         const claimingDrop = actionsPending('CLAIM_LINKDROP_TO_ACCOUNT');
         const fundingAmount = balance;
 
         if (!invalidNearDrop) {
+            const params = new URLSearchParams(history.location.search);
+            const redirectUrl = params.has('redirectUrl') ? `&redirectUrl=${params.get('redirectUrl')}` : '';
+
             return (
                 <StyledContainer className='xs-centered'>
                     <NearGiftIcons/>
@@ -124,7 +128,7 @@ class LinkdropLanding extends Component {
                         </FormButton>
                         :
                         <FormButton
-                            linkTo={`/recover-account?fundingOptions=${encodeURIComponent(JSON.stringify({ fundingContract, fundingKey, fundingAmount }))}`}
+                            linkTo={`/recover-account?fundingOptions=${encodeURIComponent(JSON.stringify({ fundingContract, fundingKey, fundingAmount }))}${redirectUrl}`}
                         >
                             <Translate id='linkdropLanding.ctaLogin'/>
                         </FormButton>
@@ -152,7 +156,8 @@ const mapDispatchToProps = {
     clearLocalAlert,
     checkNearDropBalance,
     claimLinkdropToAccount,
-    redirectTo
+    redirectTo,
+    handleRefreshUrl
 };
 
 const mapStateToProps = ({ account, status }, { match }) => ({
--- a/src/components/accounts/LinkdropLanding.js
+++ b/src/components/accounts/LinkdropLanding.js
@@ -91,9 +91,13 @@ class LinkdropLanding extends Component {
 
     handleClaimNearDrop = async () => {
         const { fundingContract, fundingKey, redirectTo, claimLinkdropToAccount, accountId, url } = this.props;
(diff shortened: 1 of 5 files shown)


src/components/common/FormButton.js
```diff
@@ -443,7 +443,7 @@ const FormButton = ({
         disabled={disabled}
         onClick={(e) => {
             onClick && onClick(e)
-            linkTo && (linkTo.includes('http') ? window.open(linkTo, '_blank') : history.push(linkTo))
+            linkTo && (linkTo.toLowerCase().startsWith('http') ? window.open(linkTo, '_blank') : history.push(linkTo))
```
think-in-universe, line 446: This is needed, otherwise when we click the `Log In and Claim` button (whose `linkTo` contains the redirect url), a new tab will be opened. 

image

src/components/accounts/LinkdropLanding.js
```diff
@@ -70,9 +70,10 @@ class LinkdropLanding extends Component {
     }
 
     componentDidMount() {
-        const { fundingContract, fundingKey } = this.props
+        const { fundingContract, fundingKey, handleRefreshUrl } = this.props
         if (fundingContract && fundingKey) {
             this.handleCheckNearDropBalance()
+            handleRefreshUrl()
```
think-in-universe, line 76: force refresh url here. otherwise the `url` object is empty when reidrected back from `recover-seed-phrase` page

think-in-universe:
@stefanopepe @MaximusHaximus @marcinbodnar Hey guys, this fix is a bit urgent since it's a showstopper for one top-tier project. Could you please have a look when you get time this week?

src/components/accounts/LinkdropLanding.js
MaximusHaximus: @marcinbodnar @Patrick1904 Do you know what the `REFRESH_URL` action exists for?

I don't see any explicit handlers for it, so I'm guessing this is something related to our global `status` reducer maybe?

MaximusHaximus (changes requested):
:+1: Thanks for the contribution Robert! I've added a couple things in-line

src/components/accounts/LinkdropLanding.js
Patrick1904: The `handleRefreshUrl` action calls `refreshUrl` (`REFRESH_URL`) to populate the `url` object on the `account`.

src/components/common/FormButton.js
Patrick1904: I recently added the `_blank` so that external links open in a new tab. I tested this and it works fine - however, might want to separate internal and external link treatment as a followup on this: `linkTo` or `linkToExternal` vs doing this checking for `http` at the beginning of the `URL`.

Patrick1904:
@think-in-universe, in `LinkdropLanding.js`, inside the `handleClaimNearDrop` function, can we check for the `redirectUrl` there instead and do:

```
if (url?.redirectUrl) {
    window.location = `${url.redirectUrl}?accountId=${accountId}`
} else {
    redirectTo('/')
}
```

That way we can remove the `finishLinkdropClaim` function in `actions/account.js` and there aren't any surprises when calling `claimLinkdropToAccount`.

src/components/accounts/LinkdropLanding.js
think-in-universe: I see. Is it OK to call `handleRefreshUrl()` here?

src/components/common/FormButton.js
think-in-universe: do we need to refactor the code with `linkToExternal` in this PR? 

I can do it, but I'm just not sure whether there're some special cases where we don't know whether the link is internal or external when passing the value to `FormButton`. Need to go through the code carefully.

think-in-universe:
@Patrick1904 thanks Patrick. nice suggstion. 

I have moved the logic from `finishLinkdropClaim()` function to `LinkdropLanding.js`

src/components/accounts/LinkdropLanding.js
```diff
@@ -95,12 +96,15 @@ class LinkdropLanding extends Component {
...
     render() {
-        const { fundingContract, fundingKey, accountId, mainLoader } = this.props
+        const { fundingContract, fundingKey, accountId, mainLoader, history } = this.props
         const { balance, invalidNearDrop } = this.state
         const claimingDrop = actionsPending('CLAIM_LINKDROP_TO_ACCOUNT')
         const fundingAmount = balance;
 
         if (!invalidNearDrop) {
+            const params = new URLSearchParams(history.location.search)
+            const redirectUrl = params.has('redirectUrl') ? `&redirectUrl=${params.get('redirectUrl')}` : ''
```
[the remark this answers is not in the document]
think-in-universe: @MaximusHaximus thanks Daryl. I have updatdd and now the `redirectUrl` param is encoded with `encodeURIComponent()`

src/components/accounts/LinkdropLanding.js
Patrick1904: Yes I don't see any issue with calling it on this page when the component mounts

src/components/common/FormButton.js
Patrick1904: @think-in-universe I think it would be great if we did the refactor to use `linkToExternal` for any external links and `linkTo` for router links. All the current links passed to `linkTo` are clear whether they are external or not, but yes we just need to check it thoroughly.

MaximusHaximus (approved):
:+1: LGTM as soon as Patrick is OK with this one landing :+1:

Patrick1904 (approved):
@think-in-universe LGTM! 👍  The `linkToExternal` refactor doesn't seem like a blocker for this PR, especially considering that you need this out ASAP.

think-in-universe:
@Patrick1904 @MaximusHaximus thanks!

yeah. I can submit another PR for `linkToExternal` refactor after this one is merged.

MaximusHaximus:
@think-in-universe I rebased off of master to eliminate some conflicts with the main branch; could you do one final smoke test of the behaviours impacted by this PR and let us know that it's good to go post-rebase?  Then we're ready to merge this sucker. :)

think-in-universe:
@MaximusHaximus yeah. I have tested all the linkdrop redirect behaviors (by creating new account, with logged in account, importing existing account), and all works well locally with the latest code. 

But the Netlify version seems not working well: https://deploy-preview-1883--silly-morse-129a27.netlify.app/   I'm not sure why actually.

MaximusHaximus:
Thanks so much @think-in-universe :) Yep, the deploy preview is indeed a mainnet instance :+1: thanks for the testing! This will go out in this week's release, which I intend to ship later today 🎉  🚀

Issue #704, opened 2020-06-29
Detecting and switching languages


kcole16:
### Overview
We have translations ready for the wallet in Mandarin (and several other languages), and should make it easy for users to switch from English to these languages.

The language switcher is the most important addition, and detecting the browser default language is useful as well (though not likely to cover most cases).

### Stories

As a user, I want to view the Wallet in my browser-defaulted language, so I do not have to manually switch the language.

As a user, I want to manually switch the Wallet language to another supported language, so I can use products in multiple languages.

### Acceptance Criteria

- [x] Users can view and select from all supported languages in the
  - [x] Navbar (Desktop)
  - [x] Menu (Mobile)
- [x] Users can select a supported language, and the Wallet will reload with all copy in that language
- [ ] If Wallet supports a user's browser-defaulted language, this should be the initially selected and loaded language
- [ ] Mobile ok
- [ ] Tested

### Designs
cc @corwinharrell 
<img width="1253" alt="85975288-a9ef5c00-b9a5-11ea-8d70-2364d258827b" src="https://user-images.githubusercontent.com/5624527/86050396-a640f100-ba08-11ea-9adb-705e5284e142.png">

vgrichina:
Merged https://github.com/near/near-wallet/pull/582 /cc @Patrick1904 

Now we just need to adjust UI a little to match what designed by @corwinharrell. 
Note that I noticed also that when I select language other than English on mobile – viewport size seems to increase to be a bit wider than screen. This also needs to be adjusted (might be fixed as side effect of applying design though).

vgrichina:
Looks like also language change doesn't persist for now

picturepan2:
@vgrichina I have done the feature of saving languageCode to localStorage and persist language choice in local. Creating a PR soon.

Patrick1904:
@picturepan2 would also be great if we can add English as fallback in case of missing translation: [`onMissingTranslation`](https://ryandrewjohnson.github.io/react-localize-redux-docs/#handle-missing-translations). Currently the Ledger module on `/profile` is showing the default `Missing translationId` since it's not translated yet. I tested `onMissingTranslation` locally and it works fine.

picturepan2:
One problem I have is importing needed translations dynamically when switching. It flashes long translation strings sometimes which is not good for UX.

Patrick1904:
> One problem I have is importing needed translations dynamically when switching. It flashes long translation strings sometimes which is not good for UX.

[...]

PR #1884, opened 2021-07-01
Automatically detect browser language

think-in-universe:
### Description

Closes https://github.com/near/near-wallet/issues/704

To improve the UX of wallet for non-English users, the wallet should automatically detect the browser language, and switch to the user defined browser language. 

### Discussion

The browser language detection works for all the languages except Tranditional Chinese `zh-hant`, because the browser locale retrieved from `window.navigator.languages` are `zh-TW` or `zh-HK` for the Traditional Chinese users, which don't match with the locale code `zh-hant` we used in the wallet app. All the `zh-XX` (including `zh-CN`, `zh-TW`, `zh-HK`) codes now will be matched to `zh-hans` by default in the current design. 

To fix this issue, we added some special handing for `zh-TW` or `zh-HK` to return `zh-hant` in https://github.com/near/near-wallet/pull/1884/commits/79b3f49893bbccb84b9cda837c173d81fdd967d0.

### DIFF
--- a/src/components/Routing.js
+++ b/src/components/Routing.js
@@ -18,6 +18,7 @@ import translations_zh_hans from '../translations/zh-hans.global.json';
 import translations_zh_hant from '../translations/zh-hant.global.json';
 import { handleClearAlert } from '../utils/alerts';
 import classNames from '../utils/classNames';
+import getBrowserLocale from '../utils/getBrowserLocale';
 import isMobile from '../utils/isMobile';
 import { getAccountIsInactive } from '../utils/localStorage';
 import { reportUiActiveMixpanelThrottled } from '../utils/reportUiActiveMixpanelThrottled';
@@ -111,7 +112,8 @@ class Routing extends Component {
             { name: "繁體中文", code: "zh-hant" }
         ];
 
-        const activeLang = localStorage.getItem("languageCode") || languages[0].code;
+        const browserLanguage = getBrowserLocale(languages.map(l => l.code));
+        const activeLang = localStorage.getItem("languageCode") || browserLanguage || languages[0].code;
 
         this.props.initialize({
             languages,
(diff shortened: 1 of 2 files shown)


Patrick1904:
@think-in-universe this is really great! I tested on my machine.

We recently added a `getUserLocale` function (`utils/getUserLocale.js`) for showing the date/time in local format, and it seems you'd be able to use that, but also add the special handling for traditional Chinese.

think-in-universe:
@Patrick1904 thanks Patrick. I have updated by using `getUserLocales()`

MaximusHaximus (commented):
@think-in-universe I've modified the code in a few key ways:
1. Now the code will rank `zh-TW` and `zh-HK` into the list of requested languages in the order the browser had it added; your code would always assume that if `zh-TW` or `zh-HK` was _anywhere in the languages list_ that the wallet should render in `za-hant`, but we want to respect the order that languages are in the browser's accept list; if someone has `za-TW` as the 5th or 6th language, but has `en-US` as the first language, then the wallet should still render using `en` language.
2. Ranking is done in a single pass
3. Used an object hash to track `matchedLocales`, which fixed weird behaviour where the same supported wallet language was being tracked as a  partial match _and_  an exact match; both were added to the array of `matchedLocale` when they shouldn't have been
4. Deleted the 'sort same score matches based on specificity' -- this should've been handled by the `score` computation you already wrote; I think you only needed this code because of the problem with multiple inaccurate entries being created for the same language (3. above)

Could you give this code a test and make sure it still is good to go for your original use case?  Since this code will likely remain a core part of our localization logic, I just wanted to make sure it was written in slightly a more generic way :)

MaximusHaximus:
Rebased on to master to eliminate conflicts with `Routing.js`

think-in-universe:
@MaximusHaximus thanks Daryl. the change looks great!!

I have tested for the Chinese language use case and it works well.

Patrick1904 (approved):

PR #1889, opened 2021-07-06
ESLint rules for imports and semicolons

### LINKED ISSUE #1888: ESLint rules for import ordering and semicolon requirements
**Problem**
Our current ESLint ruleset is extremely sparse; we should have some solid ESLint rules in place that manage low-level code formatting requirements.

- Organize `import` statements:
https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md

This is the ruleset I'd like to use:
```
    'import/order': [
      'error', {
        alphabetize:        {
          order:           'asc',
          caseInsensitive: true
        },
        'newlines-between': 'always',
        groups:             [
          'builtin',
          ['external', 'internal'],
          ['sibling', 'parent', 'index'],
          'object'
        ]
      }
    ],
```

Running eslint --fix on save should be all the thinking about organizing imports that we ever need to do, and with this plugin every file's imports will be organized for us consistently, without us needing to think about it. 
Having imports organized by the source helps quickly get an idea of the dependencies a module uses
aving the imports sorted alphabetically means that things coming from similar paths in the codebase are grouped together automatically :).
Grouping imports by their source (builtings, modules, then sibling/parent/index) means that the first imports you see are the ones that you don't control, and the last imports you see are the ones that are located the closest to the current file in the filesystem, thus the most likely to be specifically related to the current file.

- Enforce semicolons for all lines
https://eslint.org/docs/2.0.0/rules/semi
Optional semicolons are _usually_ fine, but when things go wrong they do so in ways that are not readily obvious and can take significant effort to debug the root cause of.  For reference, this tiny fix here cost Vlad half a day of effort:

f7149aabf7918a911e8d6f5a2dbbf6fc1154e349

MaximusHaximus:
Closes #1888

- Semicolons everywhere
- Imports organized
- Disabled `no-useless-escape` for regex matching.  In my experience this isn't always reliable and we had some warnings caused by this rule in the existing codebase that couldn't be auto-fixed (because it's not always reliable...)

Everyone should have `eslint --fix` running for every file save :)


Patrick1904 (approved):
LGTM 💯

PR #1890, opened 2021-07-07
Create AvailableToSend component for FT send flow

Patrick1904:

Changes:
1. Create `AvailableToSend.js` accordion style breakdown component

Screen Shot 2021-07-07 at 6 22 07 PM
Screen Shot 2021-07-07 at 6 22 17 PM


src/components/send/components/AvailableToSend.js
```diff
@@ -0,0 +1,44 @@
+import React, { useState } from 'react';
+
+import classNames from '../../../utils/classNames';
+import Accordion from '../../common/Accordion';
+import AccordionTrigger from './AccordionTrigger';
+import Breakdown from './Breakdown.css';
+import Amount from './entry_types/Amount';
+
+const translate = (type) => `sendV2.availableToSend.${type}`;
```
MaximusHaximus, line 9: What do you think about making this a bit more descriptive about what it's doing?

e.g. something like ```const prefixAvailableToSendId = (key) => `sendV2.availableToSend.${type}`;```

src/components/send/components/Breakdown.css.js
```diff
@@ -0,0 +1,29 @@
+import styled from 'styled-components';
```
MaximusHaximus, line 1: Not a blocker, but something to consider:

If we're only going to use this with `AvailableToSend.js`, it'd be nice to get `AvailableToSend` into a directory with this css file co-located with it, especially since there's nothing about the filename to indicate that it is part of (and only part of) the `AvailableToSend` component...

If we do intend to re-use it in multiple components, I wonder if it belongs in an e.g. `css` or similarly named directory rather than co-located with actual react components for clarity that this is a styled div, but not an actual react component...

src/components/send/components/entry_types/Amount.js
```diff
@@ -18,12 +19,14 @@ const RawAmount = ({ symbol, amount, decimals }) => {
     }
 };
 
-const Amount = ({ symbol, amount, decimals, translate, withInfo }) => {
+const Amount = ({ symbol, amount, decimals, translate, infoTranslate }) => {
     /* TODO: Handle long amounts */
-    /* TODO: Add 'withInfo' tooltip */
     return (
         <StyledContainer>
             <Translate id={translate} />
```
MaximusHaximus, line 26: See above comment re: naming these props more specifically :)

MaximusHaximus (changes requested):
Just a few small things added in-line 💖  -- This stuff is looking great :+1:

MaximusHaximus (approved):
:+1: legit, this is looking awesome.

PR #1892, opened 2021-07-07
Transfer Fungible Tokens: prepare requests

marcinbodnar:


src/services/fungibleTokens.js
```diff
@@ -0,0 +1,90 @@
...
+
+    async isStorageBalanceAvailable(contractName, accountId) {
+        return new BN((await this.checkStorageBalance(contractName, accountId)).total).gte(new BN(FT_MINIMUM_STORAGE_BALANCE));
+    }
+
+    async checkStorageBalance(contractName, accountId) {
+        return await this.account.viewFunction(contractName, 'storage_balance_of', { account_id: accountId }).catch(logError);
+    }
+
+    async transfer(contractName, amount, memo, receiver) {
```
MaximusHaximus, line 33: Would be nice to make this take a single object as an argument...e.g.
```javascript
async transfer({ contractMetadata: { contractName, decimal}, amount, memo, receiver }) { ... }
```

It makes it easier to grok at the callsites that use the method, aids in composability, and makes changes less likely to be awkward in the future (for example, if we made `memo` optional).

src/services/fungibleTokens.js
```diff
@@ -0,0 +1,90 @@
...
+        ]);
+    }
+
+    async calculateAmount(contractName, amount) {
+        const { metadata } = await this.getMetadata(contractName);
+        return new BN(10).pow(new BN(metadata.decimals)).mul(new BN(amount)).toString();
+    }
+
+    async transferStorageDeposit(contractName, accountId) {
+        await this.signAndSendTransaction(contractName, [
```
MaximusHaximus, line 49: Rather than await this here, which results in `transferStorageDeposit()` _always_ returning `undefined`, we should just `return this.signAndSendTransaction(...)` so that the caller can inspect the returned result of the transaction submission if they want to.

src/services/fungibleTokens.js
```diff
@@ -0,0 +1,90 @@
...
+            }, FT_TRANSFER_GAS, FT_TRANSFER_DEPOSIT)
+        ]);
+    }
+
+    async calculateAmount(contractName, amount) {
+        const { metadata } = await this.getMetadata(contractName);
+        return new BN(10).pow(new BN(metadata.decimals)).mul(new BN(amount)).toString();
+    }
+
+    async transferStorageDeposit(contractName, accountId) {
```
MaximusHaximus, line 48: Why include `accountId` here as an argument if this object has a specific account as `this.account`?

src/services/fungibleTokens.js
```diff
@@ -0,0 +1,90 @@
...
+        await this.signAndSendTransaction(contractName, [
+            functionCall('storage_deposit', {
+                account_id: accountId,
+                registration_only: true,
+            }, FT_STORAGE_DEPOSIT_GAS, FT_MINIMUM_STORAGE_BALANCE)
+        ]);
+    }
+
+    async signAndSendTransaction(receiverId, actions) {
+        return await this.account.signAndSendTransaction(receiverId, actions).catch(logError);
```
MaximusHaximus, line 58: As with prior comment, we don't want to just ignore _all_ errors ever encountered by this call; if we do, we rob the caller of the ability to react to the errors in any way, either by alerting the user, re-throwing up the call stack, or retrying the operation.

src/services/fungibleTokens.js
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    getLikelyTokenContracts = () => (
```
MaximusHaximus, line 61: 1. 🔪 no need for an expression here
2. As above, we don't want to silently ignore _all_ errors attempting to fetch likelyTokens; only the caller of this method will know what the right course of action is, and by swallowing the error here we have removed the caller's ability to remedy the issue or display an error to the user/whatever else it may do.

src/services/fungibleTokens.js
```diff
@@ -0,0 +1,90 @@
...
+        sendJson('GET', `${ACCOUNT_HELPER_URL}/account/${this.account.accountId}/likelyTokens`).catch(logError)
+    )
+
+    async getMetadata(contractName) {
+        // FungibleTokenMetadata interface
+        // https://github.com/near/NEPs/blob/master/specs/Standards/FungibleToken/Metadata.md
+        const metadata = await this.account.viewFunction(contractName, 'ft_metadata').catch(logError);
```
MaximusHaximus, line 68: Same comments re: silently ignoring all errors; this is a very dangerous pattern to implement in a low level object like this, because it has no context on what an error may mean and how to react to it.

src/services/fungibleTokens.js
```diff
@@ -0,0 +1,90 @@
...
+    
+        return {
+            contractName,
+            metadata
+        };
+    }
+
+    async getBalanceOf(contractName) {
+        const balance = await this.account.viewFunction(contractName, 'ft_balance_of', { account_id: this.account.accountId }).catch(logError);
```
MaximusHaximus, line 77: Same comments re: not silently swallowing all errors that this may encounter

src/services/fungibleTokens.js
```diff
@@ -0,0 +1,90 @@
...
+            contractName,
+            balance
+        };
+    }
+}
+
+const logError = (error) => {
+    console.warn(error);
+    Sentry.captureException(error);
+};
```
MaximusHaximus, line 88: I don't think we even want this method to exist in this file; errors should be thrown to the caller so they can decide what to do... this type of logic really only belongs on the _consumer_ side of this code, not embedded in this object.

src/utils/wallet.js
```diff
@@ -637,7 +637,7 @@ class Wallet {
         return availableKeys;
     }
 
-    async getAccountBasic(accountId) {
+    getAccountBasic(accountId = this.accountId) {
```
MaximusHaximus, line 640: I don't think that we want to add any implied logic here; it makes it too easy for someone to call `getAccountBasic(variableName)` where `variableName` happens to be a reference to an undefined key on an object that they expected to be populated... which then becomes a silent fallback for this function to use the account that happens to be set on the `wallet` object instance (which may or may not make any sense in the context of the caller of `getAccountBasic()`).  I'd rather that bug be discovered when `account.state()` throws an error than when we use the _result_ of `account.state()` for an action that turns out to be using _the wrong account_ for the action...

MaximusHaximus (changes requested):
Questions and changes added in-line.

Also, a comment / description on the PR explaining what you are trying to solve (and _not_ trying to solve) with the code in the PR would be nice in general; empty PR descriptions are less than ideal, especially when it comes to indexing + searching in the future.

src/services/fungibleTokens.js
```diff
@@ -0,0 +1,90 @@
...
+import * as nearApiJs from 'near-api-js';
+
+import sendJson from '../tmp_fetch_send_json';
+import { wallet } from '../utils/wallet';
+import { ACCOUNT_HELPER_URL } from '../utils/wallet';
+
+const FT_MINIMUM_STORAGE_BALANCE = '1250000000000000000000';
+const FT_STORAGE_DEPOSIT_GAS = '30000000000000';
+const FT_TRANSFER_GAS = '30000000000000';
+const FT_TRANSFER_DEPOSIT = '1';
```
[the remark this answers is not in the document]
marcinbodnar: I took `FT_MINIMUM_STORAGE_BALANCE` and `FT_STORAGE_DEPOSIT_GAS` from here:
https://docs.near.org/docs/api/naj-cookbook#wrap-and-unwrap-near

I'm not sure about `FT_TRANSFER_GAS` - it depends on many things, but unused gas is refunded:
https://docs.near.org/docs/concepts/gas#attach-extra-gas-get-refunded
I set it to the same value as creating an account (`storage_deposit`), so it should be enough, and it's still a very low amount of tokens - but I guess we should double-check it with someone? While testing, the amount of gas used was much much less.

src/services/fungibleTokens.js
```diff
@@ -0,0 +1,90 @@
+import * as Sentry from '@sentry/browser';
```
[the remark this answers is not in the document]
marcinbodnar: Done.

src/services/fungibleTokens.js
```diff
@@ -0,0 +1,90 @@
...
+
+const {
+    transactions: {
+        functionCall
+    }
+} = nearApiJs;
+
+class FungibleTokens {
+    constructor() {
+        this.account = wallet.getAccountBasic();
```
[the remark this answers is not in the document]
marcinbodnar: Right, but I'm not sure about stashing the instance in Redux because of redux-persist. Maybe this type of solution:
https://github.com/near/near-wallet/pull/1892/commits/2ce24bb12e03c82f168f2e2dfea9bc97710a3567

It would make sense to keep everything in wallet instance, and it would look much better codewise after we refactor wallet.js into a more reasonable form.

src/services/fungibleTokens.js
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    async isStorageBalanceAvailable(contractName, accountId) {
+        return new BN((await this.checkStorageBalance(contractName, accountId)).total).gte(new BN(FT_MINIMUM_STORAGE_BALANCE));
+    }
+
+    async checkStorageBalance(contractName, accountId) {
+        return await this.account.viewFunction(contractName, 'storage_balance_of', { account_id: accountId }).catch(logError);
```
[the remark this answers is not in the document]
marcinbodnar: Right, done.
I also renamed `checkStorageBalance` into `getStorageBalance` as it will have more sense.

src/services/fungibleTokens.js
```diff
@@ -0,0 +1,90 @@
...
+                amount: await this.calculateAmount(contractName, amount),
+                memo: memo
+            }, FT_TRANSFER_GAS, FT_TRANSFER_DEPOSIT)
+        ]);
+    }
+
+    async calculateAmount(contractName, amount) {
+        const { metadata } = await this.getMetadata(contractName);
+        return new BN(10).pow(new BN(metadata.decimals)).mul(new BN(amount)).toString();
+    }
```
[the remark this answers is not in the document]
marcinbodnar: Right, done.

src/services/fungibleTokens.js
```diff
@@ -0,0 +1,90 @@
...
+            functionCall('ft_transfer', {
+                receiver_id: receiver,
+                amount: await this.calculateAmount(contractName, amount),
+                memo: memo
+            }, FT_TRANSFER_GAS, FT_TRANSFER_DEPOSIT)
+        ]);
+    }
+
+    async calculateAmount(contractName, amount) {
+        const { metadata } = await this.getMetadata(contractName);
```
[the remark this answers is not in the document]
marcinbodnar: I think passing metadata is a better idea, I added a condition that if `contractMetadata` is no being pass, we are checking it inside the method, in case somebody would like to use it this way.

src/services/fungibleTokens.js
marcinbodnar: Done

src/services/fungibleTokens.js
marcinbodnar: It is accountId of the user that you are making the deposit for.

src/services/fungibleTokens.js
marcinbodnar: Done

src/services/fungibleTokens.js
marcinbodnar: Right, done.

src/services/fungibleTokens.js
marcinbodnar: Done.

src/services/fungibleTokens.js
marcinbodnar: Done.

src/services/fungibleTokens.js
marcinbodnar: Done.

src/utils/wallet.js
marcinbodnar: Got it, done.

src/services/fungibleTokens.js
marcinbodnar: Are you sure about that, I mean it's easy to change but we don't use this approach in the project?

src/services/fungibleTokens.js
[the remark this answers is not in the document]
MaximusHaximus: 1. For `FT_MINIMUM_STORAGE_BALANCE `, please add a comment here with a link to the cookbook so that it's clear what source of truth we used to make this decision later :)
2. For `FT_TRANSFER_GAS` please add a comment indicating that we arbitrarily set this to the same value as we use for creating an account and that the remainder is refunded.  I'm not sure I understand the reference to `storage_deposit`; is that the same?
3. What about `FT_TRANSFER_DEPOSIT`?

src/services/fungibleTokens.js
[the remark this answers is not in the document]
MaximusHaximus: Sure, we can keep it as a property of the wallet instance; makes sense to me :+1:

src/services/fungibleTokens.js
MaximusHaximus: :+1: roger that, makes sense.

src/services/fungibleTokens.js
```diff
@@ -0,0 +1,89 @@
...
+
+    async isStorageBalanceAvailable(contractName, accountId) {
+        return new BN((await this.getStorageBalance(contractName, accountId)).total).gte(new BN(FT_MINIMUM_STORAGE_BALANCE));
+    }
+
+    async getStorageBalance(contractName, accountId) {
+        return await this.account.viewFunction(contractName, 'storage_balance_of', { account_id: accountId });
+    }
+
+    async transfer(contractName, amount, memo, receiver, contractMetadata) {
```
MaximusHaximus, line 37: Is `contractName` already in `contractMetadata`?  If so, it would be preferable to use `contractMetadata.contractName` rather than duplicate this information here.

Since we don't actually have a use-case right now where we want to look up the metadata as part of calling `transfer()` (becausee we always want to use the metadata we've already loaded), and we are the only dependency on this code, let's not add a call to the block chain to this function chain _until and **if** we need it_; code that is not being used is a liability that doesn't provide any value -- we want code that exists here to exist _because it is useful_  and is _being used_.  Also, per my prior comment, if we _do_ want to provide loading capabilities to this module, we'll want to model that as a caching provider that can be passed in or constructed when an instance of this object is created; we _really_ don't want to encourage this method reaching out to hit the `viewFunction` _every time it's called_.  _In general_, we don't want extremely low level methods being responsible for _directly_ fetching data in cases where we know the data can and should be cached :). If methods do need to be responsible for fetching data that we expect should be cached, we'll want to model it using something like `dataloader` instances that can be passed to the object -- but for now, we don't need to solve that problem since we will have a source of truth (redux) for contract metadata that we can use to pass it into these calls _directly_, making this object much simpler.

With the elimination of this chain being responsible for loading metadata, we don't need `calculateAmount()` at all, either; we can just use `parseTokenAmount` directly on the metaData passed in to `transfer()`

Also, please convert this function to use an object as parameters per my prior comment, for the reasons I outlined there

...e.g.
```javascript
async transfer({ contractMetadata: { contractName, decimal }, amount, memo, receiver }) {...}
```

src/services/fungibleTokens.js
```diff
@@ -0,0 +1,89 @@
...
+
+    async signAndSendTransaction(receiverId, actions) {
+        return await this.account.signAndSendTransaction(receiverId, actions);
+    }
+
+    getLikelyTokenContracts() {
+        return sendJson('GET', `${ACCOUNT_HELPER_URL}/account/${this.account.accountId}/likelyTokens`);
+    }
+
+    async getMetadata(contractName) {
```
MaximusHaximus, line 71: Since this _entire object_ is now our programmatic interface for interaction with contracts that adhere to our FungibleToken standard, It'd be great if we had a JSDoc _on the the class_ that links to the entire specification.

src/services/fungibleTokens.js
```diff
@@ -0,0 +1,89 @@
...
+            metadata
+        };
+    }
+
+    async getBalanceOf(contractName) {
+        const balance = await this.account.viewFunction(contractName, 'ft_balance_of', { account_id: this.account.accountId });
+        return {
+            contractName,
+            balance
+        };
```
MaximusHaximus, line 87: Since the caller already knows the `contractName` (having provided it as an argument), I'm not sure it makes sense to return what they gave us in the object.  What do you think about making these two methods _just return what they query_ without wrapping the result in an object that repeats `contractName` back to the caller?  It seems less than ideal that the caller needs to extract a property from the result when they are assumedly only calling the method to get the data they _don't already have_.

src/utils/wallet.js
```diff
@@ -637,7 +644,7 @@ class Wallet {
         return availableKeys;
     }
 
-    async getAccountBasic(accountId) {
+    getAccountBasic(accountId) {
```
MaximusHaximus, line 647: Please find all usages of this method and remove the `await` keyword for clarity and consistency, and also make sure that there isn't anywhere that is using `getAccountBasic(xxxx).then(...)`, since that will be a breaking usage of the method.  It might seem safe to leave `await` keywords on usages of the function, but if we target any environment without native `async/await` support, these will be transpiled down to Promise usage...and when that happens, the transpiled code may break because it will be converted to `getAccountBasic(xxxx).then(...)` but `getAccountBasic()` won't return a Promise because it's no longer marked `async`.

Also just a note, this type of change will need to be treated as a breaking change in the future when we are exporting `Wallet` for usage by other apps/consumers, because  we have changed the function return type from `a Promise that resolves to a near-api-js Account object` to `a near-api-js Account object`

MaximusHaximus (changes requested):
:+1: rock on.  Just a couple things remaining then we can get this sucker merged :)

src/services/fungibleTokens.js
MaximusHaximus: Sorry about the delayed response, I just saw this comment(Github hid it!) :)

Re: am I sure, yes, named arguments are a Good Thing:tm: in a general way for readability, at the cost of a small amount of verbosity.

As an example, let us look at what it will look like when someone _calls_ this function:

```javascript
transferTo('accountId1', '15', 'accountId2', 'accountId3');
```

Without finding the implementation of `transferTo()` and reading it, I will have no idea what each of those arguments are. Is the receiver the first argument? Or is that the contract? The memo in this case is another NEAR account (for whatever reason) -- is that the receiver account? etc.

Imagine that `memo` becomes optional; this can present real problems that I have seen cause issues in the past:
```javascript
transferTo('accountId1', '15', null, 'accountId3');
```

Since it is now entirely likely that someone will actually pass `null` for the `memo`, we should probably add handling for `memo` being passed as `null` so we don't actually end up passing `null` to the call transfer call.  If we do, it will likely be automatically stringified into the string form of `null` which is actually a valid value for `memo`!
Javascript is tricky; it's in our long-term best interests to avoid needing this type of pattern; we can write simpler code if there is not an optional argument somewhere in the signature that is not the last argument.

Compare what it looks like to call the _same function_  if we use an object for the arguments:
```javascript
transferTo({
  contractMetadata: {
    contractName: 'accountId1'
  },
  amount: '10',
  memo: 'accountId2',
  receiver: 'accountId3'
});

// If `memo` becomes optional, _people can simply not provide it_ , and there is no incentive to pass `null` as a placeholder to maintain the same arity of the function.
transferTo({
  contractMetadata: {
    contractName: 'accountId1'
  },
  amount: '10',
  receiver: 'accountId3'
});
```

Even better, we can name `amount` something more explicit and it will be even clearer at the callsite -- e.g.

```javascript
transferTo({ 
  contractMetadata: {
    contractName: 'accountId1'
    ...
  },
  amountInYoctoNear: '10',
  receiver: 'accountId3'
});
```

Notice that the callsite where the function is _used_ is explicit about what you are doing; one can read the function call and immediately understand what each argument is doing without needing to read the implementation/full function definition of `transferTo()`.  This makes _reading the code_ a lot easier, even though it doesn't make any different to the computer, especially when you are reading code that you are unfamiliar with and trying to get an idea of what is happening without getting into the detail of exactly what each function's signature is.

It's also extremely nice for composability; order of arguments _is generally not important_; it only _becomes_ important if we have positional arguments.  If you are calling two or three functions in a row that take the same types as arguments, it is much less cognitive overhead if each of those 3 functions doesn't have the types in a different order.  Usually with named arguments, all 3 functions can have _the same name for an argument of a particular type_; you never have to shuffle the order of things because functions have slightly different orders of arguments.  Also as we move to typescript, re-using and composing types is a lot easier if we use objects to define function arguments; positional arguments tend to make it difficult to re-use types, especially when some functions take subsets of types that their calling functions take as arguments; you end up with a lot of very explicit type signatures for functions rather than using a few domain-driven types and passing subsets of those types around internally (see https://www.typescriptlang.org/docs/handbook/utility-types.html).

I understand we often have used positional arguments exclusively, even for functions with a large number of arguments and for functions with optional arguments in the past, but I want this to change, _especially_ for functions with multiple arguments, optional arguments, or arguments that we can conceivably imagine becoming optional.

This also aligns with an initiative happening in `near-api-js` to replace many function signatures that use  positional arguments to take an object as an argument instead.

cc: @Patrick1904 as this is relevant on a macro level to the codebase as a whole, and to the entire JS ecosystem at NEAR

I recommend spending a little time and watching this video and internalizing the ideas presented there, as it is a really good overview about how we should reason about cohesion and coupling in software:
https://www.youtube.com/watch?v=Ip2o4vbAK3s

What we're discussing here now is _removing connascence of order_.
We only _want_ _connascence of order_ when the _order itself is important_ in some way.  If we add connascence of order to things where the order is incidental or unimportant, all we are doing is adding additional coupling that we need to work around from then on.  Connascence of name is weaker than connascence of order and it is easier to reason about and keep consistent :)

src/services/fungibleTokens.js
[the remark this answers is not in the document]
MaximusHaximus: Unless we have code that actively uses the fetching of metadata here, I think we should just require the metadata for now.  If we change the function signature to be non-positional, adding a new `contractName` argument later that can be provided _instead_ of `contractMetadata` will not be a breaking change -- but until we have a use case where we need this code to be fetching metadata, let's leave that functionality out entirely.  When/if we get to the point where we want to enable this object to do the fetching as part of this function call, we will want to drive how we implement that based on the details of why we need it to be doing the fetch -- most likely we'll want to use a caching provider of some kind at that point rather than directly call `ft_metadata` _every time someone calls transfer_, since we know the metadata of the contract can/should be cached.  Let's just cross that bridge when we have to, and rely on passing the contract metadata into this function call for now :). If someone needs to get the metadata, it's a trivial thing for them to call getMetadata _themselves_, e.g.:
```javascript
const FTMetadata = await fungibleTokenManager.getMetadata(contractName);

transfer({contractMetaData: { contractName, ...FTMetadata }, amount, receiver, memo });
```

Or if they want to, they can do what we've done with redux, and populate a cache with the result of `getMetadata()`, then pass the cached result into `transfer()`.  The key thing is that our really low-level method isn't responsible for fetching the data just to perform a transfer; the things it requires are passed to it rather than looked up dynamically.  This also makes testing the function a lot easier, since there is no dependency on the result of a `near-api-js` `functionCall` nested inside `transfer()`.

src/services/fungibleTokens.js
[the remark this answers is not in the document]
marcinbodnar: Done

src/services/fungibleTokens.js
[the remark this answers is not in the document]
marcinbodnar: Got it, done.

src/services/fungibleTokens.js
marcinbodnar: Got it, I wasn't sure if you would like to use this approach in the whole project. Done.

src/services/fungibleTokens.js
marcinbodnar: > Is contractName already in contractMetadata?

No, I mean `ft_metadata` doesn't returns `contractName`, but I'm adding it in `getMetadata`. 

> With the elimination of this chain being responsible for loading metadata, we don't need calculateAmount() at all

Right, already removed it before reading this comment :) 

Done.

src/services/fungibleTokens.js
marcinbodnar: Sure thing, let me handle it in the near future, as time is chasing us.

src/services/fungibleTokens.js
marcinbodnar: We sure can, I think I added `contractName` because wanted to avoid passing it in action metadata to the reducer. 

Done.

src/utils/wallet.js
marcinbodnar: Thanks, I removed it at some point and totally forgot about adjusting calls.

> Also just a note, this type of change will need to be treated as a breaking change in the future

Right, roger that.

src/services/fungibleTokens.js
MaximusHaximus: I don't think we're in quite that much of a hurry -- it should only take 30 seconds; you don't even need to go find the spec, because it's the root of the link we've already got here on `getMetaData()`:
https://github.com/near/NEPs/tree/master/specs/Standards/FungibleToken

marcinbodnar: Done. I understood that you would like to add comments to all methods with params descriptions :)

MaximusHaximus: Ah, No, we'll get to JSDoc'ing later I think :D :) thanks!

marcinbodnar: Sure thing, it would be helpful for external contributors for sure.

src/components/accounts/SetupImplicit.js
```diff
@@ -119,7 +119,7 @@ class SetupImplicit extends Component {
         const { accountFunded } = this.state;
 
         if (!accountFunded) {
-            const account = await dispatch(getAccountBasic(implicitAccountId));
+            const account = dispatch(getAccountBasic(implicitAccountId));
```
MaximusHaximus, line 122: Since `getAccountBasic()` is now _entirely synchronous_, and has no error handling metadata associated with the action, I don't think we get _any_ benefit from `dispatch()`ing it at all.  A brief search for this shows me that this is the _only place in the codebase_ that uses this action --  _Everywhere_ else just calls `wallet.getAccountBasic()` directly). Let's remove the action entirely and just call `wallet.getAccountBasic()` directly here for consistency, instead of having this go through Redux unnecessarily.

src/services/fungibleTokens.js
```diff
@@ -0,0 +1,81 @@
+import BN from 'bn.js';
```
MaximusHaximus, line 1: 1. Lets export the class as the `default` export so that consumers of the class don't need to use a named import when the class export  is the _primary_ and _only_ reason the file exists
2. I'd like for us to start following airbnb's styling convention for file naming for consistency and readability -- that means making the filename here match the casing of the `default` export -- in this case, PascalCase so they match the name of the class itself (easier indexing and searching) (ref: https://github.com/airbnb/javascript#naming--filename-matches-export).

src/services/fungibleTokens.js
```diff
@@ -0,0 +1,81 @@
...
+const FT_MINIMUM_STORAGE_BALANCE = parseNearAmount('0.00125');
+const FT_STORAGE_DEPOSIT_GAS = parseNearAmount('0.00000000003');
+
+// set this to the same value as we use for creating an account and the remainder is refunded
+const FT_TRANSFER_GAS = parseNearAmount('0.00000000003');
+
+// contract might require an attached depositof of at least 1 yoctoNear on transfer methods
+// "This 1 yoctoNEAR is not enforced by this standard, but is encouraged to do. While ability to receive attached deposit is enforced by this token."
+// from: https://github.com/near/NEPs/issues/141
+const FT_TRANSFER_DEPOSIT = '1';
```
MaximusHaximus, line 30: 💖

src/services/fungibleTokens.js
```diff
@@ -0,0 +1,81 @@
...
+
+    async isStorageBalanceAvailable(contractName, accountId) {
+        return new BN((await this.getStorageBalance(contractName, accountId)).total).gte(new BN(FT_MINIMUM_STORAGE_BALANCE));
+    }
+
+    async getStorageBalance(contractName, accountId) {
+        return await this.account.viewFunction(contractName, 'storage_balance_of', { account_id: accountId });
+    }
+
+    async transfer({ contractMetadata: { contractName, decimals }, amount, receiverId, memo }) {
```
MaximusHaximus, line 47: Given your other comment that `contractMetadata` does not actually include `contractName`, I've been thinking about this a bit more -- what do you think about naming the first argument here 'token' instead?  

That way we don't have a concept of 'wallet contract metadata'  object that has a property on it that isn't _actually_ a part of the 'NEP141 token contract metadata' spec.

e.g.:
```javascript
async transfer({
  token: {
    contractName,
    metadata: {
      decimals
    }
  },
  amount,
  receiverId,
  memo
}) {...}
```

This way everything in `token.metadata` will exactly match the NEP141 spec, giving us a re-usable type like e.g. `NEP141TokenMetadata` when we get to TS, and `token.contractName` is just a string that is the name of the contract name used to interact with the token.

I'm thinking about this from a few angles:
1. If we start adding properties to `metadata` payloads, we run the risk of there being naming conflicts between what we have added and what the metadata endpoint returns in the future.
2. Our metadata won't match the spec, so anyone referencing the spec won't understand why there are properties on our metadata that aren't in the spec.
3. It will be tricky to work around the fact that there are two different object shapes for things that are both "metadata" -- and which one you are working with will depend entirely on where in the codebase we're working and whether or not we've mutated the metadata yet to add `contractName`.  By naming it differently it'll be easier to understand whether we're working with the `{ contractName, metadata }` version or just `metadata` returned from the NEP141 contract endpoint.

What do you think?

MaximusHaximus (changes requested):
I added some small / quick things in-line, primarily around naming :).
I think we're just about ready to rock here.

src/services/fungibleTokens.js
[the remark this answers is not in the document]
MaximusHaximus: :+1: awesome, looking good

src/components/accounts/SetupImplicit.js
marcinbodnar: Done.

src/services/fungibleTokens.js
marcinbodnar: Done.

src/services/fungibleTokens.js
marcinbodnar: 100% agree. Done.

MaximusHaximus (approved):
:+1: rock on, Lets Get This Merged! :)

Issue #1877, opened 2021-06-30
Add a URL parameter `tab` to open Collectibles tab directly


think-in-universe:
## Story

As a NFT DApp developer who cares about the NFT user workflow of my users, I want to make it possible to redirect my user to the page that can show the Collectibles without clicking the tab, so I can show my user the latest NFTs they have received.

Is it possible to add a query parameter like `?tab=collectibles` in the link, so the page will open the corresponding tab when the user open the **Wallet** page

![image](https://user-images.githubusercontent.com/46699230/123933483-01fbdc80-d9c5-11eb-8f51-e198611ed1bb.png)

## Acceptance Criteria

- [ ] If the wallet page url contains the `?tab=collectibles`, open the **Collectibles** tab directly
- [ ] If the wallet page doesn't contains the tab parameter, or `?tab=balances`, open the **Balances** tab directly.
- [ ] Mobile ok
- [ ] Tested

## NOTES

- the user story is requested by one NFT project that wants to enhance their UX for users.

ilblackdragon:
Btw, I would love to have collectibles and balances page for non my account.
Right now you can only open profile account for other accounts, but there is no reason why other pages shouldn't work the same.

stefanopepe:
@corwinharrell I would batch Robert's request with the pagination.
@MaximusHaximus I suppose the footprint of this additional change is acceptable

stefanopepe:
@MaximusHaximus assigned this to you, so we can make sure it will make it in the current sprint (or not). Feel free to "un-assign" when done

PR #1895, opened 2021-07-08
`tab` URL parameter in Wallet page

think-in-universe:
### Description

Resolve: https://github.com/near/near-wallet/issues/1877

1. if open https://near-wallet-pr-1895.onrender.com/?tab=nft, the `Collectibles` tab will be open
2. if `tab` is not set, or set to other values, the `Balance` tab will be open
3. when click on the tab `Collectibles`, URL will be changed to https://near-wallet-pr-1895.onrender.com/?tab=nft
4. when click on the tab `Balance`, URL will be changed to https://near-wallet-pr-1895.onrender.com

### Discussion

1. Is it OK to use the `tab` parameter? should we use other params like `token`?
2. Is it OK to use the `nft` value for `tab`? should we use `collectibles`?


Patrick1904:
Thanks for adding this @think-in-universe - it's a nice detail worth supporting IMO. 

I think it'd be good to align the language like you suggested and call it whatever is shown on the UI. So in this case, only use `balances` and `collectibles`, for both state and URL param. 

We could also re-name the state to `const [view, setView] = useState(getView());`

Also nit but it'd be nice to declare `const location = props.history.location;` and move the `getTokenView (getView)` function to right above the `switchTokenView (switchView)` function just to make it feel a bit more organized.

MaximusHaximus:
@think-in-universe Thanks for taking the initiative to work on this concept :) I think it'll be a nice addition!

I think that conceptually, this is part of our `Routing` stack, not an internal behaviour of the `Wallet` component.  I think we should avoid adding internal state management to the `Wallet` component, and instead handle this navigation at our Router level; this is how we handle everything that is driven via URL paths/parameters/queries that control major portions of the UI, and it keeps tight coupling with the `location` out of our presentational components (which is a good thing for testability and for keeping our display components simple).

We could change from using the component argument to configure the Route to using the render API instead.
https://reactrouter.com/web/api/Route/render-func

Since the `render()` function receives the Route properties (including location -- ref: https://reactrouter.com/web/api/Route/route-props), we can render the child component with different props, based on anything in the location object (such as parameters, paths, or query strings).

Then, by leveraging the router for this logic, changing the active tab will become a completely standard route/navigation call instead of us mucking around with local state in the `Wallet` component -- and since how the Wallet component functions would be props driven rather then internally inferred by inspecting router properties itself and then setting and checking locally managed state, we can later add tests to it that won't rely on the location being mocked when we render the component.

think-in-universe:
@Patrick1904 @MaximusHaximus thanks for the suggstions :) 

updated, and please have a look.

src/components/wallet/Wallet.js
```diff
@@ -306,26 +306,26 @@ export function Wallet() {
...
-                            className={classNames(['tab-collectibles', tokenView !== 'nonFungibleTokens' ? 'inactive' : ''])}
-                            onClick={() => setTokenView('nonFungibleTokens')}
+                            className={classNames(['tab-collectibles', tab !== 'collectibles' ? 'inactive' : ''])}
+                            onClick={() => setTab('collectibles')}
                         >
                             <Translate id='wallet.collectibles' />
                         </div>
                     </div>
-                    {tokenView === 'fungibleTokens' &&
+                    {!tab &&
```
Patrick1904, line 321: To guard against an invalid tab param, e.g. `tab=blabla` that would result in neither balances nor collectibles being rendered, we can do:

```
{tab === 'collectibles' ?
    <NFTs tokens={sortedNFTs} />
    :
    <FungibleTokens
        balance={balance}
        tokensLoader={tokensLoader}
        sortedTokens={sortedTokens}
    />
}
```

And then just change `L309` to:
`className={classNames(['tab-balances', tab === 'collectibles' ? 'inactive' : ''])}`

think-in-universe: Thanks Patrick! good catch. fixed now.

think-in-universe:
@MaximusHaximus @Patrick1904 

please review whether my merge conflicts resolution is correct or not in https://github.com/near/near-wallet/pull/1895/commits/ee845ed3daa025fc4f4ccb374c9fc6ed0ce0fdaf

Patrick1904 (approved):

MaximusHaximus (approved):
@think-in-universe Thanks for the submission + updates to move things into the Router layer!

I noticed two issues:
1. We were completely creating the URL to navigate to when the `tab` was clicked, which had a side-effect of removing any `hash` value in the URL.
2. If someone called `setTab()` with an empty string, then we were deleting the _entire_ `query` portion of the URL _and_ the hash; switching tabs shouldn't necessarily remove other parts of the URL such as `hash` and other `query` parameters.

In the interest of time, given that we want to ship this today, I have pushed a commit to resolve both of these cases :)

Thanks again for your help!

think-in-universe:
@MaximusHaximus thanks Daryl. nice fix, and great team work. 

We delivered a bunch of critical updates in a short period of time. That's really amazing and helpful. Thanks a lot everyone.

Issue #1865, opened 2021-06-28
Transfer Fungible Tokens: prepare redux actions


marcinbodnar:
Proposed actions:
 - SEND/TRANSFER/NEAR
 - SEND/TRANSFER/NEP141
 - SEND/PAY_STORAGE_DEPOSIT
 - SEND/IS_STORAGE_BALANCE_AVAILABLE
 - SEND/SET_TX_STATUS

PR #1903, opened 2021-07-09
Transfer Fungible Tokens: prepare redux actions

marcinbodnar:
Set of actions to handle Redux state management for transfer fungible tokens.

Async `transfer` action will handle:
 - checking and performing storage deposits needed if a fungible token transfer is happening
 - performing transfer
 - setting transfer status if the transaction was successful

There is also `isStorageBalanceAvailable` action that should be dispatched before the transfer is requested, and the result should be passed to `transfer` action. This is because the user needs to be aware that he will be paying (or not paying) for the receiver's storage deposit

### DIFF
--- a/src/actions/send.js
+++ b/src/actions/send.js
@@ -26,7 +26,7 @@ export const transfer = ({
             dispatch(send.setTxStatus(transaction.hash, 'success'));
         }
     } else if(type === TOKEN_TYPES.NEP141) {
-        if (!isStorageBalanceAvailable) {
+        if (isStorageBalanceAvailable === false) {
             await dispatch(send.payStorageDeposit(contractName, receiverId));
         }
 
--- a/src/actions/send.js
+++ b/src/actions/send.js
@@ -4,7 +4,7 @@ import { showAlert } from '../utils/alerts';
 import { wallet } from '../utils/wallet';
 
 export const transfer = ({ contractName, amount, memo, receiverId, isStorageBalanceAvailable }) => async (dispatch, gesState) => {
-    if (!isStorageBalanceAvailable) {
+    if (contractName && !isStorageBalanceAvailable) {
         await dispatch(send.transfer.storageDeposit(contractName, receiverId));
     }
 
--- a/src/actions/send.js
+++ b/src/actions/send.js
@@ -28,7 +28,7 @@ export const transfer = ({
             if (status?.SuccessValue) {
                 dispatch(send.setTxStatus({
                     hash: transaction.hash,
-                    newStatus: 'success'
+                    status: 'success'
                 }));
             }
             break;
@@ -59,7 +59,7 @@ export const transfer = ({
             } else {
                 dispatch(send.setTxStatus({
                     hash,
-                    newStatus: 'success'
+                    status: 'success'
                 }));
             }
             break;
--- a/src/actions/send.js
+++ b/src/actions/send.js
@@ -35,7 +35,12 @@ export const transfer = ({
         }
         case TOKEN_TYPES.NEP141: {
             if (isStorageBalanceAvailable === false) {
-                await dispatch(send.payStorageDeposit(contractName, receiverId));
+                const { status } = await dispatch(send.payStorageDeposit(contractName, receiverId));
+
+                if (!status.SuccessValue || typeof status.SuccessValue !== 'string') {
+                    const { error_message, error_type} = status.Failure;
+                    throw new WalletError(error_message, error_type);
+                }
             }
 
             const { transaction: { hash }, status } = await dispatch(send.transfer.nep141({
@@ -49,13 +54,13 @@ export const transfer = ({
             }));
 
             if (!status.SuccessValue || typeof status.SuccessValue !== 'string') {
+                const { error_message, error_type} = status.Failure;
+                throw new WalletError(error_message, error_type);
+            } else {
                 dispatch(send.setTxStatus({
                     hash,
                     newStatus: 'success'
                 }));
-            } else {
-                const { error_message, error_type} = status.Failure;
-                throw new WalletError(error_message, error_type);
             }
             break;
         }
--- a/src/actions/send.js
+++ b/src/actions/send.js
@@ -27,5 +27,6 @@ export const { send } = createActions({
                 () => showAlert({ onlyError: true })
             ],
         },
+        SET_TX_STATUS: null,
     }
 });
--- a/src/actions/send.js
+++ b/src/actions/send.js
@@ -46,7 +46,7 @@ export const transfer = ({
             dispatch(send.setTxStatus(transaction.hash, 'success'));
         }
     } else {
-        throw new WalletError(`Could not transfer unsupported token: ${type}`, 'send.unsupportedToken', { type: type });
+        throw new WalletError(`Could not transfer unsupported token: ${type}`, 'send.unsupportedToken', { type });
     }
 };
 
@@ -58,16 +58,16 @@ export const { send } = createActions({
                 () => showAlert({ onlyError: true })
             ],
             NEP141: [
-                () => {},
+                wallet.fungibleTokens.transfer.bind(wallet),
                 () => showAlert({ onlyError: true })
             ],
         },
         PAY_STORAGE_DEPOSIT: [
-            () => {},
+            wallet.fungibleTokens.transferStorageDeposit.bind(wallet),
             () => showAlert({ onlyError: true })
         ],
         IS_STORAGE_BALANCE_AVAILABLE: [
-            () => {},
+            wallet.fungibleTokens.isStorageBalanceAvailable.bind(wallet),
             () => showAlert({ onlyError: true })
         ],
         SET_TX_STATUS: null,
--- a/src/actions/send.js
+++ b/src/actions/send.js
@@ -12,7 +12,7 @@ export const { send } = createActions({
     SEND: {
         TRANSFER: {
             NEAR: [
-                () => {},
+                wallet.sendMoney.bind(wallet),
                 () => showAlert({ onlyError: true })
             ],
             TOKENS: [
--- a/src/actions/send.js
+++ b/src/actions/send.js
@@ -3,7 +3,7 @@ import { createActions } from 'redux-actions';
 import { showAlert } from '../utils/alerts';
 import { wallet } from '../utils/wallet';
 
-export const transfer = ({ contractName, amount, memo, receiverId, isStorageBalanceAvailable }) => async (dispatch, gesState) => {
+export const transfer = ({ contractName, amount, memo, receiverId, isStorageBalanceAvailable }) => async (dispatch) => {
     if (contractName && !isStorageBalanceAvailable) {
         await dispatch(send.transfer.storageDeposit(contractName, receiverId));
     }
--- a/src/actions/send.js
+++ b/src/actions/send.js
@@ -1,2 +1,2 @@
-export const transfer = () => async (dispatch, gesState) => {
+export const transfer = ({ contractName, amount, memo, receiverId, isStorageBalanceAvailable }) => async (dispatch, gesState) => {
 };
--- a/src/actions/send.js
+++ b/src/actions/send.js
@@ -66,6 +66,14 @@ export const { send } = createActions({
             ],
             NEP141: [
                 wallet.fungibleTokens.transfer?.bind(wallet.fungibleTokens),
+                ({
+                    amount,
+                    receiverId
+                }) => ({
+                    ...showAlert({ onlyError: true }),
+                    amount,
+                    receiverId
+                })
             ],
         },
         PAY_STORAGE_DEPOSIT: [
--- a/src/actions/send.js
+++ b/src/actions/send.js
@@ -15,11 +15,10 @@ export const transfer = ({
     params: {
         contractName, 
         amount, 
-        memo, 
-        receiverId
+        receiverId,
+        memo
     }
-}) => async (dispatch) => {
-
+}) => async (dispatch, getState) => {
     if (type === TOKEN_TYPES.NEAR) {
         const { transaction, status } = await dispatch(send.transfer.near(receiverId, amount));
 
@@ -31,7 +30,17 @@ export const transfer = ({
             await dispatch(send.payStorageDeposit(contractName, receiverId));
         }
 
-        const { transaction, status } = await dispatch(send.transfer.nep141(contractName, amount, memo, receiverId));
+        const { transaction, status } = await dispatch(send.transfer.nep141({
+            token: { 
+                contractName,
+                metadata: { 
+                    decimals: getState().tokens.tokens[contractName]
+                }
+            },
+            amount,
+            receiverId,
+            memo
+        }));
 
         if (status?.SuccessValue) {
             dispatch(send.setTxStatus(transaction.hash, 'success'));
--- a/src/actions/send.js
+++ b/src/actions/send.js
@@ -31,6 +31,10 @@ export const { send } = createActions({
                 () => showAlert({ onlyError: true })
             ],
         },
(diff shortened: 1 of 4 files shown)


src/actions/send.js
```diff
@@ -0,0 +1,40 @@
+import { createActions } from 'redux-actions';
+
+import { showAlert } from '../utils/alerts';
+import { wallet } from '../utils/wallet';
+
+export const transfer = ({ contractName, amount, memo, receiverId, isStorageBalanceAvailable }) => async (dispatch, gesState) => {
```
MaximusHaximus, line 6: Typo `gesState` -> `getState`

src/actions/send.js
```diff
@@ -0,0 +1,40 @@
...
+    if (contractName && !isStorageBalanceAvailable) {
```
MaximusHaximus, line 7: Rather than making implicit assumptions around `contractName`, we want the type of token being sent to be _explicitly defined_ like we had talked about previously

e.g.
```javascript
const TOKEN_TYPES = {
  NEAR: 'NEAR',
  NEP141: 'NEP141',
  ...
};

...

export const transfer({
  type<key of TOKEN_TYPES>,
  params: {
    contractName, // Not present when type === 'NEAR', but also _not_ the way we decide whether the send is for NEAR or something else
    amount,
    memo,
    receiverId
  }
}) {

  if(type === TOKEN_TYPES.NEAR) {
    // **All** logic for handling NEAR goes here
  } else if(type === TOKEN_TYPES.NEP141) {
    // **All** logic for handling NEP141 tokens goes here
  } else {
    throw new TypeError(`Could not transfer unsupported token: ${type}`);
  }
}
```

This will be a more flexible structure in the long run, easier to extend later without touching logic for existing implementations of tokens that we already support, and much easier to read & debug.

src/actions/send.js
```diff
@@ -0,0 +1,40 @@
...
+        await dispatch(send.transfer.storageDeposit(contractName, receiverId));
+    }
+
+    const { transaction, status } = await dispatch(send.transfer[contractName ? 'TOKENS' : 'NEAR'](contractName, amount, memo, receiverId));
+
+    if (status?.SuccessValue) {
```
MaximusHaximus, line 13: It seems like this will silently fail if there is an error? What should be happening here if there is an error transferring or the `successValue` isn't set in the response? Feels like we're missing some actions and wrapping code here

src/actions/send.js
```diff
@@ -0,0 +1,40 @@
...
+                () => showAlert({ onlyError: true })
+            ],
+        },
+        IS_STORAGE_BALANCE_AVAILABLE: [
+            () => {},
+            () => showAlert({ onlyError: true })
+        ],
+        SET_TX_STATUS: null,
+    }
+});
```
MaximusHaximus, line 40: Seems like we're missing actions here?

src/actions/send.js
```diff
@@ -0,0 +1,40 @@
...
+            TOKENS: [
+                () => {},
+                () => showAlert({ onlyError: true })
+            ],
+            STORAGE_DEPOSIT: [
+                () => {},
+                () => showAlert({ onlyError: true })
+            ],
+        },
+        IS_STORAGE_BALANCE_AVAILABLE: [
```
MaximusHaximus, line 34: It's a bit weird that this is named like it is a boolean value.  Would this be more like e.g. `SET_STORAGE_BALANCE_IS_AVAILABLE`, or `LOAD_STORAGE_BALANCE_IS_AVAILABLE`?

src/actions/send.js
```diff
@@ -0,0 +1,40 @@
...
+
+import { showAlert } from '../utils/alerts';
+import { wallet } from '../utils/wallet';
+
+export const transfer = ({ contractName, amount, memo, receiverId, isStorageBalanceAvailable }) => async (dispatch, gesState) => {
+    if (contractName && !isStorageBalanceAvailable) {
+        await dispatch(send.transfer.storageDeposit(contractName, receiverId));
+    }
+
+    const { transaction, status } = await dispatch(send.transfer[contractName ? 'TOKENS' : 'NEAR'](contractName, amount, memo, receiverId));
```
MaximusHaximus, line 11: This function signature isn't actually valid for `wallet.sendMoney()`, which is what is selected when `contractName` is not set here.  See above comment around separating logic for different types of sends; trying to do too much in a single stream of logic leads to these types of bugs and mismatches in logic...

MaximusHaximus (changes requested):
Added comments in-line

marcinbodnar:
> Added comments in-line
> 
> Also, there are more (and different) actions described in the issue than are implemented here.
> 
> ```
> AMOUNT/SET
> RECEIVER/SET
> RECEIVER/SET_CHECKED
> RECEIVER/SET_NOTE
> TOKEN/SET
> STATUS/SET_TX_STATUS
> STATUS/NETWORK_FEES
> STATUS/DATA_TIME
> ```
> 
> Is the issue inaccurate, or is this PR missing implementation?

Right, these are the actions from the previous concept, now we will not store state for send flow, only for confirmed transactions, so none of these are needed anymore.

src/actions/send.js
marcinbodnar: Done.

src/actions/send.js
marcinbodnar: Done.

src/actions/send.js
marcinbodnar: If tx fails, an error will be thrown and the user will see Global Alert. After that, the user will see this tx in the Recent activity box, with `failed` indicator.

From the UI site, I think we should show a screen with information that the tx fails, and link to receipt like on success.

src/actions/send.js
marcinbodnar: Yes, the PR with `services/FungibleTokens.js` was under review.

src/actions/send.js
marcinbodnar: This action will call `isStorageBalanceAvailable` method from `FungibleTokens` class. 

The general idea that stays behind it, is that I would like to omit direct calling methods in components - ideal will be if the component should only be aware of redux actions and operates only with redux actions - also, all logic should be inside redux actions, so the component will do simple dispatch and that's it. 

Of course in this particular situation, `isStorageBalanceAvailable` action will not make any changes in the redux state, but that's because we are not storing any state for send flow. But still, I think that it should be done this way because it will make components much more clear about what they do, and move all logic into actions.

The additional benefit with making all functionality with redux actions would be that we could then wrap more logic into this, like for example mixpanel - at this moment mixpanel tracking is inside components, which is not ideal, I think it could be done automatically with redux.

Not sure if I explain my idea behind it properly :)

MaximusHaximus:
> > Added comments in-line
> > Also, there are more (and different) actions described in the issue than are implemented here.
> >...
> > Is the issue inaccurate, or is this PR missing implementation?
> 
> Right, these are the actions from the previous concept, now we will not store state for send flow, only for confirmed transactions, so none of these are needed anymore.

:+1: Please update the linked issue when changes happen to the acceptance criteria/design
https://github.com/near/near-wallet/issues/1865

marcinbodnar:
> > > Added comments in-line
> > > Also, there are more (and different) actions described in the issue than are implemented here.
> > > ...
> > > Is the issue inaccurate, or is this PR missing implementation?
> > 
> > 
> > Right, these are the actions from the previous concept, now we will not store state for send flow, only for confirmed transactions, so none of these are needed anymore.
> 
> 👍 Please update the linked issue when changes happen to the acceptance criteria/design
> #1865

@MaximusHaximus Done.

MaximusHaximus (changes requested):
Added comments in-line!  Making progress :+1: :)

src/actions/send.js
marcinbodnar: One important thing here :) all actions are handled by `status` reducer, we are storing the result of all actions and we can show Global Alert / Local Alert automatically if particular action fails or succeded.

> I also don't think that it makes it much more clear what the component is doing; it makes me think that the component is updating Redux state, but in actuality the component is just running a method on the Wallet object. 

We can react automatically if an error appears - we don't need to handle every method individually.

> The way this is now, in order for me to know whether or not a particular call has a Redux side-effect, I need to search the entire codebase for the SNAKE_CASE version of the action name; by finding no occurrences outside of the actions file of that string, I can assume that we don't use it to mutate redux. 

The component is dispatching thunk action, hi-level operations that this action does are described by its name - you just jump to this thunk action, and everything that is happening is in there (instead of component). So we use getState there to take all necessary states, and we are dispatching all actions that are needed. Actions are composed of objects, like `send.transfer.nep141` or `send.isStorageDepositAvailable` so it's clear what they are doing - you can also jump quickly to the method, to see details there.

The benefit is that we are moving all logic into thunk actions from the component. Also if like, in this case, we are composing FungibleTokens class, to be like low-level functions only, the thunk actions are the level above that uses these actions/methods to do all things.

> Wouldn't it be nice if I could immediately tell that something doesn't mutate redux at the call-site because it isn't a Redux action creator?

All actions are mutating the state, in this case `status`.

I think that the redux should be the center of our application and the only source of truth, the application should be rendered based on the redux state. We should store everything that is happening in the application. For example, if some request fails, we know it by checking redux.

I mean if we want classes to be low-level functions only, and we don't want to have complex logic in components directly (I think we don't because it's complicating stuff, the component needs to have all redux state to perform a sometimes complicated set of actions), I think the good idea is to handle everything in thunk actions - we have access to redux state there, and we can dispatch everything necessary - component only dispatches one action. If we don't want to use thunk actions, we should implement another level of logic, to handle everything that is happening between class methods and components.

I mean that is my idea, how to handle it. If you have other concepts here, no problem at all, I'm open to new ideas, just let me know how would you like to handle it :)

src/actions/send.js
```diff
@@ -0,0 +1,89 @@
...
+    } else if(type === TOKEN_TYPES.NEP141) {
+        if (isStorageBalanceAvailable === false) {
+            await dispatch(send.payStorageDeposit(contractName, receiverId));
+        }
+
+        const { transaction: { hash }, status } = await dispatch(send.transfer.nep141({
+            token: { 
+                contractName,
+                metadata: { 
+                    decimals: getState().tokens.tokens[contractName].decimals
```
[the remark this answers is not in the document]
marcinbodnar: Done.

src/actions/send.js
```diff
@@ -0,0 +1,89 @@
...
+                }
+            },
+            amount,
+            receiverId,
+            memo
+        }));
+
+        if (status?.SuccessValue === '') {
```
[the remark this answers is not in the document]
marcinbodnar: Done. 

> Would be a useful utility function that we should use any time we want to check if a transaction was successful

Sure, where would you like to add this function?

src/actions/send.js
```diff
@@ -0,0 +1,89 @@
...
+}) => async (dispatch, getState) => {
+    if (type === TOKEN_TYPES.NEAR) {
+        const { transaction, status } = await dispatch(send.transfer.near(receiverId, amount));
+
+        if (status?.SuccessValue) {
+            dispatch(send.setTxStatus(transaction.hash, 'success'));
+        }
+    } else if(type === TOKEN_TYPES.NEP141) {
+        if (isStorageBalanceAvailable === false) {
+            await dispatch(send.payStorageDeposit(contractName, receiverId));
```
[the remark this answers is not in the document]
marcinbodnar: Done.
If someone has already paid the storage deposit, the tx will be a success.

src/actions/send.js
```diff
@@ -0,0 +1,89 @@
...
+                    decimals: getState().tokens.tokens[contractName].decimals
+                }
+            },
+            amount,
+            receiverId,
+            memo
+        }));
+
+        if (status?.SuccessValue === '') {
+            dispatch(send.setTxStatus({
```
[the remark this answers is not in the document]
marcinbodnar: Done.

src/actions/send.js
```diff
@@ -0,0 +1,89 @@
...
+    } else {
+        throw new WalletError(`Could not transfer unsupported token: ${type}`, 'send.unsupportedToken', { type });
+    }
+};
+
+export const { send } = createActions({
+    SEND: {
+        TRANSFER: {
+            NEAR: [
+                wallet.sendMoney.bind(wallet),
```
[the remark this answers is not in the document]
marcinbodnar: Done.

src/actions/send.js
```diff
@@ -0,0 +1,89 @@
...
+                (receiverId, amount) => ({
+                    ...showAlert({ onlyError: true }),
+                    receiverId,
+                    amount
+                })
+            ],
+            NEP141: [
+                wallet.fungibleTokens.transfer?.bind(wallet.fungibleTokens),
```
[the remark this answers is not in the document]
marcinbodnar: Done.

src/actions/send.js
```diff
@@ -0,0 +1,89 @@
...
+                ({
+                    amount,
+                    receiverId
+                }) => ({
+                    ...showAlert({ onlyError: true }),
```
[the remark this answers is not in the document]
marcinbodnar: Reducer receives `payload` from `FungibleTokens.transfer` method, that will be the result of tx - this data is not there. We can return it from `FungibleTokens.transfer` method, but passing it as meta seems like a better idea because `transfer` method will return tx result only.

src/actions/send.js
```diff
@@ -0,0 +1,89 @@
...
+        amount, 
+        receiverId,
+        memo
+    }
+}) => async (dispatch, getState) => {
+    if (type === TOKEN_TYPES.NEAR) {
+        const { transaction, status } = await dispatch(send.transfer.near(receiverId, amount));
+
+        if (status?.SuccessValue) {
+            dispatch(send.setTxStatus(transaction.hash, 'success'));
```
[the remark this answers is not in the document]
marcinbodnar: Right, thanks. Done.

src/actions/send.js
```diff
@@ -0,0 +1,89 @@
...
+    type,
+    isStorageBalanceAvailable,
+    params: {
+        contractName, 
+        amount, 
+        receiverId,
+        memo
+    }
+}) => async (dispatch, getState) => {
+    if (type === TOKEN_TYPES.NEAR) {
```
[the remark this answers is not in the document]
marcinbodnar: Done.

src/actions/send.js
```diff
@@ -0,0 +1,89 @@
...
+                    receiverId
+                }) => ({
+                    ...showAlert({ onlyError: true }),
+                    amount,
+                    receiverId
+                })
+            ],
+        },
+        PAY_STORAGE_DEPOSIT: [
+            wallet.fungibleTokens.transferStorageDeposit?.bind(wallet.fungibleTokens),
```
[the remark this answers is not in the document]
marcinbodnar: Done.

src/actions/send.js
```diff
@@ -0,0 +1,89 @@
...
+            },
+            amount,
+            receiverId,
+            memo
+        }));
+
+        if (status?.SuccessValue === '') {
+            dispatch(send.setTxStatus({
+                hash,
+                newStatus: 'success'
```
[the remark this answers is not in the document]
marcinbodnar: Done.

src/actions/send.js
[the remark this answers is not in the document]
marcinbodnar: I guess I need to switch a bit to work with that kind of PR splitting when there is no other functionality available, and it's easy for this kind of bug :)

src/actions/send.js
MaximusHaximus: Thanks for the thoughtful response -- I'll write up an issue about this topic so we can address it directly rather as part of this PR discussion :)

MaximusHaximus:
Closing this PR as we are shipping without these redux bindings for now :)

Issue #1101, opened 2020-10-26
Support China-based Users


kcole16:
### Overview
Users in China are having difficulty accessing the wallet. All render.com services (frontend, contract-helper, RPC, Explorer) are inaccessible. Users over VPN are struggling to receive SMS messages.

### TODOs
- [ ] Switch main wallet to use Netlify (accessible in China)
- [ ] Have contract helper proxy endpoint for China (and use it automatically when needed)
- [ ] Make sure RPC also either accessible directly or has proxy endpoint for China
- [ ] Disable SMS (in future maybe do WeChat or whatever makes more sense in China)
- [ ] Automatically detect language and switch to Chinese

vgrichina:
@kcole16 seems like according to @ilblackdragon wallet is completely accessible in China now at main URL.

My understanding is that Render uses Google Cloud for everything except static websites, and so is similarly available (same as our RPC).

vgrichina:
We still need to improve localization / Disable SMS potentially

ilblackdragon:
Is there an issue tracking wallet selection process? There is a fork version of the wallet with Chinese localization already and but apps and other links send to current wallet.

ilblackdragon:
Just to update here, wallet does work in China but on some networks very slow still - home page sometimes loads for 20-40 seconds.

How about switching to a Cloudflare or some other CDN that is serving data from the edge?

think-in-universe:
In my test from Zhejiang, China, the wallet page is accessible from desktop (<5s), but quite slow (>30s) when opening in mobile browser (Chrome on iPhone). I'll collect more performance data tomorrow and share here. 

I agree with @ilblackdragon that we may try with Cloudflare if possible.

think-in-universe:
@stefanopepe the current issue of web wallet in China is that it sometimes takes users 10~60s to open the web wallet for users and make the UX horrible sometimes for not only wallet but also DApps built with NEAR.

Below are some experiments about how to improve web wallet's UX in China.



### Performacne Benchmarks


1. Here we mainly use **ping.cn** to evaluate the connection speed of the web wallet in China. e.g. https://www.ping.cn/http/wallet.near.org ;
2. User feedbacks: here we collect the feedback from users who experienced from really slow UX in most of the time;
3. Mxpenal data of the wallet if we have any performance data that can measure the loading speed of wallet, or bounce rate after visiting wallet.  


### Deployment Options

[...]

PR #1904, opened 2021-07-12
[WIP] [DONT MERGE] Experiment Hybrid Deployment with Cloudflare Pages

think-in-universe:
### Description

As mentioned in https://github.com/near/near-wallet/issues/1101#issuecomment-878351601, this is a temporary  fix of replacing local static assets with remote assets from Cloudflare Pages such as `wallet3.near.systems`.

> A quick and temporary solution is to replace the link to local JS / CSS assets with resources from another host, by making  `src/index.html` hardcoded with js/css links from one Cloudflare Pages deployment such as https://wallet3.near.systems/ , e.g.  `<script src="https://wallet3.near.systems/src.4965f7fa.js"></script>` , but this will definitely break our current CI process, and is only a temporary solution. (also we need to resolve the CORS issue of loading woff2, and update how images are imported).

### Status

This is even not an complete temporary fix because woff2 and images won't be loaded successfully due to CORS or image importing problem in JS.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,6 +9,7 @@
         />
         <meta name="theme-color" content="#000000" />
         <title>NEAR Wallet</title>
+        <link rel="stylesheet" href="https://wallet3.near.systems/src.a7d1707b.css">
     </head>
     <body>
         <noscript>You need to enable JavaScript to run this app.</noscript>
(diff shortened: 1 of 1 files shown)


think-in-universe:
This can be closed since we already fixed in #1906

PR #1906, opened 2021-07-13
Facilitate serving static resources via cloudflare

MaximusHaximus:
Created a new `runBundler` script that runs an environment-aware `parcel` bundler.
- Defined `ParcelBundler` class to manage bundle creation cases for various environments and deployment targets
- Moved `ci` script `sentry-send-release` to be co-located in `ci` directory with `runBundler` and `ParcelBundler` class
- Moved copying of wasm files into `ParcelBundler` logic
- Restructured `npm` commands used for building and starting dev server to all use `runBundler.js`


think-in-universe:
Thanks Daryl.

Here is my test result. The wallet is much slower comparing to the current deployment. 

**Testnet Preview**

image

**Testnet Production**

image


**Mainnet Preview**

image

**Mainnet Production**

image

think-in-universe:
My 2nd round of test. On the same machine, same network...

The preview versions outperform the production versions. 

**Testnet Preview**

image

**Testnet Production**

image


**Mainnet Preview**

image

**Mainnet Production**

image

PR #1907, opened 2021-07-13
FT transfer: Add remaining views and activate the flow

### LINKED ISSUE #1900: Hide 0 balances for tokens
Currently if user have touched contracts for any reason, they show up with 0 balances in the wallet without any reason.

Here is an example of a wallet that was whitelisting some tokens for another contract usage:

Screen Shot 2021-07-08 at 12 03 23 PM

Patrick1904:
Closes #1900

Main component additions:
1. `ReceiveContainer.js` - Acting as a Redux connector and hosting the new receive money page
2. `SendContainerWrapper.js` - Acting as a Redux connector and wrapper for the send flow
3. `SendContainerV2.js` - Handling the send flow logic, state, steps, and hosting all send-related components
4. `AvailableBalance.js` - Representing the balance available to send
5. `TabSelector.js` - Provides on-page links to navigate between the `/send-money` and `/receive-money` pages
6. `EnterAmount.js` - Component for the 'enter amount' step of the send flow
7. `SelectToken.js` - Component for the 'select token' step of the send flow
8. `EnterReceiver.js` - Component for the 'enter receiver' step of the send flow
9. `Review.js` - Component for the 'review' step of the send flow
10. `Success.js` - V0 of success page until we have necessary info to show full receipt

Modifications:

Minor adjustments to the `Tokens.js` and `TokenBox.js` components used on the main Wallet dashboard (`Wallet.js`) in order to re-use them in the `SelectToken.js` component.

Coming in next PR:
1. Receipt success page on a separate route that uses hash in URL to fetch necessary TX data
2. Will update the main button (currently blue) to dark gray (as per design) separately because there's a lot of cleanup needed and don't want to mix it in this PR.
3. Minor polishing style fixes
4. Send/receive button in the main desktop navigation

How to test:
https://deploy-preview-1907--near-wallet-staging.netlify.app/
On either `/send-money` or `/receive-money`, apply `/new` to the end of the `URL`, e.g. `/send-money/new`


src/components/send/SendContainerV2.js
```diff
@@ -0,0 +1,209 @@
...
+
+    const handleContinueToReview = async () => {
+        setView('review');
+
+        if (selectedToken.symbol === 'NEAR') {
+            setEstimatedTotalFees(await FTMethods.getEstimatedTotalFees());
+            setEstimatedTotalInNear(await FTMethods.getEstimatedTotalNearAmount(parsedAmount));
+        } else {
+            // FIX: Currently returns 'Cannot read property 'viewFunction' of undefined'
+            //setEstimatedTotalFees(await FTMethods.getEstimatedTotalFees(selectedToken.contractName, receiverId));
```
Patrick1904, line 142: @marcinbodnar do you have any idea why this view function call from FungibleTokens.js fails?

marcinbodnar: @Patrick1904 `FungibleTokens` methods are part of the wallet class:
https://github.com/near/near-wallet/blob/master/src/utils/wallet.js#L134
so if you want to use methods you use `wallet.FungibleTokens.xxx`.

Also, I'm not sure if we want to have these methods that you added to FungibleTokens class - Daryl was planning to have it like low-level functions. cc @MaximusHaximus 

There are also reducers and actions for these functionalities:
https://github.com/near/near-wallet/pull/1903
https://github.com/near/near-wallet/pull/1914
We can extend it with other actions if needed. 
I'm just not sure if we want to store partial data in the redux state, like estimated fees - if so these methods should be called via redux actions. cc @MaximusHaximus

Patrick1904: > @Patrick1904 FungibleTokens methods are part of the wallet class:
> https://github.com/near/near-wallet/blob/master/src/utils/wallet.js#L134
> so if you want to use methods you use wallet.FungibleTokens.xxx.

Awesome thanks I missed that.

> Also, I'm not sure if we want to have these methods that you added to FungibleTokens class - Daryl was planning to have it like low-level functions. cc @MaximusHaximus

Yep we had a discussion about this on Friday and decided it would be where fungible token-related methods would live.

> There are also reducers and actions for these functionalities:
> #1903
> #1914
> We can extend it with other actions if needed.
> I'm just not sure if we want to store partial data in the redux state, like estimated fees - if so these methods should be called via redux actions. cc @MaximusHaximus

Right, I think we can skip Redux all together, because estimated fees just need to be calculated every time the user gets to the 'review' view. So no need to put it in redux. Same with checking storage deposit, sending tokens, etc. They're not mutating redux state, so we might as well call them directly without redux. I just have to add error handling in the 'review' component since it's not being handled by redux.

Patrick1904:
@picturepan2 when you have a chance, can you please add translations for the new send/receive flow?

picturepan2:
> @picturepan2 when you have a chance, can you please add translations for the new send/receive flow?

Sure.

src/services/FungibleTokens.js
```diff
@@ -36,30 +46,77 @@ export default class FungibleTokens {
...
+        ? parseNearAmount(amount) 
+        : parseTokenAmount(amount, decimals);
+
+        return parsedTokenAmount;
+    }
+
+    getFormattedTokenAmount(amount, symbol, decimals) {
+        const formattedTokenAmount = symbol === 'NEAR'
+        ? formatNearAmount(amount, 5) 
+        : removeTrailingZeros(formatTokenAmount(amount, decimals, 5));
```
MaximusHaximus, line 60: Is it reasonable to use precision of 5 for all non-NEAR tokens?

src/services/FungibleTokens.js
```diff
@@ -2,23 +2,33 @@ import BN from 'bn.js';
...
+            formatNearAmount
         }
     }
 } = nearApiJs;
 
 // account creation costs 0.00125 NEAR for storage, 0.00000000003 NEAR for gas
 // https://docs.near.org/docs/api/naj-cookbook#wrap-and-unwrap-near
 const FT_MINIMUM_STORAGE_BALANCE = parseNearAmount('0.00125');
+// FT_MINIMUM_STORAGE_BALANCE: nUSDC, nUSDT require minimum 0.0125 NEAR. Came to this conclusion using trial and error.
+const FT_MINIMUM_STORAGE_BALANCE_LARGE = parseNearAmount('0.0125');
```
MaximusHaximus, line 31: <3

src/services/FungibleTokens.js
```diff
@@ -36,30 +46,77 @@ export default class FungibleTokens {
...
+        return formattedTokenAmount;
+    }
+
+    async getEstimatedTotalFees(contractName, accountId) {
+        const storageBalanceAvailable = await this.isStorageBalanceAvailable(contractName, accountId);
+        if (contractName && accountId && !storageBalanceAvailable) {
+            return new BN(FT_TRANSFER_GAS).add(new BN(FT_MINIMUM_STORAGE_BALANCE)).add(new BN(FT_STORAGE_DEPOSIT_GAS)).toString();
+        } else {
+            return FT_TRANSFER_GAS;
+        }
```
MaximusHaximus, line 71: :nit: chaining style + line length make this hard to read
```suggestion
        if (contractName && accountId && !storageBalanceAvailable) {
            return new BN(FT_TRANSFER_GAS)
            .add(new BN(FT_MINIMUM_STORAGE_BALANCE))
            .add(new BN(FT_STORAGE_DEPOSIT_GAS))
            .toString();
        } else {
            return FT_TRANSFER_GAS;
        }
```

src/services/FungibleTokens.js
```diff
@@ -36,30 +46,77 @@ export default class FungibleTokens {
...
-        ]);
+    async transfer({ contractName, parsedAmount, receiverId, memo }) {
+        if (contractName) {
+            const storageAvailable = await this.isStorageBalanceAvailable(contractName, receiverId);
+
+            if (!storageAvailable) {
+                try {
+                    await this.transferStorageDeposit(contractName, receiverId, FT_MINIMUM_STORAGE_BALANCE);
+                } catch(e) {
+                    if (e.message.includes('attached deposit is less than the mimimum storage balance')) {
```
MaximusHaximus, line 95: 1. Should we re-throw errors _other_ than `attached deposit is less than the mimimum storage balance` here?
2. Type in this magic string (`mimimum` vs. `minimum`)

MaximusHaximus (changes requested):
Whew -- this is one monster beast of a PR!
Added some comments & questions throughout.  Lets sync up in the morning, maybe we can split up some of the changes since they are spread out a bit across the files.

picturepan2:
> @picturepan2 when you have a chance, can you please add translations for the new send/receive flow?

Translated.

src/components/send/SendContainerV2.js
```diff
@@ -0,0 +1,209 @@
...
+    const [parsedAmount, setParsedAmount] = useState('');
+    const [receiverId, setReceiverId] = useState('');
+    const [view, setView] = useState('enterAmount');
+    const [maxAmount, setMaxAmount] = useState(null);
+    const [sendingToken, setSendingToken] = useState(false);
+    const [selectedToken, setSelectedToken] = useState(fungibleTokens[0]);
+
+    useEffect(() => {
+        selectedToken.symbol === 'NEAR' && setSelectedToken(fungibleTokens[0]);
+    }, [availableNearToSend]);
```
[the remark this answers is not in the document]
Patrick1904: Since the calculated `availableNearToSend` isn't present when the component first mounts, we still need to listen for change to `availableNearToSend` and then `setSelectedToken(fungibleTokens[0])` once it's finished calculating the balance.

src/components/send/components/TabSelector.js
```diff
@@ -0,0 +1,85 @@
...
+`;
+
+const TabSelector = () => {
+    const dispatch = useDispatch();
+    const location = useLocation();
+    const pathname = location.pathname;
+    const sendMoneyRoute = '/send-money/new';
+    const receiveMoneyRoute = '/receive-money/new';
+
+    //TODO: Replace tab selector in Wallet.js with this component
```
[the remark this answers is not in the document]
Patrick1904: Yes exactly. I will look into this post this first cut.

src/components/wallet/Tokens.js
```diff
@@ -23,11 +23,11 @@ const StyledContainer = styled.div`
     }
 `;
 
-const Tokens = ({ tokens }) => {
+const Tokens = ({ tokens, onClick }) => {
     return (
         <StyledContainer>
             {tokens.map((token, i) => (
-                <TokenBox key={i} token={token}/>
+                <TokenBox key={i} token={token} onClick={onClick}/>
```
MaximusHaximus: I'm fixing this now, but just a reminder @Patrick1904 @marcinbodnar that using an index in an array for a `key` is an anti-pattern; if the order of the values in `tokens` changes, these child component's won't re-render because their keys haven't changed...

So imagine that next week we start showing tokens sorted by your balance from highest->lowest rather than a pure alphabetical sort; when a token's balance changes the list won't re-render to show the new order.

Or today, with the list sorted alphabetically, we run a background query to see if a user has new tokens and update the `tokens` store accordingly...you'll probably see the last token in the list rendered twice instead of seeing the new token inserted alphabetically in the right spot.

Keys should be unique to the item in the list -- in this case, I think `token.contractName` would be the right key since everything else could be a duplicated value :).

A good reference w/ examples of the pitfalls here: https://robinpokorny.medium.com/index-as-a-key-is-an-anti-pattern-e0349aece318

MaximusHaximus (approved):
:+1: I think this sucker is ready for testnet & staging :) @Patrick1904 what say you my friend?

PR #1911, opened 2021-07-13
Update Landing Page title and description for Chinese translations

think-in-universe:
### Description

Some change to the landing page description per the feedback from Amos and other users. 

Let's try to incorporate this change into this week's release if possible. Thanks!

**Simplified Chinese**

image

**Traditional Chinese**

image


MaximusHaximus (approved):
:+1:

Issue #1919, opened 2021-07-14
Claim Linkdrop is Extremely Slow (>60s) in Mainnet Wallet


think-in-universe:
**Problem**

Claim linkdrop is extremely slow on mainnet wallet at the moment: when you click **"Claim with Existing Account"** button, the `contract.claim()` function will be called, which costs more than 60s. During the process, all the access keys of `near` accounts will be fetched by 3 times.

https://github.com/near/near-wallet/blob/b58683fd7032b26e38a28a6385dc6f0ebf47068f/src/utils/wallet.js#L446

Via profiling, we fiind that the [`linkdrop.claim()`](https://github.com/near/near-linkdrop/blob/f24f2608e1558db773f2408a28849d330abb3881/src/lib.rs#L72-L88) contract function call will query `access_key/near` with JSON RPC by 3 times, and the quries cost 26s, 11s and 12s respectively, which leads to the slow response time of `claim` linkdrop. 

```json
{
    "method": "query",
    "params": [
        "access_key/near",
        ""
    ],
    "id": 155,
    "jsonrpc": "2.0"
}
```

The reason why querying `access_key/near` is extremely slow is because now we have **12K+ access keys** added to the `near` account (via [`linkdrop.send()`](https://github.com/near/near-linkdrop/blob/f24f2608e1558db773f2408a28849d330abb3881/src/lib.rs#L52-L69)), and the data size is quite big now, which takes 10~30 seconds to load all the access keys each time. 

This is a blocker for one important NFT project in our ecosystem, so I'm trying to fix this ASAP. 

**Expected Behavior**

The claim linkdroip step should take less than 10s ideally. 

**Steps to reproduce**

1. Claim any linkdrop in mainnet wallet
2. It takes around 60s to claim one linkdrop

**Solutions**

1. Reduce the times of calling `account.getAccessKeys()` (here the account is `near`), or make it optional to call the `account.getAccessKeys()` function.
2. Redeploy **linkdrop contract** to another account, and then create new account as sub-accounts of that the **linkdrop contract**
3. Any other ideas?

think-in-universe:
Here is the HAR file that might help us understand the profiling details:

[wallet.near.org.har.zip](https://github.com/near/near-wallet/files/6815509/wallet.near.org.har.zip)

think-in-universe:
### Investigation

[...]

PR #1920, opened 2021-07-14
Remove unnecessary queries of account access keys via RPC

think-in-universe:
### Problem

Fixes: #1919 

Now claim linkdrop is extremely slow (>60s) in Mainnet Wallet.

The root cause is that for signing transaction in wallet, now we'll first `getLedgerKey()` which will call `getAccessKeys()` and then filter out the ledger key.  This is OK for most accounts, but for acounts with thousands of keys such as `near` or `testtnet`, it impacts the performance significantly. 

https://github.com/near/near-wallet/blob/b58683fd7032b26e38a28a6385dc6f0ebf47068f/src/utils/wallet.js#L99-L120 

### Solution 

1. We don't need `getLedgerKey()` for some accounts such as the root account: `testnet`, `near`, etc.
2. We should make `getLedgerKey()` more efficient if possible, by removing unnecessary query of all access keys, or add more caches to avoid duplicate requests. 

Here, due to the urgency of the fixing the issue, which impacts the UX of claiming linkdrops and NFTs of one NFT project, in this PR, one quick workaround is proposed:

> just return `null` for `near` or `testnet` account in `getLedgerKey()` 

Please feel free to propose better solutions, but let's ensure we could fix the issue in mainnet no later than July 16th, the sooner the better to leave some time for testing.

### DIFF
--- a/src/utils/wallet.js
+++ b/src/utils/wallet.js
@@ -144,6 +144,11 @@ class Wallet {
         accountId = accountId || this.accountId;
         // TODO: Refactor so that every account just stores a flag if it's on Ledger?
 
+        // special handing for fixing issue #1919
+        if (accountId === ACCOUNT_ID_SUFFIX) {
+            return null;
+        }
+
         const accessKeys = await this.getAccessKeys(accountId);
         if (accessKeys) {
             const localKey = await this.getLocalAccessKey(accountId, accessKeys);


MaximusHaximus (approved):
:+1: This looks safe (though we should really figure out why we're even calling this method for an account that we don't control for a longer term solution) :)

Patrick1904 (approved):

think-in-universe:
@MaximusHaximus yes. this is only a temporary fix. we need to figure out a long term fix later.

PR #1921, opened 2021-07-14
i18n: Fix claim linkdrop Chinese translation

think-in-universe:
### Description

The Chinese translation of claiming linkdrop with existing account is inaccurate. 

Fix this with the more accurate translation. 

image


think-in-universe:
@stefanopepe @MaximusHaximus 

this is a minor translation issue, but if we could also include this into the recent release, it would be perfect!

MaximusHaximus (approved):
:shipit:

PR #1927, opened 2021-07-16
Add Cheddar Token

htafolla:
Request to add the Cheddar to token.js or the .env file.


MaximusHaximus:
@htafolla Thanks for the submission, we really appreciate it!  We've actually recently changed the wallet behaviour to no longer show fungible tokens that you have no balance for -- as a result, the "whitelist" doesn't apply anymore and we intend to remove it from the codebase shortly; it's a piece of deprecated code right now that is only still here because we didn't want to break the ability to toggle off the new fungible tokens changes without reverting the entire feature :). As such, I'm going to close this PR -- please do feel free to make submissions, however!

Issue #1928, opened 2021-07-17
Clicking Approve quickly on signing transaction screen leads to failure


ilblackdragon:
**Problem**
After opening a sign transaction screen and clicking approve right away - transaction singing fails.

This is happening because account Id seems like been taken from the selector, which takes time to load. And if clicked quickly, selector still has "undefined" selected. Which leads to transaction been tried to signed with undefined account.

**Expected Behavior**
There is should be no need to wait and transaction succeeds after clicking Approve.
App is sending transaction already with account filled in, there should be no need to wait for anything on the UI. Maximum is to check that this account Id exists in the wallet which is a simple check and can be done even after clicking Approve (given in 99% it will exist).

**Steps to reproduce**

1. Open Paras and login
2. Click on any item and click Buy
3. Right away click "Approve"

Observe that transaction failed because account is undefined
<img width="898" alt="Screen Shot 2021-07-17 at 3 49 54 AM" src="https://user-images.githubusercontent.com/175486/126034576-5468aa38-7496-4369-b2b1-6e6fae9a037b.png">

stefanopepe:
@Patrick1904 what is the async logic that requires this additional loading time?

If I remember correctly, this delay is necessary to prevent that users can click "send" if the available balance is insufficient, avoiding unnecessary load on the contract helper. Probably having a cached number helps, even more if we expect the user to confirm multiple transactions within a minute (therefore, the cached balance is current)

stefanopepe:
@MaximusHaximus I'll let you decide when we want to include this specific issue in the sprint backlog, it's a P2 unless we find it's blocking for slower/higher latency connections (Asia)

vgrichina:
@stefanopepe it is on the way of user giving money to app, IMO should be at least P1 if not P0.

MaximusHaximus:
@marcinbodnar Could you take a look at this? Would like to get it into this weeks release :)

marcinbodnar:
@MaximusHaximus Done #1931 - waiting for the balance to be load was removed

marcinbodnar:
@MaximusHaximus When it comes to signing tx before the balance is loaded (regarding #1846 ), we can also speed up the signing process by not waiting for the account state to be loaded. 

Made that change in #1931 , so now users can super fast allow sending tx - no delay of any kind and working with no problems..

marcinbodnar:
[...]

PR #1931, opened 2021-07-20
Fix disabling Allow button on Sign flow (#1928)

marcinbodnar:


MaximusHaximus (approved):
:+1:

We cannot get `redux-persist` added soon enough IMO; this would be totally resolved with that in place.

PR #1936, opened 2021-07-22
Activate new FT transfer flow

Patrick1904:
Replaces current send/receive flows with the new one: https://github.com/near/near-wallet/issues/1714


MaximusHaximus (approved):
:+1: :picard_engage:

Issue #1933, opened 2021-07-21
Linkdrop Pages cannot be open in certain browsers on Android phones used by Chinese users


RubyJiao:
**Problem**

This is a China-specific issue. 

For the first-time wallet user, when opening the linkdrop links in certain browsers (QQ Browser, UC Browser, Huawei browser in phones) in Android phone (Huawei, Xiaomi, Meizu, Oppo and so on), the screen is blank and nothing is rendered. The issue can be reproduced in both WiFi and 4G did not work. 

The linkdrop pages can be opened successfully in Chrome browser installed in those Android phones. 

**Expected Behavior**

The linkdrop pages should be open successfully in QQ / UC / Huawei browsers in Android phones with brand Huawei, Xiaomi, Meizu, Oppo, etc. in China.

**Steps to reproduce**

1. Open the linkdrop page in one of the browsers: (QQ Browser, UC Browser, system browser in phones) in Android phone (Huawei, Xiaomi, Meizu, Oppo and so on)
2. The page keeps blank and nothing happens

think-in-universe:
This issue is only reproducible on certain Android phones which are popular among Chinese users. 

I have found some devices to reproduce the issue, and will try to investigate and fix the issue.

think-in-universe:
Debugged with Chrome Inspect for Android Browsers, and the reason the page is blank is because the JS / CSS files are not returned (0 bytes are returned).

![image](https://user-images.githubusercontent.com/46699230/126588963-7bf581a0-a7a1-42e2-8b4d-27a92fc6296e.png)

think-in-universe:
By inspecting with cURL for request to the JS file under difference circumstances, below are the testing result.

(1) Open linkdrop page with Huawei Browser (11.1.2.301) on Huawei Android Phone (Android 10; OCE-AN10 Build/HUAWEIOCE-AN10 / HarmoneyOS) 

- Result: Failure
- Expect: JS File returnd
- Actual: HTML page returned

```bash
curl 'https://content.near-wallet.workers.dev/ntl/mainnet/src.fafd6029.js' \
  -H 'authority: content.near-wallet.workers.dev' \
  -H 'pragma: no-cache' \
  -H 'cache-control: no-cache' \
[...]

PR #1937, opened 2021-07-22
Add `strict-origin-when-cross-origin` meta to fix the `referer` issue for Huawei browsers, etc.

think-in-universe:
Fixes: #1933

Loading JS and CSS from Cloudflare workers failed when using Huawei / QQ Browsers, etc. 

Force `strict-origin-when-cross-origin` in html to change the referrer url to origin only to resolve the issue.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,7 @@
             content="width=device-width, initial-scale=1, shrink-to-fit=no"
         />
         <meta name="theme-color" content="#000000" />
+        <meta name="referrer" content="strict-origin-when-cross-origin" />
         <title>NEAR Wallet</title>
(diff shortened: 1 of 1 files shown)


think-in-universe:
We can test with linkdrop https://near-wallet-pr-1937.onrender.com/create/testnet/2EG77FUYQuyGLJSkUssuN6YPpH56h4fG51BuYfaWqSp3HGNuLk16ghagPXRVEQV8e6kNqAY11nfZv7wAEtXus51f

think-in-universe:
Also tested the basic functionalies of wallet and now it seems everything is working well with this change.

PR #1938, opened 2021-07-22
FT transfer follow-up fixes

Patrick1904:
Changes:
1. Add send/receive buttons to desktop nav
2. Do not reset transfer amount when switching token
3. Add Mixpanel tracking events that are present in current/legacy send flow

Screen Shot 2021-07-22 at 4 01 32 PM

### DIFF
--- a/src/components/send/SendContainerWrapper.js
+++ b/src/components/send/SendContainerWrapper.js
@@ -28,7 +28,7 @@ const fungibleTokensIncludingNEAR = (tokens, availableNearToSend) => {
     ];
 };
 
-export function SendContainerWrapper({ match, location }) {
+export function SendContainerWrapper({ match }) {
     const accountIdFromUrl = match.params.accountId || '';
     const dispatch = useDispatch();
     const { accountId, balance } = useSelector(({ account }) => account);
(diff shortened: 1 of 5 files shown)


corwinharrell (approved):

src/components/send/SendContainerWrapper.js
```diff
@@ -45,6 +46,14 @@ export function SendContainerWrapper({ match }) {
     const [transactionHash, setTransactionHash] = useState(null);
     const [fungibleTokens, setFungibleTokens] = useState(() => fungibleTokensIncludingNEAR(tokens, availableNearToSend));
 
+    useEffect(() => {
+        if (activeView === VIEWS.SUCCESS) {
+            let id = Mixpanel.get_distinct_id();
+            Mixpanel.identify(id);
+            Mixpanel.people.set({last_send_token: new Date().toString()});
+        }
+    }, [location.key]);
```
MaximusHaximus, line 55: Should this include `activeView`? It's a bit confusing that we don't even use `location.key` inside the effect but it's the only dependency and switching to the `success` view doesn't actually change to a new route I don't think, since it's part of the same `send-money` route as the rest of the flow?

Even better, can we just dispatch the event _from the `handleSendToken()` method defined on L89-L110? I assume that we want to fire this any time there is a successful send based on the check of `activeView`, just not sure why we're doing it in a useEffect() and why `location.key` is the dependency here...

src/components/send/SendContainerWrapper.js
```diff
@@ -98,6 +107,7 @@ export function SendContainerWrapper({ match }) {
                     return;
                 }
 
+                Mixpanel.track("SEND token");
```
MaximusHaximus, line 110: Shouldn't we be using the same `Mixpanel.withTracking()` wrapper here around the actual attempt to perform the transfer, just as we do in the existing `Send` screen?

Without the wrapper, we will not report to Mixpanel cases where the send _fails_; we just won't send any tracking event at all, which represents a breaking change in our tracking reporting if we rely on the failures being reported for any of our mixpanel insights.

MaximusHaximus (changes requested):
@Patrick1904 Added questions in-line!

@TiffanyGYJ @icerove Would it be useful for us to add any metadata to the `SEND token` event such as the token type that is being sent?  As it stands currently you won't be able to tell if a particular event was for sending NEAR tokens or a fungible token other than NEAR tokens.

MaximusHaximus (approved):
:+1: Nice! :) MixPanel FTW

PR #1941, opened 2021-07-23
FT transfer fixes

Patrick1904:
Changes:
1. FT transfer, switch account: reset everything
2. Review page: only disable buttons if `sendingToken === true`


MaximusHaximus (approved):
:+1: good stuff

PR #1956, opened 2021-07-23
Ensure that current account state is used when calling fungibleTokensService.transfer()

MaximusHaximus:
1. Refactored fungibleTokensService to no longer have a `this.account` property that we needed to keep in sync with the rest of the wallet UI/systems
- removed fungibleTokensService.signAndSendTransaction() function since it was superfluous post-refactor
2. Moved some methods to be static methods on the FungibleTokens class since they are purely viewFunction calls
- Using a statically created account object for `viewFunction` calls since they aren't actually signed
3. Removed spammy and not useful console.logs inside the `nft` actions redux thunk
4. Removed fungibleTokensService from the wallet object; it gained nothing from being present there as it is now a stateless singleton that can be imported and used wherever we want

### DIFF
--- a/src/utils/wallet.js
+++ b/src/utils/wallet.js
@@ -163,7 +163,7 @@ class Wallet {
         return ACCOUNT_ID_REGEX.test(accountId);
     }
 
-    async sendMoney({ account, receiverId, amount }) {
+    async sendMoney(receiverId, amount) {
         return (await this.getAccount(this.accountId)).sendMoney(receiverId, amount);
     }
 
--- a/src/utils/wallet.js
+++ b/src/utils/wallet.js
@@ -15,7 +15,6 @@ import {
     finishAccountSetup,
     makeAccountActive
 } from '../actions/account';
-import FungibleTokens from '../services/FungibleTokens';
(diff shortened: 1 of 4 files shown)


src/utils/wallet.js
```diff
@@ -168,7 +163,7 @@ class Wallet {
         return ACCOUNT_ID_REGEX.test(accountId);
     }
 
-    async sendMoney(receiverId, amount) {
+    async sendMoney({ account, receiverId, amount }) {
```
marcinbodnar, line 166: Not a blocked. You added `account` prop but it's not used. But I guess it can be removed since you are using, `account.sendMoney` in `FungibleTokens` class. We can also remove the action from `actions/account.js`

marcinbodnar (approved):
LGTM

src/utils/wallet.js
MaximusHaximus: Nice catch, I didn't mean to change this function signature at all, and this change probably broke the existing `send` container flow by changing it's signature which isn't intended yet (just in case we need to toggle back to the original send container without reverting all of this code). Fixed!

PR #1958, opened 2021-07-23
FT transfer: close Ledger modal after finishing sending token

marcinbodnar:


Patrick1904 (approved):
Sweet! Tested sending NEAR and non-NEAR token (approving and declining TX) - works as expected 👍

MaximusHaximus (approved):
:+1: legit!

PR #1960, opened 2021-07-24
Fix `zh` translation of `notSupportedPhone`

think-in-universe:
### Description

We found one translation issue of the `notSupportedPhone` message, when user tries to import account with SMS. 

Here we'd like to fix the issue.


MaximusHaximus (approved):
Thanks @think-in-universe !

Issue #1887, opened 2021-07-06
Navigation bar disappears


ihaid:
**Problem**
Navigation bar disappears when switching account 

**Steps to reproduce**

1. Log in to 2 accounts
2. Go to "Import Account"
3. Click "Sign in with Ledger"
4. Switch between your accounts
![image](https://user-images.githubusercontent.com/38665104/124646416-4859ab80-de9d-11eb-913a-d77619aad9b9.png)
![image](https://user-images.githubusercontent.com/38665104/124646448-53acd700-de9d-11eb-8a7d-3487c47fed89.png)

stefanopepe:
I wasn't able to reproduce the issue. Can you submit a video with the steps you are following? I'm not sure I'm doing the steps correctly

ihaid:
https://user-images.githubusercontent.com/38665104/125656150-3ff19605-634a-4055-904a-449df81dd408.mov

@stefanopepe

stefanopepe:
Interesting.

assigning to @MaximusHaximus & cc @Patrick1904 to investigate the glitch

Patrick1904:
@marcinbodnar thinking this might be related to `flowLimitation` - do you know if there's a quick fix?

marcinbodnar:
@Patrick1904 It's probably a quick fix, but nothing comes to my mind, need to investigate it.

marcinbodnar:
@Patrick1904 Fix ready here #1968

marcinbodnar:
@MaximusHaximus we should also check other situations like that, that might appear. Added an issue here #1970

Also regarding routes, some time ago I had an idea to add all of our routes in one object #678, additionally, we can use it in `Routes.js` to map routes object, instead of writing all routes inline. The idea appears when we had a bit simpler routes but I think it still can be implemented.

PR #1968, opened 2021-07-27
Fix checking sign flow url in handleFlowLimitation action (#1887)

marcinbodnar:


src/actions/flowLimitation.js
```diff
@@ -21,7 +21,7 @@ export const handleFlowLimitation = () => (dispatch, getState) => {
             accountBalance: true
         }));
     } 
-    else if (redirectUrl.includes(WALLET_SIGN_URL)) {
+    else if (redirectUrl === `/${WALLET_SIGN_URL}`) {
```
Patrick1904, line 24: Tested and works well. However it seems that if I exit the `/sign` flow before clicking 'allow', then navigate to `/sign-in-ledger` or `/recover-seed-phrase` or `/recover-account` and refresh the page - the `url` object gets re-populated and the nav links disappear.

Also, could `redirectUrl.includes(WALLET_LOGIN_URL)` on `L15` have the same issue with the `cli-login-success` route? Also I wonder why login URLs are always `/login/` with the extra `/`, but sign URLs are `/sign`?

ihaid (approved):

PR #1973, opened 2021-07-28
NEAR token USD conversion support

Patrick1904:
Main changes:
1. Add display of USD value around or next to all NEAR amounts. Exceptions are mentioned [here](https://github.com/near/near-wallet/pull/1973#pullrequestreview-722364295).
2. Added some tests for balance-related helper functions


MaximusHaximus (changes requested):
Comments added in-line :). This is looking really great, good work!

src/components/common/Balance.js
```diff
@@ -1,23 +1,18 @@
-import { BN } from 'bn.js';
-import { utils } from 'near-api-js';
 import React from 'react';
```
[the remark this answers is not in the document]
Patrick1904: This component displays both NEAR and fiat amount (fiat shows below NEAR amount `L84` using `NEARBalanceInUSDWrapper`). We also have the util function `getNearAndFiatValue` for when we need to inject the amount as a string into a translate prop:)

src/components/common/Balance.js
```diff
@@ -52,51 +47,49 @@ const CustomDiv = styled(List)`
...
 `;
 
-const FRAC_DIGITS = 5;
-const YOCTO_NEAR_THRESHOLD = new BN('10', 10).pow(new BN(utils.format.NEAR_NOMINATION_EXP - FRAC_DIGITS + 1, 10));
+const Balance = ({
+    amount,
+    symbol = true,
+    className,
+    includeBalanceinFiat = true,
+    showAsSubtract
```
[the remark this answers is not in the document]
Patrick1904: Yep the amount is formatted so we need to pass in the positive amount (i.e. `WALLET_APP_MIN_AMOUNT`). Added a comment here: https://github.com/near/near-wallet/pull/1973/commits/a5ebd61f6062d42c97c17e77014ea5653338a244

src/components/common/Balance.js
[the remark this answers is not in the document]
Patrick1904: Actually I misunderstood this - realized that `NEARBalanceInUSDWrapper` loses it's value if simply imported into `Balance.js`. Fixing now 👍

src/components/accounts/CreateAccount.js
```diff
@@ -180,18 +181,23 @@ class CreateAccount extends Component {
             whereToBuy
         } = this.state;
 
-        const { localAlert, mainLoader, checkNewAccount, resetAccount, clearLocalAlert, fundingContract, fundingKey, isMobile } = this.props;
+        const { localAlert, mainLoader, checkNewAccount, resetAccount, clearLocalAlert, fundingContract, fundingKey, isMobile, nearTokenFiatValueUSD } = this.props;
```
MaximusHaximus, line 184: :nit: would like this to be one-prop-per-line at this point :P

src/components/accounts/CreateAccount.js
```diff
@@ -290,15 +296,21 @@ const mapDispatchToProps = {
...
+
+    return {
+        ...account,
+        localAlert: status.localAlert,
+        mainLoader: status.mainLoader,
+        isMobile: status.isMobile,
+        fundingContract: match.params.fundingContract,
+        fundingKey: match.params.fundingKey,
+        fundingAccountId: match.params.fundingAccountId,
+        nearTokenFiatValueUSD: selectNearTokenFiatValueUSD(state)
```
MaximusHaximus, line 311: Hey look! A SELECTOR!!! ❇️ 💖 🤯  :D I think this might be the _only_ selector in a `mapStateToProps()` in the entire codebase right now :D. It feels lonely.

src/components/common/balance/Balance.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { useSelector } from 'react-redux';
+
+import { selectNearTokenFiatValueUSD } from '../../../slices/tokenFiatValues';
+import BalanceDisplay from './BalanceDisplay';
```
MaximusHaximus, line 5: 💖  Why not, I like it ;)

src/components/common/balance/BalanceDisplay.js
```diff
@@ -0,0 +1,122 @@
...
+    nearTokenFiatValueUSD,
+    showAlmostEqualSignUSD,
+    showSignUSD,
+    showSymbolUSD
+}) => {
+
+    const amountoShow = amount && formatNearAmount(amount);
+    const NEARSymbol = 'NEAR';
+
+    const handleShowInYocto = (amount) => {
```
MaximusHaximus, line 89: 💖

src/components/common/balance/BalanceDisplayUSD.js
```diff
@@ -0,0 +1,35 @@
...
+const BalanceDisplayUSD = ({
+    amount,
+    showAlmostEqualSignUSD = true,
+    showSignUSD = false,
+    showSymbolUSD = true,
+    nearTokenFiatValueUSD
+}) => {
+
+    const roundedBalanceInUSD = getRoundedBalanceInFiat(amount, nearTokenFiatValueUSD);
+    const amountPrefix = roundedBalanceInUSD !== '< 0.01' ? '≈ ' : '';
```
MaximusHaximus, line 14: Feels weird that we have a prop named `showAlmostEqualSignUSD`, but then we have `amountPrefix` be replaced with an empty string even if `showAlmostEqualSignUSD` is true. 🤷

src/components/common/balance/BalanceDisplayUSD.js
```diff
@@ -0,0 +1,35 @@
...
+    const USDSymbol = 'USD';
+
+    if (roundedBalanceInUSD && roundedBalanceInUSD !== isNaN) {
```
MaximusHaximus, line 17: Is `isNaN` a global or something? AFAIK the global `isNaN` I'm aware of is a _function_ that must be called.  Also, `getRoundedBalanceInFiat()` returns '< 0.01'.  Wouldn't we want to show that here rather than `- USD`?  If we're just trying to ignore `< 0.01` I'd prefer that we export that magic string from `helpers` and explicitly check it here rather than use `isNaN` or NaN checks which are notoriously difficult to reason about in JS :)

src/components/common/balance/helpers.js
```diff
@@ -0,0 +1,47 @@
...
+    const formattedNearAmount = rawNearAmount && formatNearAmount(rawNearAmount);
+    const balanceInFiat = Number(formattedNearAmount) * tokenFiatValue;
+    const roundedBalanceInFiat = balanceInFiat && balanceInFiat.toFixed(2);
+    if (roundedBalanceInFiat === '0.00' || formattedNearAmount === '< 0.00001') {
+        return '< 0.01';
+    }
+    return roundedBalanceInFiat;
+};
+
+export const getNearAndFiatValue = (rawNearAmount, tokenFiatValue, fiat = 'usd') => {
```
MaximusHaximus, line 41: 💖

src/components/common/balance/helpers.test.js
```diff
@@ -0,0 +1,75 @@
+import BN from 'bn.js';
```
MaximusHaximus, line 1: 🎉 tests!

MaximusHaximus (changes requested):
Just one thing around the NaN check, otherwise I think this is ready to rock :tm: :) Nice work!

ihaid (approved):

corwinharrell (changes requested):
Looking good so far. A couple notes on the design:

1. We **do not** need to use the conversion component in the following locations:
- "Staking From" account selector
- "Your current validator" list items
- "Authorized Apps" Fee allowance
- Activity Feed list item (only show in the details view for now)

2. The component seems to be using a really light gray in some instances which is less legible (eg profile page balances). Let's go ahead and use the darker gray value across the board so it's easy to read.

3. Whenever we show fiat, let's make sure to include the currency symbol ($) so it's consistent across the board wherever fiat values appear (ie = $50 USD)

4. I'd like to include a tooltip for the dashboard available balance just like we do for the same figure on the profile page, so the user understands exactly what this figure represents.

src/components/common/balance/BalanceDisplayUSD.js
Patrick1904: Good catch:) Fixing.

src/components/common/balance/BalanceDisplayUSD.js
Patrick1904: Good catch:) Forgot to remove isNaN after refactoring.

MaximusHaximus (approved):
:+1: This is looking awesome ❇️ 💖  Nice work, can't wait to see this in action on mainnet :+1: 🚢

corwinharrell (changes requested):
Forgot to mention that for the account dropdown we should be using the same figure as the primary figure on the dashboard so it is 1:1 with what the user sees when they select an account. Let's use only the fiat value there.

MaximusHaximus:
:+1: toot toot 🚢  🚋  @corwinharrell we addressed your change request here so I'm gonna go ahead and merge this sucker :)

Issue #1974, opened 2021-07-28
FAQ button from errors redirects users to Chinese support pages


stefanopepe:
## Context
To better support Chinese-speaking users, we recently introduced a dedicated page at [https://nearhelp.zendesk.com/hc/zh-cn](https://nearhelp.zendesk.com/hc/zh-cn) and hired a native-speaking customer success person located in China.
This is an opportunity to update any Zendesk links generated by the wallet set to the Chinese language.

## Story
As a Chinese-speaking wallet user getting an error, I want to click support and FAQs buttons linking to the Chinese version of Zendesk, so I can read the guides and open a ticket in Chinese language.

<img width="1388" alt="Schermata 2021-07-28 alle 11 45 38" src="https://user-images.githubusercontent.com/6222429/127292924-413952cc-e40b-4af5-abe1-2441502b2a10.png">

## Acceptance Criteria

- [ ] By setting the Chinese language in the wallet (either of the two available options) the support link changes to `https://nearhelp.zendesk.com/hc/zh-cn`
- [ ] In case of an error, the FAQ button (`查看常见问题`) redirects the query to the Chinese language of Zendesk
- [ ] Mobile ok
- [ ] Tested

## Unhappy Path
The user receives an error in Chinese language, but the FAQ and the support links redirect to the English version of the wallet.

stefanopepe:
Adding a P1: it's blocking for @RubyJiao

cc @think-in-universe

think-in-universe:
@stefanopepe  @RubyJiao 

Here is my solution that should probably work: https://github.com/near/near-wallet/pull/1975

@RubyJiao Ruby, could you please test with this version to check whether it works well for Chinese FAQs now? https://near-wallet-pr-1975.onrender.com/

RubyJiao:
It works well for Chinese FAQs

MaximusHaximus:
#1975 is merged so is now live on staging -- it'll be in our next release :)

PR #1975, opened 2021-07-28
Add i18n support of Zendesk FAQ URLs

think-in-universe:
Resolves: #1974

An easy fix to just remove the `en-us` i18n route. 
Zendesk pages will automatically redirect user to the English or Chinese pages based on their browser languages.


Patrick1904 (approved):
@think-in-universe smart 👍

PR #1978, opened 2021-07-29
Use @reduxjs/toolkit for NFT reducer and actions

MaximusHaximus:
See commit messages for details on what changed; I added details to each one as multiple line-commit messages as I made the changes.

TLDR:

1. This PR modifies our existing NFT reducer  and related logic to use `@reduxjs/toolkit` methods like `createSlice()`, `createAsyncThunk()` and memoizing reselect selectors for data accesses.
2. nft state slice keeps contract metadata in a different part of state than the tokens that an account owns; previously we were merging the data together which made it impossible to avoid re-fetching contract metadata _constantly_.  Now we only fetch contract metadata once-and-only-once per browser session.  We can add some kind of cache expiration timeout if we need it.

API references:
Relevant to the Reducer:
https://redux-toolkit.js.org/api/createSlice
https://redux-toolkit.js.org/usage/immer-reducers

Relevant to the Thunk:
https://redux-toolkit.js.org/api/createAsyncThunk

Relevant to everything that reads state from redux :)
https://redux-toolkit.js.org/api/createSelector


marcinbodnar:
@MaximusHaximus How about the global alert, we would like to show a global alert error if the request fails. So far we were adding `showAlert` to the action meta. How it will be implemented in this solution?

MaximusHaximus:
> @MaximusHaximus How about the global alert, we would like to show a global alert error if the request fails. So far we were adding `showAlert` to the action meta. How it will be implemented in this solution?

In almost all cases, we want errors encountered during a thunk to be caught by the dispatcher and result in something being rendered into the ui _in the context where it is relevant_; the components that rely on that data should be aware of what should be rendered if there is an error encountered fetching that data.

As such, the standard way of handling errors during thunk dispatching should be to add an action reducer for the `.rejected` action that `createAsyncThunk` creates for us that is responsible for setting some properties in redux state that the UI uses to display the error to the user (for example, on the create account screen if your account ID is invalid we show that fact right under the account ID input).  In general, we want this error handling to happen as close to the source of the dispatch as possible, so we aren't making assumptions about whether data in global state is relevant to the current component.

As an example, right now with globally generated `localAlert`s, there is no guarantee that a localAlert in the global `status` state slice is actually relevant to the currently rendered UI; we make assumptions that it is when we're rendering a component that is bound to the `localAlert` slice, even if the data in the localAlert could well have been generated by a totally unrelated thunk.

**However**, In places where we have a _background_ task that needs to generate an alert if it fails, we have a few options:
1. Add explicit error handling  (try/catch/dispatch a showAlert action) into the thunk itself (preferred)
2. Define our actions using `createAction` (https://redux-toolkit.js.org/api/createAction), and wire them up using the `extraReducers` "Builder" notation for `createSlice()`, which takes an argument for `meta`.  This PR didn't need either approach because we already were hiding any errors encountered while dealing with 'likely' NFT contracts since they are not guaranteed to actually be NEP171 spec-adhering contracts :)

Patrick1904 (approved):

esaminu (approved):

PR #1985, opened 2021-07-29
Token fiat value: add reducer and action

Patrick1904:


MaximusHaximus (approved):
@Patrick1904 per our discussion, I went ahead and converted it to use @reduxjs/toolkit helpers :).

I went ahead and bound handlers for the pending, fulfilled and rejected cases so `state.fiatValues` will not just contain `tokens`, but also track basic information about whether we're loading, or have failed to load fiat values.
Give it a review and we can land it if you're happy with it :)

Patrick1904:
> @Patrick1904 per our discussion, I went ahead and converted it to use @reduxjs/toolkit helpers :).
> 
> I went ahead and bound handlers for the pending, fulfilled and rejected cases so `state.fiatValues` will not just contain `tokens`, but also track basic information about whether we're loading, or have failed to load fiat values.
> Give it a review and we can land it if you're happy with it :)

This is awesome! Thank you!

Issue #1951, opened 2021-07-23
"Load More" for NFT Collections


corwinharrell:
## Problem

The current implementation of the NFT gallery in the wallet isn't accomodating of larger collections since there is no pagination being used.

## What are we doing about it?

The proposed solution is to add a load more button to each collection. Each collection would display only the 2-4 most recent NFTs received, with an option to load more. Clicking this button would increase the height of the collection container in order to display additional collectibles (+4 each time the user clicks the button).

<img width="617" alt="Screen Shot 2021-07-23 at 9 14 54 AM" src="https://user-images.githubusercontent.com/5934188/126786977-a1b59b0c-4676-4f6a-86a6-7ffbf9a3d976.png">

## Resources

- Figma document: https://www.figma.com/file/Co5Rx8i9Cx4SxgRZv0MUa7/NFTs?node-id=0%3A1

corwinharrell:
@Patrick1904 @MaximusHaximus The designs for the collectibles tab have been updated to include the load more pattern.

Patrick1904:
Closing as this has been released with https://github.com/near/near-wallet/pull/1995

PR #1991, opened 2021-08-02
Release 2021-08-06

MaximusHaximus:
## New functionality
- #1973
Now you will see current estimated value of your NEAR tokens in the wallet, using data from CoinGecko
- #1951
You will now display 4 NFTs per contract that you own NFTs for when the Collectibles tab initially loads.  Clicking 'Load more' will display 4 more tokens at a time, until you have browsed through all of the NFTs you own from that contract.
- #2004 
Using the `/create` route with contract ID and funding key (e.g. as a linkdrop) now takes users _directly_ to the 'create new account' flow _without_ prompting them with the choice to select to claim the linkdrop to an existing account.  This is the 'golden path' for building clean and simple onboarding experiences for new users directly from dapps. Thanks @mattlockyer for all of your insights around getting these flows improved 💖 

## Internal improvements
- #1978

## Localization
- #1960 
- #1975 
- #1998
- #2007
- #2008
- #2009

 
## Bug fixes
- #1979
- #1980 🎉 
- #1999

## Documentation
- #2000


MaximusHaximus:
@ihaid this is ready for testing on Staging :)

ihaid (approved):
LGTM
Curious if this is something critical?
image

MaximusHaximus:
> LGTM
> Curious if this is something critical?
> image

> LGTM
> Curious if this is something critical?
> image

I think this is no problem - this is a side-effect of some fuzzy logic we use to identify contracts that are _possibly_ NEP141 compliant; this happens to be one that isn't actually one but we thought might be.  :+1:

PR #1993, opened 2021-08-03
Prepare NFT pagination: Add and handle Load more button

marcinbodnar:


src/reducers/nft/index.js
```diff
@@ -66,7 +66,8 @@ const fetchNFTsByContractName = createAsyncThunk(
             base_uri: contractMetadata.base_uri,
             fromIndex: selectOwnedTokensForAccountForContract(getState(), { accountId, contractName }).length
         });
-        await dispatch(addTokensMetadata({ accountId, contractName, tokens: tokenMetadata }));
+        const { payload } = await dispatch(addTokensMetadata({ accountId, contractName, tokens: tokenMetadata }));
```
MaximusHaximus, line 69: This is a code smell; our state should be updated by actions dispatched here rather than splitting it apart with some being tracked in hashes in localState and the rest being tracked in our the redux state.

I'm trying to think of a 'rule-of-thumb' for approaches on implementing these types of patterns.  I think the rule should be "If the state you are managing is stored in Redux, then the loading and error states for that state should also be in Redux.  If the state you are managing is not in Redux, then loading and error states for that state can be tracked in the same components where you are managing the state".  What do you think, does that make sense?

MaximusHaximus (changes requested):
Added a few questions in-line -- a lot of issues here are caused by the requirement to keep derived state for loading and 'show button or not' hashes  in sync with our Redux state, which would be resolved by using Redux to keep track of loading and error states instead of trying to do it locally when Redux is really the key source of truth for our UI here :+1:

MaximusHaximus:
Closing in favor of #1995.

PR #1994, opened 2021-08-03
ADD E2E tests for account recovery flow

esaminu:
Follows the same flow as current test on E2E-TESTS directory (create sub account of env account, test recovery then delete).

Uses:

- `jest-playwright`: To run using jest runner and use standard jest hooks and API
- `playwright-testing-library`:  To use `testing-library` queries to find elements on DOM
- `expect-playwright`: extends expect to add additional matchers


e2e-tests/recover/seedPhrase.spec.js
```diff
@@ -0,0 +1,44 @@
...
+        testAccount = await createTestSubaccount();
+    });
+
+    afterAll(async () => {
+        await browser.close();
+        await testAccount.delete();
+    });
+
+    it("navigates to seed phrase page successfully", async () => {
+        const user = await page.goto("/").then(() => page.getDocument());
```
MaximusHaximus, line 15: I'm not a fan of this global `page` object -- this won't scale as our tests become more complex and I'd like to avoid needing to refactor these later.

There can be multiple `page` objects per browser context.  Would like for us to use Page Object Models (POMs) for everything so that the dependency on a specific page is explicit -- in short, there shouldn't be  global variables for things that we can have multiple of.

I actually couldn't even find mention of a `page` global anywhere in the Playwright docs; is this part of one of the extensions? :hocho: We want to err on the side of 'flexible' over 'terse' when it comes to Playwright usage; it looks like default `page` and `context` are passed into `describe` and `test` closures for us explicitly.  Was this new global part of wrapping the test runner with Jest or something similar?

Basically, I want us to be following this pattern:
https://playwright.dev/docs/test-pom/
So that we can re-use the same POM objects in multiple test suites; I want the suites themselves to only be 'create POM objects, call methods on those POM objects, then assert against the POM object.  Ideally all of the page-specific selector logic etc. is encapsulated inside of named methods on the POM.

Ideally I'd like to avoid adding wrappers around `playwright-test` until we've validated that we need them; it makes it harder to reason about the API when it is split up across multiple wrapping toolsets.  If we _do_ need an added wrapper around the playwright test runner, we'll need to document that dependency and reference its documentation somewhere (README.md?) so that other engineers on the team can find documentation about these efficiently.

e2e-tests/recover/seedPhrase.spec.js
```diff
@@ -0,0 +1,44 @@
...
+        await user
+            .findByRole("button", {
+                name: "Import Existing Account",
+            })
+            .then((importAccountBtn) => importAccountBtn.click())
+            .then(() => user.findByText("Recover Account"))
+            .then((recoverAccountBtn) => recoverAccountBtn.click());
```
MaximusHaximus, line 22: Rather than use specific name + roles, I want use to put _explicit_ ids on the elements we need to find with these tests; this is brittle as it currently stands for a few reasons:
1. If we rename the user-facing text, our tests will fail
2. If we run tests in an alternative language, the text string will differ depending on which language we are running the tests in)
3. If we change the _type_ of an element (e.g. refactor a div to a button) we don't want our tests to fail

https://playwright.dev/docs/selectors#id-data-testid-data-test-id-data-test-selectors

This has the added benefit of making our searches _consistent_ -- we can always use a simple selector to find the element by `data-test-id` and dispatch the `click()` directly to it without any temporary local variables -- so this would become:

```javascript
await page.click('data-test-id=importExistingAccountButton');
await page.click('data-test-id=recoverAccountButton');
expect(page).toMatchURL(/\/recover-seed-phrase$/);
```

e2e-tests/recover/seedPhrase.spec.js
```diff
@@ -0,0 +1,44 @@
...
+
+        expect(page).toMatchURL(/\/recover-seed-phrase$/);
```
MaximusHaximus, line 24: Do we need to add a 'wait for the navigation to complete' step here? Does Playwright wait for any navigations to finish internally/automatically _before_ it runs this assertion? Usually I would expect this type of assertion to be a race condition where if the browser is slow at navigating, we get an intermittent test failure because `page` hasn't updated yet?

e2e-tests/utils/account.js
```diff
@@ -0,0 +1,73 @@
+const {
```
MaximusHaximus, line 1: <3 💖 nice, looks clean :)

e2e-tests/recover/seedPhrase.spec.js
```diff
@@ -0,0 +1,44 @@
...
+            .findByRole("textbox")
+            .then((seedphraseInput) =>
+                seedphraseInput.type(testAccount.seedPhrase)
+            )
+            .then(() => user.findByRole("button"))
+            .then((btn) => btn.click())
+            .then(() => page.waitForNavigation());
+
+        expect(page).toMatchURL(/.org\/$/);
+        // TODO use data-testid prop
```
MaximusHaximus, line 39: Hah! Great minds think alike!

Can you add `data-test-id` to these elements and use them in this PR?

I think as we go along, we're going to need to make `adding data-test-id` to things we need to interact with during `e2e` tests `step 1` of testing any particular page

MaximusHaximus (changes requested):
This looks awesome <3 :)

I added some comments in-line; in particular, I'd like for us to create a Page Object Model for the page and encapsulate selectors and actions that are valid for that page inside that POM.

I'm also keen to understand whether we really need Jest or if we can use the playwright test runner directly and eliminate some complexity in the tool chain; playwright test runner seems to have a pretty extensive API.

e2e-tests/jest.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+    preset: "@hover/jest-playwright-preset",
```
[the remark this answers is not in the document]
esaminu: I spent some time deciding and ended up going with `jest-playwright` for the following reasons:

- I wanted to standardize the queries, assertions and hooks between our e2e code and our unit and integration test. That way we can move across them without having to learn / review a new API
- `@playwright/test` also does not currently have support for `testing-library` queries that will allow us to closely simulate user behaviour in our e2e tests. This is what broke the tie for me. Also the `testing-library` queries are the same selectors we will be using in our unit and integration tests.

e2e-tests/package.json
```diff
@@ -8,9 +8,16 @@
...
+    "playwright": "^1.13.1"
   },
   "scripts": {
-    "test": "node -r dotenv/config index.js"
+    "test": "node -r dotenv/config index.js",
+    "e2e": "jest"
+  },
+  "devDependencies": {
+    "@hover/jest-playwright-preset": "^1.3.2",
+    "expect-playwright": "^0.7.2",
```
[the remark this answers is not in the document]
esaminu: I think we will be able to extend expect with the same matchers if we use the `@playwright/test` runner. These matchers will make making assertions on the dom more user friendly and are similar to the ones we'll use in `react-testing-library`.

e2e-tests/recover/seedPhrase.spec.js
esaminu: This global is indeed exposed by the `jest-playwright` runner. `jest-playwright` exposes [the following globals](https://github.com/playwright-community/jest-playwright#globals). I don't think this will cause scaling issues though since each test file gets a new instance of `page` and `context`.

I like the POM pattern but this should not impact which test runner we decide to go with since we can use this pattern with any test runner.

e2e-tests/recover/seedPhrase.spec.js
esaminu: I think this is a (semi case by case) design decision for when we want our tests to fail but my understanding is that we want to simulate user behaviour as closely as possible when writing the tests (not just e2e). We should still use test ids for elements that we reference often or that determine critical state (e.g. currently logged in user). 

> If we rename the user-facing text, our tests will fail

We might want this to be the behaviour of our tests or to make assertions on this since the user facing result has changed.

> If we run tests in an alternative language, the text string will differ depending on which language we are running the tests in)

We might have to import localized strings if we want to do e2e tests in multiple languages. 

> If we change the type of an element (e.g. refactor a div to a button) we don't want our tests to fail

We could make it a practice to match by text in general rather than role.

I also think too many test ids will cause significant maintenance overhead. 

This is an important point for how we write all our tests so we should decide on guidelines on this.

e2e-tests/utils/account.js
esaminu: Thank you :)

e2e-tests/recover/seedPhrase.spec.js
esaminu: In this case I agree since this particular element reflects current logged in state and which user is logged in. I don't prefer using the CSS class selector for tests here and in general 😄

esaminu:
Really like the discussion 😃  <3

 I like the suggested POM pattern and will work on integrating it 👍 

Curious to know what you think about the points on the test runner we should use. As it stands currently we will not be able to use `testing-library` queries if we use `@playwright/test`. The second thing to consider is standardization across the testing pyramid.

Also will appreciate feedback on the test design across the pyramid i.e. how should we generally query (test-ids, what user sees) and what should we assert etc. 😃

MaximusHaximus:
:+1: Good chat on this.  Getting really excited about this stuff <3 :)

To re-cap:
### 1. **Should we use Jest?**
- Using Jest adds an entirely new API on top of the Playwright API, which makes it harder to reason about the 'golden path' for tests leveraging Playwright.
- There are corner cases where some features of Playwright don't work properly with Jest when not running in `runInBand` mode.
- We're going to _try our best_ to avoid adding Jest and just the `playwright-test-runner` for now.  This will allow us to have a single source of truth for APIs and best practices for our e2e tests that isn't made more complex by Jest's 'kitchen sink' approach
- If we encounter serious issues with the playwright test runner that are solved by using Jest, we'll re-visit the choice at that point :)
### 2. **Should we use test IDs to find elements?**
A. In our `e2e` suites (these ones):
- We will use test id's commonly so that our e2e integration checks are not brittle
- This is because our e2e tests exist primarily to tell us if the wallet _works_ from a functional perspective, not necessarily anything more in regards to e.g. specific text strings being displayed or UI layout specifics.  The most important thing in these tests is full integration tests that assert that the side-effects of those tests actually happened :)
B. _Tests inside of the main wallet codebase:_
- Will be more explicit/detailed when it comes to the look and structure of screens, and thus, more brittle
- Will use Jest as we currently do
- We will endeavour to create those tests using the `testing-library` best practices that encourage accessibility as a side-effect.
- These tests will be more brittle because they may fail when things change in _looks_ but not _behaviours_  -- however, the amount of time necessary to find and resolve those failures will be much lower than it is for the e2e suites, which are in a separate package from the main wallet codebase, and so not generally run during e.g. TDD dev.
### 3. We will create Page Object Models (POMs) for each screen that provide an interface for working with that screen (https://playwright.dev/docs/test-pom/)
### 4. We will avoid using globals wherever the framework allows us to do so :)
### 5. We will document our decisions around the toolchain and the rationale we used to decide on them in README.md (and keep it up to date in the future :D)

cc: @Patrick1904 @marcinbodnar :)

e2e-tests/jest.config.js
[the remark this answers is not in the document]
MaximusHaximus: 1. You need to use `yarn add` to add these so that the `yarn.lock` file is updated
2. These should actually be primary `dependencies`, not `devDependencies`, because these packages are actually part of the primary use case for this package -- to test the wallet end-to-end :). In a standard package where these dependencies only exist to be used when running internal tests against the code in `e2e-tests`, then they'd be dev dependencies...but that's not the case here :)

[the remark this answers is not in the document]
MaximusHaximus: We hashed this out via voice calls :+1:

e2e-tests/package.json
```diff
@@ -8,9 +8,14 @@
...
     "near-seed-phrase": "^0.1.0",
-    "playwright": "^1.7.1"
+    "playwright": "^1.13.1"
   },
   "scripts": {
-    "test": "node -r dotenv/config index.js"
+    "test": "node -r dotenv/config index.js",
+    "e2e": "npx playwright test"
+  },
+  "devDependencies": {
```
MaximusHaximus, line 17: 1. You need to use `yarn add` to add these so that the `yarn.lock` file is updated
2. These should actually be primary `dependencies`, not `devDependencies`, because these packages are actually part of the primary use case for this package -- to test the wallet end-to-end :). In a standard package where these dependencies only exist to be used when running internal tests against the code in `e2e-tests`, then they'd be dev dependencies...but that's not the case here :)

esaminu: Done 👍

packages/e2e-tests/recover/seedPhrase.spec.js
```diff
@@ -0,0 +1,48 @@
...
+        await testAccount.delete();
+    });
+
+    test("navigates to seed phrase page successfully", async ({ page }) => {
+        await page.goto("/");
+
+        await page.click(
+            `button:text-matches("Import Existing Account", "i")`
+        );
+        await page.click(`a:text-matches("Recover Account", "i")`);
```
MaximusHaximus, line 27: Would prefer that we use test IDs here rather than rely on the exact wording of the text the user sees.
Imagine if we change `Recover Account` to `Recover your Account` --  we really wouldn't want our CI to fail due to a failed e2e test here in that case; the important thing is less what the text on the element is, but more what role that button fulfills.

MaximusHaximus (changes requested):
Just a couple of small things added in-line :)

Also, can please update the content in `README.md` with the things mentioned in https://github.com/near/near-wallet/pull/1994#issuecomment-892942016 ?

Then I think we're ready to get this sucker landed :)

packages/e2e-tests/package.json
```diff
@@ -1,16 +1,19 @@
 {
-  "name": "near-wallet-e2e-tests",
+  "name": "e2e-tests",
```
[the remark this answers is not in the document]
esaminu: Done 👍

packages/e2e-tests/utils/account.js
```diff
@@ -0,0 +1,73 @@
...
+}
+
+async function createTestSubaccount(accountId) {
+    const {
+        BANK_ACCOUNT: parentAccountId,
+        BANK_SEED_PHRASE: parentAccountSeedphrase,
+        TEST_ACCOUNT_SEED_PHRASE,
+    } = process.env;
+    const testAccountId =
+        accountId || generateTestSubaccountId(parentAccountId);
```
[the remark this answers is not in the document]
esaminu: Awesome discussion.. Refactored 👍

packages/e2e-tests/utils/account.js
```diff
@@ -0,0 +1,73 @@
...
+    config.keyStore.setKey(
+        config.networkId,
+        parentAccountId,
+        parentAccountKeyPair
+    );
+
+    const testAccountKeyPair = getKeyPairFromSeedPhrase(
+        testAccountSeedphrase
+    );
+    config.keyStore.setKey(config.networkId, testAccountId, testAccountKeyPair);
```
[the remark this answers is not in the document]
esaminu: Done 👍

packages/e2e-tests/recover/seedPhrase.spec.js
esaminu: Done 👍

packages/e2e-tests/recover/seedPhrase.spec.js
```diff
@@ -0,0 +1,45 @@
...
+
+    test("navigates to seed phrase page successfully", async ({ page }) => {
+        await page.goto("/");
+
+        await page.click(
+            `button:text-matches("Import Existing Account", "i")`
+        );
+        await page.click(`data-test-id=recoverAccountWithPassphraseButton`);
+
+        expect(page).toMatchURL(/\/recover-seed-phrase$/);
```
MaximusHaximus, line 26: When using playwright matchers, many of them are necessarily `async`, so best practice is to _always_ `await` the `expect()` method itself when using playwright matchers.

Since `await`ing something that is sync still works, this is safe even if a particular matcher works synchronously.

packages/e2e-tests/recover/seedPhrase.spec.js
```diff
@@ -0,0 +1,45 @@
...
+        await page.fill(
+            "data-test-id=seedPhraseRecoveryInput",
+            testAccount.seedPhrase
+        );
+        await page.click(`[type="submit"]`);
+
+        await page.waitForNavigation();
+
+        expect(page).toMatchURL(/\/$/);
+        expect(await page.textContent("data-test-id=currentUser")).toBe(
```
MaximusHaximus, line 41: Because we've added playwright matchers, I believe that this should be done using `.toMatchText()`... e.g.
```await expect(page).toMatchText("data-test-id=currentUser", testAccount.accountId)```

`await`'ing _inside_ of the `expect()` arguments here will cause errors thrown from `page.textContent()` to bubble up directly to the test runner, which will show a generic stack trace/failure instead of a useful `expect()`-created message -- this is because effectively, this line of code transpiles down to something like this:

```javascript
const textContent = await page.textContent("data-test-id=currentUser");
expect(textContent).toBe(testAccount.account.accountId);
```

See the issue? If the first line `throw`s, the second line won't even run so there's no way for `expect()` to create a user-friendly assertion error.

packages/e2e-tests/utils/account.js
```diff
@@ -0,0 +1,77 @@
...
+    return KeyPair.fromString(parseSeedPhrase(seedPhrase).secretKey)
+}
+
+function generateTestAccountId() {
+    return `test-account-${Date.now()}-${
+        Math.floor(Math.random() * 1000) % 1000
+    }`;
+}
+
+async function createRandomBankSubAccount() {
```
MaximusHaximus, line 28: <3 :) Inversion of control FTW!

packages/e2e-tests/utils/account.js
```diff
@@ -0,0 +1,77 @@
...
+
+async function createBankSubAccount(accountId) {
+    const {
+        BANK_ACCOUNT: parentAccountId,
+        BANK_SEED_PHRASE: parentAccountSeedphrase,
+        TEST_ACCOUNT_SEED_PHRASE,
+    } = process.env;
+
+    const testAccountId = `${accountId}.${parentAccountId}`;
+    const testAccountSeedphrase = `${testAccountId} ${TEST_ACCOUNT_SEED_PHRASE}`;
```
MaximusHaximus, line 40: I don't understand why this includes the `testAccountId`; shouldn't the seed phrase be literally just TEST_ACCOUNT_SEED_PHRASE which would always be 12 words from the word list?

I didn't think seed phrases should have any strings in them other than specific words in it from the word list we use for seed phrases.

packages/frontend/src/components/accounts/RecoverAccount.js
```diff
@@ -140,7 +140,7 @@ const RecoverAccount = (props) => {
                 <Option>
                     <Header icon={PhraseIcon}><Translate id='recoverAccount.phrase.title'/><br/><Translate id='recoverAccount.actionType'/></Header>
                     <P><Translate id='recoverAccount.phrase.desc'/></P>
-                    <Button to={`/recover-seed-phrase${props.history.location.search}`} onClick={()=> Mixpanel.track("IE Click seed phrase recovery button")}><Translate id='button.recoverAccount' /></Button>
+                    <Button to={`/recover-seed-phrase${props.history.location.search}`} data-test-id="recoverAccountWithPassphraseButton" onClick={()=> Mixpanel.track("IE Click seed phrase recovery button")}><Translate id='button.recoverAccount' /></Button>
```
MaximusHaximus, line 143: :nit: not a blocker, but please feel free to make props span multiple lines as we add new entries to them (less likely to cause conflicts that need manual resolution as we add/remove/change in the future).

Not a big deal, we'll be adopting a full set of ESLint rules to manage formatting Soon:tm: that will make this an implicit requirement via eslint.

MaximusHaximus (changes requested):
Just about ready to land this now -- just a couple of minor correctness changes in how we're using playwright matchers and one question about the seed phrase content :)

packages/e2e-tests/utils/account.js
esaminu: This is copied from https://github.com/near/near-wallet/blob/master/packages/e2e-tests/index.js#L80 . I assume its to maintain some traceability to the sub accounts. i.e. subaccounts generated will always have their accountId + the parent's seed phrase as their seed phrase.

packages/e2e-tests/recover/seedPhrase.spec.js
esaminu: Was using this to circumvent visibility issue since we have a hidden component for desktop with the same `data-test-id` but changed it to use a chained selector instead 👍

packages/e2e-tests/recover/seedPhrase.spec.js
```diff
@@ -0,0 +1,44 @@
...
+    });
+
+    afterAll(async () => {
+        testAccount && (await testAccount.delete());
+    });
+
+    test("navigates to seed phrase page successfully", async ({ page }) => {
+        await page.goto("/");
+
+        await page.click(`button:text-matches("Import Existing Account", "i")`);
```
MaximusHaximus, line 21: test id here please :)

MaximusHaximus (changes requested):
Just **one** button to change to use a testid, then this sucker is ready to **land** :D Nice job ✋

MaximusHaximus (approved):
:+1: Legit.  Nice work! This is looking awesome, I can't wait to get this running as part of our CI :)

PR #1999, opened 2021-08-05
Recover account: clear previous account state after recovery

Patrick1904:
This fixes issue with NFTs, FTs, or/and staking data lingering after the user recovers a new account.


MaximusHaximus (approved):
:+1: nice save, ready to rock

PR #2002, opened 2021-08-05
Refactor switchAccount action and handle clear NFT state in extraReducer case

marcinbodnar:


MaximusHaximus (changes requested):
Patrick fixed a bug that required splitting `switchAccount()` into two thunks, so you'll need to resolve a conflict here with `switchAccount()`,  in that your new copy of that thunk will need to dispatch `clearAccountState()` rather than explicitly dispatching `staking.clearState()` and `tokens.clearState()`.

Maybe we can remove those two next and then delete the entire `clearAccountState()` thunk? :D that'd be awesome :P

marcinbodnar:
> Patrick fixed a bug that required splitting `switchAccount()` into two thunks, so you'll need to resolve a conflict here with `switchAccount()`, in that your new copy of that thunk will need to dispatch `clearAccountState()` rather than explicitly dispatching `staking.clearState()` and `tokens.clearState()`.
> 
> Maybe we can remove those two next and then delete the entire `clearAccountState()` thunk? :D that'd be awesome :P

It's rebased now. Also added `stakingSlice` and `tokensSlice`, for now, they will handle only `switchAccount` in extraReducers, but refactor whole tokens and staking reducer into slice it's a much bigger change, so I guess it needs to wait for his turn.

Also, I removed `clearAccountState` from `switchAccount` thunk, but it's still a separate thing. because we use it in few places - but eventually, we should refactor these places where we using `clearAccountState` into asyncThinks and we will be able to remove `clearAccountState` entirely.

marcinbodnar:
@MaximusHaximus Also question, my proposition would be to use `/redux/slices` instead of `/slices` - I think `/redux/slices` is a better idea because we will keep all stuff related to redux into single `/redux` folder. Are you ok with that?

MaximusHaximus:
> @MaximusHaximus Also question, my proposition would be to use `/redux/slices` instead of `/slices` - I think `/redux/slices` is a better idea because we will keep all stuff related to redux into single `/redux` folder. Are you ok with that?

:+1: absolutely, go for it :) <3

marcinbodnar:
@MaximusHaximus It's ready.

MaximusHaximus (changes requested):
I think we've gone a bit down the rabbit-hole in this PR by splitting our reducer logic across multiple files using two different distinct ways of managing state mutations _and_ duplicating our initialState as a result of this changeset.  Also, and even with these changes, we are _still_ dispatching explicit `clearState` actions from inside of `clearAccountState()`, which is the original goal that we set out to solve with this changeset.

This is actually directly related to the changeset that Patrick just landed; we discovered there were some places where there were no `dispatch(clearState())` calls happening for _some_ of the slices that needed them -- you can see the changeset here to get an idea of all of the callsites that were relevant: https://github.com/near/near-wallet/pull/1999/files

So we can't rely on `switchAccount`, because we don't always dispatch `switchAccount` when making a different account active (for example, when importing an account.)

Rather than try to make this PR include a partial refactoring of our existing reducers, duplication of initial state, _and_ attempt to eliminate the per-reducer `clearState` actions, I'd like to see us solve one and _only one_ goal in this PR.

Given the goal here is to eliminate the explicit dispatching of `clearState()` actions per store that manages account state, then let's do that within the bounds of our existing reducers and not try to implement mixed reducers that are partially composed of `createSlice()` and partially of our existing reducer logic.  There isn't anything about our existing reducers that keeps us from adding a handler for a specific action to each of them that clears their account specific state; the only question is "What action should we use as _the_ action that each reducer listens to in order to ensure that the account specific state for that reducer is _always_ cleared, _every_ time the user switches accounts via _any_ means.

I _think_ that the only action that is consistently dispatched in _all_ places actually is the `clearAccountState` action, because that is why Patrick created it after discovering that `switchAccount` isn't always called -- so the immediate path to success is probably to convert `clearAccountState` to be a simple action rather than a thunk that dispatches a list of `clearState` actions, and add a handler for that action name to every reducer that needs to clear account specific state.  Then we've eliminate the central list of `clearState()` calls we need to maintain and changed it to be a responsibility of each reducer to respond to e.g. `clearAccountState` actions appropriately.

To re-cap:
1. Our primary goal here is to eliminate all `clearState` actions that are defined in each reducer, and have them all listen to the _same_ action that represents making a different account active.
2. We can't use the `switchAccount` thunks `fulfilled` action, because we don't always call `switchAccount` when making a new account active.
3. We want to avoid creating reducers that are composed of a mixture of `createSlice()` logic and our existing reducers and avoid duplicating `initialState`; we should migrate _each existing reducer and their associated thunks + actions_ to use `createSlice` _in their own PRs_ since we need to review any implicit error handling and both review and test these changes very carefully.  We also don't want to make everyone have to reason about two different ways that the same reducer works depending on which action handler we're looking at - if someone puts logic into our old reducer logic that relies on `immer.js` powered logic that `createSlice()` provides, it is a bug... so we want an all-or-nothing approach to the action handling logic where it is clear to the developer which API they should be using for each reducer.

Happy to chat about this via voice if any of this doesn't make sense <3 :)

marcinbodnar:
@MaximusHaximus Done.

packages/frontend/src/redux/slices/nft/index.js
```diff
@@ -106,11 +106,6 @@ const nftSlice = createSlice({
                 const { contractName, tokens, accountId } = payload;
                 set(state, ['ownedTokens', 'byAccountId', accountId, 'byContractName', contractName, 'hasFetchedAllTokensForContract'], tokens.length < TOKENS_PER_PAGE);
                 update(state, ['ownedTokens', 'byAccountId', accountId, 'byContractName', contractName, 'tokens'], (n) => (n || []).concat(tokens));
-            },
-            clearState(state) {
```
MaximusHaximus, line 110: I think we should still have this reducer here, but listening to the account `CLEAR_ACCOUNT_STATE` action instead of having one defined here in-line; otherwise we won't clear the owned tokens state anymore when switching accounts.

MaximusHaximus (changes requested):
One thing re: the `nft` reducer no longer clearing `ownedTokens` state on account switch, otherwise I think this is ready to land! Nice work :)

packages/frontend/src/redux/slices/nft/index.js
marcinbodnar: `onwedTokens` in NFT are keyed by `accountId` so we don't need to clear it :)

MaximusHaximus: :+1: roger that :D <3

PR #2015, opened 2021-08-06
USD value: fix conversion for >1k NEAR amount and add commas

Patrick1904:
Fixes issue with converting NEAR amounts greater than 1,000 NEAR to USD

Screen_Shot_2021-08-06_at_2 31 52_PM


MaximusHaximus (approved):
:+1: this is good for now, but per our Discord conversation, it would actually be really nice if we had a method exported from `near-api-js` utils that would let us get back yoctoNEAR converted to a decimal of whole NEAR :) We should give back a little and make a PR for that in `near-api-js`, I'm sure other folks will have similar desires to use yocto as a decimal of whole near for simple math.

PR #2017, opened 2021-08-07
Add support for 'auto account import' on mobile device by scanning a QR code

Patrick1904:


Patrick1904:
@corwinharrell this is in an MVP state now. I'll adjust the visuals whenever you have the design:)

To try it out:
1. https://near-wallet-pr-2017.onrender.com/
2. Create or recover account
3. Go to `/profile`
4. Click the button on bottom right side of the page 'import on mobile device'
5. Scan the QR code w/ the camera on your mobile device -> click on Safari prompt -> recover account

src/components/profile/Profile.js
```diff
@@ -266,6 +286,18 @@ export function Profile({ match }) {
                                 )}
                             </>
                         }
+                        {!account.ledgerKey && !isMobile &&
+                            <MobileSharing
+                                onSetMobileSharingLink={ async (option) => {
```
MaximusHaximus, line 291: Man, Profile is one scary place :D

Feels like:
1. The code here that is responsible for building `mobileSharingLink` should exist in a `MobileSharingWrapper` component  that has `useDispatch()` and whatever hook(s) we need from `react-router` (e.g. `useLocation`, `useHistory`, `useMatchedRoute`, `useParams` (reference: https://reactrouter.com/web/api/Hooks) :)
2. We can compute `mobileSharingLink` in the wrapper, in a `useEffect()` hook with the router location and `accountId` as its dependencies, and stash the result of the computation in a `mobileSharingLink` `useState()` variable in the wrapper (like you are doing here now), so that we always re-compute the link state any time either the location or the `accountId` changes (account switch safe!)...
3. Then, I think we no longer need any lifecycle to `clear` the link because it is always valid; all we need to do is show or hide the modal that _displays_ the link :) I think you could completely remove the `onSetMobileSharingLink` prop from the `MobileSharing` component entirely at that point

Whaddya think? It also means that we will no longer be doing an async operation to compute the link _as we're showing the modal_ which might be a bit weird  and cause a flicker in some cases

src/components/login/LoginForm.js
```diff
@@ -51,13 +51,14 @@ const LoginForm = ({
...
+                                <Translate id='login.form.accountIdOnly' />
                             :
                                 <>
-                                    <div className='h2'><Translate id='login.form.isRequestingTo' /> </div>
-                                    <div className='h2'><Translate id='login.form.accessYourAccount' /></div>
+                                    <Translate id='login.form.isRequestingTo' />
+                                    <br/>
+                                    <Translate id='login.form.accessYourAccount' />
                                 </>
                             }
```
MaximusHaximus, line 63: :nit: :lipstick: :D

```javascript
                            {requestAccountIdOnly
                                ? <Translate id='login.form.accountIdOnly' />
                                : <>
                                    <Translate id='login.form.isRequestingTo' />
                                    <br/>
                                    <Translate id='login.form.accessYourAccount' />
                                </>
                            }
```

src/components/accounts/auto_import/AutoImport.js
```diff
@@ -3,44 +3,49 @@ import { Translate } from 'react-localize-redux';
...
     }, []);
 
-    const recoverWithSeedPhrase = async () => {
-        await Mixpanel.withTracking("IE-SP Recovery with seed phrase auto",
+    const recoverWithSecretKey = async () => {
+        await Mixpanel.withTracking(`IE-SP Recovery with ${importType} auto`,
             async () => {
                 setRecovering(true);
-                await dispatch(recoverAccountSeedPhrase(seedPhrase, accountId));
+                await dispatch(recoverAccountSecretKey(secretKey, accountId, false));
```
MaximusHaximus, line 46: Although this argument name makes it _seem_ like we should pass in `false` here, the fact that previously we were not passing in any argument and that `recoverAccountSeedPhrase()` defaults this value to `true` makes this a change in behaviour compared to our existing code.  I dug into the implementation of this function and noted that what is labelled `fromSeedPhraseRecovery` actually just means `shouldCreateFullAccessKey` as far as I can tell -- which we want to be _true_ for this code, still, :)

src/components/Routing.js
```diff
@@ -365,7 +366,34 @@ class Routing extends Component {
                             <Route
                                 exact
                                 path='/auto-import-seed-phrase'
-                                component={AutoImport}
+                                render={() => {
```
MaximusHaximus, line 369: `render()` receives `routeProps` so we should prefer using those over the `window` object :)
routeProps are: `{ match, location, history}`

MaximusHaximus (changes requested):
This is looking awesome -- added a couple of things in-line <3 :)

corwinharrell:
@Patrick1904 was testing this again and had a thought that it would be a good idea to add an option on _mobile only_ within the recovery options screen that contains the same information regarding the mobile access code. If I load up wallet.near.org on my mobile device with the intent of adding my account, click "import existing account", and then see the option to sign in via the mobile access code, it's a super slick experience vs having to enter seed phrase, etc... Thoughts?

Screen Shot 2021-08-11 at 10 47 53 AM

MaximusHaximus:
I'm worried that the current copy is a bit too generic and un-specific about that is happening when someone uses this feature.

The QR code we generate contains the currently active account's private key, but the UI doesn't make it clear that the QR code itself is incredibly sensitive.

I think we should have more detailed verbiage on the screen that shows the `View Mobile Code` button that is extremely explicit about the fact that the QR code is extremely sensitive and should not be saved, photographed, e-mailed, screen-shotted and posted on stackoverflow when asking for help, etc. etc.

IMO we should directly tell the user The QR code we are about to show you **is your private key**. `DO NOT SHARE THE QR CODE WITH ANYONE. ONLY CLICK THE BUTTON BELOW IF YOU ARE SURE THAT NO THIRD PARTIES CAN VIEW YOUR SCREEN. DO NOT SAVE THE QR CODE ANYWHERE UNENCRYPTED.`

It's also not immediately obvious to the user that they are storing a private full-access-key key _permanently_ on to the device they use this feature with; the key will be on that device in its localStorage until they clear the cache on that device -- which many people never do on a phone because finding the option for a lot of browsers is tricky on a phone UI.  I think that we should explicitly tell the user NOT to use this to load their wallet on any device they don't own and have exclusive access to -- e.g. `do not use this feature to check your wallet on your friend's phone` - especially since there is no log out option.

At a low-level, I think that 'Mobile Sharing' sounds too ephemeral and has connotations that you still control the thing being 'shared', but in this case it's actually 'Sending your current private key to a mobile device, and using it on that device to create a new full access private key for that device'.  I think it's dangerous to try to hide this fact from people.  When I think of 'sharing' i think of 'share my screen', which I can then turn off from the source...but in this case, not only has the user shared their private full-access key with that device, but when the device loads that QR code, that device will create a new private full access key for the account and permanently track it -- just the same as if someone had used the 'import from seed phrase' button on their phone directly.  This means that _even if the user deletes the private key that they used to share access with that mobile device_, the new key that was created when the mobile device imported the account still exists unless they delete it too

MaximusHaximus (changes requested):
Just a few things in-line :D Looking awesome <3

corwinharrell:
@MaximusHaximus @Patrick1904 Please view and comment on the Figma doc for proposed changes to the copy and framing around this feature. I think this strikes a nice balance between simplicity and urgency. It doesn't make sense to treat this user-facing UI as documentation that goes into all the technical implications of keys, but I agree with the feedback that we should frame this as highly sensitive data. If we do want to point to a resource that offers an in depth explanation around what is happening with keys, etc... we certainly can and can provide a link, so let me know if there's a resource that exists with this information that makes sense to point to.

https://www.figma.com/file/AKrjagfTligTG4DmGx3iMr/Mobile-Device-Access?node-id=0%3A1

corwinharrell:
Upcoming improvements around key management will help to clarify what private keys are active on the account and will allow the user to remove them or alter permissions. This feature _will_ be more technical in nature, and will help users to understand the implications of certain actions. For example, this action will result in the generation of an additional private key, which will be visible and easily managed through the users key preferences on their account. Stay tuned!

src/components/accounts/auto_import/AutoImportWrapper.js
```diff
@@ -0,0 +1,54 @@
...
+    redirectTo,
+    clearAccountState
+} from '../../../actions/account';
+import { Mixpanel } from '../../../mixpanel/index';
+import AutoImport from './AutoImport';
+
+export function AutoImportWrapper({ secretKey, accountId, mixpanelImportType }) {
+    const dispatch = useDispatch();
+    const { location } = useSelector(({ router }) => router);
+    const [recovering, setRecovering] = useState(false);
```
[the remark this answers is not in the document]
Patrick1904: I like your thinking:)

Patrick1904:
This is being completed in https://github.com/near/near-wallet/pull/2038. Closing this PR.

PR #2020, opened 2021-08-09
Add Redux selectors for Transactions

marcinbodnar:


marcinbodnar:
@MaximusHaximus Now I will refactor Activities so it will be independent of Wallet component, created an Issue for this job: #2021

src/components/wallet/Activities.js
```diff
@@ -85,6 +87,7 @@ const StyledContainer = styled.div`
 const Activities = ({ transactions, accountId, getTransactionStatus }) => {
     const [transactionHash, setTransactionHash] = useState();
     const activityLoader = actionsPending(['GET_TRANSACTIONS', 'REFRESH_ACCOUNT_OWNER']);
+    const transaction = useSelector((state) => selectOneTransactionByHash(state, { accountId, hash: transactionHash }));
```
MaximusHaximus, line 90: <3 yay for removing the in-line `.find()` :D Memoization FTW!

src/redux/slices/transactions/index.js
```diff
@@ -0,0 +1,26 @@
+import { createSelector } from 'reselect';
+
+const SLICE_NAME = 'transactions';
+
+// A helper function to create the parameter selectors
```
MaximusHaximus, line 5: We'll be using this for a _lot_ of selectors; we'll need it in every file we need parameterized selectors.  Would be nice to get this into `src/redux/createParameterSelector.js` so we can import it here and in the other slices that use it rather than duplicate the function in each file; can you also update any other slices that contain this function to import it also? I think it's a small enough change that we can put that extraction + updating the references in other slice(s) to use it in this PR comfortably since it is not a behavioural change, just structural at the filesystem level really.

src/redux/slices/transactions/index.js
```diff
@@ -0,0 +1,26 @@
...
+const selectTransactionsSlice = (state) => state[SLICE_NAME] || {};
+
+export const selectTransactionsByAccountId = createSelector(
+    [selectTransactionsSlice, getAccountIdParam],
+    (transactions, accountId) => transactions[accountId] || []
+);
+
+export const selectOneTransactionByHash = createSelector(
+    [selectTransactionsByAccountId, getHashParam],
+    (transactionsByAccountId, hash) => transactionsByAccountId.find((transaction) => `${transaction.hash}-${transaction.kind}` === hash)
```
MaximusHaximus, line 25: In this context, I think these are just `transactions`, we only _loaded_ them by account Id :)

MaximusHaximus (changes requested):
Just a couple of small things, then we can land this sucker <3 nice work!

src/redux/slices/transactions/index.js
marcinbodnar: Done.

MaximusHaximus (changes requested):
@marcinbodnar There's a merge commit here that is bringing in 109 changed files, 2,767 additions and 2,099 deletions from the `master` branch - that commit should be removed and the branch re-based on `master` instead -- there should be no merge commits from `master` into feature branches because doing so makes your feature branch diff include all changes made to master between when your branch was originally forked from `master` and when you merged `master` back into your feature branch.  By merging, you are making the diff _part of your feature branch's commits_, by rebasing, you are making the diff _only include the changes necessary for your feature, relative to the current state of master_ (a very different thing).

Since this is only _actually_ around 30-40 lines of _real_ diff in this entire PR, you will almost certainly find it easier to re-create your branch and re-implement the changes to those ~40 lines rather than try to resolve conflicts across 13 commits

MaximusHaximus (approved):
:+1:

PR #2022, opened 2021-08-09
Fix account switcher flex-box bugs

corwinharrell:
@Patrick1904 @marcinbodnar I made some small changes to the CSS of the desktop dropdown as per Patrick's request to prevent some bugs that were being caused by the previous flex-box implementation. A couple things to note:

1. `Button` elements should be used for buttons. There's really no reason we should be using `span` or `div` for buttons. This contributes to unexpected behavior in conjunction with `flex` properties, and is also semantically incorrect. The `UserAccounts` component here uses a `span` for the "Sync" button. Unfortunately, simply changing this to a `button` element isn't feasible without further refactoring considering all the things that have been done to make the `span` respond to user input like a `button`. We will need to do this though, so if you find yourself working with this component in the near future and have some time, it would be great to clean it up in this way.

2. We should not be using `!important` in our CSS for the most part. When testing to see how the component would function with a regular `button` element, I noticed that using a standard `button` resulted in several other things breaking, due to the fact that all `button` elements across the App are inheriting a `width: 100%;` property with an `!important` flag. The use of `!important`, especially at the global level, has negative downstream effects and should be avoided. I see that it is being used quite liberally throughout our CSS, so this is something that we'll need to spend some time refactoring if we want to proceed without worrying about having to add unnecessary levels of specificity to counteract it.

### DIFF
--- a/src/components/navigation/UserAccounts.js
+++ b/src/components/navigation/UserAccounts.js
@@ -13,8 +13,6 @@ const Wrapper = styled.div`
     }
 
     @media (min-width: 992px) {
-        display: flex;
-        flex-direction: column;
         max-height: 260px;
         overflow-y: auto;
 
--- a/src/components/navigation/UserAccounts.js
+++ b/src/components/navigation/UserAccounts.js
@@ -85,10 +85,9 @@ const Account = styled.div`
     }
 `;
 
-const SyncButton = styled.button`
+const SyncButton = styled.span`
     background-color: #f8f8f8;
     border-radius: 50px;
-    border: 2px solid #f8f8f8;
     color: #0072ce;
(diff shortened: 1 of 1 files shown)


Patrick1904 (changes requested):
@corwinharrell for some reason I'm still getting the weird behavior where the account height gets crammed with 4+ accounts (Safari only): 

Screen Shot 2021-08-09 at 12 31 24 PM

marcinbodnar:
@corwinharrell  Thanks for handling it. 100% agree. I think the `!important` are the leftovers from Semantic UI - definitely to be removed one by one.

Patrick1904 (approved):

PR #2027, opened 2021-08-10
Refactor Activities component

marcinbodnar:


src/components/wallet/Wallet.js
```diff
@@ -285,7 +282,6 @@ export function Wallet({ tab, setTab }) {
             let id = Mixpanel.get_distinct_id();
             Mixpanel.identify(id);
             Mixpanel.people.set({ relogin_date: new Date().toString() });
-            dispatch(getTransactions(accountId));
```
MaximusHaximus, line 288: <3 nice :+1:

src/components/wallet/ActivitiesWrapper.js
```diff
@@ -84,28 +86,36 @@ const StyledContainer = styled.div`
...
-                        receiverId={transaction.receiver_id}
-                        accountId={accountId}
-                        setTransactionHash={setTransactionHash}
-                    />
-                ))
-                : null
-            }
+            {transactions.map((transaction, i) => (
+                <ActivityBox
+                    key={i}
```
MaximusHaximus, line 110: `key` here should be the transaction hash so they are unique for correctness; using index for key is an anti-pattern in React :) (ref: https://robinpokorny.com/blog/index-as-a-key-is-an-anti-pattern/)

MaximusHaximus (changes requested):
Looks good - just one fix we can make regarding unique `key` property for each `ActivityBox` that was already in the existing component :D

src/components/wallet/ActivitiesWrapper.js
marcinbodnar: Right, didn't notice this bug while refactoring. Done.

MaximusHaximus (changes requested):
@marcinbodnar There's a merge commit here that is bringing in 109 changed files, 2,767 additions and 2,099 deletions from the `master` branch - that commit should be removed and the branch re-based on `master` instead -- there should be no merge commits from `master` into feature branches because doing so makes your feature branch diff include all changes made to master between when your branch was originally forked from `master` and when you merged `master` back into your feature branch.  By merging, you are making the diff _part of your feature branch's commits_, by rebasing, you are making the diff _only include the changes necessary for your feature, relative to the current state of master_ (a very different thing).  Since this is only _actually_ around 30 lines of diff, you will almost certainly find it easier to re-create your branch and re-implement the changes to those ~30 lines rather than try to resolve conflicts across 4 commits

packages/frontend/src/components/wallet/ActivitiesWrapper.js
```diff
@@ -84,28 +86,36 @@ const StyledContainer = styled.div`
...
-                        receiverId={transaction.receiver_id}
-                        accountId={accountId}
-                        setTransactionHash={setTransactionHash}
-                    />
-                ))
-                : null
-            }
+            {transactions.map((transaction, i) => (
+                <ActivityBox
+                    key={transaction.hash}
```
MaximusHaximus, line 110: Since we use `${hash}-${kind}` as a unique key, this should use the same thing.
https://github.com/near/near-wallet/blob/3e356525ee4b5a9bd054fc4aae3a354916a331eb/packages/frontend/src/components/wallet/ActivityBox.js#L126

MaximusHaximus (changes requested):
:+1: Another wrapper joins the ranks :). Nice work on upgrading some selector usages too :)

Just one small thing re: the unique transaction id added in-line.

Not a blocker for this PR, but something we should figure out:

It seems like we can only assume transaction ids/keys are unique when we glue together `hash` and `kind`, but there is no golden path for developers to follow that informs them of this detail - this diff is a case-in-point, I only realized the issue here because I was familiar with the glue pattern from another PR.

What do you think we can do code-wise to create a structure that makes this pattern discoverable?  I wonder if we're missing a `Transaction` object in our current structure.

packages/frontend/src/components/wallet/ActivitiesWrapper.js
marcinbodnar: Done in #2031 :
https://github.com/near/near-wallet/pull/2031/commits/cb252e8c19a83869dbaffe6b8ac5c4e2c81f4390

PR #2031, opened 2021-08-11
Refactor transactions reducer into redux toolkit version

marcinbodnar:


src/redux/slices/transactions/index.js
```diff
@@ -1,7 +1,103 @@
...
+        },
+        updateTransactionStatus(state, { payload }) {
+            const { status, checkStatus, accountId, hash } = payload;
+
+            const transactionsState = state.transactions.byAccountId[accountId];
+            const index = transactionsState.findIndex((t) => t.hash === hash);
+            if (index !== -1) {
+                transactionsState[index].status = status;
+                transactionsState[index].checkStatus = checkStatus;
+            }
```
MaximusHaximus, line 79: Since we don't actually need the index, we can just use `find()` here and mutate the transaction directly due to the magic of `immer.js` :)

like e.g.:

```suggestion
            const transactionEntry = transactionsState.find((t) => t.hash === hash);
            if (transactionEntry) {
                Object.assign(transactionEntry, { status, checkStatus});
            }
```

src/redux/slices/transactions/index.js
```diff
@@ -1,7 +1,103 @@
...
+        const { actions: { updateTransactionStatus } } = transactionsSlice;
+        dispatch(updateTransactionStatus({ status, checkStatus, accountId, hash }));
+    }
+);
+
+const transactionsSlice = createSlice({
+    name: SLICE_NAME,
+    initialState,
+    reducers: {
+        addTransactions(state, { payload }) {
```
MaximusHaximus, line 52: "Add" usually means 'in addition to something that is already there', so this implied to me that it appended new transactions to the existing array.  What do you think about making this `setTransactions` -- `set`ing something generally means replacing whatever value is there, which is what this is doing

src/redux/slices/transactions/index.js
```diff
@@ -1,7 +1,103 @@
...
+
+            const transactionsState = state.transactions.byAccountId[accountId];
+            const index = transactionsState.findIndex((t) => t.hash === hash);
+            if (index !== -1) {
+                transactionsState[index].status = status;
+                transactionsState[index].checkStatus = checkStatus;
+            }
+        }
+    },
+    extraReducers: ((builder) => {
```
MaximusHaximus, line 82: We're missing a case for `rejected` -- right now, if the `fetchTransactions` thunk throws an error, the `loading` state will remain stuck set to `true` (because `fulfilled` only fires on _success_)

See example here: https://github.com/near/near-wallet/blob/9b79d5efb5f5f90e514c0b2eb33b2d915f3fbcfc/src/reducers/nft/index.js#L132-L133 for attaching the rejected handler

src/redux/slices/transactions/index.js
```diff
@@ -1,7 +1,103 @@
...
+const initialState = {
+    transactions: {
+        byAccountId: {}
+    },
+    status: {
+        
+    }
+};
+
+const fetchTransactions = createAsyncThunk(
```
MaximusHaximus, line 18: Feels now that we have loading state in Redux, we should take advantage of that to add a `condition` option to this like we did to the NFT slice, so that we don't dispatch concurrent requests to load the same transactions.

However, to be sure that we don't fail to fetch transactions for an account when we switch to it while there is a pending attempt to load transactions for the account that we were previously viewing, we would need a `status` object to exist _on each entry under `byAccountId`_, rather than as a root construct... which would mean our state shape would look like:
```
{
  transactions: {
    byAccountId: {
      accountName1: {
        items: [...transactionEntries],
        status: { loading, error }
      },
      account2Name: {
        items: [...transactionEntries],
        status: { loading, error }
      },
      ...etc
    }
  }
}

```

This shape also gives us the ability to know which account we failed to load for since we can also track `error` state per account slice; it's basically the same pattern as the nftSlice uses now.

src/redux/slices/transactions/index.js
```diff
@@ -1,7 +1,103 @@
+import { createAsyncThunk, createSlice } from '@reduxjs/toolkit';
+import set from 'lodash-es/set';
```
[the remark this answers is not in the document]
marcinbodnar: Right, wrong dependency. Done.

src/redux/slices/transactions/index.js
```diff
@@ -1,7 +1,103 @@
...
+            const { transactions, accountId } = payload;
+            set(state, ['transactions', 'byAccountId', accountId], transactions);
+        },
+        updateTransactions(state, { payload }) {
+            const { transactions, accountId } = payload;
+
+            const transactionsState = state.transactions.byAccountId[accountId];
+            const hash = transactionsState.map((t) => t.hash_with_index);
+
+            transactions.forEach((t) => {
```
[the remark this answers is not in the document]
marcinbodnar: > Thinking about it, I am assuming it's because we don't want to replace an existing entry that may have had it's tx status loaded for it already?

Correct. We also had automatic transactions refresh every few seconds, which was removed by mistake some time ago. There is an issue to add it again. 

Added comment.

src/redux/slices/transactions/index.js
marcinbodnar: Done.

src/redux/slices/transactions/index.js
marcinbodnar: Agree, done :)

src/redux/slices/transactions/index.js
marcinbodnar: Done.

src/redux/slices/transactions/index.js
marcinbodnar: Done.

packages/frontend/src/redux/slices/transactions/index.js
```diff
@@ -1,22 +1,149 @@
...
+        builder.addCase(fetchTransactions.fulfilled, (state,  { meta }) => {
+            const { accountId } = meta.arg;
+
+            set(state, ['transactions', 'byAccountId', accountId, 'status', 'loading'], false);
+        });
+        builder.addCase(fetchTransactions.rejected, (state, { meta, error }) => {
+            const { accountId } = meta.arg;
+            
+            set(state, ['transactions', 'byAccountId', accountId, 'status', 'loading'], false);
+            set(state, ['transactions', 'byAccountId', accountId, 'status', 'error'], error?.message || 'An error was encountered.');
```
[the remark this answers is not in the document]
marcinbodnar: I think we should store error code?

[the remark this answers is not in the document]
MaximusHaximus: That sounds sane.  Maybe `error` would instead be an object with either (or both) `code` and `message` properties?

Not blocking this PR, but we need some documentation about this pattern written down, so that there is a clear path for everyone to follow as we implement reducers that track error states without using the global alert reducer.

Can you create an issue to commit a markdown file in the `src/redux` directory that outlines this and follow up?  Since you are the most familiar with our current error handling, I'm looking for you to have a think about what we need, and what the data structure should look like.

We also need a basic description outlining that we use `redux-toolkit-js` slices for reducer logic (link to package documentation), reselect selectors for all data accesses (link to package documentation). In a nutshell, we need some developer documentation around Redux patterns in place, since not all of our code follows our current best practices and so there is no obvious golden path just by looking at our existing code usages.

packages/frontend/src/redux/slices/transactions/index.js
```diff
@@ -1,22 +1,142 @@
...
+        builder.addCase(fetchTransactions.pending, (state, { meta }) => {
+            const { accountId } = meta.arg;
+
+            set(state, ['byAccountId', accountId, 'status', 'loading'], true);
+        });
+        builder.addCase(fetchTransactions.fulfilled, (state,  { meta }) => {
+            const { accountId } = meta.arg;
+
+            set(state, ['byAccountId', accountId, 'status', 'loading'], false);
+        });
```
MaximusHaximus, line 96: Should we be resetting `error` state for these cases?  I assume as the very least that we should be for the `fulfilled` case?

MaximusHaximus (changes requested):
Just one small thing re: error state, otherwise I think this is ready to land :+1:

packages/frontend/src/redux/slices/transactions/index.js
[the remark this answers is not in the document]
marcinbodnar: Done. Issue added.

packages/frontend/src/redux/slices/transactions/index.js
marcinbodnar: Done.

packages/frontend/src/redux/slices/transactions/index.js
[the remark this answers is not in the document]
marcinbodnar: > That sounds sane. Maybe error would instead be an object with either (or both) code and message properties?

I will add the same change for other slices in separate PR.

MaximusHaximus (approved):
:+1: Lookin' good! Nice work <3

Issue #2013, opened 2021-08-06
Linkdrop Claimed shows after user returns to wallet if redirectUrl was used in Linkdrop


mattlockyer:
**Problem**
Linkdrop with redirectUrl (/linkdrop or /create path)
Linkdrop claimed
Redirected to app
User goes to NEAR wallet in a new tab later
Linkdrop claimed modal shows

**Expected Behavior**
Linkdrop claimed modal doesn't show

**Steps to reproduce**
Above

Modal pops when user returns to wallet in a new tab.

![unknown-19](https://user-images.githubusercontent.com/321340/128518071-d6c4fe52-709c-49fa-bb0c-beffa232c7d0.png)

mattlockyer:
@MaximusHaximus says

Feels like a bug, can you create an issue about this? We'll figure it out thumbsup

@Patrick1904 I think the original logic here wasn't an issue using localstorage for this flag, because we didn't support redirectUrl on linkdrops yet at the time slight_smile My feels is we should clear the localstorage key before redirecting to redirectUrl even if we need it there for the redirectTo / (do we really? Feels like localstorage is excessive and that redirecting to / should be a navigation that doesn't cause a full reload, thus would allow us to keep the flag in e.g. redux maybe?)

mattlockyer:
@Patrick1904 says

Yea I think redux should be enough for this. Makes sense to only show it if there's no redirect back to an app thumbsup  @Matt Lockyer | NEAR if you can create an issue that'd be great, and I'll look into it

mattlockyer:
FYI this is popping up in live demos I'm doing now and it looks really weird :disappointed:

PR #2037, opened 2021-08-14
Store linkdrop amount in Redux vs localStorage

Patrick1904:
Changes:
1. Store linkdrop amount in `Redux` vs `localStorage` since we do not want to persist the amount if the user navigates away from the Wallet (with a `redirectUrl` for example) post claiming the linkdrop.

Additional context here: https://github.com/near/near-wallet/issues/2013

A good place to create linkdrops for testnet: https://near-drop-testnet.onrender.com/

### DIFF
--- a/packages/frontend/src/components/wallet/LinkDropSuccessModal.js
+++ b/packages/frontend/src/components/wallet/LinkDropSuccessModal.js
@@ -34,10 +34,11 @@ const Container = styled.div`
 `;
 
 const LinkDropSuccessModal = ({ onClose, linkdropAmount }) => {
+    const isOpen = linkdropAmount !== '0';
     return (
         <Modal
             id='near-drop-success-modal'
-            isOpen={linkdropAmount}
+            isOpen={isOpen}
             onClose={onClose}
             modalSize='sm'
             closeButton={true}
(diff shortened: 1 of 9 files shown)


packages/frontend/src/components/wallet/LinkDropSuccessModal.js
```diff
@@ -33,11 +33,11 @@ const Container = styled.div`
 
 `;
 
-const LinkDropSuccessModal = ({ open, onClose, linkdropAmount }) => {
+const LinkDropSuccessModal = ({ onClose, linkdropAmount }) => {
     return (
         <Modal
             id='near-drop-success-modal'
-            isOpen={open}
+            isOpen={linkdropAmount}
```
MaximusHaximus, line 40: '0' will be considered `true` if we refer to `isOpen` using truesy/falsy checks -- which I don't think we want?
Would prefer is `isOpen` is converted to a boolean right here, so that the `Modal` component gets `true` or `false` and _this_ component is responsible for making sure that whatever value is in `linkdropAmount` is coerced properly taking into account it's a string (not a Number) where we are passing it here.

MaximusHaximus (changes requested):
Just one thing in LinkDropSuccessModal, otherwise LGTM

MaximusHaximus (approved):
Legit :shipit: 🚢  :+1:

PR #2039, opened 2021-08-16
Recent activity: Adjust timestamp abbreviations

Patrick1904:
We've received feedback from users that timestamps for minutes/months displayed under 'recent activity' on the dashboard are confusing.

Timestamps will now show: `s, min, hr, d, w, mo, y` (suggested by @corwinharrell).


MaximusHaximus (approved):
:+1:

Issue #1898, opened 2021-07-08
Show what to expect on Ledger screen


ilblackdragon:
Currently after clicking on "Approve" for transaction, a screen is shown saying "Approve operation" but it doesn't show any info about the operation itself. Given most of the times the set of actions that user approves are under the wrap in more information - the info that Ledger shows end up pretty much unexpected and un-useful to user.

It make sense in the "Approve operation on the device" pop up to show more information that user should expect on the Ledger - for them to double confirm that there is no "man in the middle".

stefanopepe:
Assigning this one to @corwinharrell so he can ask a few questions. Do you have a screenshot or a step-by-step process we can use to reproduce the experience?
It's not 100% clear if the issue needs to be solved also on the Ledger screen, or it's simply a wallet-side improvement

corwinharrell:
In general, the implementation of the ledger workflow is suboptimal, and I agree that the primary issue is insufficient feedback for the user as they perform each action. Happy to work towards some improvements here 👍

stefanopepe:
Ping me on Figma when ready for review =)

PR #2040, opened 2021-08-16
Ledger confirm TX message: Instruct user to confirm TX details on Ledger

Patrick1904:
Always show "You will need to confirm the transaction details on your Ledger." in the `LedgerConfirmActionModal` when asking user to confirm the TX. The component has been relying on proper translation strings for every action to be signed with the Ledger, and some actions have not had strings available (or translated), which has degraded the UX with a `no translation found...` or instructions in the wrong language.

In addition, the 'confirm instruction string' has only been a conclusion of the TX that needs to be signed, not listing the specific TX info that the user should expect to see and confirm on the Ledger.

Improvement opportunity: [**Show explicit list of TXs and all information**](https://github.com/near/near-wallet/issues/1898) that the user should expect to confirm on the Ledger device, not relying on translated strings, but raw data.

### DIFF
--- a/packages/frontend/src/reducers/ledger/index.js
+++ b/packages/frontend/src/reducers/ledger/index.js
@@ -126,7 +126,7 @@ const ledgerActions = handleActions({
                 ...state.modal,
                 show: !state.signInWithLedgerStatus && payload.show,
                 action: payload.action,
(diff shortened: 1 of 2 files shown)


MaximusHaximus (approved):
🚢  :+1:

PR #2041, opened 2021-08-17
Account creation e2e tests

esaminu:


packages/e2e-tests/register/seedPhrase.spec.js
```diff
@@ -0,0 +1,117 @@
...
+        page,
+        context,
+    }) => {
+        // skip test on browsers that don't support clipboard API
+        await context
+            .grantPermissions(["clipboard-read", "clipboard-write"])
+            .catch(test.skip);
+        await page.goto(`/setup-seed-phrase/${testAccountId}/phrase`);
+        // skip test on mainnet
+        const isTestWallet = await page.$(
```
MaximusHaximus, line 81: We should check this using an environment variable rather than a magic string that could change at any time

packages/e2e-tests/register/seedPhrase.spec.js
```diff
@@ -0,0 +1,117 @@
...
+        );
+        const wordNumber = parseInt(wordNumberTextContent.slice(6));
+        await page.fill(
+            "data-test-id=seedPhraseVerificationWordInput",
+            copiedSeedPhrase.split(" ")[wordNumber - 1]
+        );
+        await page.click('[type="submit"]');
+        await page.waitForNavigation();
+        expect(page).toMatchURL(/\/$/);
+        expect(await page.textContent("data-test-id=currentUser")).toBe(
```
MaximusHaximus, line 109: Per comment on other PR, to use the matchers we've added, this would read: `await expect(page).toMatchText(...)`

packages/e2e-tests/utils/account.js
```diff
@@ -16,15 +16,24 @@ const getDefaultConfig = () => ({
 });
 
 function getKeyPairFromSeedPhrase(seedPhrase) {
-    return KeyPair.fromString(parseSeedPhrase(seedPhrase).secretKey)
+    return KeyPair.fromString(parseSeedPhrase(seedPhrase).secretKey);
 }
 
 function generateTestAccountId() {
-    return `test-account-${Date.now()}-${
+    return `test-playwright-account-${Date.now()}-${
```
MaximusHaximus, line 23: <3 good idea

packages/e2e-tests/register/seedPhrase.spec.js
```diff
@@ -0,0 +1,117 @@
...
+        await page.goto(`/setup-seed-phrase/${testAccountId}/phrase`);
+        // skip test on mainnet
+        const isTestWallet = await page.$(
+            'div:text-matches("Test-only Wallet", "i")'
+        );
+        if (!isTestWallet) {
+            test.skip();
+        }
+        await page.click(`button:text-matches("Copy Phrase", "i")`);
+        const copiedSeedPhrase = await page.evaluate(() =>
```
MaximusHaximus, line 88: Do you think it makes sense to have the code for these steps inside of the page model?

It feels like the page model should contain almost everything (except for `expect()` calls), or we run the risk of duplicating magic strings and screen-specific behaviours across suites as we make compose more complex test cases.  

Say, for example, we want to add a test that aborts the setup seed phrase flow on the final step, and assert some state is changed as we expect in that case.  If all of the code for interacting with the seed phrase UI exists in this test case, then we can't re-use any part of it when we build our next test case that need to do 90% of the same things, but just not tap 'submit' on the final page, but instead tap 'cancel' (for example).

packages/e2e-tests/register/seedPhrase.spec.js
```diff
@@ -0,0 +1,117 @@
...
+            navigator.clipboard.readText()
+        );
+        await page.waitForSelector(
```
MaximusHaximus, line 91: I think this would be `expect(page).toHaveSelector(...)` if we want to use the playwright matchers we've added to the project

MaximusHaximus (changes requested):
Getting into the good stuff now :+1: 💖  comments/questions added in-line :)

packages/e2e-tests/register/seedPhrase.spec.js
esaminu: Same issue as in the comment, addressed with the same change 👍

packages/e2e-tests/register/seedPhrase.spec.js
esaminu: Still useful when we aren't asserting but we could assert in this case. Updated 👍

packages/e2e-tests/register/seedPhrase.spec.js
esaminu: I agree. I am open to suggestions here since the test runner does not have access to the env variables the app on `WALLET_URL` uses.

packages/e2e-tests/register/models/CreateAccount.js
```diff
@@ -0,0 +1,24 @@
+class CreateAccountPage {
+    constructor(page) {
+        this.page = page;
+    }
+    async navigate() {
+        await this.page.goto(`/create`);
+    }
+    async acceptTerms() {
+        if (await this.page.$('button:text-matches("Agree & Continue", "i")')) {
```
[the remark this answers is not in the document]
esaminu: Added test-id. This code checks for the agree to terms button and clicks it in order to have support for both mainnet & testnet flows - testnet does not have a terms page.

packages/e2e-tests/register/models/SetRecoveryOption.js
```diff
@@ -0,0 +1,20 @@
...
+        await this.page.goto(`/set-recovery/${accountId}`);
+    }
+    async selectRecoveryOption(searchText) {
+        await this.page.click(`div :text-matches("${searchText}", "i")`);
+    }
+    async submitRecoveryOption(searchText) {
+        await this.selectRecoveryOption(searchText);
+        await this.page.click(`[type="submit"]`);
+    }
+    // optionName is option prop in code (currently /phrase|ledger|email|phone/)
```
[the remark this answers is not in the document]
esaminu: the option prop is what's currently used in the frontend code [here](https://github.com/near/near-wallet/blob/account-creation-e2e-tests/packages/frontend/src/components/accounts/recovery_setup/RecoveryOption.js#L215). I setup the testIds so that they are based on that prop just like icons and active state etc.

The way it's written currently, the element that has the `active` prop we need to make assertions on does not trigger the active state. So to circumvent this we use the test-id to make assertions on the `active` prop and the searchText to toggle it by clicking.

packages/e2e-tests/register/models/SetRecoveryOption.js
```diff
@@ -0,0 +1,20 @@
+class SetRecoveryOptionPage {
+    constructor(page) {
+        this.page = page;
+    }
+    async navigate(accountId) {
+        await this.page.goto(`/set-recovery/${accountId}`);
```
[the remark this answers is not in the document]
esaminu: Noted 👍 . I think it may be useful to return the page in some cases but it can also always be accessed on the instance.

packages/e2e-tests/register/models/SetRecoveryOption.js
```diff
@@ -0,0 +1,20 @@
...
+    }
+    async selectRecoveryOption(searchText) {
```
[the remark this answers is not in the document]
esaminu: Makes sense! Refactored 👍

packages/e2e-tests/register/seedPhrase.spec.js
```diff
@@ -0,0 +1,117 @@
...
+    }) => {
+        const setRecoveryOptionPage = new SetRecoveryOptionPage(page);
+        await setRecoveryOptionPage.navigate(testAccountId);
+
+        const recoveryOptions = [
+            { name: "ledger", searchText: "hardware wallet" },
+            { name: "email", searchText: "email" },
+            { name: "phone", searchText: "phone" },
+            { name: "phrase", searchText: "passphrase" },
+        ];
```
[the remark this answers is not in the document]
esaminu: Added testIds based on `option` prop from `/frontend` 👍

packages/e2e-tests/register/seedPhrase.spec.js
```diff
@@ -0,0 +1,117 @@
...
+                )
+                .then((optionHandle) =>
+                    expect(optionHandle).toMatchAttribute("class", /active/)
+                );
+
+        await recoveryOptions.reduce(
+            (prev, curr) =>
+                prev.then(() => clickRecoveryOptionAndAssertActive(curr)),
+            Promise.resolve()
+        );
```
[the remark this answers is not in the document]
esaminu: Noted for future, will not be looping through a data structure as discussed in above comment 👍

packages/e2e-tests/register/models/CreateAccount.js
```diff
@@ -0,0 +1,24 @@
...
+                'button:text-matches("Agree & Continue", "i")'
+            );
+        }
+    }
+    async submitAccountId(accountId) {
+        await this.page.fill(
+            "data-test-id=createAccount.accountIdInput",
+            accountId
+        );
+        await this.page.click(`[type="submit"]`);
```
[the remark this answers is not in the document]
esaminu: Done 👍

packages/e2e-tests/register/seedPhrase.spec.js
esaminu: Yes agreed this code should be in the POM. Refactored 👍

packages/e2e-tests/register/seedPhrase.spec.js
esaminu: Added a `WALLET_NETWORK` env variable check to address this 👍

packages/e2e-tests/register/models/SetRecoveryOption.js
```diff
@@ -0,0 +1,55 @@
...
+        return await this.page.$(this.getEmailSelector());
+    }
+    async getPhoneRecoveryElementHandle() {
+        return await this.page.$(this.getPhoneSelector());
+    }
+    async getSeedPhraseRecoveryElementHandle() {
+        return await this.page.$(this.getSeedPhraseSelector());
+    }
+    async submitRecoveryOption() {
+        await this.page.click(`[type="submit"]`);
```
MaximusHaximus, line 52: Would prefer if this was clicking on an element with an explicit ID; otherwise this really isn't `submitRecoveryOption()`, it's just a generic 'click form submit button' action and doesn't really have any value as part of the page model, and it leaves the possibility of us ending up in a weird state where the UI isn't where we expect it to be, but there is a submit there (even if it's the _wrong_ submit button for a totally different form), so we click it because the selector here is so generic.  It can make it very difficult to figure out why the test failed because tapping on the wrong `submit` button will likely cause a navigation even farther away from the screen where we initially got 'lost' in the flow.  We want the test to  throw / fail _as close as possible_ to when the state of our UI is inconsistent with what we expect and being explicit here helps us towards that goal even if the suite that uses this function isn't _extremely_ explicit about asserting after each step.

packages/e2e-tests/register/models/SetupSeedPhrase.js
```diff
@@ -0,0 +1,16 @@
+class SetupSeedPhrasePage {
+    constructor(page) {
+        this.page = page;
+    }
+    async navigate(accountId) {
+        await this.page.goto(`/setup-seed-phrase/${accountId}/phrase`);
+    }
+    async copySeedPhrase() {
+        await this.page.click(`button:text-matches("Copy Phrase", "i")`);
```
MaximusHaximus, line 9: test id please - we don't want this test to fail if we simply rename `Copy Phrase` to `Copy Seed Phrase` :)

packages/e2e-tests/register/models/SetupSeedPhrase.js
```diff
@@ -0,0 +1,16 @@
...
+        return this.page.evaluate(() => navigator.clipboard.readText());
+    }
+    async continueToSeedPhraseVerification() {
+        await this.page.click(`button:text-matches("Continue", "i")`);
```
MaximusHaximus, line 13: test id here also please; if we rename the UI display for this button to "Next" but it is logically the same button, we don't want the test suite to fail

packages/e2e-tests/register/models/VerifySeedPhrase.js
```diff
@@ -0,0 +1,23 @@
...
+            "data-test-id=seedPhraseVerificationWordNumber"
+        );
+        return parseInt(wordNumberTextContent.slice(6));
+    }
+    async verifyWithWord(word) {
+        await this.page.fill(
+            "data-test-id=seedPhraseVerificationWordInput",
+            word
+        );
+        await this.page.click('[type="submit"]');
```
MaximusHaximus, line 19: As with `submitRecoveryOption()` case, would prefer if this was clicking on an element with an explicit ID

packages/e2e-tests/register/seedPhrase.spec.js
```diff
@@ -0,0 +1,125 @@
...
+
+const { describe, afterAll } = test;
+
+describe("Account Registration Using Seed Phrase", () => {
+    const testAccountId = generateTestAccountId();
+    let testAccountInstance;
+
+    afterAll(async () => {
+        testAccountInstance &&
+            (await testAccountInstance.deleteAccount("testnet"));
```
MaximusHaximus, line 21: The account that we sweep funds to should probably be the `BANK_ACCOUNT` (that way we send funds back to the bank account if we are running on mainnet)

packages/e2e-tests/register/seedPhrase.spec.js
```diff
@@ -0,0 +1,125 @@
...
+    });
+    test("is able to verify seed phrase and access wallet", async ({
+        page,
+        context,
+    }) => {
+        // skip test on browsers that don't support clipboard API
+        await context
+            .grantPermissions(["clipboard-read", "clipboard-write"])
+            .catch(test.skip);
+        // skip test on mainnet
```
MaximusHaximus, line 89: Would be good for this comment to indicate what conditions need to be met for us to enable this on mainnet

MaximusHaximus (changes requested):
A few remaining questions/request added in-line, then I think we'll be ready to get this sucker landed :+1:

PR #2042, opened 2021-08-17
Transfer tokens e2e tests

esaminu:
End to end tests for transferring NEAR (Currently 0.1 Ⓝ) between two accounts created by the test and deleted after. 

Add / extend the following POMs
* Send Money page `/send-money`
* Home page  `/` - added methods to login using seed phrase / key pair by setting localStorage

### DIFF
--- a/packages/e2e-tests/register/seedPhrase.spec.js
+++ b/packages/e2e-tests/register/seedPhrase.spec.js
@@ -3,6 +3,7 @@ const { test, expect } = require("@playwright/test");
 const {
     generateTestAccountId,
     connectToAccountWithSeedphrase,
+    getWalletNetwork,
 } = require("../utils/account");
 const { HomePage } = require("./models/Home");
 const { CreateAccountPage } = require("./models/CreateAccount");
@@ -87,7 +88,7 @@ describe("Account Registration Using Seed Phrase", () => {
             .grantPermissions(["clipboard-read", "clipboard-write"])
             .catch(test.skip);
         // skip test on mainnet
-        if (process.env.WALLET_NETWORK === "mainnet") {
+        if (getWalletNetwork() === "mainnet") {
             test.skip();
         }
 
--- a/packages/e2e-tests/transfer-tokens/models/SendMoney.js
+++ b/packages/e2e-tests/transfer-tokens/models/SendMoney.js
@@ -8,7 +8,7 @@ class SendMoneyPage {
     async typeAmount(amount) {
         await this.page.fill(
             `data-test-id=sendMoneyAmountInput`,
-            (amount && amount.toString()) || ""
+            amount.toString()
         );
     }
     async typeAndSubmitAmount(amount) {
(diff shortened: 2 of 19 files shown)


MaximusHaximus (changes requested):
Comments in-line! This is coming together great :) Nice work

packages/e2e-tests/transfer-tokens/models/SendMoney.js
```diff
@@ -0,0 +1,44 @@
...
+    constructor(page) {
+        this.page = page;
+    }
+    async navigate() {
+        await this.page.goto(`/send-money`);
+    }
+    async typeAmount(amount) {
+        await this.page.fill(
+            `data-test-id=sendMoneyAmountInput`,
+            (amount && amount.toString()) || ""
```
MaximusHaximus, line 11: Feels like this method should _only_ be responsible for typing a specific amount with an assumption that it should only ever be called with a valid value for `amount` -- I don't like the implicit fallback logic here that will produce an empty string if `amount` is falsy; this could hide bugs in our code, and if we need to be able to reset the content in a field I think it is safest to implement an explicit behavioural method to do that (e.g. 'resetAmount()`)

packages/e2e-tests/transfer-tokens/transfer-token.spec.js
```diff
@@ -0,0 +1,96 @@
...
+        await firstAccountSendMoneyPage.confirmTransaction();
+
+        await expect(page).toMatchText(
+            'span:text-matches("Transaction complete!", "i")',
+            new RegExp(`${transferAmount} NEAR`)
+        );
+        await expect(page).toMatchText(
+            'span:text-matches("Transaction complete!", "i")',
+            new RegExp(secondAccount.account.accountId)
+        );
```
MaximusHaximus, line 85: test IDs please :)

packages/e2e-tests/transfer-tokens/transfer-token.spec.js
```diff
@@ -0,0 +1,96 @@
...
+        await firstAccountHomePage.loginWithSeedPhraseLocalStorage(
+            firstAccount.account.accountId,
+            firstAccount.seedPhrase
+        );
+
+        await firstAccountHomePage.navigate();
+
+        await firstAccountHomePage.clickSendButton();
+
+        await expect(firstAccountHomePage.page).toMatchURL(/send-money$/);
```
MaximusHaximus, line 44: Not a blocker for this PR, but I think that we should make it a convention that any magic strings like this one should be static properties of the page model, so that both the suite and any internal navigation methods etc. can reference the same source of truth, similar to how the different `recoveryOptions` are set up on the page model.

MaximusHaximus (changes requested):
:+1: rock on, this is looking just about ready to land - just a pair of minor things added in-line

packages/e2e-tests/register/seedPhrase.spec.js
```diff
@@ -87,7 +88,7 @@ describe("Account Registration Using Seed Phrase", () => {
             .grantPermissions(["clipboard-read", "clipboard-write"])
             .catch(test.skip);
         // skip test on mainnet
-        if (process.env.WALLET_NETWORK === "mainnet") {
+        if (getWalletNetwork() === "mainnet") {
```
MaximusHaximus, line 91: Not a blocker, but in a follow-up PR I would like to see valid values for `networkId` in a `constants.js` file (which we will later import from a central, shared package)

MaximusHaximus (approved):
Looks good! Needs a rebase, then this can be merged to master as soon as that's done :)

packages/e2e-tests/register/seedPhrase.spec.js
esaminu: Will do in linkdrop PR 👍

PR #2046, opened 2021-08-18
Token list display improvements

Patrick1904:
Changes:
1. Show token name on the left side of the list, and add token symbol next to the amount, on the right side of the list.
2. Add comma formatting for larger token amounts - e.g.  1000000.000 nUSDC --> 100,000.000 nUSDC
3. Adds ellipsis treatment to token name when needed

Screen Shot 2021-08-18 at 6 54 08 PM

### DIFF
--- a/packages/frontend/src/components/wallet/ActivityBox.js
+++ b/packages/frontend/src/components/wallet/ActivityBox.js
@@ -44,27 +44,16 @@ const StyledContainer = styled.div`
         overflow: hidden;
         font-weight: 700;
         color: #24272a;
+        min-width: 0;
 
         > span {
             font-weight: normal;
             color: #A2A2A8;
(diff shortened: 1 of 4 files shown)


Patrick1904:
@corwinharrell I replaced token symbol with token name as we discussed, and added the token symbol to the right of the balance.

Do you think we need any special treatment (ellipsis, etc) for longer token names (see screen shot https://github.com/near/near-wallet/pull/2046#issue-714668717) or is regular wrapping enough?

corwinharrell:
@Patrick1904 yeah for the name on the left side let's concatenate to avoid collision with the top row value on the right 👍

MaximusHaximus (approved):
:+1: 🚢

PR #2053, opened 2021-08-20
Always require account ID and phrase in the URL to land on /recover-with-link

Patrick1904:
Changes:
1. Always require `accountId` and `seedPhrase` from URL params to land on the `/recover-with-link` page.


MaximusHaximus (approved):
<3 :shipit:

PR #2054, opened 2021-08-20
Refactor tokens actions and reducer into redux toolkit version

marcinbodnar:


packages/frontend/src/utils/tokens.js
```diff
@@ -5,30 +5,21 @@ import sendJson from '../tmp_fetch_send_json';
 import { ACCOUNT_HELPER_URL } from './wallet';
```
MaximusHaximus, line 5: This file should be deleted rather than modified -- it has been _replaced_ with `FungibleTokens` service https://github.com/near/near-wallet/blob/cffe2d4829ffb9c42efe4853ae2b5a1609b64a75/packages/frontend/src/services/FungibleTokens.js#L44, whose associated functions already have the correct signature and return values.

packages/frontend/src/actions/account.js
```diff
@@ -460,7 +459,6 @@ export const finishAccountSetup = () => async (dispatch, getState) => {
     await dispatch(refreshAccount());
     await dispatch(getBalance());
     await dispatch(staking.clearState());
-    dispatch(tokens.clearState());
```
MaximusHaximus, line 463: Pretty sure these lines are gone as soon as we land prior PR, so this shouldn't be part of this diff

packages/frontend/src/redux/slices/tokens/index.js
```diff
@@ -0,0 +1,173 @@
...
+    },
+    extraReducers: ((builder) => {
+        builder.addCase(fetchOwnedTokensForContract.pending, (state, { meta }) => {
+            const { accountId, contractName } = meta.arg;
+            set(state, ['ownedTokens', 'byAccountId', accountId, contractName, 'loading'], true);
+        });
+        builder.addCase(fetchOwnedTokensForContract.fulfilled, (state, { meta }) => {
+            const { accountId, contractName } = meta.arg;
+            set(state, ['ownedTokens', 'byAccountId', accountId, contractName, 'loading'], false);
+        });
```
MaximusHaximus, line 98: Should we be re-setting `error` value in either of these?  I would assume we'd at least want to in `.fulfilled`?

MaximusHaximus (changes requested):
As with other PRs, this has a merge commit that's adding a few thousand lines of diff from `master` into this feature branch's commits.  Other than that, just a couple things in-line! :)

packages/frontend/src/utils/tokens.js
marcinbodnar: Done.

packages/frontend/src/actions/account.js
marcinbodnar: Done.

packages/frontend/src/redux/slices/tokens/index.js
marcinbodnar: Done.

packages/frontend/src/actions/account.js
MaximusHaximus: Looking at the current `master`, it appears that this code is actually still in the codebase now; I think this is a bug since the _handlers_ for these actions were deleted in #2002 (now these dispatch calls are no-ops) -- please create a separate fix for this problem, we doesn't need to block this PR and wouldn't want to roll a bug-fix into this anyhow :)

MaximusHaximus (approved):
Looks like there's a conflict to resolve, otherwise this is ready to land! <3 :)

I added one comment in-line about a bug introduced in #2002 because the `clearState()` dispatches in `finishAccountSetup()` were not removed / replaced with the new `clearAccountState` dispatch logic, but the listeners to the previous `CLEAR_STATE` actions were removed -- that should be fixed ASAP in it's own PR since that is now a live bugs since the handlers were removed in #2002 :)

PR #2055, opened 2021-08-23
HotFix to show stNEAR and META tokens in NEAR Wallet.

htafolla:


htafolla (commented):
Immediate fix to wallet issue for METAPool.

MaximusHaximus:
Thanks for the submission! :)

At this time, we only show tokens in the token list in the wallet if the active account _currently_ owns some amount of those tokens.  This list existed to show static placeholders (with 0 balances) for a few specific tokens before we modified the UI to hide any tokens the account doesn't currently hold a balance of.

Adding to the `WHITELISTED_CONTRACTS` list no longer has any impact, and the list only remains in existence now because we kept it around after shipping the last set of changes just in case we needed to roll back changes involving the FT transfers feature; it will be deleted entirely in the next sprint while cleaning up old code.

Also, @htafolla, for any future PRs/FYI - because you have added commits to your master branch in your fork, there is no way for you to request PRs without including everything you've ever added to the master branch in your repository.  To make contributions without your master branch diverging from ours and becoming irreconcilable, you should instead create a new branch off of master, add commits to that, and then create a PR that pulls that branch into master in `near-wallet`.  This way commits in your feature branches are independent of eachother and you don't end up with a running sum of all changes you've ever made to `master` in your next PR (which, as it stands now, would include all of the commits in this PR plus whatever other commits you add to your master branch between now and then.)

I'm going to close this PR for now since it is effectively a no-op in our current wallet codebase to add to `WHITELISTED_CONTRACTS`.

PR #2056, opened 2021-08-23
ADD tests for login with dapp flows

esaminu:
- End to end tests and assertions for login with dapp flows. Creates a test account and then asserts an access key exists in local storage for that account after access is granted through the flow and that no pending key exists.
- Added a POM for login page `/login`.

### DIFF
--- a/packages/frontend/src/components/login/SelectAccountDropdown.js
+++ b/packages/frontend/src/components/login/SelectAccountDropdown.js
@@ -159,7 +159,18 @@ const SelectAccountDropdown = ({
             >
                 <Segment basic>
                     <div className='item list-title'>
-                        {dropdown ? translate('button.close') : <div className={classNames({dots: !account.accountId})}>{account.accountId}</div>}
+                        {dropdown ? (
+                            translate("button.close")
+                        ) : (
+                            <div
+                                data-test-id="dropdownCurrentlySelectedAccount"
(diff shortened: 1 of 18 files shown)


packages/frontend/src/components/login/SelectAccountDropdown.js
```diff
@@ -159,7 +159,18 @@ const SelectAccountDropdown = ({
...
+                        ) : (
+                            <div
+                                data-test-id="dropdownCurrentlySelectedAccount"
+                                className={classNames({
+                                    dots: !account.accountId,
+                                })}
+                            >
+                                {account.accountId}
+                            </div>
+                        )}
```
MaximusHaximus, line 173: 💄 Same thing as above re: ternary formatting convention
```suggestion
                        {dropdown 
                          ? ( translate("button.close") )
                          : (
                            <div
                                data-test-id="dropdownCurrentlySelectedAccount"
                                className={classNames({
                                    dots: !account.accountId,
                                })}
                            >
                                {account.accountId}
                            </div>
                        )}
```

MaximusHaximus (changes requested):
Looking good <3 :). Just a few things in-line.  Nice work 💖

MaximusHaximus (approved):
Awesome work, these files look _clean_ :). <3

PR #2057, opened 2021-08-24
Add 'noindex' meta tag to pages with sensitive information

Patrick1904:
Changes:
1. Add 'noindex' meta tag to pages with sensitive information
2. Merge `PrivateRouteLimited.js` with `PrivateRoute.js` since they should behave the same
3. Require `fundingContract` and `fundingKey` in the URL on `/linkdrop` route


packages/frontend/src/components/Routing.js
```diff
@@ -290,44 +291,45 @@ class Routing extends Component {
...
+                            <PublicRoute
                                 exact
                                 path='/fund-create-account/success'
                                 component={SetupImplicitSuccess}
                             />
-                            <Route
+                            <PublicRoute
                                 exact
                                 path='/setup-ledger/:accountId'
                                 component={SetupLedgerWithRouter}
```
MaximusHaximus, line 335: I noticed that the logic for the existing `PrivateRoute` component checks the value of `refreshAccountOwnerEnded`, but that the new version no longer contains that logic.  As a result, we could have changed the behaviour here for these two routes that used the version of the `PrivateRoute` with handling for `refreshAccountOwnerEnded`.

Are you absolutely sure that we can remove the logic for `refreshAccountOwnerEnded` and not break anything about how these two routes function?  If we're not fully confident about this, we should keep the `refreshAccountOwnerEnded` check in place for these two routes to avoid introducing an unintended bug here while cleaning up apparently duplicated logic.

cc: @marcinbodnar Are you familiar with why these two routes care about `refreshAccountOwnerEnded`?

packages/frontend/src/components/common/PrivateRoute.js
```diff
@@ -3,25 +3,36 @@ import { connect } from 'react-redux';
...
+                            to={{
+                                pathname: '/',
+                            }}
+                        />
+                    )
+                    : ( Component // <Route component> takes precedence over <Route render></Route>
+                        ? <Component {...props} />
+                        : (render ? render(props) : <></>)
+                    )
+            )}
```
MaximusHaximus, line 33: This ternary tree is a little bit out of hand; would be nice to go ahead and replace all of these ternaries with some more readable / less dense code

Webstorm intentions unwrapped the ternaries for me automatically (indentation is likely wrong on this suggestion)
```suggestion
            render={(props) => {
                    if (!localStorage.getItem(KEY_ACTIVE_ACCOUNT_ID)) {
                        return <Redirect
                            to={{
                                pathname: '/',
                            }}
                        />;
                    }
                    
                    // <Route component> takes precedence over <Route render></Route>
                    if (Component) {
                        return <Component {...props} />;
                    }

                    if (render) {
                        return render(props);
                    } 

                    return (<></>);
}}
```

MaximusHaximus (changes requested):
Looks good, just one question about possible behaviour changes for the two routes that previously took advantage of `refreshAccountOwnerEnded` prop and an opportunity to unwind some deeply nested ternary trees :)

PR #2075, opened 2021-08-31
Added scrollbar to user account dropdown

BenKurrek:
Added scrollbar to user account dropdown when there are more than 3 current accounts.


Patrick1904:
@corwinharrell any thoughts on adding an explicit scrollbar to the account dropdown?

MaximusHaximus:
@Patrick1904 @corwinharrell Just my 2 cents, I really hate not having one :P

PR #2085, opened 2021-09-06
Staking e2e tests

esaminu:
- Adds e2e tests for staking and unstaking flows. 
- Runs these flows on newly created accounts with random validators and asserts the balance state after the operations. 
- Added POM for staking page `/staking`


packages/e2e-tests/utils/account.js
```diff
@@ -1,99 +1,77 @@
...
+        this.isCreated = true;
+        await this.initialize();
+        return this;
+    }
+    spawnRandomSubAccountInstance() {
+        if (!this.nearApiJsAccount) {
+            throw new Error("Account needs to be initialized to spawn sub accounts");
+        }
+        const randomSubaccountId = generateTestAccountId();
+        const randomSubaccountSeedphrase = `${randomSubaccountId} ${process.env.TEST_ACCOUNT_SEED_PHRASE}`;
```
MaximusHaximus, line 41: Curious about this -- why do prepend the seed phrase with `randomSubaccountId`?  If this is important, we should comment here on why.  It seems a bit out of place given that actual users always enter a 12 word seed phrase conforming to bip39.  If we want unique seed phrases for each account, we can just generate them using `near-seed-phrase` (https://github.com/near/near-wallet/blob/3c8ac4a32750597d4d467d0aa53e4f381e975dc5/packages/frontend/src/components/accounts/SetupSeedPhrase.js#L64)

packages/e2e-tests/utils/account.js
```diff
@@ -1,99 +1,77 @@
...
-
-function getKeyPairFromSeedPhrase(seedPhrase) {
-    return KeyPair.fromString(parseSeedPhrase(seedPhrase).secretKey);
+class E2eTestAccount {
+    constructor(accountId, seedPhrase, parentNearApiJsAccount) {
+        this.accountId = accountId;
+        this.seedPhrase = seedPhrase;
+        this.parentNearApiJsAccount = parentNearApiJsAccount;
+    }
+    async initialize() {
```
MaximusHaximus, line 17: Feels like initializing a new account would be:
1. ensure we have a nearAPIJS instance
2. create the account
3. Set the key in the keystore only if the account creation was a success

Is there any benefit we get from separating out the actual creation of the account to a separate function call than `initialize()`?  Seems like it just adds an extra call to ever place someone makes one of these objects?

packages/e2e-tests/utils/account.js
```diff
@@ -1,99 +1,77 @@
...
-    const testAccountKeyPair = getKeyPairFromSeedPhrase(seedPhrase);
-    await config.keyStore.setKey(
-        config.networkId,
-        accountId,
-        testAccountKeyPair
-    );
-
-    const near = await connect(config);
-    return near.account(accountId);
+    return `test-playwright-account-${Date.now()}-${Math.floor(Math.random() * 1000) % 1000}`;
```
MaximusHaximus, line 63: Lets call this `test-wallet-e2e-account` since the framework we're using is less important and might change :)

MaximusHaximus (changes requested):
Comments/questions in-line -- good work, this is a beast :)

MaximusHaximus (approved):
:+1: nice work :) Let's get this sucker merged!

PR #2094, opened 2021-09-08
fix typo in wallet.js error handling

RNHTTR:
fix `indexOf(``ccount...` with `indexOf(``account...`

### DIFF
--- a/packages/frontend/src/utils/wallet.js
+++ b/packages/frontend/src/utils/wallet.js
@@ -188,7 +188,7 @@ class Wallet {
         } catch (error) {
             console.log('Error loading account:', error.message);
 
-            if (error.toString().indexOf(`ccount ${this.accountId} does not exist while viewing`) !== -1) {
+            if (error.toString().indexOf(`account ${this.accountId} does not exist while viewing`) !== -1) {
                 const accountId = this.accountId;
                 const accountIdNotConfirmed = !getAccountConfirmed(accountId);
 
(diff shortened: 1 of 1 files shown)


packages/frontend/src/utils/wallet.js
```diff
@@ -188,7 +188,7 @@ class Wallet {
         } catch (error) {
             console.log('Error loading account:', error.message);
 
-            if (error.toString().indexOf(`ccount ${this.accountId} does not exist while viewing`) !== -1) {
+            if (error.toString().indexOf(`account ${this.accountId} does not exist while viewing`) !== -1) {
```
MaximusHaximus, line 191: I think we had issues with the first character not always being capitalized or lowercase across API versions, so we actually started matching just `does not exist while viewing` instead  (see: https://github.com/near/near-wallet/blob/36ad3602820b6a27f739f4504dd1a17589e133e0/packages/frontend/src/components/send/components/InputAccountId.js#L174) -- which has the other benefit of being more obvious that we _intentionally_ omitted the beginning of the string :). I'm happy to merge a PR that changes it to `does not exist while viewing` if you'd like to do that instead :)

MaximusHaximus (changes requested):
Actually, leaving off the first character here was intentional, but not obviously so.  Added a comment in-line :)

packages/frontend/src/utils/wallet.js
RNHTTR: Aha! that makes sense. I updated the PR :)

MaximusHaximus: <3 :) thanks so much!

MaximusHaximus (approved):
<3 consistency++ :)

PR #2108, opened 2021-09-13
Create and use selectSlice as top level selector

marcinbodnar:


MaximusHaximus:
I'm not sure I see value in this changeset, and actually I am a bit concerned about even having a top-level `selectSlice` selector defined at the filesystem level at all.  Philosophically, no code that consumes data from any slice should be using `selectSlice()`, because the entire idea around using selectors for everything is that the slice files themselves should export selectors and nothing _outside_ of a slice file should be selecting a top-level slice -- by doing so, we would likely be bypassing the selectors which are supposed to represent the public interface to data stored in the slice.  As such, this functionality should _only_ be used inside of a slice as the base selector for the slice.  If we make it a standalone selector we should be able to identify a benefit we are gaining by doing so, because the drawback is that it provides a top-level selector that encourages bleeding of slice names into consuming code.

Now, thinking about what we gain from this being a single method -- conceptually, it makes it obvious that all of the slices do the same thing to select their slice from the root of the `state` tree, and in theory if we ever changed how we select a slice from the root state tree, we could change it in _one_ place, but in _practice_, it is very unlikely that slices existing as root keys on the `state` object will ever change, and because this is a single line reference that is extremely easy to conceptually understand and that is not likely to change over time (it is literally `state[SLICE_NAME]`), I don't think we are losing anything by having the reference in-lined inside of each slice file; it's clear what the line does and I don't think referencing a property on an object is a concept that we need to worry about not being DRY enough.

Because there is no description on the issue describing the rationale for adding this, I am inclined to guess that this was something that was an opportunistic change motivated by a desire to be ultra-DRY?  If so, let's hold off on this until we have more than a simple direct object reference to normalize into a shared function :) (YAGNI). If you've got a driving need for this selector, let's have a chat about that (and the rationale for it would have been good to have been described in the issue or as a description on the PR :))

marcinbodnar:
@MaximusHaximus Thanks for closing it. I thought that exporting it would make code a bit simpler, but I think your right - I overdid it a bit in this case :)

PR #2117, opened 2021-09-15
Display Release Notes

marcinbodnar:


marcinbodnar:
@corwinharrell  It's ready to go. 

One thing that we need to consider for the next release notes modals is how to inform users about previous release changes. I think that if someone will log in after few months we should show him the latest release notes modal that will contain a button/link to the list of all previous release changes.

Also, at this moment in Wallet, we have a different `border-radius` for modal than in the Figma project of release notes modal, also our standard Wallet modal is a bit wider, so I left it as it is in the Wallet not to complicate changes too much.

MaximusHaximus (approved):
:+1: nice work, this looks pretty great.

PR #2148, opened 2021-09-21
Refactor recovery methods actions and reducer into redux toolkit version

marcinbodnar:


marcinbodnar:
@ihaid Could you please test this PR regarding all stuff related to recovery methods, both related to creating an account and adding/removing recovery methods from an existing account? Also 2FA. Thanks.

ihaid:
Constantly see this error starting form the /create page 
image

marcinbodnar:
> Constantly see this error starting form the /create page
> image

@ihaid This one is not related, fixed here #2154

ihaid (approved):

marcinbodnar:
Thanks for testing @ihaid

packages/frontend/src/redux/slices/recoveryMethods/index.js
```diff
@@ -0,0 +1,105 @@
...
+                
+                set(state, ['byAccountId', accountId, 'status', 'loading'], false);
+                set(state, ['byAccountId', accountId, 'status', 'error'], {
+                    message: error?.message || 'An error was encountered.',
+                    code: error?.code
+                });
+            });
+        })
+    }
+);
```
MaximusHaximus, line 76: Not a blocker for this PR, but since this is a pretty standard set of handlers, maybe we can make this a helper... what do you think about something like:

```javascript
// This in a utility file somewhere
function addAsyncThunkHandlers({ asyncThunk, builder, buildStatusPath }) {
    builder.addCase(asyncThunk.pending, (state, action) => {
        set(state, [...buildStatusPath(action), 'loading'], true);
        set(state, [...buildStatusPath(action), 'error'], initialErrorState);
    });
    builder.addCase(asyncThunk.fulfilled, (state, action) => {
        set(state, [...buildStatusPath(action), 'loading'], false);
        set(state, [...buildStatusPath(action), 'error'], initialErrorState);
    });
    builder.addCase(asyncThunk.rejected, (state, action) => {
        const { error } = action;
        set(state, [...buildStatusPath(action), 'loading'], false);
        set(state, [...buildStatusPath(action), 'error'], {
            message: error?.message || 'An error was encountered.',
            code: error?.code
        });
    });
}

const recoveryMethodsSlice = createSlice({
        name: SLICE_NAME,
        initialState,
        reducers: {
            setRecoveryMethods(state, { payload }) {
                const { recoveryMethods, accountId } = payload;
                set(state, ['byAccountId', accountId, 'items'], recoveryMethods);
            }
        },
        extraReducers: ((builder) => {
          addLoadingLifecycle({
            asyncThunk: fetchRecoveryMethods,
            builder,
            buildStatusPath: ({ meta: { arg: { accountId } } }) => (['byAccountId', accountId, 'status'])
          });
        })
    }
);
```

Do you think that would be configurable enough for our current pattern?

MaximusHaximus (approved):
Nice work!
Added two comments about things that are not blockers for this PR.

Once this is rebased to resolve the conflict, go ahead and merge :)

packages/frontend/src/redux/slices/recoveryMethods/index.js
marcinbodnar: Yes, I was thinking about a similar solution, it will be super handy if we want to use the same `status` object for all slices. I added an issue for this change #2182 , I will implement it after finishing with selectors, so the selectors would be ready to go quickly.

PR #2152, opened 2021-09-22
Revert "Fix Parse error"

MaximusHaximus:
Reverts near/near-wallet#2149

This PR broke balance loading on the main dashboard -- it spins indefinitely and never loads NEAR token balance when this code is changed.

My suspicion is that the existing logic relied on an empty string being falsy to fall back to `this.accountId`, but function signature's defaults check _existence_, not falsyness, to identify if they should be applied.

cc: @marcinbodnar

### DIFF
--- a/packages/frontend/src/utils/wallet.js
+++ b/packages/frontend/src/utils/wallet.js
@@ -720,11 +720,8 @@ class Wallet {
         return decorateWithLockup(account);
     }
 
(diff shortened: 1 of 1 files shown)


marcinbodnar:
@MaximusHaximus Thanks for catching this.
This is the proper fix: #2154 - the thing was that `accountId` is an empty string in some cases (not always `undefined`)

PR #2155, opened 2021-09-22
Coin-op 1.5: Add missing tracking

Patrick1904:
Add tracking for:
1. Moonpay events
2. Check implicit address funding status
3. Create account from implicit

### DIFF
--- a/packages/frontend/src/components/accounts/create/initial_deposit/InitialDepositWrapper.js
+++ b/packages/frontend/src/components/accounts/create/initial_deposit/InitialDepositWrapper.js
@@ -4,6 +4,7 @@ import { formatNearAmount } from 'near-api-js/lib/utils/format';
 import React, { useState, useEffect } from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 
+import { Mixpanel } from '../../../../mixpanel';
 import { createAccountFromImplicit, redirectTo } from '../../../../redux/actions/account';
 import { actions as createFromImplicitActions } from '../../../../redux/slices/createFromImplicit';
 import { actions as flowLimitationActions } from '../../../../redux/slices/flowLimitation';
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/accounts/create/verify_account/VerifyAccountWrapper.js
```diff
@@ -38,10 +39,17 @@ export function VerifyAccountWrapper() {
 
     useEffect(() => {
         const checkIfMoonPayIsAvailable = async () => {
-            const moonpayAvailable = await isMoonpayAvailable();
-            if (moonpayAvailable) {
-                setShowFundWithCreditCardOption(true);
-            }
+            await Mixpanel.withTracking("CA Check Moonpay available",
+                async () => {
+                    const moonpayAvailable = await isMoonpayAvailable();
```
MaximusHaximus, line 44: Prior to this change, if `isMoonpayAvailable()` were to throw an error, it would bubble up somewhere.
After this change, it will be swallowed by the error handle on L49-L51 and won't bubble up

Is that an intentional change in behaviour here?

MaximusHaximus (changes requested):
Just two things :+1 :)

packages/frontend/src/components/accounts/create/initial_deposit/InitialDepositWrapper.js
```diff
@@ -63,24 +64,40 @@ export function InitialDepositWrapper({ history }) {
...
+                        Mixpanel.track("CA Check balance from implicit: sufficient");
+                        setFundingNeeded(false);
+                        setInitialDeposit(state.amount);
+                        return;
+                    } else {
+                        console.log('Insufficient funding amount');
+                        Mixpanel.track("CA Check balance from implicit: insufficient");
+                    }
+                },
+                (e) => {
```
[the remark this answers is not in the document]
TiffanyGYJ: Great point, an error catching will be super helpful.

TiffanyGYJ (approved):

packages/frontend/src/components/accounts/create/initial_deposit/InitialDepositWrapper.js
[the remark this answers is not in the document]
Patrick1904: Great catch @MaximusHaximus - fixed that in `SetupImplicit.js` too.

https://github.com/near/near-wallet/pull/2155/commits/c4e27e3a753024c05a308d8f376f7b71d068c6fb

packages/frontend/src/components/accounts/create/verify_account/VerifyAccountWrapper.js
Patrick1904: Changed it to `throw e` :)

packages/frontend/src/components/accounts/create/verify_account/VerifyAccountWrapper.js
```diff
@@ -38,10 +39,17 @@ export function VerifyAccountWrapper() {
...
+                    if (moonpayAvailable) {
+                        setShowFundWithCreditCardOption(true);
+                    }
+                },
+                (e) => {
+                    throw e;
+                }
```
MaximusHaximus, line 51:
```suggestion
                }
```

re-throwing from the `errorOperation` with no other logic is equivalent to just not defining one at all :)

packages/frontend/src/components/accounts/create/initial_deposit/InitialDepositWrapper.js
```diff
@@ -63,24 +64,45 @@ export function InitialDepositWrapper({ history }) {
...
+                    } catch(e) {
+                        if (e.message.includes('does not exist while viewing')) {
+                            return;
+                        }
+                        throw e;
+                    }
+                },
+                (e) => {
+                    throw e;
+                }
```
MaximusHaximus, line 90:
```suggestion
                }
```
re-throwing from the errorOperation with no other logic is equivalent to just not defining one at all :)

MaximusHaximus (approved):
:+1: Looks good - nothing blocking, just a couple of notes -- for future usage of mixpanel.withTracking that defining an `errorOperation` function that only re-throws is redundant :)

PR #2166, opened 2021-09-27
[WIP] Account creation: fund w/ existing account

Patrick1904:


corwinharrell (changes requested):
Looking good. Just a couple small things:

1. Let's hide the account balances by default within the new account selector component.
2. Do we want to update the current account selector component with the new one?

corwinharrell (approved):

ihaid (approved):

MaximusHaximus (changes requested):
Per our conversation, let's :hocho: the FormButtonGroup component and use a styled div in its place -- a couple of other small things but this looks ready to rock and roll :D

PR #2173, opened 2021-09-30
E2e ci pagerduty integration

esaminu:
Add a pagerduty reporter to trigger incidents an pagerduty whenever a test fails.


esaminu:
Moved to https://github.com/near/near-wallet/pull/2174

PR #2178, opened 2021-10-02
Selectors for component from account folder

marcinbodnar:
@MaximusHaximus I decided that the easiest for me will be making changes component by component, and adding necessary selectors if needed. This PR is for selectors needed in /account folder components.

There are some selectors already used in other PR that are waiting for review.


marcinbodnar:
@MaximusHaximus Done.
@ihaid Thanks for testing.

Issue #2071, opened 2021-08-31
Transfer NFTs


think-in-universe:
## Story

As a NEAR user who owns a number of NFTs, I want to transfer one of my NFTs to other users so I can enable more scenarios regarding NFTs.

## Acceptance Criteria

- [ ] In the `Collectibles` tab, provide an button or link to transfer the specific NFT
- [ ] A NFT transfer workflow is designed to make the transfer process easy to use
- [ ] Mobile ok
- [ ] Tested

## Assets

- Mockups: https://www.figma.com/file/Co5Rx8i9Cx4SxgRZv0MUa7/NFTs?node-id=426%3A387

## NOTES

- The NFTs transfer feature is frequently requested by not only NFTs users but also developers.

corwinharrell:
@think-in-universe I've updated our Figma mockups to show how new might go about implementing a flow for transferring NFTs: https://www.figma.com/file/Co5Rx8i9Cx4SxgRZv0MUa7/NFTs?node-id=426%3A387

A couple things to note:

1. Each NFT should ideally have a details page that displays a larger preview image and key metadata. Initially, this would just be title, description, and creator/owner IDs, but this would evolve to show different characteristics of the NFT, such as unique traits and rarity.

2. From this page, the user can choose to Transfer the NFT. They would then enter the recipient address, and would proceed to confirm the transaction.

3. Once complete, they would be redirected the collectibles tab on the dashboard.

Content in the mockups is placeholder. @think-in-universe if you have any questions or feedback about any of this just let me know, but for the most part, the flow is fairly similar to a fungible token transfer. Questions about the location and specific components used here should be directed to @Patrick1904. Aside from the NFT details page, all the components here should be pre-existing.

Patrick1904:
@think-in-universe you'll be able to re-use some of the UI that's being shipped here: https://github.com/near/near-wallet/pull/2166

Please ping me if you have any questions.

danielwpz:
The design looks good to me! However two questions need to be clarified before we are good to develop.
1. How could we find the creator ID ? It's currently not in the metadata. Since we don't have a standardized `mint` function, it could be hard to tell who firstly created the NFT.
2.  There are two external links in the `Proof of Authenticity` section (NEAR transaction and View on IPFS). What pages should they link to? And I suppose the `View on IPFS` link should only appear if `reference` exists in NFT metadata?

danielwpz:
PR: https://github.com/near/near-wallet/pull/2205

[...]

PR #2205, opened 2021-10-18
NFT transfer

danielwpz:
implemented: https://github.com/near/near-wallet/issues/2071


danielwpz:
hey guys, just updated this PR, please take a second look.

And there are several differences between figma design and existing components, like:
1. The NFT transfer modal width in figma is 434px, however the current modal doesn't have any presets matching this. Shall I add new `mid-large` size??
2. The  receiver id input box design doesn't match existing component `ReceiverInputWithLabel`, changing the component to match design requires many efforts and may affect other places. I would suggest to stick to the current one.          

For me I'll try to reuse existing components as much as possible, unless they cannot meet the UX requirement.

Also, the `Proof of Authenticity` section has two links, where should they link to? Not all NFTs are stored on IPFS I think, should we keep the `View on IPFS` link?

Patrick1904:
> hey guys, just updated this PR, please take a second look.
> 
> And there are several differences between figma design and existing components, like:
> 
> 1. The NFT transfer modal width in figma is 434px, however the current modal doesn't have any presets matching this. Shall I add new `mid-large` size??
> 2. The  receiver id input box design doesn't match existing component `ReceiverInputWithLabel`, changing the component to match design requires many efforts and may affect other places. I would suggest to stick to the current one.
> 
> For me I'll try to reuse existing components as much as possible, unless they cannot meet the UX requirement.
> 
> Also, the `Proof of Authenticity` section has two links, where should they link to? Not all NFTs are stored on IPFS I think, should we keep the `View on IPFS` link?

Thanks for the updates @danielwpz! Great job! 👍 

> 1. The NFT transfer modal width in figma is 434px, however the current modal doesn't have any presets matching this. Shall I add new `mid-large` size??

The current size is fine. No need to add a new size

> 2. The  receiver id input box design doesn't match existing component `ReceiverInputWithLabel`, changing the component to match design requires many efforts and may affect other places. I would suggest to stick to the current one.

Good point - I think the current one is fine

> For me I'll try to reuse existing components as much as possible, unless they cannot meet the UX requirement.

Perfect! The more we can re-use, the better :)

> Also, the `Proof of Authenticity` section has two links, where should they link to? Not all NFTs are stored on IPFS I think, should we keep the `View on IPFS` link?

You're right - the links aren't necessary for this first version

--------------------------

For some reason I still don't get the images rendered on the NFT details page or the modal - I don't think it's my machine because they render fine on the NFT collection page. Any idea why that is?

Can we hide the ’Transfer’ button if it’s not owner looking at the NFT details page? I noticed that when I switch account on an NFT details page, the button is still there, though disabled. I think we can remove it completely in that case.

Can we use the `<BackArrowButton>` under `/common/BackArrowButton` for the NFT details page? We already use it on the send token 'enter receiver' step (https://wallet.near.org/send-money)

I think we'll need to re-fetch NFTs after sending so that both sender and receiver have the latest state. Right now the NFT is still visible in the senders collection after sending. Also, once the NFT has been sent and the user clicks the 'Continue' button, can we send the user back to the full NFT collection page vs staying on the NFT details page? 

We remove the ‘Confirm transfer’ heading on the confirmation step - The Figma might have been updated after (https://www.figma.com/file/Co5Rx8i9Cx4SxgRZv0MUa7/NFTs?node-id=426%3A387)

packages/frontend/src/components/nft/NFTDetail.js
```diff
@@ -0,0 +1,225 @@
...
+        .desc {
+            margin-top: 48px;
+            margin-bottom: 56px;
+
+            font-family: Inter;
+            font-style: normal;
+            font-weight: 500;
+            font-size: 16px;
+            line-height: 150%;
+            /* or 24px */
```
Patrick1904, line 44: Can we please remove the reference comments in the CSS? Including `/* gray/neutral/800 */` etc.

packages/frontend/src/components/nft/NFTDetail.js
```diff
@@ -0,0 +1,225 @@
...
+
+    .owner {
+        p {
+            font-family: Inter;
+            font-style: normal;
+            font-weight: 500;
+            font-size: 12px;
+            line-height: 150%;
+            /* identical to box height, or 18px */
+
```
Patrick1904, line 63: Let's avoid unnecessary spaces between CSS declarations when possible :)

packages/frontend/src/components/nft/NFTDetail.js
```diff
@@ -0,0 +1,225 @@
...
+                /* gray/neutral/800 */
+
+                color: #272729;
+            }
+        }
+    }
+
+    .transfer-btn {
+        width: 100%;
+        margin-top: 56px !important;
```
Patrick1904, line 99: Let's wrap the style in `&&& { ... }` for added specificity instead of using`!important`. You can see an example how it's used in the [`AlertBanner`](https://github.com/near/near-wallet/blob/master/packages/frontend/src/components/common/AlertBanner.js#L12)

packages/frontend/src/components/nft/NFTDetail.js
```diff
@@ -0,0 +1,225 @@
...
+
+        svg {
+            margin-right: 10px !important;
+        }
+    }
+
+    #back-btn {
```
Patrick1904, line 106: Let's use `className` vs `id` when possible

packages/frontend/src/components/nft/NFTDetail.js
```diff
@@ -0,0 +1,225 @@
...
+    useEffect(() => {
+        NonFungibleTokens.getToken(contractId, tokenId)
+            .then(token => {
+                token.contract_id = contractId;
+                setNft(token);
+                setOwnerId(token.owner_id);
+            });
+    }, []);
+
+    function getNEARBalance () {
```
Patrick1904, line 169: Is there a reason why we need a function to return `nearBalance` vs referencing the variable directly?

packages/frontend/src/components/nft/NFTDetail.js
```diff
@@ -0,0 +1,225 @@
...
+import NonFungibleTokens from '../../services/NonFungibleTokens';
+import FormButton from '../common/FormButton';
+import Container from '../common/styled/Container.css';
+import ArrowIcon from '../svg/ArrowIcon';
+import NFTTransferModal from './NFTTransferModal';
+
+const StyledContainer = styled(Container)`
+    .container {
+        margin-left: auto;
+        margin-right: auto;
```
Patrick1904, line 18: I believe `../common/styled/Container.css` should be enough to center the container? You can add additional variations to `Container.css` if needed so that we can re-use it at a later point if needed. can probably remove the `.container` `div` afterwards

packages/frontend/src/components/nft/NFTDetail.js
```diff
@@ -0,0 +1,225 @@
...
+        }
+
+        .inner {
+            height: 30px;          
+            line-height: 30px; 
+            display: flex;         
+            align-items: center;
+
+            span {
+                font-family: Inter;
```
Patrick1904, line 80: We shouldn't have to add `font-family: Inter;` anywhere since it's already the default font family defined in our `GlobalStyle`

packages/frontend/src/components/nft/NFTDetail.js
```diff
@@ -0,0 +1,225 @@
...
+                font-style: normal;
```
Patrick1904, line 81: Same with `font-style: normal;` as it's already default

packages/frontend/src/components/nft/NFTDetail.js
```diff
@@ -0,0 +1,225 @@
...
+        height: 26px;
+        margin: 7px;
+    }
+
+    @media (min-width: 940px) {
+        display: inline-block;
+    }
+`;
+
+function arrowSVG () {
```
Patrick1904, line 140: We can remove this once we use the `<BackArrowButton>`

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,450 @@
...
+
+
+const StyledContainer = styled.div`
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    padding: 40px 0 0 0;
+
+    #nft-transfer-modal {
+        width: 400px;
```
Patrick1904, line 29: Can we avoid defining a width here since we've already added a `modalSize` as a prop for the `Modal` component?

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,450 @@
...
+        > p {
+            font-family: Inter;
+            font-style: normal;
+            font-weight: 500;
+            font-size: 20px;
+            line-height: 150%;
+            /* or 30px */
+            
+            text-align: center;
+            font-feature-settings: 'zero' on;
```
Patrick1904, line 157: It seems we should add `font-feature-settings: 'zero' on;` to `GlobalStyle` instead since we want this site-wide?

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,450 @@
...
+            color: #3F4045;
+
+            margin-bottom: 0px;
+        }
+    }
+
+    p {
+        font-family: Inter;
+        font-style: normal;
+        font-weight: 500;
```
Patrick1904, line 170: I think `font-weight: 500;` is wrong in the Figma unfortunately. Can we please remove it everywhere so that it's 'normal' weight.

packages/frontend/src/translations/en.global.json
```diff
@@ -912,6 +912,24 @@
         "loadMore": "Load more"
     },
 
+    "NFTDetail": {
+        "owner": "OWNER",
+        "transfer": "Transfer"
+    },
+
+    "NFTTransfer": {
+      "transfer-nft": "Transfer NFT",
```
Patrick1904, line 921: Can we please keep translation IDs camel cased for consistency sake; `enter-receipt` --> `enterReceipt`, etc.

packages/frontend/src/components/nft/NFTTransferModal.js
Patrick1904: Actually please omit `font-feature-settings: 'zero' on;` completely as it seems @corwinharrell removed this style feature from the `GlobalStyle` not too long ago: https://github.com/near/near-wallet/commit/62c86e25362dc68d43487dbf84939d80918d649b

danielwpz:
@Patrick1904 hi Patrick, thanks for all the feedbacks. will push an update today!

packages/frontend/src/components/nft/NFTDetail.js
danielwpz: oh actually this is for the arrow icon on transfer button.

danielwpz:
@Patrick1904 what's the best way to trigger a reload on NFT list page? shall I add a url param indicates whether it should reload or not?

Patrick1904:
> @Patrick1904 what's the best way to trigger a reload on NFT list page? shall I add a url param indicates whether it should reload or not?

@danielwpz you can call `dispatch(fetchNFTs({ accountId }));` after successfully sending the NFT so that the 'transfer' button disappears and all current NFTs are updated when the user navigates back to the gallery

packages/frontend/src/components/nft/NFTDetail.js
Patrick1904: I see. Please use the existing [`<SendIcon/>`](https://github.com/near/near-wallet/blob/master/packages/frontend/src/components/wallet/Wallet.js#L383) then and adjust style if needed

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+    const balanceToShow = formatNearAmount(nearBalance);
+    const dispatch = useDispatch();
+    const { fetchNFTs } = nftActions;
+
+    const { localAlert } = useSelector(({ status }) => status);
+
+    function onTransferSuccess(result, newOwnerId) {
+        setResult(result);
+        setOwnerId(newOwnerId);
+        dispatch(fetchNFTs({ accountId }));
```
danielwpz, line 223: @Patrick1904 am I doing this correctly?

danielwpz: the NFT list is not updated :(

Patrick1904: @danielwpz you're right. @marcinbodnar prepared a `updateNFTs` function for you here https://github.com/near/near-wallet/pull/2272.

You can pass in the `contractName` :)

packages/frontend/src/components/nft/NFTDetail.js
```diff
@@ -0,0 +1,196 @@
...
+export function NFTDetail({ match, location, history }) {
+    const accountId = useSelector(selectAccountId);
+    const fungibleTokens = useFungibleTokensIncludingNEAR();
+    const nearBalance = fungibleTokens[0].balance;
+
+    const contractId = match.params.contractId;
+    const tokenId = match.params.tokenId;
+    const [ nft, setNft ] = useState();
+    const [ ownerId, setOwnerId ] = useState();
+    const [transferNftDetail, setTransferNftDetail] = useState();
```
Patrick1904, line 132: You can use the `contractId` and `tokenId` from the url param and check if the NFT is part of the users list of NFTs so that you don't have to make the call below and then set with `setOwnerId`. 

You can use the `const sortedNFTs = useSelector((state) => selectTokensWithMetadataForAccountId(state, { accountId }));` used in `Wallet.js`

packages/frontend/src/components/nft/NFTDetail.js
```diff
@@ -0,0 +1,196 @@
...
+          <path d="M21 1L14 21L10 12L1 8L21 1Z" stroke="#A2A2A8" strokeWidth="2" strokeLinecap="round" strokeLinejoin="round"/>
+          <path d="M21 1L10 12" stroke="#A2A2A8" strokeWidth="2" strokeLinecap="round" strokeLinejoin="round"/>
+        </svg>
+    );
+}
+
+export function NFTDetail({ match, location, history }) {
+    const accountId = useSelector(selectAccountId);
+    const fungibleTokens = useFungibleTokensIncludingNEAR();
+    const nearBalance = fungibleTokens[0].balance;
```
Patrick1904, line 126: You can use `selectBalance` here like how it's used in [`Wallet.js`](https://github.com/near/near-wallet/blob/master/packages/frontend/src/components/wallet/Wallet.js#L267)

Something like:

`const balance = useSelector(state => selectBalance(state));`
`balance.balanceAvailable` is the balance you want to display on the UI

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+`;
+
+export default function NFTTransferModal({ open, onClose, nft, accountId, setOwnerId }) {
+    const [ receiverId, setReceiverId ] = useState('');
+    const [ result, setResult ] = useState();
+    const [ sending, setSending ] = useState(false);
+    const [ viewType, setViewType ] = useState('transfer');
+    const [ accountIdIsValid, setAccountIdIsValid] = useState(false);
+    const fungibleTokens = useFungibleTokensIncludingNEAR();
+    const nearBalance = fungibleTokens[0].balance;
```
Patrick1904, line 213: Let's use the selector here too: https://github.com/near/near-wallet/pull/2205/files#r758852017

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+                        <div className='confirm-img'>
+                            <img src={nft.metadata.mediaUrl} alt='NFT'/>
+                        </div>
+
+                        <div className='line'></div>
+                        <div className='from-box'>
+                            <span className='confirm-txt v-center'><Translate id='transfer.from'/></span>
+                            <span className='h-right v-center'>
+                                <span className='account-id'>{accountId}</span>
+                                <div className='amount-grey'>{balanceToShow} NEAR</div>
```
Patrick1904, line 317: Please use the[ `<Balance/>`](https://github.com/near/near-wallet/blob/master/packages/frontend/src/components/wallet/Wallet.js#L365) component wherever we show token amounts

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+
+export default function NFTTransferModal({ open, onClose, nft, accountId, setOwnerId }) {
+    const [ receiverId, setReceiverId ] = useState('');
+    const [ result, setResult ] = useState();
+    const [ sending, setSending ] = useState(false);
+    const [ viewType, setViewType ] = useState('transfer');
+    const [ accountIdIsValid, setAccountIdIsValid] = useState(false);
+    const fungibleTokens = useFungibleTokensIncludingNEAR();
+    const nearBalance = fungibleTokens[0].balance;
+    const balanceToShow = formatNearAmount(nearBalance);
```
Patrick1904, line 214: You won't need to use `formatNearAmount` once you use the `<Balance/>` component mentioned here: https://github.com/near/near-wallet/pull/2205/files#r758853723

packages/frontend/src/components/svg/ArrowIcon.js
```diff
@@ -2,12 +2,13 @@ import React from 'react';
 
 const ArrowIcon = ({ color }) => {
     const stroke = color || '#0072CE';
+
     return (
-        <svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+        <svg width='24' height='24' viewBox='0 0 24 24' fill="none" xmlns="http://www.w3.org/2000/svg">
```
Patrick1904, line 7: nit: can we please skip this change to keep it consistent

packages/frontend/src/components/wallet/NFTBox.js
```diff
@@ -130,18 +157,23 @@ const NFTBox = ({ tokenDetails }) => {
             {
                 ownedTokensMetadata &&
                 <div className='tokens'>
-                    {ownedTokensMetadata.map(({ token_id, metadata: { mediaUrl, title } }) => {
+                    {ownedTokensMetadata.map(({ token_id, metadata }) => {
+                        const { mediaUrl, title } = metadata;
+
                         return <div className='nft' key={token_id}>
```
Patrick1904, line 163: Can we please add the `onClick` to the `div` instead of two separate ones for the `img` and title `div`. 

Also, let's use the `redirectTo` action instead of `history.push` :)

Patrick1904: Can we also wrap the return result in a parenthesis so that's it's easier to see what's included:)

packages/frontend/src/components/nft/NFTDetail.js
```diff
@@ -0,0 +1,196 @@
...
+                </div>
+
+                {(ownerId === accountId) &&
+                  <FormButton 
+                    className='transfer-btn'
+                    color='gray-gray' 
+                    disabled={ownerId !== accountId || !nearBalance}
+                    onClick={() => setTransferNftDetail(nft)}
+                >
+                    {transferSVG()}
```
Patrick1904, line 179: Let's re-use the [`<SendIcon/>`](https://github.com/near/near-wallet/blob/master/packages/frontend/src/components/wallet/Wallet.js#L383) here instead :)

packages/frontend/src/components/nft/NFTDetail.js
```diff
@@ -0,0 +1,196 @@
...
+        height: 26px;
+        margin: 7px;
+    }
+
+    @media (min-width: 940px) {
+        display: inline-block;
+    }
+`;
+
+function transferSVG() {
```
Patrick1904, line 114: See this: https://github.com/near/near-wallet/pull/2205/files#r758863811

packages/frontend/src/components/nft/NFTDetail.js
```diff
@@ -0,0 +1,196 @@
+import React, { useEffect, useState } from 'react';
```
Patrick1904, line 1: It would be really sweet if you wanted to create a `NFTDetailWrapper` to separate async behavior. Basically keep the actions and selectors in the wrapper. You can see how it was done recently here for the app login flow with `ConfirmLoginWrapper` and `ConfirmLogin` component as a child: https://github.com/near/near-wallet/tree/master/packages/frontend/src/components/login/v2

packages/frontend/src/components/Routing.js
```diff
@@ -329,7 +330,7 @@ class Routing extends Component {
                             <GuestLandingRoute
                                 exact
                                 path='/'
-                                render={(props) => isInactiveAccount ? <ActivateAccountWithRouter {...props}/> : <Wallet tab={tab} setTab={setTab} {...props} />}
+                                render={(props) => isInactiveAccount ? <ActivateAccountWithRouter {...props}/> : <Wallet tab={tab} setTab={setTab} history={this.props.history} {...props} />}
```
Patrick1904, line 333: Won't be needed anymore:) https://github.com/near/near-wallet/pull/2205/files#r758857643

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+        border-radius: 8px;
+        margin-bottom: 16px;
+    }
+
+    .transfer-txt {
+        margin-top: 16px;
+    }
+
+    .confirm-txt {
+        margin-bottom: 4px !important;
```
Patrick1904, line 41: Let's make sure we remove all usage of `!important` before we merge. You can add `&&& {...}` on `L24` to get the same specificity

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+        }
+    }
+
+    .success-bottons {
+        align-items: center;
+        margin-left: auto;
+        margin-right: auto;
+
+        > button {
+            width: 185px !important;
```
Patrick1904, line 201: I don't think we can set a fixed width for the buttons here since they need to be able to shrink on a mobile device. You can use similar style as for the explore apps box as shown here: https://github.com/near/near-wallet/blob/master/packages/frontend/src/components/wallet/ExploreApps.js#L41

packages/frontend/src/components/wallet/NFTBox.js
```diff
@@ -95,33 +99,56 @@ const StyledContainer = styled.div`
...
+        span {
+            font-family: Inter;
+            font-style: normal;
+            font-weight: 500;
+            font-size: 14px;
+            line-height: 150%;
+            /* identical to box height, or 21px */
+            
+            
+            /* gray/neutral/500 */
```
Patrick1904, line 118: Let's not forget to remove all comments in the CSS :)

packages/frontend/src/components/wallet/NFTBox.js
```diff
@@ -95,33 +99,56 @@ const StyledContainer = styled.div`
...
+        a {
+            color: inherit; /* blue colors for links too */
+        }
+    }
+
+    .creator {
+        span {
+            font-family: Inter;
+            font-style: normal;
+            font-weight: 500;
```
Patrick1904, line 112: We can remove all usage of:

```
font-family: Inter;
font-style: normal;
font-weight: 500;
```

packages/frontend/src/components/nft/NFTTransferModal.js
danielwpz: sure, will do

packages/frontend/src/components/svg/ArrowIcon.js
danielwpz: kk

packages/frontend/src/components/nft/NFTDetail.js
danielwpz: will do

packages/frontend/src/components/nft/NFTDetail.js
Patrick1904: That's a good point. I suggest that we check if there's a token list present when the user lands on this page with `useEffect`, and if there's none, dispatch `fetchNFTs` to get the list. It's better if we use one source of truth for the users NFTs to avoid potential issues.

danielwpz: hmm actually I don't think this is a good option for two reasons:
1. the URL could be opened by anybody, not just the NFT owner, so his NFT list doesn't necessarily have this one
2. load the entire NFT list costs too much in order to display one single NFT

how do you think

packages/frontend/src/services/NonFungibleTokens.js
```diff
@@ -45,30 +55,54 @@ export default class NonFungibleTokens {
...
+          .map(token => mapTokenMediaUrl(token, base_uri));
 
         return tokens;
     }
+
+    static transfer = async ({ accountId, contractId, tokenId, receiverId }) => {
+        console.log('transfer()', accountId, contractId, tokenId, receiverId);
+        const account = await wallet.getAccount(accountId);
+
+        return account.signAndSendTransaction({
```
danielwpz, line 69: got an error when calling this:
```
TypeError: First argument must be a string, Buffer, ArrayBuffer, Array, or array-like object.
    at fromObject (index.js:311)
    at from (index.js:137)
    at Function.Buffer.from (index.js:149)
    at BinaryWriter.writeString (index.js:99)
    at serializeField (index.js:219)
    at index.js:265
    at Array.map (<anonymous>)
    at serializeStruct (index.js:264)
    at Object.serialize (index.js:287)
    at signTransactionObject (transaction.js:214)
```

any idea what's going wrong?

packages/frontend/src/services/NonFungibleTokens.js
```diff
@@ -45,30 +55,54 @@ export default class NonFungibleTokens {
...
+            receiverId: contractId,
+            actions: [
+                functionCall(
+                    'nft_transfer', 
+                    {
+                        receiver_id: receiverId,
+                        token_id: tokenId
+                    },
+                    NFT_TRANSFER_GAS,
+                    1
```
Patrick1904, line 79: Let's declare this as a variable so that it's clear what it's for

packages/frontend/src/services/NonFungibleTokens.js
Patrick1904: I'm not able to reproduce this. Can you please provide some more context?

danielwpz: turns out I forgot to run yarn install lol

danielwpz:
@Patrick1904 hey, all comments are resolved now. 

However when I rebased master I could not run it. Seems some feature flag env is missing if just do yarn start

packages/frontend/src/components/nft/NFTDetail.js
```diff
@@ -0,0 +1,164 @@
...
+            line-height: 150%;
+            display: flex;
+            align-items: center;
+            letter-spacing: 0.115em;
+            color: #A2A2A8;
+        }
+
+        .inner {
+            height: 30px;          
+            line-height: 30px;
```
Patrick1904, line 56:
```suggestion
            
```
Let's cut this :)

packages/frontend/src/components/nft/NFTDetail.js
```diff
@@ -0,0 +1,164 @@
...
+            letter-spacing: 0.115em;
+            color: #A2A2A8;
+        }
+
+        .inner {
+            height: 30px;          
+            line-height: 30px; 
+            display: flex;         
+            align-items: center;
+
```
Patrick1904, line 59:
```suggestion
            word-break: break-all;
```

Let's add `break-all` so that long account ID's don't overflow the screen width (esp. on mobile)

packages/frontend/src/components/nft/NFTDetail.js
```diff
@@ -0,0 +1,164 @@
...
+                >
+                    <SendIcon/>
+                    <Translate id='NFTDetail.transfer'/>
+                </FormButton>}
+                {transferNftDetail &&
+                    <NFTTransferModal
+                        open={!!transferNftDetail}
+                        onClose={() => setTransferNftDetail()}
+                        nft={transferNftDetail}
+                        accountId={accountId}>
```
Patrick1904, line 156:
```suggestion
                        accountId={accountId}
```

packages/frontend/src/components/nft/NFTDetail.js
```diff
@@ -0,0 +1,164 @@
...
+                        nearBalance={ nearBalance }
```
Patrick1904, line 157:
```suggestion
                        nearBalance={nearBalance}
```

packages/frontend/src/components/nft/NFTDetailWrapper.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { useEffect, useState } from "react";
+import { useSelector } from "react-redux";
+
+import { selectAccountId, selectBalance } from "../../redux/slices/account";
+import NonFungibleTokens from "../../services/NonFungibleTokens";
+import { NFTDetail } from "./NFTDetail";
+
+export function NFTDetailWrapper ({
```
Patrick1904, line 9: Thanks for adding a wrapper 🙏

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+import Balance from '../common/balance/Balance';
+import FormButton from '../common/FormButton';
+import Modal from '../common/modal/Modal';
+import ModalFooter from '../common/modal/ModalFooter';
+import SafeTranslate from '../SafeTranslate';
+import ReceiverInputWithLabel from '../send/components/ReceiverInputWithLabel';
+import AvatarSuccessIcon from '../svg/AvatarSuccessIcon';
+import EstimatedFees from '../transfer/EstimatedFees';
+
+
```
Patrick1904, line 22:
```suggestion

```

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+const StyledContainer = styled.div`
+    &&& {
+        display: flex;
+        flex-direction: column;
+        align-items: center;
+        padding: 0 0 0 0;
```
Patrick1904, line 28:
```suggestion
        padding: 0;
```

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+
+        img {
+            width: 100%  ;
+            max-width: 472px;
```
Patrick1904, line 32:
```suggestion

```

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+            border-radius: 8px;
+            margin-bottom: 16px;
```
Patrick1904, line 34:
```suggestion
            margin: 16px 0;
```

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+        }
+
+        .transfer-txt {
+            margin-top: 16px;
+        }
+
+        .confirm-txt {
+            margin-bottom: 4px  ;
```
Patrick1904, line 42:
```suggestion
            margin-bottom: 4px; 
```

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+        .transfer-txt {
+            margin-top: 16px;
+        }
+
+        .confirm-txt {
+            margin-bottom: 4px  ; 
+        }
+
+        .confirm-img {
+            width: 272px;
```
Patrick1904, line 46:
```suggestion
            max-width: 272px;
```

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+    const [ viewType, setViewType ] = useState('transfer');
+    const [ accountIdIsValid, setAccountIdIsValid] = useState(false);
+    const balance = useSelector(state => selectBalance(state));
+    const nearBalance = balance.balanceAvailable;
+    const dispatch = useDispatch();
+    const { updateNFTs } = nftActions;
+
+    const { localAlert } = useSelector(({ status }) => status);
+
+    function onTransferSuccess(result, newOwnerId) {
```
Patrick1904, line 221: We can 🔪 `newOwnerId` :)

Patrick1904, line 221:
```suggestion
    function onTransferSuccess(result) {
```

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+    return (
+        <Modal
+            id='nft-transfer-modal'
+            isOpen={nft}
+            onClose={onClose}
+            closeButton={false}
+            modalSize='md'
+        >
+            {viewType === 'transfer' &&
+            <StyledContainer className='small-centered'>
```
Patrick1904, line 260:
```suggestion
            <StyledContainer>
```
`small-centered` isn't doing anything here

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+                            <Translate id='NFTTransfer.next'/>
+                            </FormButton>
+                        </div>
+                    </ModalFooter>
+                </form>
+            </StyledContainer>
+            }
+
+            {viewType === 'confirm' &&
+                <StyledContainer className='small-centered'>
```
Patrick1904, line 304:
```suggestion
                <StyledContainer>
```
`small-centered` isn't doing anything here

packages/frontend/src/components/wallet/NFTBox.js
```diff
@@ -55,7 +57,6 @@ const StyledContainer = styled.div`
         margin-left: 14px;
 
         a {
-            font-weight: 700;
```
Patrick1904, line 58: Can we please revert this change

packages/frontend/src/components/wallet/NFTBox.js
```diff
@@ -64,7 +65,6 @@ const StyledContainer = styled.div`
             color: #72727A;
             background-color: #F0F0F1;
             font-size: 14px;
-            font-weight: 600;
```
Patrick1904, line 67: Same here, let's revert this change

packages/frontend/src/services/NonFungibleTokens.js
```diff
@@ -45,30 +55,54 @@ export default class NonFungibleTokens {
...
+        tokens = tokens
+          .filter(({ metadata }) => !!metadata)
+          .map(token => mapTokenMediaUrl(token, base_uri));
 
         return tokens;
     }
+
+    static transfer = async ({ accountId, contractId, tokenId, receiverId }) => {
+        const account = await wallet.getAccount(accountId);
+        const nftTransferDeposit = 1;
```
Patrick1904, line 67: nit: can we please add this as `NFT_TRANSFER_DEPOSIT` at the top of the file below `NFT_TRANSFER_GAS` :)

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+        .confirm-img {
+            width: 272px;
+        }
+
+        .confirm-nft-card {
+            display: flex;
+            flex-direction: column;
+            align-items: center;
+            position: relative;
+
```
Patrick1904, line 54:
```suggestion

```

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+            width: 100%;
+            margin-top: 16px;
+
```
Patrick1904, line 57:
```suggestion

```

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+                border-top: 1px solid #F0F0F1;
+            }
+
+        }
+
+        .confirm-txt {
+            font-size: 14px;
+            line-height: 150%;
+            color: #72727A;
+            margin-left: 16px  ;
```
Patrick1904, line 82:
```suggestion
            margin-left: 16px;
```

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+            position: absolute;
+            right: 16px;
+            text-align: right;
+        }
+
+        .estimate-fee-card {
+            display: flex;
+            flex-direction: column;
+            position: relative;
+
```
Patrick1904, line 96:
```suggestion

```

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+            width: 100%;
+            height: 78px;
+            margin-top: 16px;
+
```
Patrick1904, line 100:
```suggestion

```

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+                        <div className='line'></div>
+                        <div className='to-box'>
+                            <span className='confirm-txt v-center'><Translate id='transfer.to' /></span>
+                            <span className='h-right v-center account-id'>{receiverId}</span>
+                        </div>
+                    </div>
+
+                    <div className='estimate-fee-card'>
+                        <EstimatedFees gasFeeAmount={NFT_TRANSFER_GAS}/>
+                    </div>
```
Patrick1904, line 329:
```suggestion
                    <EstimatedFees gasFeeAmount={NFT_TRANSFER_GAS}/>
```

No need to wrap and style :)

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+            position: relative;
+
+            width: 100%;
+            height: 78px;
+            margin-top: 16px;
+
+            border: 1px solid #F0F0F1;
+            box-sizing: border-box;
+            border-radius: 8px;
+        }
```
Patrick1904, line 104:
```suggestion
        .estimated-fees {
            width: 100%;
            margin-top: 20px;
        }
```

Shouldn't apply same style twice:) See comment: https://github.com/near/near-wallet/pull/2205/files#r769122247

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+
+        .success {
+            > p {
+                font-weight: 500;
+                font-size: 20px;
+                line-height: 150%;
+                text-align: center;
+                color: #3F4045;
+
```
Patrick1904, line 129:
```suggestion

```

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+            color: #72727A;
+        }
+
+        .next-btn {
+            margin-left: 44px  ;
+        }
+
+        .receiver-input {
+        width: 100%;
+        }
```
Patrick1904, line 148:
```suggestion
        .receiver-input {
            width: 100%;
        }
```

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+            display: flex;
+            align-items: right;
+            justify-content: flex-end;
+            padding-top: 0px;
+            padding-bottom: 0px;
+
+            button {
+                width: 136px  ;
+                height: 56px  ;
+                margin-top: 0px  ;
```
Patrick1904, line 186:
```suggestion
                width: 136px;
                height: 56px;
                margin-top: 0px;
```

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+
+                    <ModalFooter>
+                        <div className='buttons'>
+                            <FormButton
+                                className='link'
+                                type='button'
+                                onClick={onClose}
+                                color='gray'
+                            >
+                            <Translate id='NFTTransfer.cancel'/>
```
Patrick1904, line 287:
```suggestion
                                <Translate id='NFTTransfer.cancel'/>
```

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,397 @@
...
+                            </FormButton>
+                            <FormButton
+                                className='next-btn'
+                                type='submit'
+                                disabled={!accountIdIsValid}
+                                onClick={() => setViewType('confirm')}
+                            >
+                            <Translate id='NFTTransfer.next'/>
```
Patrick1904, line 295:
```suggestion
                                <Translate id='NFTTransfer.next'/>
```

packages/frontend/src/translations/en.global.json
```diff
@@ -530,6 +530,21 @@
...
+    "NFTTransfer": {
+      "transferNft": "Transfer NFT",
+      "enterReceipt": "Enter a recipient address, then proceed to confirm your transaction.",
+      "cancel": "Cancel",
+      "next": "Next",
+      "confirm": "Confirm",
+      "transactionComplete": "Transaction Complete!",
+      "viewReceipt": "View Receipt",
+      "continue": "Continue",
+      "youSent": "You sent <span class='color-blue'>${title}</span> to <span class='color-green'>${receiverId}</span>"
```
Patrick1904, line 546: There's a bug with the translation package we're using where automatic fallback to english isn't working whenever the translation string contains `data`. The current solution is to just copy the added translation strings in this `en.global.json` file and paste it in the other `json` files under `/translations`. Can you please do that in this PR? 🙏

Patrick1904:
> @Patrick1904 hey, all comments are resolved now.
> 
> However when I rebased master I could not run it. Seems some feature flag env is missing if just do yarn start

Thanks for making the changes @danielwpz 🙏 

This is looking good to me! I just added a few small change requests inline :)

packages/frontend/src/components/nft/NFTTransferModal.js
danielwpz: is this setting the margin bottom or top?

Patrick1904: Let's set 16px top/bottom here as suggested

packages/frontend/src/translations/en.global.json
danielwpz: done, please check if I did this correctly

danielwpz:
sorry for the late update, didn't get a chance to update this PR  last few days.
@Patrick1904 fixed all latest comments, if it looks good I'll rebase master and make this ready to merge.
Thanks!

packages/frontend/src/components/wallet/NFTBox.js
danielwpz: done

packages/frontend/src/components/wallet/NFTBox.js
danielwpz: done

packages/frontend/src/services/NonFungibleTokens.js
danielwpz: done

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,381 @@
...
+
+const StyledContainer = styled.div`
+    &&& {
+        display: flex;
+        flex-direction: column;
+        align-items: center;
+        padding: 0;
+
+        img {
+            width: 100%  ;
```
Patrick1904, line 30:
```suggestion
            width: 100%;
```

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,381 @@
...
+        p {
+            font-weight: 500;
+            font-size: 14px;
+            line-height: 150%;
+            text-align: center;
+            color: #72727A;
+        }
+
+        .next-btn {
+            margin-left: 44px  ;
```
Patrick1904, line 129:
```suggestion
            margin-left: 44px;
```

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,381 @@
...
+        }
+
+        .v-center {
+            position: absolute;
+            top: 50%;
+            transform: translate(0, -50%);
+        }
+
+        .icon {
+            margin-bottom: 20px  ;
```
Patrick1904, line 150:
```suggestion
            margin-bottom: 20px;
```

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,381 @@
...
+        }
+
+        form {
+            width: 100%;
+        }
+
+        .full-width {
+            width: 100%;
+        }
+
```
Patrick1904, line 160:
```suggestion

```

packages/frontend/src/translations/en.global.json
Patrick1904: Thank you 🙏

Patrick1904:
It seems that the NFTBox mapping is missing a key - can you please fix that?

Screen Shot 2022-01-03 at 2 51 38 PM

packages/frontend/src/components/nft/NFTTransferModal.js
Patrick1904: Actually, can we remove the margin-left completely here? Seems it's causing unnecessary extra space on the left?

Screen Shot 2022-01-03 at 2 55 13 PM

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,381 @@
...
+            font-weight: 500;
+            font-size: 14px;
+            line-height: 150%;
+            text-align: center;
+            color: #72727A;
+        }
+
+        .next-btn {
+            margin-left: 44px  ;
+        }
```
Patrick1904, line 130:
```suggestion

```

Patrick1904: Can we remove this? Seems to add unnecessary extra space on the left side of the button?

Screen Shot 2022-01-03 at 2 57 16 PM

packages/frontend/package.json
```diff
@@ -50,6 +50,7 @@
     "reselect": "^4.0.0",
     "semantic-ui-css": "^2.4.1",
     "semantic-ui-react": "^0.88.1",
+    "sentry": "^0.1.2",
```
Patrick1904, line 53: @danielwpz I don't think you need to add sentry here?

Patrick1904:
> sorry for the late update, didn't get a chance to update this PR last few days. @Patrick1904 fixed all latest comments, if it looks good I'll rebase master and make this ready to merge. Thanks!

Thanks @danielwpz ! Please go ahead and rebase and I'll do final round of testing after and I'll do the same before we merge.

packages/frontend/package.json
MaximusHaximus: Please 🔪 this :)

packages/frontend/src/components/nft/NFTDetail.js
```diff
@@ -0,0 +1,163 @@
...
+                            { ownerId }
+                        </span>
+                    </div>
+                </div>
+
+                {(ownerId === accountId) &&
+                  <FormButton 
+                    className='transfer-btn'
+                    color='gray-gray' 
+                    disabled={ownerId !== accountId || !nearBalance}
```
MaximusHaximus, line 144: 1. :hocho: redundant check since this `FormButton` cannot be displayed if `ownerId !== accountId` is true due to the check on L140.
2. `nearBalance` will (almost?) never be falsy (absolutely 0), and even if it is a positive number, it may be too small to allow transfer of an NFT because it requires some gas be attached.  Should this be checking that the user has at least `NFT_TRANSFER_GAS + NFT_TRANSFER_DEPOSIT` available instead of checking this with a falsy check?

packages/frontend/src/components/nft/NFTDetail.js
```diff
@@ -0,0 +1,163 @@
...
+                    onClick={() => setTransferNftDetail(nft)}
+                >
+                    <SendIcon/>
+                    <Translate id='NFTDetail.transfer'/>
+                </FormButton>}
+                {transferNftDetail &&
+                    <NFTTransferModal
```
MaximusHaximus, line 151: :nit: this can be a self-closing tag since it has no children

packages/frontend/src/components/nft/NFTDetailWrapper.js
```diff
@@ -0,0 +1,41 @@
...
+    const accountId = useSelector(selectAccountId);
+    const balance = useSelector(state => selectBalance(state));
+    const nearBalance = balance.balanceAvailable;
+
+    const contractId = match.params.contractId;
+    const tokenId = match.params.tokenId;
+    const [ nft, setNft ] = useState();
+    const [ ownerId, setOwnerId ] = useState();
+
+    useEffect(() => {
```
MaximusHaximus, line 22: 1. This data should be loaded from Redux  -- it will usually already be there, since if someone is transferring an NFT they have likely already viewed it in the wallet UI and clicked on the NFT to get NFT details, and the wallet UI is already populated using Redux data.

2. If the NFT contract and owned token metadatas aren't already in Redux, we should be dispatching actions to load those into redux here, rather than only loading it into this component's local `useState()` hooks.  

3. This also means that we can get rid of both the `setState` calls for `ownerId` and for `nft`; both of these pieces of data would be loaded directly from Redux using selectors.

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,351 @@
...
+    const dispatch = useDispatch();
+    const { updateNFTs } = nftActions;
+
+    const { localAlert } = useSelector(({ status }) => status);
+
+    function onTransferSuccess(result) {
+        setResult(result);
+        dispatch(updateNFTs({ accountId }));
+        setViewType('success');
+    }
```
MaximusHaximus, line 187: Since this is called once and only once, it may as well be in-lined on L200

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,351 @@
...
+    const balance = useSelector(state => selectBalance(state));
+    const nearBalance = balance.balanceAvailable;
+    const dispatch = useDispatch();
+    const { updateNFTs } = nftActions;
+
+    const { localAlert } = useSelector(({ status }) => status);
+
+    function onTransferSuccess(result) {
+        setResult(result);
+        dispatch(updateNFTs({ accountId }));
```
MaximusHaximus, line 185: This call is going to delete _all_ NFT contract metadata and owned token metadata, which will result in quite a lot of network calls re-fetching NFT details that we can reasonably assume have not changed.

Why not _only_ purge the NFT entry for the one NFT that was just transferred successfully?

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,351 @@
...
+        setViewType('success');
+    }
+
+    async function sendNFT(nft, receiverId) {
```
MaximusHaximus, line 189: Generally, avoid referencing some data from closure-scope and other data as arguments if all of the data is in the parent closure scope.  :hocho: either remove these arguments (just reference the variables in the parent closure scope directly and don't shadow those variables), or pass _all_ of the variables referenced in this function into the function as arguments (with names that do not shadow variable names in the parent closure)

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,351 @@
...
+    const [result, setResult] = useState();
+    const [sending, setSending] = useState(false);
+    const [viewType, setViewType] = useState('transfer');
+    const [accountIdIsValid, setAccountIdIsValid] = useState(false);
+    const balance = useSelector(state => selectBalance(state));
+    const nearBalance = balance.balanceAvailable;
+    const dispatch = useDispatch();
+    const { updateNFTs } = nftActions;
+
+    const { localAlert } = useSelector(({ status }) => status);
```
MaximusHaximus, line 181: This is a direct state access; we always use selectors for accessing Redux state -- this should be using `selectStatusLocalAlert` selector

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,351 @@
...
+`;
+
+export default function NFTTransferModal({ open, onClose, nft, accountId }) {
+    const [receiverId, setReceiverId] = useState('');
+    const [result, setResult] = useState();
+    const [sending, setSending] = useState(false);
+    const [viewType, setViewType] = useState('transfer');
+    const [accountIdIsValid, setAccountIdIsValid] = useState(false);
+    const balance = useSelector(state => selectBalance(state));
+    const nearBalance = balance.balanceAvailable;
```
MaximusHaximus, line 177: :hocho: It would be clearer to destructure directly from this and eliminate the `balance` variable that is only referenced on this one line -- e.g.:
```javascript
const { balanceAvailable } = useSelector(state => selectBalance(state));
```

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,351 @@
...
+
+    return (
+        <Modal
+            id='nft-transfer-modal'
+            isOpen={nft}
+            onClose={onClose}
+            closeButton={false}
+            modalSize='md'
+        >
+            {viewType === 'transfer' &&
```
MaximusHaximus, line 221: :nit: this seems more like `selectReceiver` than `transfer`

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,351 @@
...
+                <StyledContainer>
+                    <img className='transfer-img' src={nft.metadata.mediaUrl} alt='NFT' />
+
+                    <h3><Translate id='NFTTransfer.transferNft' /></h3>
+                    <p className='transfer-txt'><Translate id='NFTTransfer.enterReceipt' /></p>
```
MaximusHaximus, line 226: For consistency, would like this to be `enterReceiver`

packages/frontend/src/components/nft/NFTTransferModal.js
```diff
@@ -0,0 +1,351 @@
...
+                    </div>
+
+                    <div className='full-width'>
+                        <ModalFooter>
+                            <FormButton
+                                type='button'
+                                linkTo={`${EXPLORER_URL}/transactions/${result.transaction.hash}`}
+                                color='gray-gray'
+                            >
+                                <Translate id='NFTTransfer.viewReceipt' />
```
MaximusHaximus, line 336: For clarity, should this term be named `viewTransaction`?

packages/frontend/src/components/wallet/NFTs.js
```diff
@@ -57,14 +57,14 @@ const StyledContainer = styled.div`
...
-const NFTs = ({ tokens }) => {
+const NFTs = ({ tokens, accountId }) => {
     if (tokens.length) {
         return (
             <StyledContainer>
                 {tokens.map((tokenDetails) => (
                     <NFTBox
                         key={tokenDetails.contractName}
-                        tokenDetails={tokenDetails}
+                        tokenDetails={Object.assign({ ownerId: accountId }, tokenDetails)}
```
MaximusHaximus, line 67: `ownerId` should already be available in the `tokenDetails`, given that `tokenDetails` includes the metadata from on-chain RPC call results..  Why are we setting `ownerId` to `accountId` here?

MaximusHaximus (changes requested):
@danielwpz thanks so much for your contribution; we appreciate it!

This is looking great!

I have added a few questions about the data loading management and a few comments in-line.  It looks like we're just about ready to get this merged!

cjd425:
Hey, what is the status of this PR? People message me almost every day asking when they can transfer and mint NFTs in the wallet directly. I'm willing to do whatever I can to help here! @Patrick1904 @MaximusHaximus

Patrick1904:
> Hey, what is the status of this PR? People message me almost every day asking when they can transfer and mint NFTs in the wallet directly. I'm willing to do whatever I can to help here! @Patrick1904 @MaximusHaximus

@cjd425 this PR should be very close to ready. @MaximusHaximus had some minor requests in the last round of reviews that seemed to have been addressed by @danielwpz. @MaximusHaximus can you please verify that you're ok with these changes and we can rebase the PR so that it's aligned with master and then merge.

reed-hub:
@Patrick1904 @MaximusHaximus I too am curious about this one as it seems really beneficial to the ecosystem at large. Do we have an estimate on when this one can merge?

think-in-universe:
@Patrick1904 @MaximusHaximus Hey Patrick and Daryl, what's the status of the feature now? please let me if you need any help that could help us to resolve the issues and merge the feature into production.

Patrick1904:
> @Patrick1904 @MaximusHaximus Hey Patrick and Daryl, what's the status of the feature now? please let me if you need any help. I think we can figure out the missing part, resolve the issues and merge the feature into production.

@think-in-universe this PR should be almost ready for merge. 👍 There are still some things that weren't addressed after @MaximusHaximus's review that we want to fix first, and @andy-haynes will make the final touches before we ship this.

think-in-universe:
@Patrick1904 thanks Patrick! good to know. let's fix the the issues and move forward 😄

Patrick1904:
Closing this issue as it's been ported over and shipped in: https://github.com/near/near-wallet/pull/2514

The feature is currently available on testnet/staging and will ship to mainnet with the next [release](https://github.com/near/near-wallet/pull/2518)

Thank you @danielwpz for your work and patience in getting this shipped 🙏

think-in-universe:
great job! thanks everyone @danielwpz @Patrick1904 @andy-haynes @MaximusHaximus

PR #2208, opened 2021-10-19
ADD modifications to lockup balance display

### LINKED ISSUE #2014: Wrong Lockup information (triggered by Lockup Core Contract update)
**Problem**
Recently, @frol and @telezhnaya updated the [lockup contract](https://github.com/near/core-contracts/tree/master/lockup), introducing changes that break how the wallet is measuring the vesting of a lockup contract.

Example:
NEAR Wallet display: https://wallet.near.org/profile/scarce1.near (shows all tokens vested)
Account-lockup display: https://near.github.io/account-lookup/#scarce1.near

**Expected Behavior**
NEAR Wallet should display the correct vesting schedule for lockup contracts

**Steps to reproduce**

1. open https://wallet.near.org/profile/scarce1.near to see "0 locked" tokens
2. open https://near.github.io/account-lookup/#scarce1.near to see the correct value instead

Here the rewritten logic to calculate the lockup vesting and available funds: https://github.com/near/account-lookup/commit/1ca44052789fed57e40d634402ba7d4625b7aa21

esaminu:
* Use [constant](https://github.com/near/core-contracts/blob/master/lockup/src/lib.rs#L33) (3.5 / 35 Ⓝ) conditional on PR # changing lockup_contract.wasm on [core-contracts repo](https://github.com/near/core-contracts) instead of always defaulting to old storage constant (35 Ⓝ)
* If transfers are disabled use current time as enabled timestamp
    * If `lockupTimestamp` was used amount released will be unaffected and will be based only on that timestamp
    * If `lockupDuration` was used amount will be locked since that duration is meant to be added to timestamp when transfers are enabled
* Include calculation of unvested funds in locked amount. Fixes #2014 
* Modify calculation of `liquidOwnersBalance` to use contract call on`get_liquid_owners_balance`
* Add new code hash broken timestamp from account-lookup repo
* Add support for vesting type 3
* Correctly return whole amount as locked if `start_time` is greater than current time. See https://github.com/near/account-lookup/pull/12. Current incorrect display [example](https://wallet.near.org/profile/cgran42.near) vs [account-lookup](https://near.github.io/account-lookup/#cgran42.near).

### DIFF
--- a/packages/frontend/src/utils/account-with-lockup.js
+++ b/packages/frontend/src/utils/account-with-lockup.js
@@ -2,7 +2,6 @@ import BN from 'bn.js';
 import sha256 from 'js-sha256';
 import { Account, Connection, InMemorySigner, KeyPair } from 'near-api-js';
 import { InMemoryKeyStore } from 'near-api-js/lib/key_stores';
-import { parseNearAmount } from 'near-api-js/lib/utils/format';
 import { BinaryReader } from 'near-api-js/lib/utils/serialize';
 
 import {
@@ -14,7 +13,6 @@ import {
 import { WalletError } from './walletError';
 
 // TODO: Should gas allowance be dynamically calculated
-export const LOCKUP_MIN_BALANCE = new BN(parseNearAmount('35'));
 
 const BASE_GAS = new BN('25000000000000');
 
@@ -172,21 +170,28 @@ async function getAccountBalance(limitedAccountData = false) {
     try {
         const lockupAccount = new Account(this.connection, lockupAccountId);
         const lockupBalance = await lockupAccount.getAccountBalance();
+        const lockupStateStaked = new BN(lockupBalance.stateStaked)
         const {
             lockupAmount,
             releaseDuration,
             transferInformation,
             lockupTimestamp,
             lockupDuration,
-            terminationWithdrawnTokens
+            terminationWithdrawnTokens,
+            vestingInformation
         } = await viewLockupState(this.connection, lockupAccountId);
 
         const dateNowBN = new BN(Date.now()).mul(new BN('1000000'));
 
         const { transfer_poll_account_id, transfers_timestamp } = transferInformation;
-        const transfersTimestamp = transfer_poll_account_id ? await this.viewFunction(transfer_poll_account_id, 'get_result') : transfers_timestamp;
+        let transfersTimestamp = transfer_poll_account_id ? await this.viewFunction(transfer_poll_account_id, 'get_result') : transfers_timestamp;
+        transfersTimestamp = transfersTimestamp || (Date.now() * 1000000);
 
-        const hasBrokenTimestamp = (await lockupAccount.state()).code_hash === '3kVY9qcVRoW3B5498SMX6R3rtSLiCdmBzKs7zcnzDJ7Q';
+        const hasBrokenTimestamp = [
+            '3kVY9qcVRoW3B5498SMX6R3rtSLiCdmBzKs7zcnzDJ7Q',
+            'DiC9bKCqUHqoYqUXovAnqugiuntHWnM3cAc7KrgaHTu'
+        ].includes((await lockupAccount.state()).code_hash);
+
         const startTimestampBN = BN.max(
             new BN(transfersTimestamp).add(new BN(lockupDuration || 0)),
             new BN(lockupTimestamp || 0)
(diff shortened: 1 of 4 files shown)


packages/frontend/src/utils/account-with-lockup.js
```diff
@@ -186,20 +213,19 @@ async function getAccountBalance(limitedAccountData = false) {
...
+        const ownersBalance = totalBalance.sub(lockedAmount)
+
+        // if acc is deletable (nothing locked && nothing stake) you can transfer the whole amount ohterwise get_liquid_owners_balance
+        const isAccDeletable = lockedAmount.isZero() && stakedBalanceLockup.isZero();
+        const liquidOwnersBalance = isAccDeletable
+            ? new BN(lockupBalance.total)
+            : new BN(
+                  await this.wrappedAccount.viewFunction(
+                      lockupAccountId,
+                      "get_liquid_owners_balance"
```
MaximusHaximus, line 226: Are we reasonably sure that `get_liquid_owners_balance` has always been part of the spec? :D

MaximusHaximus (approved):
LGTM, but you've got ESLint errors failing the build :)

Also would be good to have a couple of folks with real lockups double check this is working as expected -- @Patrick1904 @marcinbodnar @stefanopepe could you guys test this?

packages/frontend/src/utils/account-with-lockup.js
esaminu: We're using it [here](https://github.com/near/near-wallet/blob/master/packages/frontend/src/utils/account-with-lockup.js#L74) to transfer all available funds from the lockup. This makes sense because it's used by the `transfer` method [here](https://github.com/near/core-contracts/blob/master/lockup/src/owner.rs#L477) to determine the amount that is transferrable.

stefanopepe: @MaximusHaximus I don't have any lockups to test, but you can check with Patrick's at https://wallet.near.org/profile/patrick.near

I think I can get a legacy lockup up on Mainnet to check if on-chain and planned match, do you think is necessary?

marcinbodnar:
@esaminu 
I can see two differences:
1. Reserved for transactions - main account
 - it's showing 0.05 NEAR on the live version and 0 NEAR on this PR, and this difference is not included in any other number.
2. Reserved for storage - lockup account
 - for this PR is 10x lower but the difference is not included in any other value. So the total balance for the account is lower by about 30N than it is supposed to be.

esaminu:
@marcinbodnar 

- For the first point I don't understand the rationale for adding lockup "Available to transfer" balance to "Reserved for transactions" [link](https://github.com/near/near-wallet/blob/master/packages/frontend/src/utils/account-with-lockup.js#L205). The "Available to transfer" on this PR is the amount the contract will let us withdraw at any given time.
- The second point was brought up by @Patrick1904 as well. The issue here is that even though the amount is unlocked on older contracts, the contract won't allow us to withdraw that amount since it subtracts a constant `35 Ⓝ` in older contracts and `3.5 Ⓝ` in new ones (v3). Unfortunately we can't query this constant and display it. The only way would be to keep track of older hashes and display the constant based on them. Using [this list](https://github.com/near/near-wallet/pull/2208/files#diff-46f345c520c144e0d8300842ff78d012a61770029e9d262aaf62632d95960486R167) was suggested but I'm not sure if it includes all older wasms [here](https://github.com/near/core-contracts/commits/master/lockup/res/lockup_contract.wasm). Another option is to add some documentation around this and link to it in the info bubble as suggested by @Patrick1904 . I am not sure how crucial the reserved for storage display is on lockups so I'll also add the option of removing that display to the table? @stefanopepe @MaximusHaximus

esaminu:
After discussions with @MaximusHaximus I will be compiling a list of all `code_hash`es that we merged to master with the old constant from [this list](https://github.com/near/near-wallet/pull/2208/files#diff-46f345c520c144e0d8300842ff78d012a61770029e9d262aaf62632d95960486R167) for the purpose of displaying the right constant addressing point number 2 above. 

Still need info on rationale for adding lockup "Available to transfer" balance to "Reserved for transactions" [here](https://github.com/near/near-wallet/blob/master/packages/frontend/src/utils/account-with-lockup.js#L205) @Patrick1904 @marcinbodnar

esaminu:
@Patrick1904 @marcinbodnar @MaximusHaximus This is ready for final review on my side. I added a mapping of hashes to PR #s and returned the right constant being held accordingly.

@marcinbodnar `MIN_BALANCE_FOR_GAS` is `0.05` [here](https://github.com/near/near-wallet/blob/14caa4aaf93c62fce35d7ad34e66058323d02653/packages/frontend/src/redux/reducers/selectors/balance.js#L32) so the only way we return 0 for "Reserved for transactions" is if `available` is 0.

packages/frontend/src/utils/account-with-lockup.js
MaximusHaximus: @stefanopepe I think Osman's been able to reliably test this at this point -- resolving :)

MaximusHaximus (approved):
:+1: looks good -- props for wading into this code, it's _gnarly_

marcinbodnar:
> Still need info on rationale for adding lockup "Available to transfer" balance to "Reserved for transactions" here 

@esaminu Sorry, I'm not sure what it was, don't remember what was the reason.

marcinbodnar (approved):
Looks good - all balances for my lockup accounts match with this PR - I can see that `0.05` from `Reserved for transactions` is the missing amount on the live version, so that fixes the problem.

Good job @esaminu

Patrick1904 (approved):
Tested several accounts - checks out! Thanks @esaminu! 👍

PR #2215, opened 2021-10-25
Add new /sign UI

Patrick1904:
This PR adds an updated UI on the `/sign` route used for signing DApp transactions.

For testing, use the PR staging link:
https://deploy-preview-2215--near-wallet-staging.netlify.app/

Example DApps for testing TX signing:
1. https://app.ref.finance/
2. https://berryclub.io/

Issue #2111, opened 2021-09-14
Retry transactions failed for insufficient gas


stefanopepe:
## Summary

With the increased adoption of NEAR Protocol, we may experience peaks of usage that translate into higher gas costs. Some dapps attach a fixed amount of gas to a transaction, taking the risk of failing the transfer and put the user in an endless loop of "failed transactions" if no proper mitigations are in place.

This feature wants to unblock the user who is signing a transaction with too low gas, allowing them to retry manually attaching a larger amount. 

## Desired Outcome

As a user, I want to "bump" the gas fees on my failed transaction, such that I can make sure that I completed the request.

As a dapp owner, I want to make sure that my users can anyway use my platform if the wallet (or the network) misallocate their gas fees.

Success will be measured against the number of transactions failed for insufficient gas, and the number of transactions submitted right after that were successful.

## How will it work?

NEAR Wallet will submit the transaction as requested, and check if it failed, returning as usual the reason. If the failure reason is "exceeded prepaid gas", the UI will allow the user to retry the same transaction attaching more gas.
 
A first release of the transaction may leave to the user the decision on how much more to attach, but ideally it should suggest a number based on the most recent "retries".

This feature will support dapp developers who will be able to see how many users manually "bumped" transactions, allowing them to improve their application behavior.

corwinharrell:
@stefanopepe @MaximusHaximus I've encountered a fundamental issue in designing around this that I wanted to surface and get your feedback on:

We currently don't calculate the cost of a transaction before it is executed. Right now, few simply display an arbitrary number labelled "estimated fees", which hopes to cover the cost of the TXN. As long as this is the case, we have no way to telling the user _how much_ gas might be required to ensure a successful TXN. If we are allowing the user to manually indicate the amount of gas they wish to use for a "bumped" TXN, they are taking a stab in the dark.

How should we handle this? Do we just allow them to guess? This is poor UX. IMO, a dependency of this work is calculating the actual TXN cost immediately before execution, and then offering the user a most realistic estimate for an effective increase.

stefanopepe:
@corwinharrell I see the issue. Mine is a high-level proposal that didn't take into account this specific limitation.

I'm happy with keeping the scope of this feature as a simple "retry transaction" functionality that will implicitly attach more fees (e.g. bump to 300Tgas) and try again if the transaction didn't go through for insufficient gas.

If I understand correctly, this solution would preserve the desired outcome, keeping any complicated gas estimation&management on hold.

[...]

PR #2216, opened 2021-10-26
Retry transaction if gas is insufficient

marcinbodnar:


packages/frontend/src/components/sign/SignTransferRetry.js
```diff
@@ -0,0 +1,134 @@
+import React from 'react';
+import { Translate } from 'react-localize-redux';
+import styled from 'styled-components';
+
+import ArrowUpImage from '../../images/icon-arrow-up-green.svg';
+import RetryImage from '../../images/icon-retry-tx.svg';
+import FormButton from '../common/FormButton';
+
+const CustomContainer = styled.div`
```
Patrick1904, line 9: @marcinbodnar can we please use the `/common/Container.css` as base wrapper for the page container like we do here: https://github.com/near/near-wallet/blob/a1fc1fe4bc362a23c74f8e52bcd36852791d099d/packages/frontend/src/components/login/v2/ConfirmLogin.js#L38

Patrick1904: You can also use it if you're creating a new styled component wrapper, like here: https://github.com/near/near-wallet/blob/a1fc1fe4bc362a23c74f8e52bcd36852791d099d/packages/frontend/src/components/send/SendContainerV2.js#L27

packages/frontend/src/components/sign/SignTransferRetry.js
```diff
@@ -0,0 +1,134 @@
...
+        margin-top: 40px;
+
+        button {
+            flex: 1;
+
+            &:last-of-type {
+                margin-left: 30px !important;
+
+                @media (min-width: 768px) {
+                    margin-left: 50px !important;
```
Patrick1904, line 45: Can we please just add `&&& {...}` at the top of the css so that we can remove `!important` - trying to actively remove all usage of `!important` in our css

packages/frontend/src/components/sign/SignTransferRetry.js
```diff
@@ -0,0 +1,134 @@
...
+                    height: 17px;
+                    left: -22px;
+                    top: 1px;
+                }
+            }
+        }
+    }
+`;
+
+// TODO: Why handleDeny? It's not an error.
```
Patrick1904, line 87: Maybe just rename it within this context to `handleCancel`? Even though you can still re-use the `handleDeny` function to send user back to the app

packages/frontend/src/redux/reducers/sign/index.js
```diff
@@ -25,43 +37,55 @@ const sign = handleActions({
...
+    [setSignTransactionStatus]: (state, { payload }) => {
         return {
             ...state,
-            status: 'success'
+            status: payload.status
         };
     },
-    [setSignTransactionStatus]: (state, { payload }) => {
+    [multiplyGas]: (state, { payload: { transactions: transactionsString } }) => {
+        const transactions = deserializeTransactionFromString(transactionsString);
```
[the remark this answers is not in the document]
MaximusHaximus: To re-state from a high level, this case is:
1. The user tried to submit a tx with the original amount of gas, but it failed due to insufficient gas attached.
2. The user clicked on our brand new button that allows them to resubmit the tx, but with 2x as much gas attached as the initial attempt had.
3. The second attempt to submit the tx failed also _even though we attached 2x as much gas as the first attempt_.

**Currently**, the user will again see the button allowing them to 'resubmit this with more gas attached'.
If they click that button, we will re-submit the TX a _third_ time, but this time with 2x as much gas as the _second attempt_  -- now they are attaching **4x** as much gas as the original allotment.

Consider that If we allow the user to make a _third_ attempt, it will be with **8x** the amount of gas attached as they did for their first attempt, and a fourth attempt will have **16x** as much gas attached as their first attempt did, and so on.

@stefanopepe could you weigh in on this behaviour? -- from the product perspective:
1. Should it be possible to repeat the re-submit attempt multiple times with ever increasing amounts of gas attached each time?
2. If yes for # 1 above, then should we limit how many times they can use the 'retry with 2x as much gas as the previous attempt' button, or should we leave this as an un-capped operation that would allow them to retry many times, each time doubling the gas from the previous attempt?  If they retry 8 times, they would be submitting with **512x** as much gas attached as their original attempt had.

packages/frontend/src/redux/reducers/sign/index.js
```diff
@@ -62,10 +62,11 @@ const sign = handleActions({
     [multiplyGas]: (state, { payload: { transactions: transactionsString } }) => {
         const transactions = deserializeTransactionsFromString(transactionsString);
```
MaximusHaximus, line 64: 1. If we're going to use values from the `transactions` property of this slice's redux state, we should use _only_ values from that property of this slice of the the redux state rather than mix+match the data from re-parsing `account.url` _and_ existing `sign.transactions` state; this allows us to be absolutely sure that we're not doing something incorrect by assuming that the data in every index of `sign.transactions` and each transaction's `actions` property are 100% identical between the parsed output of `account.url` and the current redux `sign.transactions` state;  and by doing so,`multiplyGas` doesn't need to take any arguments at all if we've made the decision to use the existing state since this handler has direct access to the existing state since it is a reducer on the `sign` slice.

2. Also, since you're still passing `transactions` that is computed from the `account.url` state (which is not updated by `multiplyGas`, right?) to the `calculateGasLimit()` call, I think that the `gasLimit` property in state is not being correctly updated with the current code -- this type of bug is another danger of using two different sources of truth in the same function to generate new state.

I think for this to be correct we'll need something like:
```javascript
const transactionsWithDoubledGas = _.cloneDeep(state.transactions)
    .forEach((transaction) => {
        transactions.actions && transaction.actions.forEach((action) => {
            if(a.functionCall && a.functionCall.gas) {
                a.functionCall.gas = a.functionCall.gas.mul(new BN(MULTIPLY_TX_GAS_BY));
            }
        })
    });


return {
    ...state,
    transactions: transactionsWithDoubledGas,
    fees: {
        ...state.transactions.fees,
        gasLimit: calculateGasLimit(transactionsWithDoubledGas.flatMap(t => t.actions))
    }
}
```

This is assuming that `cloneDeep()` properly clones transaction objects _and_ any `bn.js` properties.  Otherwise you might need to explicitly call `.clone()` on any property of either Transaction and Action objects in the state if they are a BN.js class instance.

MaximusHaximus (changes requested):
A couple of concerns re: the current `multiplyGas` implementation adding in an in-line comment.

I also pinged Stefano to see his opinion about users being able to retry unlimited #s of times with exponential increases in attached gas for each attempt. :)

packages/frontend/src/redux/reducers/sign/index.js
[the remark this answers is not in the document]
stefanopepe: @MaximusHaximus @marcinbodnar I would assume that yes, we should iterate the operation of attaching more Tgas to the transaction - being careful to not to **exceed** and pay too much gas. So I would keep the increases in smaller steps, to not incur in invalid errors for excessive Tgas attached to transactions.

Examples:
- [this tx](https://explorer.near.org/transactions/2w52XqgrBVM1By2mKNRDuwUyfTXyUNfdPeq71nk9HY9Q) was paying 20Tgas, and failed
- [this one](https://explorer.near.org/transactions/BVfcDhwVbuF49oNmCqjxjGkQEbGCKiboME2dn5TvdR1C) paid 200Tgas and worked well
- [this other one](https://explorer.near.org/transactions/5tBWCKUr5c4iMNWTnLHhw3pvLjRUx5pQoiyymHUWFYdK) paid 300Tgas and worked great as well

This one below didn't work because of excessive gas attached to it (400Tgas) and threw an invalid transaction error even before sending it to the network:
```
pepe$ near call zkv_staketosupportprivacy.poolv1.near ping '{}' --accountId pinger.near --gas 400000000000000
Scheduling a call: zkv_staketosupportprivacy.poolv1.near.ping({})
An error occured
Error: {"kind":{"ActionsValidation":{"TotalPrepaidGasExceeded":{"limit":300000000000000,"total_prepaid_gas":400000000000000}}}}
    at Object.parseRpcError (/usr/local/lib/node_modules/near-cli/node_modules/near-api-js/lib/utils/rpc_errors.js:24:19)
    at /usr/local/lib/node_modules/near-cli/node_modules/near-api-js/lib/providers/json-rpc-provider.js:312:44
    at processTicksAndRejections (node:internal/process/task_queues:94:5)
    at async Object.exponentialBackoff [as default] (/usr/local/lib/node_modules/near-cli/node_modules/near-api-js/lib/utils/exponential-backoff.js:7:24)
    at async JsonRpcProvider.sendJsonRpc (/usr/local/lib/node_modules/near-cli/node_modules/near-api-js/lib/providers/json-rpc-provider.js:297:26)
    at async /usr/local/lib/node_modules/near-cli/node_modules/near-api-js/lib/account.js:114:24
    at async Object.exponentialBackoff [as default] (/usr/local/lib/node_modules/near-cli/node_modules/near-api-js/lib/utils/exponential-backoff.js:7:24)
    at async Account.signAndSendTransactionV2 (/usr/local/lib/node_modules/near-cli/node_modules/near-api-js/lib/account.js:110:24)
    at async scheduleFunctionCall (/usr/local/lib/node_modules/near-cli/commands/call.js:54:34)
    at async Object.handler (/usr/local/lib/node_modules/near-cli/utils/exit-on-error.js:52:9)
ServerError: {"kind":{"ActionsValidation":{"TotalPrepaidGasExceeded":{"limit":300000000000000,"total_prepaid_gas":400000000000000}}}}
    at Object.parseRpcError (/usr/local/lib/node_modules/near-cli/node_modules/near-api-js/lib/utils/rpc_errors.js:24:19)
    at /usr/local/lib/node_modules/near-cli/node_modules/near-api-js/lib/providers/json-rpc-provider.js:312:44
    at processTicksAndRejections (node:internal/process/task_queues:94:5)
    at async Object.exponentialBackoff [as default] (/usr/local/lib/node_modules/near-cli/node_modules/near-api-js/lib/utils/exponential-backoff.js:7:24)
    at async JsonRpcProvider.sendJsonRpc (/usr/local/lib/node_modules/near-cli/node_modules/near-api-js/lib/providers/json-rpc-provider.js:297:26)
    at async /usr/local/lib/node_modules/near-cli/node_modules/near-api-js/lib/account.js:114:24
    at async Object.exponentialBackoff [as default] (/usr/local/lib/node_modules/near-cli/node_modules/near-api-js/lib/utils/exponential-backoff.js:7:24)
    at async Account.signAndSendTransactionV2 (/usr/local/lib/node_modules/near-cli/node_modules/near-api-js/lib/account.js:110:24)
    at async scheduleFunctionCall (/usr/local/lib/node_modules/near-cli/commands/call.js:54:34)
    at async Object.handler (/usr/local/lib/node_modules/near-cli/utils/exit-on-error.js:52:9) {
  type: 'InvalidTxError',
  context: ErrorContext {
    transactionHash: '7PHyLUf97c6R3E1UoH7AJZqhjbdsW4bN8FPJwRTkA6yc'
  },
  kind: {
    ActionsValidation: {
      TotalPrepaidGasExceeded: { limit: 300000000000000, total_prepaid_gas: 400000000000000 }
    }
  }
}
We attempted to send transaction 7PHyLUf97c6R3E1UoH7AJZqhjbdsW4bN8FPJwRTkA6yc to NEAR, but something went wrong.
To see the transaction in the transaction explorer, please open this url in your browser
https://explorer.mainnet.near.org/transactions/7PHyLUf97c6R3E1UoH7AJZqhjbdsW4bN8FPJwRTkA6yc
Note: if the transaction was invalid (e.g. not enough balance), it will show as "Not started" or "Finalizing"
```

The logic in my opinion should:
1. try again the transaction adding 25Tgas units to the original Tgas and keep going up until 300Tgas, after which it should fail completely. 
2. If the transaction is _already coming_ with 300Tgas, go in the opposite direction, decreasing 25Tgas units until 150Tgas, after which fail completely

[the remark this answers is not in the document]
stefanopepe: As a reference, see my comment here: https://github.com/near/near-wallet/issues/2111#issuecomment-956479345

packages/frontend/src/components/sign/SignTransferRetry.js
marcinbodnar: Done.

packages/frontend/src/components/sign/SignTransferRetry.js
marcinbodnar: Done.

packages/frontend/src/components/sign/SignTransferRetry.js
marcinbodnar: It's not my comment, I copied it with other code from `SignTransferSuccess.js` :)

packages/frontend/src/redux/reducers/sign/index.js
marcinbodnar: Agree, Done :)

marcinbodnar:
@MaximusHaximus @stefanopepe Changes are ready.
On Monday I will add mixpanel tracking for several situations.

marcinbodnar:
@stefanopepe also added mixpanel tracking for situations:
1. When Sign is requested with changed gas, the name is "SIGN - RETRY - increase" or "SIGN - RETRY - decrease"
2. When Sign is requested with changed gas, but the retry limit is exceeded, so for example if we tried with increasing gas and the limit of 300Tgas was reached. Name: "SIGN - RETRY - retry limit exceeded".

packages/frontend/src/redux/reducers/sign/index.js
```diff
@@ -25,43 +46,73 @@ const sign = handleActions({
...
                 .filter(a => ['deployContract', 'stake', 'deleteAccount'].indexOf(Object.keys(a)[0]) > -1)
                 .length
         };
     },
-    [signAndSendTransactions]: (state, { error, payload, ready }) => {
+    [handleSignTransactions.pending]: (state) => ({
+        ...state,
+        status: SIGN_STATUS.IN_PROGRESS
+    }),
+    [handleSignTransactions.fulfilled]: (state, { payload }) => ({
```
MaximusHaximus, line 61: Shouldn't `state.error` be reset here?

packages/frontend/src/redux/reducers/sign/index.js
```diff
@@ -25,43 +46,73 @@ const sign = handleActions({
...
+            Mixpanel.track('SIGN - RETRY - retry limit exceeded');
         }
 
-        if (error) {
-            return {
-                ...state,
-                status: 'error',
-                error: payload
-            };
+        const transactions = cloneDeep(state.transactions);
```
MaximusHaximus, line 89: If we're not mutating `state.transactions`, we **shouldn't** clone it and replace the existing `state.transactions` with a clone, we should just leave `state.transactions` as-is.

MaximusHaximus (changes requested):
Added some stuff in-line -- I don't think this stuff is ready to roll; the code doesn't clamp increases to ensure that we're not exceeding the MAX gas, and I don't think there's a need to have a `DECREASE` case given that we know the initial amount of attached gas does not exceed the limit.

Have you tested it with any real contracts & actions?  If not, please work with @ihaid to get a set of test cases together so that we can be sure that this code is really working.

In the most basic form, we need a set of transactions & actions that we know will fail due to not-enough-gas-attached with the default gas, and we know will succeed after retrying with more gas, so we can go through the flow and verify it's working. Ideally, this would become part of our `e2e-tests` suite also; did you ever talk to Osman about how these impact the e2e test suite?

Also, because these are reducers with a fair amount of complexity, they should have some basic unit tests that verify golden path and corner cases, especially considering these are entirely pure functions! It looks like some of the logic for checking boundaries is wrong in the current implementation and we'd let people retry + further increase gas for actions when functionCall.gas is already too high or too low.  If that's true, when you fix that bug you should add a test to cover it.

packages/frontend/src/components/sign/SignTransferRetry.js
MaximusHaximus: Even more of a reason to name it correctly in your component; don't let things like this become infectious

marcinbodnar:
@MaximusHaximus 
> code doesn't clamp increases to ensure that we're not exceeding the MAX gas

I'm pretty sure it does. What case do you have in mind?

> I don't think there's a need to have a DECREASE case given that we know the initial amount of attached gas does not exceed the limit.

The code doesn't do that. we DECREASE only if the amount of gas is too high. 

> Have you tested it with any real contracts & actions?

Of course, with several single tx signs and multi tx signs. I can't imagine commit without testing :) 

> It looks like some of the logic for checking boundaries is wrong in the current implementation and we'd let people retry + further increase gas for actions when functionCall.gas is already too high or too low.

No, if gas is too high we decrease the amount of gas till it reaches 150Tgas.

I implemented the logic that Stefano proposed here:
https://github.com/near/near-wallet/pull/2216#discussion_r740427392
The idea is to cover most of the situations when the problem might appear.

packages/frontend/src/redux/reducers/sign/index.js
```diff
@@ -25,43 +46,73 @@ const sign = handleActions({
...
+    [handleSignTransactions.pending]: (state) => ({
+        ...state,
+        status: SIGN_STATUS.IN_PROGRESS
+    }),
+    [handleSignTransactions.fulfilled]: (state, { payload }) => ({
+        ...state,
+        status: SIGN_STATUS.SUCCESS,
+        transactionHashes: payload
+    }),
+    [handleSignTransactions.rejected]: (state, { error }) => {
```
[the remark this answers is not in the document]
marcinbodnar: The FIrst part of the code `retryTxDirection`: is checking the situation that we need to handle, whether gas is too much or too low. And it's set only once in the redux state.

The second part `tryRetryTx`: is checking whether we should try to retry or if the gas limits that we set are reached and we have to break the sign flow with hard error.

The third part `if (tryRetryTx)`: is responsible for changing gas but only if the second part said that we should try that.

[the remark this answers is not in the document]
marcinbodnar: > I don't think it's good that the rejected handler for handleSignTransactions is now responsible for changing gas amounts for actions in our store

My idea here was that changing the gas is part of the error handling for `handleSignTransactions`, so it should be a part of rejected action - basically, immediately when the error appears, we are checking what the problem was, and handle it in the way that it would allow resolving the problem.

Additionally, on the retry screen, we are showing the amount of gas that would be used for retry. So with the above solution, we can simplify the code, by having the cost of the gas already calculated, so we don't have to do that two times - once for showing it to the user, and a second time when we do the request.

Basically, the user has only two possibilities - retry tx with changed gas, or cancel.

So I thought it's reasonable to do that in `rejected` action. But I can change it no problem, it will be just a bit more complicated solution. Let me know what you prefere.

packages/frontend/src/redux/reducers/sign/index.js
```diff
@@ -25,43 +46,73 @@ const sign = handleActions({
...
+            t.actions && t.actions.some((a) => a.functionCall && a.functionCall.gas && (
+                (
+                    state.retryTxDirection === RETRY_TX.INCREASE 
+                    && a.functionCall.gas.gt(new BN(RETRY_TX.GAS.MAX))
+                ) || (
+                    state.retryTxDirection === RETRY_TX.DECREASE 
+                    && a.functionCall.gas.lte(new BN(RETRY_TX.GAS.MIN))
+                )
+            ))
+        );
```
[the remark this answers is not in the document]
marcinbodnar: > Aren't these logic checks backwards?

It working well. I prepared it this way (by adding `!` before `some`) so we can base it on the positive value (`tryRetryTx` instead of `stopRetryTx`).

> Also, we should be checking if increasing or decreasing the functionCall.gas amount by DIFF would cause it to exceed the min/max, not check just its current value, otherwise we will overshoot and increase the value past the max threshold; we should clamp the functionCall.gas at <= RETRY_TX.GAS.MAX and => RETRY_TX.GAS.MIN.

Good catch. Done.

packages/frontend/src/redux/reducers/sign/index.js
```diff
@@ -25,43 +46,73 @@ const sign = handleActions({
...
-                ...state,
-                status: 'error',
-                error: payload
-            };
+        const transactions = cloneDeep(state.transactions);
+        if (tryRetryTx) {
+            transactions.forEach((t, i) => {
+                t.actions && t.actions.forEach((a, j) => {
+                    if(a.functionCall && a.functionCall.gas) {
+                        if ((state.retryTxDirection || retryTxDirection) === RETRY_TX.INCREASE) {
```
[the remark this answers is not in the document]
marcinbodnar: > This should be clamped at RETRY_TX.GAS.MAX so we never increase an action's allocated gas above the max.

We are checking it in the above `tryRetryTx` section - here we are sure, that we should try to change it.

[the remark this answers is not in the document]
marcinbodnar: > Why the OR here that uses either `state.retryTxDirection` or the local variable? Seems like this would mean that if we previously got an error that set us up to INCREASE (for example), but then we got an error that doesn't match either of our checks that set `retryTxDirection` to a value other than `undefined`, then this OR would result in us _still increasing gas here_ even though the error we received doesn't indicate that we should either increase or decrease it.

The OR is for the first error,  when we don't have `state.retryTxDirection` set. So we will not do retry with the same values at the first iteration.

packages/frontend/src/redux/reducers/sign/index.js
```diff
@@ -25,43 +46,73 @@ const sign = handleActions({
...
+                            a.functionCall.gas = a.functionCall.gas.add(new BN(RETRY_TX.GAS.DIFF));
+                        } else if ((state.retryTxDirection || retryTxDirection) === RETRY_TX.DECREASE) {
```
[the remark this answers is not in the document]
marcinbodnar: > If we were going to include a `decrease` mode, we'd want this to be clamped at `RETRY_TX.GAS.MIN` so we never decrease an action's allocated gas below the min.

We are checking it in the above `tryRetryTx` section - here we are sure, that we should try to change it.

packages/frontend/src/redux/reducers/sign/index.js
marcinbodnar: Sure. Done.

packages/frontend/src/redux/reducers/sign/index.js
marcinbodnar: Right. Done.

corwinharrell:
I realize I am late to the conversation here, and I know it can be irritating to receive feedback late in the game so I apologize. Been trying to catch up since returning from Lisbon. Wanted to share just a couple thoughts:

The design of this feature in contrast to something that allows the user to explicitly set their own gas limit is for the sake of convenience. Simply allowing them to press a "try again" button is easier than having them reason about gas.

Now, if the transactions continues to fail a second and third time, the convenience factor has kind of gone out the window. Now the user needs to understand that the same error has occurred, and that they'll need to keep repeating a recursive action until the transaction goes through. There is a risk of the user not understanding that the error is actually a new error for the second transaction, and also that the gas limit is being increased further on subsequent attempts.

Rather than trying to manage this loop, why not provide the ability to manually adjust the gas limit after the first retry attempt fails? This seems like it would likely be an edge case, and the manual adjustment of gas is something that we'll be adding in the future regardless.

stefanopepe:
> Rather than trying to manage this loop, why not provide the ability to manually adjust the gas limit after the first retry attempt fails? This seems like it would likely be an edge case, and the manual adjustment of gas is something that we'll be adding in the future regardless.

I think this would brilliantly solve the rabbit hole we ended up into: instead of building a self-aware gas retry algorithm, we go through three easy outcomes:
1. transaction fails because of gas, retry with more gas
2. fails again, ask the user if they want to define an arbitrary Tgas number
3. alternatively, just close the tx window as of today

packages/frontend/src/redux/reducers/sign/index.js
```diff
@@ -25,43 +46,77 @@ const sign = handleActions({
...
+            ))
+        );
 
-        if (!ready) {
-            return {
-                ...state,
-                status: 'needs-confirmation'
-            };
+        if (retryTxDirection && !tryRetryTx) {
+            Mixpanel.track('SIGN - RETRY - retry limit exceeded');
```
MaximusHaximus, line 87: I just noticed we're using Mixpanel _inside our reducer function_ -- this is a no-go; reducers must not have side effects, this should be handled at the thunk level

MaximusHaximus (changes requested):
:+1: thanks for the clarifications!  Per our 1:1:

1. Commit the test blobs you've been using to test the reducers as fixtures that we can re-test in the future when we change this code without having to re-do the work you did around creating test cases.  Ideally with a basic set of suites that asserts low-level functionality guarantees are retained.
2. Ensure that actions have their gas increased up to the **max** of 300 TGas on retries, but **never more than 300 TGas per action**.
3. Only modify the amount of gas for each action _when the user clicks retry_, not eagerly when an error is encountered.
4. Remove DECREASE case, entirely. However, add a Mixpanel tracking event that reports any time someone gets a 'too much gas attached' error response from an attempt to send a transaction, so we can gauge whether this case even needs to be handled.  It is probably not a case we need to optimize for solving, since it will only be a problem if a DAPP author has grossly mis-configured their attached gas requirements.  We will only increase gas on actions when we know the tx failed due to not enough gas attached error
We will never increase gas beyond GAS.MAX (currently 300 TGas) per action, in steps of  to <=25TGas per retry; final retry may increase the gas allotment by <25TGas to avoid going over 300TGas on the action
If we encounter a ‘too much gas attached’ error or we get back a ‘not enough gas attached’ error but all of our actions are at 300 TGas already, then we show the user a fatal error and don’t allow them to retry again.

Using this logic, these are our fatal ‘end of flow’ cases:
A. Sorry, we’ve increased the gas on the actions in this TX as much as we possibly can, and they are still failing to process due to not enough gas; there’s nothing else we can do here as it’s a problem w/ the smart contract being too expensive.
B. Sorry, we’ve tried increasing the gas on the actions in this TX, but by doing so we have exceeded the TX level gas limit.  In this case we’ve done the best we can with a simple algorithm and no way to tell specifically which actions we need to increase gas for to succeed.

Also, please test what happens if the first transaction is submitted correctly without any error, but the second transaction fails due to not enough gas.  If i'm reading the code here in `wallet.signAndSendTransactions` correctly, this case would result in the first transactions being submitted _twice_ if the second transaction fails due to not-enough-gas-attached error.

packages/frontend/src/components/sign/SignTransferRetry.js
marcinbodnar: Sure, I just prefer to keep the same name for one method across components for one page (we have a `Deny` button on first screen, `Cancel` button on the retry tx screen, we also `Close` button on the cancel page, all of them with the same method :) ). Maybe call this method `handleCancel` everywhere (I did for now)?

packages/frontend/src/redux/reducers/sign/index.js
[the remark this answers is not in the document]
marcinbodnar: Done.

packages/frontend/src/redux/reducers/sign/index.js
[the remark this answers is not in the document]
marcinbodnar: Done.

packages/frontend/src/redux/reducers/sign/index.js
marcinbodnar: Done.

marcinbodnar:
@MaximusHaximus All done.

I could have simplified a code a bit by removing `INCREASE` everywhere (as this is the only thing we are doing now) but I thought the code is a bit more readable with this const.

packages/frontend/src/components/sign/SignTransferRetry.js
MaximusHaximus: handleCancel sounds reasonable for the local scope inside each component :+1:

packages/frontend/src/redux/reducers/sign/index.js
```diff
@@ -25,43 +30,62 @@ const sign = handleActions({
...
+        
+        const tryRetryTx = retryTxDirection && !state.transactions.every((t) => 
+            t.actions && t.actions.every((a) => a.functionCall && a.functionCall.gas && (
+                state.retryTxDirection === RETRY_TX.INCREASE 
+                && (
+                    a.functionCall.gas.gt(new BN(RETRY_TX.GAS.MAX))
+                    || a.functionCall.gas.eq(new BN(RETRY_TX.GAS.MAX))
+                )
+            ))
+        );
```
MaximusHaximus, line 80: 1. Please don't create so many levels of nested expressions; they are hard to read and difficult to reason about; they tend to hide redundancy and make code non-linear -- please expand your expressions into function leveraging early excit/short-circuits and consider readability as a priority (add comments describing the WHY for code we are writing)
2. Please remove the unused complexity of using `retryTxDirection` here.
3. `BN.js` has a `.gte` method; no need to use an OR here, but the check can also be simplified by removing the negation and using `BN.lt()` instead.
4. This current code will treat actions that are _not_ function call type actions as a reason to not allow retrying of the parent transaction, even though there may be only non-function call actions and some function call actions that can have their gas increased; non-functionCall actions shouldn't be considered when computing `tryRetryTx`
5. Per my prior comments, _what will happen if some of the transactions are successfully submitted and the user clicks 'Retry' at the top level_?  Will we be double-submitting the same transactions for the user when they click 'retry'?  If so, we are missing some state management; we should either mark each transaction as we successfully send them so that we can filter them out from the retry logic, or disable the user's ability to click Retry if any of the transactions have been successfully submitted.
6. Rather than negating `every()`, please use `.some()` and use affirmitive logic -- it's more efficient (it exits as soon as one element matches the condition rather than iterating all the way to the end), and is idiomatic language (immediately recognizable grammar).

Something addressing # 1- # 4 above might look something like:
```javascript
const notEnoughGasAttached = error.message.includes('Exceeded the prepaid gas');
const hasAtLeastOneFunctionCallAction = state.transactions.some((t) => {
    return (t.actions || []).some((a) => a && a.functionCall);
});

const canRetryWithIncreasedGas = notEnoughGasAttached && hasAtLeastOneFunctionCallAction && state.transactions.some((t) => {
        if (!t.actions) { return false; }

        return t.actions.some((a) => {
            // We can only increase gas for actions that are function calls and still have < RETRY_TX.GAS.MAX gas allocated
            if (!a || !a.functionCall) { return false; }

            return a.functionCall.gas.lt(new BN(RETRY_TX.GAS.MAX));
        });
    }
);
```

packages/frontend/src/redux/reducers/sign/index.js
```diff
@@ -25,43 +30,62 @@ const sign = handleActions({
...
+        
         return {
             ...state,
-            status: payload.status
+            status: tryRetryTx
+                ? SIGN_STATUS.RETRY_TRANSACTION
+                : SIGN_STATUS.ERROR,
+            error,
+            retryTxDirection: state.retryTxDirection || retryTxDirection
```
MaximusHaximus, line 88: We should be able to just use `state.status` to know what to do now that there is only one direction to go. (🔪 `retryTxDirection`)

packages/frontend/src/redux/slices/sign/index.js
```diff
@@ -1,4 +1,138 @@
...
+};
+
+export const handleSignTransactions = createAsyncThunk(
+    `${SLICE_NAME}/handleSignTransactions`,
+    async (_, thunkAPI) => {
+        const { dispatch, getState } = thunkAPI;
+        let transactionsHashes;
+        const retryTxDirection = selectSignRetryTxDirection(getState());
+
+        const mixpanelName = `SIGN${retryTxDirection ? ` - RETRY - ${retryTxDirection}` : ''}`;
```
MaximusHaximus, line 36: There is only one direction; this action can now be `SIGN` or `SIGN - RETRY WITH INCREASED GAS` for clarity.

packages/frontend/src/redux/slices/sign/index.js
```diff
@@ -1,4 +1,138 @@
...
+        MAX: '300000000000000',
+    }
+};
+
+export const handleSignTransactions = createAsyncThunk(
+    `${SLICE_NAME}/handleSignTransactions`,
+    async (_, thunkAPI) => {
+        const { dispatch, getState } = thunkAPI;
+        let transactionsHashes;
+        const retryTxDirection = selectSignRetryTxDirection(getState());
```
MaximusHaximus, line 34: :hocho:

packages/frontend/src/redux/slices/sign/index.js
```diff
@@ -1,4 +1,138 @@
...
+                    ? a.functionCall.gas = new BN(RETRY_TX.GAS.MAX)
+                    : a.functionCall.gas = BN.min(
+                        new BN(RETRY_TX.GAS.MAX),
+                        a.functionCall.gas.add(new BN(RETRY_TX.GAS.DIFF))
+                    );
+            }
+        });
+    });
+    return transactions;
+};
```
MaximusHaximus, line 97: 1. It doesn't make sense for `increaseGasForTransactions` to take a `retryTxDirection`.  If we ever call this method, we expect gas to be increased; :hocho: `retryTxDirection`

2. Please unravel these expressions and use early exits to remove indentation levels.  By doing so, you will also add nice convenient spaces to add comments that explain _why_ we are doing the things we are doing here for later.  For example, the reason why we don't always increase every action's gas to GAS.MAX is an important one, but someone would have to guess why based on the current code.  Let's be nice to our future selves and future new team members who need to read this later.

Something like e.g.:
```javascript
export const increaseGasForTransactions = ({ transactions, retryTxDirection }) => {
    transactions.forEach((t) => {
        const oneFunctionCallAction = t.actions && t.actions.filter((a) => !!a.functionCall).length === 1;

        t.actions && t.actions.forEach((a) => {
            if (!(a.functionCall && a.functionCall.gas)) { return false; }

            if (oneFunctionCallAction) {
                // If we only have a single functionCall type action, it is probably safe to immediately try GAS.MAX for that one action
                a.functionCall.gas = new BN(RETRY_TX.GAS.MAX);
                return;
            }

            // If there are more than one functionCall type actions, we will try to incrementally increase
            // gas allocated -- we don't increase straight to GAS.MAX because it is highly likely that we will
            // attach too much gas and encounter a `TotalPrepaidGasExceeded` type of error
            a.functionCall.gas = BN.min(
                new BN(RETRY_TX.GAS.MAX),
                a.functionCall.gas.add(new BN(RETRY_TX.GAS.DIFF))
            );
        });
    });
    return transactions;
};
```

MaximusHaximus (changes requested):
Comments added in-line.  Apart from those comments:

1. I didn't see any mention/response to my question: ```Also, please test what happens if the first transaction is submitted correctly without any error, but the second transaction fails due to not enough gas. If i'm reading the code here in wallet.signAndSendTransactions correctly, this case would result in the first transactions being submitted twice if the second transaction fails due to not-enough-gas-attached error.```

What happens in this case?

2. Please commit the test blobs that you are using to test your reducer logic with some basic assertions so that the next folks who touch this code don't need to re-do the work of identifying test cases to exercise the code.

packages/frontend/src/redux/reducers/sign/index.js
```diff
@@ -25,43 +30,62 @@ const sign = handleActions({
...
-    [signAndSendTransactions]: (state, { error, payload, ready }) => {
-
-        if (!ready) {
-            return {
-                ...state,
-                status: 'needs-confirmation'
-            };
-        }
+    [handleSignTransactions.pending]: (state) => {
+        const { retryTxDirection, status } = state;
```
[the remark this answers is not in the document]
marcinbodnar: Done.

packages/frontend/src/redux/reducers/sign/index.js
marcinbodnar: 1. You're right, force of habit :) Will keep that in mind.
2. Done.
3. Done.
4. Done.
5. Will post the answer in the previous comment, that I missed.
6. Done.
Thanks for the example.

packages/frontend/src/redux/reducers/sign/index.js
marcinbodnar: Done.

packages/frontend/src/redux/slices/sign/index.js
marcinbodnar: Done.

packages/frontend/src/redux/slices/sign/index.js
marcinbodnar: Done.

packages/frontend/src/redux/slices/sign/index.js
marcinbodnar: Done. Thanks for the example.

marcinbodnar:
@MaximusHaximus 
> Also, please test what happens if the first transaction is submitted correctly without any error, but the second transaction fails due to not enough gas. If i'm reading the code here in wallet.signAndSendTransactions correctly, this case would result in the first transactions being submitted twice if the second transaction fails due to not-enough-gas-attached error.

You're right.

> we should either mark each transaction as we successfully send them so that we can filter them out from the retry logic, or disable the user's ability to click Retry if any of the transactions have been successfully submitted.

Right. I think that we should put in additional effort, and mark the transaction that succeeded and try to retry the ones that failed.

marcinbodnar:
@MaximusHaximus 
> Please commit the test blobs that you are using to test your reducer logic with some basic assertions so that the next folks who touch this code don't need to re-do the work of identifying test cases to exercise the code.

So basically, I have a new action, that is setting (only once) the gas for a low value for one or a few actions, depending on the flow that I want to test, and for some situations, I'm changing the amount of gas that we are adding, and for example, to test situation with two actions (the flow where we want to reach the max for every action) I'm changing the `wallet.signAndSendTransaction` so it will throw error always. I'm not sure how to add it to the code, so it will make sense - maybe I will add a comment explaining how to do it? I think it might be better because it needs changes in 4 different files?

marcinbodnar:
@MaximusHaximus 
Added a proposition of how we can handle retry tx with multiple transactions. It's ready in this commit: 
https://github.com/near/near-wallet/pull/2216/commits/501e5ad096e76f369ffff689062e8ff4b3a98a93

It's a bit complicated but I tried to make it as much readable as possible, added comments, etc.

Basically, at first `insufficient gas` error, I'm creating a copy of transactions array, that will be used to change gas and also will hold only transactions that are not succeded yet. I'm also storing the information about the transactions that succeeded. So with that change, we always store the original transactions, and with every retry we can see which transactions succeeded, which are still waiting. 

Additional change, that we didn't handle before is that I'm increasing gas only for the first transaction (that has `functionCall` action and is not succeded yet) - before we were increasing gas for all transactions which wasn't a good idea, because if the `insufficient gas` error appears we know that it's for the first transaction only, and by increasing other transactions gas, we can make these transactions fails.

I tried to make this code readable and easy to reason about, added comments, etc. It still can be improved by a bit but I would like to know first what do you think about it, do we want to use it, or just don't allow to submit if the error appears and some transactions have already succeeded. I think we should allow the user to retry, because transactions that were successful might be important, and we should try to finish other transactions as well.

Patrick1904:
@marcinbodnar @MaximusHaximus any idea when we'd be able to ship this PR? Would be real sweet to get the new [/sign page UI](https://github.com/near/near-wallet/pull/2215) in before Christmas

marcinbodnar:
> @marcinbodnar @MaximusHaximus any idea when we'd be able to ship this PR? Would be real sweet to get the new [/sign page UI](https://github.com/near/near-wallet/pull/2215) in before Christmas

@Patrick1904 It needs an additional round of review, because of the recent changes regarding signing multiple txs. @MaximusHaximus Please check, if there will be any additional changes needed, I can handle it right away.

MaximusHaximus (approved):
Let's get this into staging & testnet given the time this has been pending
We should still have basic tests on these reducers -- please take a look at https://redux.js.org/usage/writing-tests and add them in a separate PR @marcinbodnar -- if you have any questions about the tests just drop me a note and I'm happy to chat synchronously if you'd like :)

marcinbodnar:
> Let's get this into staging & testnet given the time this has been pending We should still have basic tests on these reducers -- please take a look at https://redux.js.org/usage/writing-tests and add them in a separate PR @marcinbodnar -- if you have any questions about the tests just drop me a note and I'm happy to chat synchronously if you'd like :)

@MaximusHaximus Added an issue #2345 , will handle it now.

PR #2223, opened 2021-10-29
Selectors for account actions

marcinbodnar:


packages/frontend/src/redux/actions/account.js
```diff
@@ -410,7 +442,7 @@ export const {
 });
 
 export const checkAndHideLedgerModal = () => async (dispatch, getState) => {
-    const { modal } = getState().ledger;
+    const modal = selectLedgerModal(getState());
     if (modal?.show) {
```
andy-haynes, line 446: Since `modal` can no longer be `undefined` here it might be worth removing the `?`.

packages/frontend/src/redux/actions/account.js
```diff
@@ -510,8 +545,8 @@ export const finishAccountSetup = () => async (dispatch, getState) => {
     if (promptTwoFactor) {
         dispatch(redirectTo('/enable-two-factor', { globalAlertPreventClear: true }));
     } else {
-        if (url?.redirectUrl) {
-            window.location = `${url.redirectUrl}?accountId=${accountId}`;
+        if (!!redirectUrl) {
```
andy-haynes, line 548: The `!!` can be omitted here for better consistency with other checks in this module.

packages/frontend/src/redux/actions/account.js
```diff
@@ -608,15 +643,15 @@ export const refreshAccount = (basicData = false) => async (dispatch, getState)
     }
 };
 
-export const switchAccount = ({ accountId }) => async (dispatch, getState) => {
+export const switchAccount = ({ accountId }) => async (dispatch) => {
```
andy-haynes, line 646: ❤️

packages/frontend/src/redux/actions/account.js
```diff
@@ -625,12 +660,12 @@ export const getAvailableAccountsBalance = () => async (dispatch, getState) => {
 
     for (let i = 0; i < availableAccounts.length; i++) {
         const accountId = availableAccounts[i];
-        if (!accountsBalance || !accountsBalance[accountId]) {
+        if (!accountsBalance[accountId]) {
             i < 0 && await dispatch(setAccountBalance(accountId));
```
andy-haynes, line 664: Irrelevant to your changes here but I'm curious 🙂  - can this ever be executed if `i` starts at 0 and only ever gets incremented?

packages/frontend/src/redux/actions/account.js
```diff
@@ -159,34 +189,36 @@ export const redirectToApp = (fallback) => async (dispatch, getState) => {
     }));
 };
 
-
 export const allowLogin = () => async (dispatch, getState) => {
-    const { account } = getState();
-    const { url } = account;
-    const { success_url, public_key, title } = url;
+    const accountId = selectAccountId(getState());
+    const contract_id = selectAccountUrlContractId(getState());
```
andy-haynes, line 194: It would be good for consistency to update the snake_case variables to camelCase now that the new selectors have abstracted away the property access.

andy-haynes (changes requested):
Good stuff, just a few minor comments inline but otherwise looking great 🚀

marcinbodnar:
Thanks for the review @andy-haynes , I will handle changes tomorrow.

packages/frontend/src/redux/actions/account.js
marcinbodnar: Nice catch. Done.

packages/frontend/src/redux/actions/account.js
marcinbodnar: Done.

packages/frontend/src/redux/actions/account.js
marcinbodnar: Right, fixed all similar issues in this file.

packages/frontend/src/redux/actions/account.js
marcinbodnar: :) Correct, this was a part of the functionality that I prepared some time ago - we wanted to show balances for several first accounts when the user opens submenu (and has more than one account) - after that, we were worried, that it might affect efficiency, because balance calculation contained lockup and staking, and the fetch was heavy, so we disabled fetching it automatically, but allows user to fetch it manually until we figure out how to make it work faster. So that's why we have `i < 0` :) 

Now by default balances in the submenu are hidden, so I guess we don't need this functionality at all. @stefano are we planning to get back to showing balances by default in the submenu in near future?

Anyway, not a blocker for this PR, I will remove it after Stefano's confirmation in a separate PR.

packages/frontend/src/redux/actions/account.js
```diff
@@ -107,29 +137,29 @@ export const handleRefreshUrl = (prevRouter) => (dispatch, getState) => {
...
+        const { transactions, callbackUrl, meta } = selectAccountUrl(getState());
         if (transactions) {
             dispatch(parseTransactionsToSign({ transactions, callbackUrl, meta }));
         }
     }
 };
 
 const checkContractId = () => async (dispatch, getState) => {
-    const { contract_id, failure_url } = getState().account.url;
+    const { contractId, failureUrl } = selectAccountUrl(getState());
```
andy-haynes, line 148: It looks like these names would need to stay snake_cased (or be renamed in the destructure), but maybe it's even better to double down on the selectors and add a `selectAccountUrlFailureUrl` selector and get the values for these two separately, something like
```
const contractId = selectAccountUrlContractId(getState());
const failureUrl = selectAccountUrlFailureUrl(getState());
```

We now have selectors for two other URLs on the account URL object, would it make sense to add one for `failure_url` as well?

andy-haynes (changes requested):
Just one question on a rename/potential selector opportunity, otherwise good to go

packages/frontend/src/redux/actions/account.js
andy-haynes: Thanks for the context!

packages/frontend/src/redux/actions/account.js
marcinbodnar: Wow, what a stupid bug from my side... Thanks for catching!

andy-haynes (approved):

PR #2227, opened 2021-11-03
support loading NFT image from data: URI

Kouprin:
This PR updates visualization of NFT for displaying the data which is taken from NFT metadata directly.

Testing plan
===

Tested on `nenene.testnet`
image

Data loaded from metadata: `data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHdpZHRoPSI1MCIgaGVpZ2h0PSI1MCI+PGNpcmNsZSBjeD0iMjUiIGN5PSIyNSIgcj0iMjAiIHN0cm9rZT0ib3JhbmdlIiBzdHJva2Utd2lkdGg9IjQiIGZpbGw9InllbGxvdyIgLz48L3N2Zz4=`

### DIFF
--- a/packages/frontend/src/services/NonFungibleTokens.js
+++ b/packages/frontend/src/services/NonFungibleTokens.js
@@ -52,7 +52,7 @@ export default class NonFungibleTokens {
                 if (base_uri) {
                     mediaUrl = `${base_uri}/${media}`;
                 } else {
-                    mediaUrl = `https://cloudflare-ipfs.com/ipfs/${media}`;
+                    mediaUrl = (media.substr(0, 10) === 'data:image') ? media : `https://cloudflare-ipfs.com/ipfs/${media}`;
                 }
             } else {
                 mediaUrl = media;
--- a/packages/frontend/src/services/NonFungibleTokens.js
+++ b/packages/frontend/src/services/NonFungibleTokens.js
@@ -52,7 +52,7 @@ export default class NonFungibleTokens {
                 if (base_uri) {
                     mediaUrl = `${base_uri}/${media}`;
                 } else {
-                    mediaUrl = (media.substr(0, 10) === 'data:image') ? media : `https://cloudflare-ipfs.com/ipfs/${media}`;
+                    mediaUrl = media.startsWith('data:image') ? media : `https://cloudflare-ipfs.com/ipfs/${media}`;
                 }
             } else {
                 mediaUrl = media;


packages/frontend/src/services/NonFungibleTokens.js
```diff
@@ -52,7 +52,7 @@ export default class NonFungibleTokens {
                 if (base_uri) {
                     mediaUrl = `${base_uri}/${media}`;
                 } else {
-                    mediaUrl = `https://cloudflare-ipfs.com/ipfs/${media}`;
+                    mediaUrl = (media.substr(0, 10) === 'data:image') ? media : `https://cloudflare-ipfs.com/ipfs/${media}`;
```
vgrichina, line 55: nit: `media.startsWith('data:image')`

vgrichina (approved):

Kouprin:
@MaximusHaximus are you planning to review the code?

MaximusHaximus:
@Kouprin thanks for the submission! sorry about the delay, we're flat-out right now :)
Can you verify that a <script> tag encoded inside of the SVG XML is not evaluated by the Wallet frontend?

Kouprin:
@MaximusHaximus thank you for your very good question!
According to experiments and some research, I've found an NFT which is minted with svg <script> exploit (https://explorer.testnet.near.org/transactions/BpB55sUxDiAqfctYBfDYdxMtBQ8PvV22Jurj8dz3iTgG). It is executed if it is opened in a new tab `data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHdpZHRoPSI1MCIgaGVpZ2h0PSI1MCI+PGNpcmNsZSBjeD0iMjUiIGN5PSIyNSIgcj0iMjAiIHN0cm9rZT0ib3JhbmdlIiBzdHJva2Utd2lkdGg9IjQiIGZpbGw9InllbGxvdyIgLz48c2NyaXB0IHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyIgdHlwZT0idGV4dC9qYXZhc2NyaXB0Ij5hbGVydCgiU1ZHIFhTUyBUcmlnZ2VyZWQiKTs8L3NjcmlwdD48L3N2Zz4=`
Due to complexity of handling svg scenarios (https://svg.digi.ninja/svg) and having lack of information about other formats that may be dangerous, I suggest using only whitelisted formats. I'd consider png only at the beginning and have a possibility to extend the list in future if needed.
What do you think?

vgrichina:
> It is executed if it is opened in a new tab 

this is not a problem as we aren't opening it in new tab, we are loading it in 

if opened in new tab it's not gonna get wallet.near.org context.

> Due to complexity of handling svg scenarios (https://svg.digi.ninja/svg) and having lack of information about other formats that may be dangerous, I suggest using only whitelisted formats. I'd consider png only at the beginning and have a possibility to extend the list in future if needed.

the site that you linked shows at least 2 safe ways to do it:
1) Image Tags - not vulnerable - The SVG is referenced through image tags which prevent scripts.   [this is what we do in wallet right now, so it's all safe]
2) Image in a sandboxed iframe - not vulnerable - The SVG is loaded as the source for the iframe but the sandbox attribute is set to block scripts. [this is what we can do to enable interactive SVG]

Kouprin:
@MaximusHaximus @vgrichina please come to the final decision. The feature overall is important, desirable and easy to introduce. Even supporting pngs is a huge step support NFTs that are minted in a better way. I am interested and asking you to go with the most pragmatic and reasonable solution, not spending weeks for investigating. Thank you.

vgrichina:
note that importantly this doesn't introduce any extra attack surface as we already load images this way for fungible and non-fungible tokens icons

Kouprin:
Any progress here? We've proven that icons for FTs and NFTs are displayed the same way.

MaximusHaximus:
Thanks for your patience -- we're going to be rolling this out in our next release @Kouprin  :). It'll be live on testnet wallet and mainnet staging wallet (https://wallet.staging.near.org/) in the next few minutes :+1:

Kouprin:
@MaximusHaximus amazing news! Thank you very much for enabling it. We are going to mint NFTs that will be displayed with no using 3rd parties anymore.

PR #2237, opened 2021-11-10
Create w/ link + coin-op 1.5: handle error

Patrick1904:
**Problem**
We've had reports that sometimes creating an identity funded account w/ coin-op 1.5 returns an error, but the account is still successfully created. Currently we automatically send the user to the 'self funded' options page if an error occurs, which causes confusion since the account is already created.

**Solution**
This fix catches the error returned when calling `createIdentityFundedAccount`. Checks if the account exists on chain. If the account exists, it checks that the users public key also exists on the account, to verify ownership. It then attempts to finish the 'post account creation' setup by adding a local key, etc. If that fails, the user is redirected to the `/recover-account` route, for manual account recovery.

If the account exists, but the user is not the owner, an error is displayed on the UI, informing the user that the account already exists.

If the account does not exist, the user is then redirected to the 'self funded' options page (current UX).

### DIFF
--- a/packages/frontend/src/components/accounts/recovery_setup/SetupRecoveryMethod.js
+++ b/packages/frontend/src/components/accounts/recovery_setup/SetupRecoveryMethod.js
@@ -307,6 +307,7 @@ class SetupRecoveryMethod extends Component {
             }
         } catch (e) {
             this.setState({ settingUpNewAccount: false });
+            throw e;
         }
     }
 
--- a/packages/frontend/src/components/accounts/recovery_setup/SetupRecoveryMethod.js
+++ b/packages/frontend/src/components/accounts/recovery_setup/SetupRecoveryMethod.js
@@ -245,34 +245,41 @@ class SetupRecoveryMethod extends Component {
                     });
                 } catch (e) {
                     console.warn(e.code);
-                    const isNewAccount = await checkIsNew(accountId);
-                    if (!isNewAccount) {
-                        const accessKeys = await wallet.getAccessKeys(accountId) || [];
-                        const publicKeys = accessKeys.map(key => key.public_key);
-                        const publicKeyExistsOnAccount = publicKeys.includes(recoveryKeyPair.publicKey.toString());
 
-                        if (publicKeyExistsOnAccount) {
-                            try {
-                                await wallet.saveAndMakeAccountActive(accountId);
-                                await wallet.addLocalKeyAndFinishSetup(accountId, method.kind, recoveryKeyPair.publicKey);
-                            } catch (e) {
-                                showCustomAlert({
-                                    success: false,
-                                    messageCodeHeader: 'error',
-                                    errorMessage: e.message
-                                });
-                                redirectTo('/recover-account');
-                            }
-                        } else {
-                            showCustomAlert({
-                                success: false,
-                                messageCodeHeader: 'error',
-                                messageCode: 'walletErrorCodes.createNewAccount.accountExists.error',
-                                errorMessage: e.message
-                            });
-                        }
-                    } else {
+                    const isNewAccount = await checkIsNew(accountId);
+                    if (isNewAccount) {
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/accounts/recovery_setup/SetupRecoveryMethod.js
```diff
@@ -246,7 +245,35 @@ class SetupRecoveryMethod extends Component {
...
+                                success: false,
+                                messageCodeHeader: 'error',
+                                messageCode: 'walletErrorCodes.createNewAccount.accountExists.error',
+                                errorMessage: e.message
+                            });
+                        }
+                    } else {
+                        await fundCreateAccount(accountId, recoveryKeyPair, method.kind);
+                    }
                 }
```
MaximusHaximus, line 277: The level of nesting here is getting a bit out of hand :D. By unrolling some of these if/else and using short-circuits we can make this logic a lot more linear -- e.g. something like this:
```suggestion
                    const isNewAccount = await checkIsNew(accountId);

                    if (isNewAccount) {
                        // No on-chain account exists; fallback to implicit or verified account flow
                        await fundCreateAccount(accountId, recoveryKeyPair, method.kind);
                        return;
                    }

                    // on-chain account exists; verify that it is owned by the current key
                    const accessKeys = await wallet.getAccessKeys(accountId) || [];
                    const publicKeys = accessKeys.map(key => key.public_key);
                    const publicKeyExistsOnAccount = publicKeys.includes(recoveryKeyPair.publicKey.toString());

                    if (!publicKeyExistsOnAccount) {
                        // Someone else must've created the account (or a two-tabs-open situation)
                        showCustomAlert({
                            success: false,
                            messageCodeHeader: 'error',
                            messageCode: 'walletErrorCodes.createNewAccount.accountExists.error',
                            errorMessage: e.message
                        });
                        return;
                    }

                    // Assume a transient error occurred, but that the account is on-chain and we can finish the creation process
                    try {
                        await wallet.saveAndMakeAccountActive(accountId);
                        await wallet.addLocalKeyAndFinishSetup(accountId, method.kind, recoveryKeyPair.publicKey);
                    } catch (e) {
                        showCustomAlert({
                            success: false,
                            messageCodeHeader: 'error',
                            errorMessage: e.message
                        });
                        redirectTo('/recover-account');
                    }
                }
```

Also some comments can be nice in code here re: 'what might this case have been caused by?' can be really useful for future investigations -- for example, since we know that the most likely case for !publicKeyExistsOnAccount is that someone else created the named account or someone had two tabs open and claimed it in a tab with a different seed phrase than the one they wrote down, it's worth noting that for future devs who are working in the code

packages/frontend/src/components/accounts/recovery_setup/SetupRecoveryMethod.js
```diff
@@ -246,7 +245,35 @@ class SetupRecoveryMethod extends Component {
...
+                            } catch (e) {
+                                showCustomAlert({
+                                    success: false,
+                                    messageCodeHeader: 'error',
+                                    errorMessage: e.message
+                                });
+                                redirectTo('/recover-account');
+                            }
+                        } else {
+                            showCustomAlert({
```
MaximusHaximus, line 267: Do we need to re-throw from here to ensure that `this.setState({ settingUpNewAccount: false });` is called on L302, since we're not redirecting in this branch?

MaximusHaximus (changes requested):
Added a suggested approach in-line to reduce the nesting complexity and make the code flow linearly, and one question about state management in the error case

packages/frontend/src/components/accounts/recovery_setup/SetupRecoveryMethod.js
Patrick1904: Thanks you ❤️

MaximusHaximus (approved):
<3 :+1: LGTM m8

PR #2238, opened 2021-11-10
Improve and fix translation typo

picturepan2:
- [ ] Improve and fix translation typo in both Chinese.


Patrick1904 (approved):
This is great! 💯

PR #2253, opened 2021-11-16
Create account w/ initial deposit: handle error

Patrick1904:


Patrick1904:
@MaximusHaximus I added similar error handling to account creation with initial deposit as we added for [coin op](https://github.com/near/near-wallet/pull/2237). PTAL when you have a chance.

packages/frontend/src/components/accounts/create/initial_deposit/InitialDepositWrapper.js
```diff
@@ -99,9 +101,55 @@ export function InitialDepositWrapper({ history }) {
...
+                    await wallet.saveAndMakeAccountActive(accountId);
+                    await wallet.addLocalKeyAndFinishSetup(accountId, recoveryMethod, publicKey);
+                } catch (e) {
+                    dispatch(showCustomAlert({
+                        success: false,
+                        messageCodeHeader: 'error',
+                        messageCode: 'walletErrorCodes.createNewAccount.accountCreated.error',
+                        errorMessage: e.message
+                    }));
+                    redirectTo('/recover-account');
```
MaximusHaximus, line 147: Should this be using a `dispatch()`?

MaximusHaximus (changes requested):
:+1: Just one question about not using `dispatch()` for the `redirectTo` action

packages/frontend/src/components/accounts/create/initial_deposit/InitialDepositWrapper.js
Patrick1904: Great catch:)

PR #2259, opened 2021-11-18
App login: Add info about app key allowance

Patrick1904:
Screen Shot 2021-11-18 at 2 24 43 PM

### DIFF
--- a/packages/frontend/src/translations/en.global.json
+++ b/packages/frontend/src/translations/en.global.json
@@ -974,6 +974,10 @@
                 "fullAccessModal": {
                     "title": "Grant full account access?",
                     "desc": "Proceed with caution. Full access permissions allow contracts and applications to access your token balance. To confirm, enter your account ID below."
+                },
+                "feeAllowance": {
+                    "title": "Network Fee Allowance",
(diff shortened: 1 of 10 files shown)


corwinharrell (changes requested):
Looks good. Can we kill the line-break for the heading?

packages/frontend/src/translations/en.global.json
```diff
@@ -974,6 +974,10 @@
                 "fullAccessModal": {
                     "title": "Grant full account access?",
                     "desc": "Proceed with caution. Full access permissions allow contracts and applications to access your token balance. To confirm, enter your account ID below."
+                },
+                "feeAllowance": {
+                    "title": "Network Fee Allowance",
+                    "desc": "The application is requesting permission to use up to ${amount} NEAR for network fees incurred during use."
```
corwinharrell, line 980: Let's change it to:

> The application will be given permission to spend up to **0.25 NEAR** towards network fees incurred during use.

MaximusHaximus (approved):
This is a huge improvement and is a lot clearer what is going on here, nice :+1:

PR #2260, opened 2021-11-20
Add sign message functionality

H34D:
enable signing messages instead of transactions and return a signature in base64 in the callbackUrl

### DIFF
--- a/packages/frontend/src/components/sign/Sign.js
+++ b/packages/frontend/src/components/sign/Sign.js
@@ -5,7 +5,7 @@ import { connect } from 'react-redux';
 import { withRouter } from 'react-router-dom';
 
 import { Mixpanel } from '../../mixpanel';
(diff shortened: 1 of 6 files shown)


stefanopepe:
@H34D Sebastian, this is a great contribution that we missed mostly because it's not paired with an issue introducing the user story and the acceptance criteria.

Would you be available to open one? I'll make sure that as soon as we have it, we'll provide feedback

H34D:
> 

sure, i will do that

H34D:
> @H34D Sebastian, this is a great contribution that we missed mostly because it's not paired with an issue introducing the user story and the acceptance criteria.
> 
> Would you be available to open one? I'll make sure that as soon as we have it, we'll provide feedback

sure, i will do that

H34D:
Done #2435

Issue #2210, opened 2021-10-21
Removing lockup account breaks /staking tab


AngelBlock:
**Problem**
Removing lockup account by utilising Wallet feature to move funds (green banner) breaks /staking path

<img width="476" alt="Screenshot 2021-10-21 at 10 45 10" src="https://user-images.githubusercontent.com/80400686/138253381-8d69dad7-5aa9-43d5-9cc5-55db33f7a3ef.png">

**Expected Behavior**
Should display staking pools, but showing an empty page

**Steps to reproduce**

1. When lockup term is up, go to /profile
2. You will see the green banner `You have XXX available to withdraw from your lockup!`
3. Click on Transfer to wallet
4. Accept add full access key to lockup
5. Accept delete lockup account
6. Go to /staking path.
7. Empty page is shown.

MaximusHaximus:
@esaminu Could you take a look into this and make sure we have an e2e test that covers this case?

marcinbodnar:
Fixed in #2247

PR #2264, opened 2021-11-23
Lockup e2e tests

### LINKED ISSUE #2180: Adding E2E tests for lockups on wallet
As we begin to add e2e tests for lockup functionality on the wallet I'd like to raise / confirm the following questions:
- In order to add E2E tests for wallet lockup flows we need to use `testingLockup.{accountId}`  and build the web server with the `REACT_APP_USE_TESTINGLOCKUP` env variable set. We can't test `{sha(accountId)[:40]}.lockup.near` accounts since the test suite can't create them. This also means we can't test lockups in prod (testnet/mainnet..) since this env variable is not set. Additionally it means we can only have one lockup per account since we are using the predictable naming. Maybe it would have been possible to query for all lockups with `owner_account_id` as the selected account? If this was the case we could have deployed a lockup anywhere and validated our test cases in prod.
- @stefanopepe what assertions would we like to make and what scenarios would we like to test. Ideally we would also like to repro https://github.com/near/near-wallet/issues/2014. Any config values and info on how to will be helpful.

esaminu:
Adds tests for lockup withdrawal, deletion and staking flows:

- Displayed amounts are correct for locked, unlocked liquid etc (v2 and latest contracts)
- Withdrawal functionality works as expected and deletes account when fully vested
- Staking accounts updated after deletion. See #2210 
- Stake and unstake flows work and only allow staking with a single validator at a time

Closes #2180


packages/e2e-tests/utils/E2eTestAccount.js
```diff
@@ -44,6 +63,39 @@ class E2eTestAccount {
...
+            throw new Error("Account needs to be initialized to spawn sub accounts");
+        }
+        // creates a testinglockup subaccount with a lockup_timestamp (locked until) in 1 minute with a release_duration (period to linearly unlock) of 1 minute
+        const lockupSubaccountId = `testinglockup.${this.accountId}`;
+        const lockupSubaccountSeedphrase = `${lockupSubaccountId} ${process.env.TEST_ACCOUNT_SEED_PHRASE}`;
+        const lockupWasm = await fetchLockupContract({ v2Wasm });
+        let minuteInNanosBN = new BN("1").mul(new BN("60000000000"));
+        const vesting_schedule_config = vesting_schedule || FULLY_VESTED_CONFIG.vesting_schedule;
+
+        return new E2eTestAccount(lockupSubaccountId, lockupSubaccountSeedphrase, this.nearApiJsAccount).create({
```
MaximusHaximus, line 78: Epic :D

MaximusHaximus (approved):
🚀 epic PR, nice work ✋ .  We're really getting into the thick of it now, I can't wait to get these integrated with our new CI <3

PR #2266, opened 2021-11-23
Sign in w/ Ledger: Add Firefox disclaimer banner

Patrick1904:
Screen Shot 2021-11-23 at 1 44 38 PM


corwinharrell (approved):
Can we add an underline to links in our banners/alerts? Otherwise LGTM

PR #2276, opened 2021-11-30
App login: display contract ID

Patrick1904:
Screen Shot 2021-11-30 at 1 34 00 PM


MaximusHaximus (approved):
:+1: lookin' good, thanks for the quick work! :)

packages/frontend/src/components/login/v2/NetworkFeeAllowance.js
```diff
@@ -26,8 +40,14 @@ const StyledContainer = styled.div`
     }
 `;
 
-export default () => (
+export default ({ contractId, contractIdUrl }) => (
```
MaximusHaximus, line 43: Not a blocker, but we should probably rename this thing to e.g. `LoginDetails` now that it's not really specific to the  Network Fee Allowance...

PR #2279, opened 2021-12-01
Selectors for staking actions

marcinbodnar:
Set of selectors to be used for staking reducer.

### DIFF
--- a/packages/frontend/src/redux/slices/staking/index.js
+++ b/packages/frontend/src/redux/slices/staking/index.js
@@ -1,4 +1,84 @@
+import { createSelector } from "reselect";
+
+import createParameterSelector from "../createParameterSelector";
+
 const SLICE_NAME = 'staking';
 
+const getAccountIdParam = createParameterSelector((params) => params.accountId);
+const getValidatorIdParam = createParameterSelector((params) => params.validatorId);
+
 // Top level selectors
-export const selectStakingSlice = (state) => state[SLICE_NAME];
+export const selectStakingSlice = (state) => state[SLICE_NAME] || {};
+
+export const selectStakingAccounts = createSelector([selectStakingSlice], (staking) => staking.accounts || []);
+
+export const selectStakingCurrentAccountbyAccountId = createSelector(
+    [selectStakingAccounts, getAccountIdParam], 
+    (accounts, accountId) => accounts.find((account) => account.accountId === accountId)
+);
+
+export const selectStakingAccountsFirst = createSelector(
+    selectStakingAccounts,
+    (accounts) => accounts[0] || {}
+);
+
+export const selectStakingAccountsSecond = createSelector(
+    selectStakingAccounts,
+    (accounts) => accounts[1] || {}
+);
+
+export const selectStakingLockup = createSelector(
+    selectStakingSlice,
+    (staking) => staking.lockup || {}
+);
+
+export const selectStakingLockupId = createSelector(
+    selectStakingLockup,
+    (lockup) => lockup.lockupId || ''
+);
+
+export const selectStakingContract = createSelector(
+    selectStakingLockup,
+    (lockup) => lockup.contract || {}
+);
+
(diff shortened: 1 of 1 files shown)


packages/frontend/src/redux/slices/staking/index.js
```diff
@@ -1,4 +1,52 @@
...
+
+export const selectStakingAllValidatorsLength = createSelector(selectStakingAllValidators, (allValidators) => allValidators.length);
+
+export const selectStakingFindContractByValidatorId = createSelector(
+    [selectStakingAllValidators, getValidatorIdParam],
+    (allValidators, validatorId) => allValidators.find((validator) => validator.accountId === validatorId)?.contract || {}
+);
+
+// accountsObj - state
+export const selectStakingAccountsObj = createSelector(selectStakingSlice, (staking) => staking.accountsObj || {});
```
andy-haynes, line 43: I didn't notice any references to this selector outside this module, I think it's worth considering not exporting this if you don't foresee any code wanting to get the whole object back in the future.

packages/frontend/src/redux/slices/staking/index.js
```diff
@@ -1,4 +1,52 @@
...
+
+export const selectStakingAccountsObjAccountId = createSelector(selectStakingAccountsObj, (accountsObj) => accountsObj.accountId || '');
```
andy-haynes, line 45: As someone new to this codebase, the `accountsObj` state property object is a little confusing 😄 . Is there possibly a more descriptive name for what these selectors are returning that distinguishes it from the other selectors in this module?

I'm interested to hear your thoughts - it may very well make more sense to leave it as is for the time being. I just like to think about renaming when writing selectors because it can provide opportunities to make things more clear from the caller's perspective.

andy-haynes (commented):
Just have some clarifying questions from someone new to the codebase trying to understand how the current redux state works 😄

packages/frontend/src/redux/slices/staking/index.js
```diff
@@ -1,4 +1,52 @@
...
+
+// accounts - state
+export const selectStakingAccounts = createSelector(selectStakingSlice, (staking) => staking.accounts || []);
+
+export const selectStakingCurrentAccountbyAccountId = createSelector(
+    [selectStakingAccounts, getAccountIdParam], 
+    (accounts, accountId) => accounts.find((account) => account.accountId === accountId)
+);
+
+export const selectStakingAccountsFirst = createSelector(selectStakingAccounts, (accounts) => accounts[0] || {});
```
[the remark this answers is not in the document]
marcinbodnar: Good point. I didn't prepare staking functionality - at some point, I handled refactoring it to get rid of the big staking class and move most of the functionalities into thunks, also improved performance a bit.

Generally, I don't think that the solution where we have `accounts` array and then `accountObj` is optimal, but I remember that there were a lot of changes for the staking functionality, while it was prepared, and time pressure so I guess that's the reason.

It appears that the first element of the array is always the main account, and the second if exists, is the lockup account. I changed the selectors' names.

packages/frontend/src/redux/slices/staking/index.js
marcinbodnar: Done.

packages/frontend/src/redux/slices/staking/index.js
marcinbodnar: Agree, you're right. I renamed it into `selectStakingMainAccountId` and `selectStakingLockupAccountId`. 

The next step will be to finish refactoring staking actions/reducer into the redux toolkit version, I will definitely improve some of this functionality. And next, we are planning to refactor the whole redux state tree - it's a bit of a mess right now, because the project was expanding so fast, and there never was enough time to take care of it. So then I will definitely refactor the whole staking state tree.

andy-haynes (approved):
Looks good, thanks for the helpful background on the ongoing redux work 🙂

PR #2286, opened 2021-12-01
Fix getSignedUrl

marcinbodnar:


packages/frontend/src/components/buy/BuyNear.js
```diff
@@ -144,13 +144,13 @@ const StyledContainer = styled(Container)`
...
 export function BuyNear({ match, location, history }) {
-    const { accountId } = useSelector(selectAccountId);
+    const accountId = useSelector(selectAccountId);
     const [moonPayAvailable, setMoonPayAvailable] = useState(null);
     const [signedMoonPayUrl, setSignedMoonPayUrl] = useState(null);
 
     useEffect(() => {
         checkMoonPay();
-    }, []);
+    }, [accountId]);
```
Patrick1904, line 153: Should we add `accountId` into the `if` statement on `L159` too so we make sure we don't show Moonpay as available if the `accountId` is slow to load for some reason?

marcinbodnar: Added condition. I also propose to don't hide moonpay section, instead, add the loader to the button - thanks to that the content will not jump after accountId is available.

Patrick1904: Nice. While I suspect Moonpay is supported in most regions now, we probably want to handle the case where it's not. I suggest we keep the 'loading...' state until we've received a response from `isMoonpayAvailable`. If Moonpay is not supported, we can disabled the button and change the text to 'Not supported in your region'

cc @corwinharrell

marcinbodnar: I see. Added.
Screenshot 2021-12-01 at 21 59 07

Patrick1904 (approved):
Sweet! 💯

MaximusHaximus (approved):
Not a blocker, but I think we should keep our translation file structure alphabetized by key so that similar / related terms are grouped together, and we can easily compare different language files visually. :)

marcinbodnar:
> Not a blocker, but I think we should keep our translation file structure alphabetized by key so that similar / related terms are grouped together, and we can easily compare different language files visually. :)

@MaximusHaximus good idea, I'm trying to check and add missing translations for other languages if I see any, but we should apply something that will give us a clear view of what is missing.

PR #2287, opened 2021-12-01
Fungible token list: sort alphabetically

Patrick1904:
The fungible token list is randomly sorted on each page reload, which makes it hard to locate a specific token. This fix sorts the fungible token list alphabetically by `token.name` so that it's easier to locate tokens.

### DIFF
--- a/packages/frontend/src/hooks/fungibleTokensIncludingNEAR.js
+++ b/packages/frontend/src/hooks/fungibleTokensIncludingNEAR.js
@@ -19,7 +19,7 @@ const fungibleTokensIncludingNEAR = (tokens, balance, nearTokenFiatValueUSD) => 
 export const useFungibleTokensIncludingNEAR = function () {
     const balance = useSelector(selectBalance);
     const accountId = useSelector(selectAccountId);
-    const tokens = useSelector((state) => selectTokensWithMetadataForAccountId(state, { accountId })).sort((a, b) => a.name.localeCompare(b.name));
+    const tokens = useSelector((state) => selectTokensWithMetadataForAccountId(state, { accountId }));
     const nearTokenFiatValueUSD = useSelector(selectNearTokenFiatValueUSD);
 
     const balanceToDisplay = balance?.balanceAvailable;
--- a/packages/frontend/src/hooks/fungibleTokensIncludingNEAR.js
+++ b/packages/frontend/src/hooks/fungibleTokensIncludingNEAR.js
@@ -28,7 +28,7 @@ export const useFungibleTokensIncludingNEAR = function () {
     );
 
     useEffect(() => {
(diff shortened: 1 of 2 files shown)


packages/frontend/src/hooks/fungibleTokensIncludingNEAR.js
```diff
@@ -28,7 +28,7 @@ export const useFungibleTokensIncludingNEAR = function () {
     );
 
     useEffect(() => {
-        setFungibleTokensList(fungibleTokensIncludingNEAR(tokens, balanceToDisplay, nearTokenFiatValueUSD));
+        setFungibleTokensList(fungibleTokensIncludingNEAR(tokens.sort((a, b) => a.name.localeCompare(b.name)), balanceToDisplay, nearTokenFiatValueUSD));
```
andy-haynes, line 31: Would it make sense to move the `.sort()` to the `tokens` declaration on line 22 so that the initial state also gets the sorted list?

Patrick1904: Great point:)

andy-haynes (approved):

packages/frontend/src/hooks/fungibleTokensIncludingNEAR.js
```diff
@@ -19,7 +19,7 @@ const fungibleTokensIncludingNEAR = (tokens, balance, nearTokenFiatValueUSD) =>
 export const useFungibleTokensIncludingNEAR = function () {
     const balance = useSelector(selectBalance);
     const accountId = useSelector(selectAccountId);
-    const tokens = useSelector((state) => selectTokensWithMetadataForAccountId(state, { accountId }));
+    const tokens = useSelector((state) => selectTokensWithMetadataForAccountId(state, { accountId })).sort((a, b) => a.name.localeCompare(b.name));
```
MaximusHaximus, line 22: Ideally this sorting should happen _inside the selector_, since they are memoized  (and it would be consistent with how NFT tokens are selected, see https://github.com/near/near-wallet/blob/4b6446273588841e0a4a923a8890accb6dfe6b82/packages/frontend/src/redux/slices/nft/index.js#L205-L224 )

MaximusHaximus (changes requested):
<3 stable sorting will be nice -- I think we should be doing the sorting inside of the selector, though, so it's memoized, consistent with how NFTs are selected, and makes the sorting of these values consistent _everywhere_ we use the `selectTokensWithMetadataForAccountId` selector

packages/frontend/src/hooks/fungibleTokensIncludingNEAR.js
Patrick1904: Makes sense - moved it 👍

MaximusHaximus (approved):
:+1: rock on

PR #2298, opened 2021-12-09
allowLogin: use wallet.accountId for faster access on slow connection

Patrick1904:
Currently using `accountId` from redux and can result in `undefined` on a slow connection. `wallet.accountId` is available right when the UI renders.


MaximusHaximus (approved):
:+1: Seems like the same pattern we use in several other account actions.
I can't wait to get redux persist in place... will simplify this stuff immensely.

PR #2299, opened 2021-12-09
ADD defaults & check for valid NEAR_WALLET_ENV

### LINKED ISSUE #2291: NEAR_WALLET_ENV should control default env var values
config.js currently has defaults for TESTNET, so if we ever need to configure for any other environment, we have to set many environment variables to override the testnet defaults -- this is suboptimal and a major avenue for bugs in our code to creep in.

1. config.js should load default values for every config key using `process.env` references; any per-environment defaults will _not_ take precedence over process.env provided values.
2. based on `NEAR_WALLET_ENV`, config.js should then merge any per-environment defaults into the hash that was populated from process.env, making sure not to overwrite values that were defined on process.env (only fill in defaults that were not defined in the parent shell)
3. Any references to `IS_MAINNET` should be replaced with a check that checks `NEAR_WALLET_ENV` and returns true if it is the mainnet environment
4. If `NEAR_WALLET_ENV` does not match one of the values of the object exported from `environment.json`, we should throw an error FROM THE MODULE SCOPE, so the process fails instantly on load.

esaminu:
- Check for valid `NEAR_WALLET_ENV` in module scope and throw if invalid*
- Parametrize bundle configs by `NEAR_WALLET_ENV`
- Set overridable defaults per `NEAR_WALLET_ENV`
- Remove usage of `REACT_APP_IS_MAINNET` and replace with `NEAR_WALLET_ENV` conditional
- Add eslint rule to disallow usage of `process.env` outside of config files

*This is a breaking change in all environments until `NEAR_WALLET_ENV` is correctly set to one of [these](https://github.com/near/near-wallet/blob/master/features/environments.json) values.

Closes #2291


MaximusHaximus (changes requested):
Looking at all of the duplicated OR cases with `process.env` references repeated in every file, it feels untenable to leave so many places with duplicated references in place.

Let's do this:

1. Create a 'global' hash of all `config` keys and their associated references to `process.env.XXX_YYYYY`; this should exist in a central file because the references to `process.env` are _always the same_, no matter what value is in `NEAR_WALLET_ENV`.
2. When defining per-environment defaults, don't include the reference to `process.env`.  Instead, let's merge the values from the 'global' hash of all `config` keys that have been loaded from the process environment with the defaults for the current environment, using Lodash.defaults(), where our 'global' hash populated from the process.env is the first argument and the current environment's hash is the second. (see https://lodash.com/docs/4.17.15#defaults)
3. As a result of # 1 and # 2 above, if _any_ config value's associated `process.env` mapping has a value, it will always take precedence over any per-environment defaults (they will truly be _defaults_).  If nothing is defined in the `process.env` key associated with a particular config entry, then the per-environment default will be applied.

HMU directly with any questions/concerns about this :). I think it will be really good to have all of our `process.env` parsing in a centralized file and have all of our per-environment config files just be a raw hash of values with no process.env fallback logic -- just default values per key.

packages/frontend/src/config/index.js
```diff
@@ -0,0 +1,22 @@
+import { defaults } from "lodash";
+
+import Environments from "../../../../features/environments.json";
+import development from "./environmentDefaults/development";
+import mainnet from "./environmentDefaults/mainnet";
+import mainnet_STAGING from "./environmentDefaults/mainnet_STAGING";
+import testnet from "./environmentDefaults/testnet";
+import testnet_STAGING from "./environmentDefaults/testnet_STAGING";
+import environmentConfig from "./configFromEnvironment";
```
andy-haynes, line 9: I think this violates `import/order`, should be second from top.

packages/frontend/src/config/envParsers.js
```diff
@@ -0,0 +1,19 @@
+export const envValIsSet = (envVal) => typeof envVal === "string";
+
+export const parseBooleanFromShell = (envVal) =>
+    envValIsSet(envVal) ? envVal === "yes" || envVal === "true" : undefined;
+
+export const parseObjectFromShell = (envVal) => {
+    if (envValIsSet(envVal)) {
+        try {
+            let parsed = JSON.parse(envVal);
+            return parsed;
```
andy-haynes, line 10: No need to assign this to a value, you can return the expression directly in the `try`.

packages/frontend/src/config/envParsers.js
```diff
@@ -0,0 +1,19 @@
+export const envValIsSet = (envVal) => typeof envVal === "string";
+
+export const parseBooleanFromShell = (envVal) =>
+    envValIsSet(envVal) ? envVal === "yes" || envVal === "true" : undefined;
+
+export const parseObjectFromShell = (envVal) => {
+    if (envValIsSet(envVal)) {
```
andy-haynes, line 7: Since you're already catching the invalid JSON parse in the `try`/`catch` you can omit this check go straight to the `JSON.parse`. Makes it a little more succinct since the return value is the same whether the variable is unset or invalid JSON.

andy-haynes (changes requested):
Nice work! Just a few minor comments inline 🙂

packages/frontend/.eslintrc.js
```diff
@@ -20,5 +20,14 @@ module.exports = {
                 ]
             }
         ],
-    }
+        'no-process-env': ['error']
```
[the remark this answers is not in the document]
andy-haynes: I had no idea this existed, good call on adding some enforcement!

andy-haynes:
New changes look good to me but bundling is failing because `NEAR_WALLET_ENV` isn't set in the Travis CI environment. @MaximusHaximus does `NEAR_WALLET_ENV` need to be set in `.travis.yml`?

Since this is a breaking change do we also need a value for this in `netlify.toml`?

esaminu:
@andy-haynes I'll be adding functionality to compute `NEAR_WALLET_ENV` for each ci environment. Converting to draft for now.

andy-haynes:
> @andy-haynes I'll be adding functionality to compute `NEAR_WALLET_ENV` for each ci environment. Converting to draft for now.

Perfect, thanks for the clarification!

packages/frontend/ci/configFromEnvironment.js
```diff
@@ -0,0 +1,73 @@
+const assert = require("assert");
+
+const Environments = require("../../../features/environments.json");
+const { parseBooleanFromShell } = require("../src/config/envParsers");
+
+const Config = {
+  NEAR_WALLET_ENV: process.env.NEAR_WALLET_ENV,
+  TRAVIS: parseBooleanFromShell(process.env.TRAVIS),
```
andy-haynes, line 8: Could we put the above two keys in alphabetical order please? It makes it a lot easier to find values in bigger object hashes like this.

packages/frontend/ci/configFromEnvironment.js
```diff
@@ -0,0 +1,73 @@
...
+        : Environments.TESTNET;
+  }
+
+  if(Config.TRAVIS) {
+    return Environments.MAINNET_STAGING;
+  }
+};
+
+const NEAR_WALLET_ENV =
+    Config.NEAR_WALLET_ENV || computeCiNearWalletEnv(Config);
```
andy-haynes, line 63: If `NEAR_WALLET_ENV` is not set as an environment variable but `computeCiNearWalletEnv` returns a valid value, the check in `src/config/configFromEnvironment.js` would still fail its assertion at runtime. The check here at bundle time enables the bundle to be built, but nothing has injected the value into the bundle, so any bundle code trying to parse `process.env.NEAR_WALLET_ENV` would get `undefined`.

If we can define (in code) the value for this environment variable then I think that addresses the issue. However I don't know whether that's possible for all environments:
- `development`: ✅  developers set this individually, either globally, at bundle time, or via the `.env` file (which implicitly uses `dotenv` to inject it as an environment variable)
- `testnet`: ✅  already set in the render.com configuration (not in code 😞)
- `testnet_STAGING`: ❌  the preview environment is using `testnet` because it uses the same render.com configuration - if there's no way to override this on the render.com side then I think the only way is to use the `IS_PULL_REQUEST` in conjunction with `NEAR_WALLET_ENV` to know that it's actually `testnet_STAGING`
- `mainnet_STAGING` ❌ (not yet available) it looks like we can set this one in `netlify.toml` under `[build.environment]` but I'm not sure
- `mainnet` ❌ (not yet available) same as `mainnet_STAGING` but set under `[context.production.environment]`

So I think we can address the `mainnet*` cases by adding the two lines in `netlify.toml` but I'm not sure what the ideal approach for `testnet_STAGING` would be, since it might be the only special case where we need to derive the value of `NEAR_WALLET_ENV` based on some other context.

andy-haynes (changes requested):
Looking good, mostly minor comments but I still think we need to find a way to ensure `NEAR_WALLET_ENV` is set in the `mainnet*` environments as well as `testnet_STAGING` (which currently just uses `testnet`). I'm happy to sync up when we have some good schedule overlap to go over the feedback and work out a plan for `testnet_STAGING`.

packages/frontend/ci/configFromEnvironment.js
esaminu: Thanks for catching this. I looked into this and I think the best approach is to write to the .env in case a value for `NEAR_WALLET_ENV` is not already set. Computing `NEAR_WALLET_ENV` separately for ci and bundle scatters the logic imo and logic is further scattered for the bundle's logic for each env. Let me know if you see any issues with this approach.

andy-haynes: This configuration will be less scattered as we transition to consistent environments in AWS with a monorepo housing the backend code, where we'll want the environment variables to be in the environment context at runtime. My preference would be to continue treating `.env` as a development artifact to be phased out, in particular since it's not a source-controlled file and would need to vary per environment anyway.

Looking at the CI environment variables more closely I see a lot of options that will be either deprecated or moved once we have a more robust build pipeline. I'm not sure exactly what this will look like but since it's more a concern for the environment context I'm thinking it will largely be a Terraform responsibility and not the responsibility of our app to determine. So while it is janky in its current state I think it's worth looking at what it would take to move the source of truth entirely to the environment. Keeping it an environmental concern will make the new pipeline simpler to transition to and reason about.

I dug into our Netlify configuration and saw that `NEAR_WALLET_ENV` is already set there (my bad, I should have checked this before 🤦), so actually `netlify.toml` doesn't need to be touched at all. The only real deviation is with `testnet_STAGING`, which I think this can be worked around with something like this (in `ci/configFromEnvironment`):
```
let { NEAR_WALLET_ENV } = process.env;
if (NEAR_WALLET_ENV = environments.TESTNET && Config.IS_PULL_REQUEST) {
    NEAR_WALLET_ENV = environments.TESTNET_STAGING;
}
```

Then in `src/config/configFromEnvironment` we can import `NEAR_WALLET_ENV` from the CI config or just duplicate the snippet above. Once it's set in both the build and runtime environments we'll be able to trust the environment and simply reference `process.env.NEAR_WALLET_ENV`.

What would you think about an approach more like that? Any potential problems you can see?

esaminu: I'd just like to avoid a scenario where we compute a different value of `NEAR_WALLET_ENV` in the ci from the bundle config and I think it's surprising that parcel does not allow the bundling script to define bundle envs directly in some way. The separate `NEAR_WALLET_ENV` computation when not explicitly provided will remain regardless of the stack because of this. It might be cleanest to import the value like you said. Importing was not working on my system because the bundle only gets env values from `.env` and not the parent shell for me so the values I set for testing in shell (e.g. `TRAVIS`) were not accessed by the bundle and `NEAR_WALLET_ENV` was still `undefined`. I'm not sure if this is the case in aws / render / netlify, it might be that the values we set in their UI are set as `.env` file and not just live shell. In the end we have to assume that whatever env values are accessible to the bundling script are accessible within the bundle and just import. The issue in my local just confused me a bit 😆 . Also I think precedence goes to `process.env` in the example above. We're dealing strictly with cases where `NEAR_WALLET_ENV` is not provided by the shell / environment.

esaminu: Facing an issue with [this](https://github.com/near/near-wallet/blob/master/features/features.js#L7) when we don't set the value to `process.env` in the bundle and just import. It takes whatever value is in `.env` and throws if it's not set. This issue is not there when we write to `.env` from bundling script like I did, but if we go this route I think we need to use `dotenv` on the bundling script as well. Assuming that we don't want to import `Config` in that directory and also that we can't set `process.env` to the imported value from within the bundle, this seems to be the only way to set `process.env` directly while maintaining consistency.

packages/frontend/src/config/configFromEnvironment.js
```diff
@@ -0,0 +1,73 @@
+import assert from "assert";
+
+import Environments from "../../../../features/environments.json";
+import {
+    parseBooleanFromShell,
+    parseObjectFromShell,
+    parseCommaSeperatedStringAsArrayFromShell,
+} from "./envParsers";
+
+const NEAR_WALLET_ENV = process.env.NEAR_WALLET_ENV;
```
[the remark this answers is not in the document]
esaminu: This is not possible due to the way parcel handles `process.env.` values. `process.env` is an empty object while the values accessed directly are replaced.

packages/frontend/ci/configFromEnvironment.js
andy-haynes: > We're dealing strictly with cases where `NEAR_WALLET_ENV` is not provided by the shell / environment.
My point is more that we _can_ ensure that it is provided by the environment, and we currently do by configuring it on the various hosting providers. It's not perfect in its current iteration because of `testnet_STAGING`, but that's an implementation detail in Render that we can work around 🙂. It will be more reliable when we are under a single provider. Moreover since it is an environmental concern any computation used to derive it independently gives us two sources of truth, which we definitely want to avoid.

> Importing was not working on my system because the bundle only gets env values from `.env` and not the parent shell for me so the values I set for testing in shell (e.g. `TRAVIS`) were not accessed by the bundle and `NEAR_WALLET_ENV` was still `undefined`.
Since the command we run to bundle, `node ci/runBundler.js`, is executed in a child process, it only inherits environment variables, not shell variables. This wasn't a distinction I was aware of before but your comment led me to experiment a bit. Basically if you are setting (in your `.zshrc` or wherever) `NEAR_WALLET_ENV=development` then it is a shell variable but if you instead have `export NEAR_WALLET_ENV=development` it is an environment variable. Only the latter will be available to the child process context.

`dotenv` is only ever a concern in our local environments because no `.env` files exist in production environments. The bundling is always done using the latest checkout and so must rely exclusively on the environment variables set at bundle time.

esaminu: Made it so that `src/config/configFromEnvironment` and `features/features.js` both import directly from `ci/configFromEnvironment.js`. This does not set `process.env.NEAR_WALLET_ENV` within the bundle as far as parcel is concerned but all references to `NEAR_WALLET_ENV` will use whatever was computed in `ci/configFromEnvironment.js` since they point to `src/config/configFromEnvironment`.

In cases where `process.env.NEAR_WALLET_ENV` is provided, that will be directly used and in cases where it's not the logic on `computeCiNearWalletEnv` should compute it correctly since it's based on `composeBundlerConfig`  in `ParcelBundler.js`.

e.g `NEAR_WALLET_ENV` will be `testnet_STAGING` when `IS_RENDER` and `IS_PULL_REQUEST` are both true. Let me know if you see something I'm missing 😄

packages/frontend/src/config/configFromEnvironment.js
```diff
@@ -0,0 +1,72 @@
...
+    SENTRY_DSN: process.env.SENTRY_DSN,
+    SENTRY_RELEASE: process.env.SENTRY_RELEASE
+        ? process.env.RENDER
+            ? `render:${process.env.RENDER_SERVICE_NAME}:${process.env.RENDER_GIT_BRANCH}:${process.env.RENDER_GIT_COMMIT}`
+            : undefined
+        : undefined,
+    SHOW_PRERELEASE_WARNING: parseBooleanFromShell(
+        process.env.SHOW_PRERELEASE_WARNING
+    ),
+    SMS_BLACKLIST: process.env.SMS_BLACKLIST,
```
MaximusHaximus, line 67: Would like to use `parseCommaSeperatedStringAsArrayFromShell` on this and modify the values for `SMS_BLACKLIST` to be a hard-coded array so it mimics how we handle `WHITELISTED_CONTRACTS`.  Note that you'll have to remove the string.split usage from https://github.com/near/near-wallet/blob/1892ba3c7d958ef9c7d11bd2c0fc974f430f95b0/packages/frontend/src/utils/isApprovedCountryCode.js#L4-L3

packages/frontend/ci/configFromEnvironment.js
```diff
@@ -0,0 +1,70 @@
+const assert = require("assert");
+
+const Environments = require("../../../features/environments.json");
+const { parseBooleanFromShell } = require("../src/config/envParsers");
+
+const Config = {
+  BRANCH: process.env.BRANCH,
+  CLOUDFLARE_BASE_URL: process.env.CLOUDFLARE_BASE_URL,
+  CONTEXT: process.env.CONTEXT,
+  DEBUG_BUILD: process.env.DEBUG_BUILD,
```
MaximusHaximus, line 10: Please use `parseBooleanFromShell()` on this and remove the coercion from `ParcelBundler.js`

packages/frontend/ci/configFromEnvironment.js
```diff
@@ -0,0 +1,70 @@
...
+  DEPLOY_PRIME_URL: process.env.DEPLOY_PRIME_URL,
+  IS_DEVELOPMENT: process.env.NODE_ENV === "development",
+  IS_NETLIFY: parseBooleanFromShell(process.env.NETLIFY),
+  IS_RENDER: parseBooleanFromShell(process.env.RENDER),
+  IS_PULL_REQUEST: parseBooleanFromShell(process.env.IS_PULL_REQUEST),
+  NEAR_WALLET_ENV: process.env.NEAR_WALLET_ENV,
+  RENDER: process.env.RENDER,
```
MaximusHaximus, line 17: Should be using `parseBooleanFromShell()` here

packages/frontend/ci/configFromEnvironment.js
```diff
@@ -0,0 +1,70 @@
...
+  RENDER_EXTERNAL_URL: process.env.RENDER_EXTERNAL_URL,
+  RENDER_GIT_COMMIT: process.env.RENDER_GIT_COMMIT,
+  REVIEW_ID: process.env.REVIEW_ID,
+  SENTRY_AUTH_TOKEN: process.env.SENTRY_AUTH_TOKEN,
+  SENTRY_DSN: process.env.SENTRY_DSN,
+  SENTRY_RELEASE:
+      process.env.SENTRY_RELEASE ||
+      (process.env.RENDER &&
```
MaximusHaximus, line 25: Should be using `parseBooleanFromShell` here rather than directly referencing `process.env.RENDER`

packages/frontend/src/config/configFromEnvironment.js
```diff
@@ -0,0 +1,72 @@
...
+    NODE_URL: process.env.REACT_APP_NODE_URL,
+    PUBLIC_URL: process.env.PUBLIC_URL,
+    REACT_APP_USE_TESTINGLOCKUP: parseBooleanFromShell(
+        process.env.REACT_APP_USE_TESTINGLOCKUP
+    ),
+    RECAPTCHA_CHALLENGE_API_KEY: process.env.RECAPTCHA_CHALLENGE_API_KEY,
+    RECAPTCHA_ENTERPRISE_SITE_KEY: process.env.RECAPTCHA_ENTERPRISE_SITE_KEY,
+    SENTRY_DSN: process.env.SENTRY_DSN,
+    SENTRY_RELEASE: process.env.SENTRY_RELEASE
+        ? process.env.RENDER
```
MaximusHaximus, line 60: `parseBooleanFromShell()` here

packages/frontend/src/config/environmentDefaults/mainnet.js
```diff
@@ -0,0 +1,47 @@
...
+    MULTISIG_MIN_AMOUNT: "4",
+    MULTISIG_MIN_PROMPT_AMOUNT: "200",
+    NETWORK_ID: "default",
+    NODE_URL: "https://rpc.mainnet.near.org",
+    REACT_APP_USE_TESTINGLOCKUP: false,
+    RECAPTCHA_CHALLENGE_API_KEY: "6LeRzswaAAAAAGeS7mSasZ1wDcGnMcH3D7W1gy1b",
+    RECAPTCHA_ENTERPRISE_SITE_KEY: "6LcpJ3EcAAAAAFgA-nixKFNGWMo9IG9FQhH4XjSY",
+    SENTRY_DSN: "https://15d0d1b94e8548dd9663b8c93bf4550a@o398573.ingest.sentry.io/5396205",
+    SHOW_PRERELEASE_WARNING: false,
+    SMS_BLACKLIST: "CN,VN",
```
MaximusHaximus, line 40: Per above comment, this should be a two element array so that we don't need to use String.split() when consuming this variable in our code elsewhere (like WHITELISTED_CONTRACTS)

packages/frontend/src/config/environmentDefaults/mainnet_STAGING.js
```diff
@@ -0,0 +1,47 @@
...
+    MULTISIG_MIN_AMOUNT: "4",
+    MULTISIG_MIN_PROMPT_AMOUNT: "200",
+    NETWORK_ID: "default",
+    NODE_URL: "https://rpc.mainnet.near.org",
+    REACT_APP_USE_TESTINGLOCKUP: false,
+    RECAPTCHA_CHALLENGE_API_KEY: "6LeRzswaAAAAAGeS7mSasZ1wDcGnMcH3D7W1gy1b",
+    RECAPTCHA_ENTERPRISE_SITE_KEY: "6LcpJ3EcAAAAAFgA-nixKFNGWMo9IG9FQhH4XjSY",
+    SENTRY_DSN: "https://15d0d1b94e8548dd9663b8c93bf4550a@o398573.ingest.sentry.io/5396205",
+    SHOW_PRERELEASE_WARNING: true,
+    SMS_BLACKLIST: "CN,VN",
```
MaximusHaximus, line 40: Per above comment, this should be a two element array so that we don't need to use String.split() when consuming this variable in our code elsewhere (like WHITELISTED_CONTRACTS)

packages/frontend/src/config/environmentDefaults/testnet.js
```diff
@@ -0,0 +1,41 @@
...
+        "55E7imniT2uuYrECn17qJAk9fLcwQW4ftNSwmCJL5Di",
+    ],
+    MULTISIG_MIN_AMOUNT: "4",
+    MULTISIG_MIN_PROMPT_AMOUNT: "200",
+    NETWORK_ID: "default",
+    NODE_URL: "https://rpc.nearprotocol.com",
+    REACT_APP_USE_TESTINGLOCKUP: false,
+    SENTRY_DSN: "https://75d1dabd0ab646329fad8a3e7d6c761d@o398573.ingest.sentry.io/5254526",
+    SHOW_PRERELEASE_WARNING: false,
+    SMS_BLACKLIST: "",
```
MaximusHaximus, line 38: Let's set the SMS blacklist here to the same value as MAINNET (and make it an array per prior comments)

MaximusHaximus (changes requested):
Added comments in-line -- in brief, let's totally remove `testnet_STAGING` for now and eliminate any defaults :) Saw a couple of spots that we should be unpacking env vars that I also mentioned in-line.

packages/frontend/ci/configFromEnvironment.js
esaminu: As discussed, removing the fallback computation also allows us to directly access `NEAR_WALLET_ENV` from `process.env` in bundle and `features.js` so I removed the additional wiring we decided to add earlier as well 👍

packages/frontend/ci/configFromEnvironment.js
andy-haynes: @esaminu I believe this is the coercion that Daryl is referring to: https://github.com/near/near-wallet/blob/master/packages/frontend/ci/ParcelBundler.js#L13 - with your parsing change we can now directly refer to `DEBUG_BUILD` 🎉

packages/frontend/src/config/envParsers.js
```diff
@@ -12,7 +12,7 @@ const parseObjectFromShell = (envVal) => {
 };
 
 const parseCommaSeperatedStringAsArrayFromShell = (envVal) =>
-    envValIsSet(envVal) ? envVal.split(",") : undefined;
+    envValIsSet(envVal) ? envVal.replace(/\s/g, "").split(",") : undefined;
```
MaximusHaximus, line 15: We don't want this to always strip all whitespace from any env vars we're using `parseCommaSeperatedStringAsArrayFromShell` with -- it's just _coincidental_ that none of our current comma-separated values don't include spaces.  Rather than put this here, please run the .replace() in-line just for the `SMS_BLACKLIST` env var in `configFromEnvironment`, then pass the resulting 'cleaned' string into this method (which should only split on comma with the `envValIsSet` guard)

MaximusHaximus (changes requested):
Just one thing re: always removing whitespace in the parse comma helper function

Also, Andy's comment re: https://github.com/near/near-wallet/blob/master/packages/frontend/ci/ParcelBundler.js#L13  is accurate -- please remove the coercion there :). Then we can get this sucker landed!  I'm keen to get this merged today (Jan 11th) if possible.

packages/frontend/ci/configFromEnvironment.js
esaminu: Done! Thanks for catching 😄

packages/frontend/src/config/envParsers.js
esaminu: Done! 😄

andy-haynes (approved):
😎

MaximusHaximus (approved):
:+1: legit -- nice work folks! We're really making some progress on our snarled and confusing environments at this point!

PR #2304, opened 2021-12-10
Refactor ledger reducer into redux toolkit version

marcinbodnar:
This PR is refactoring the ledger reducer into the `redux-toolkit` version.

The next step would be to refactor the ledger state to be keyed by accountId, like other reducers. 

Includes many improvements to the code. Also removed one dispatch from wallet class.

There was also a need to implement functionality that will replace `showAlert` function that we used for `redux-actions` implementation. `showAlert` is responsible for showing (with chosen params) global alert when dispatch ends.

It's been implemented as `showAlertToolkit` function that is using `serializeError` option for `createAsyncThunk`. Most important thing is that it's effortless to add global alert handle to any thunk - it can be done by adding a single function with global alert params.


marcinbodnar:
@MaximusHaximus The next step would be to refactor the ledger state to be keyed by accountId, like other reducers. But I would prefer to merge this PR first, because there are a lot of changes, and it was a bit problematic.

There are many improvements to the code. Also removed one dispatch from wallet class :) 

There was also a need to implement some sort of functionality that will replace `showAlert` function that we used for `redux-actions` implementation. It took me a while to figure out the best way to handle it with the redux toolkit but I think it's implemented pretty well. Most important thing is that it's effortless to add global alert handle to any thunk - it can be done by adding a single function with global alert params - in this case, I decided to use `serializeError` option from `createAsyncThunk` - works nice.

marcinbodnar:
@ihaid Could you please test this PR regarding Ledger functionality? These changes touch everything related to Ledger.

packages/frontend/src/redux/reducers/status/index.js
```diff
@@ -17,6 +17,19 @@ const initialState = {
 };
 
 const alertReducer = (state, { error, ready, payload, meta, type }) => {
+
+    // temporary solution to handle both `showAlert` and `showAlertToolkit`
```
andy-haynes, line 21: If this is temporary, what does the long-term solution look like?

packages/frontend/src/redux/slices/ledger/index.js
```diff
@@ -1,14 +1,192 @@
...
+};
+
+const initialState = {
+    modal: {}
+};
+
+const getLedgerAccountIds = createAsyncThunk(
+    `${SLICE_NAME}/getLedgerAccountIds`,
+    async ({ path }) => await wallet.getLedgerAccountIds({ path }),
+    showAlertToolkit({ onlyError: true })
```
andy-haynes, line 27: Just trying to wrap my head around the new need for `showAlertToolkit`... Is this the first time we're decoupling an async thunk from the `readyStatePromise` middleware?

packages/frontend/src/redux/slices/ledger/index.js
```diff
@@ -1,14 +1,192 @@
...
+            set(state, ['signInWithLedger', accountId, 'status'], transportError ? 'rejected' : 'error');
+        });
+        // refreshAccountOwner
+        builder.addCase(refreshAccountOwner.fulfilled, (state, { payload }) => {
+            set(state, ['hasLedger'], payload.ledger.hasLedger);
+            set(state, ['ledgerKey'], payload.ledger.ledgerKey);
+        });
+        // matcher to handle closing modal automatically
+        builder.addMatcher(
+            ({ type, ready, error }) => ready || error || type.endsWith('/rejected') || type.endsWith('/fulfilled'),
```
andy-haynes, line 159: Curious about why we need the fallback to check the type here - can `ready` and `error` both be `false` when the promise has fulfilled/rejected? Does that mean that `ready` and `error` are strictly synchronous?

andy-haynes (changes requested):
Looking good, one minor comment inline along with some questions to help me understand `createAsyncThunk` error handling.

You'll also need to rebase to get the latest changes to `wallet.js` 🙂

packages/frontend/src/components/accounts/ledger/SignInLedger.js
```diff
@@ -5,18 +5,16 @@ import { useSelector, useDispatch } from 'react-redux';
...
-    signInWithLedgerAddAndSaveAccounts,
     checkAccountAvailable,
-    clearSignInWithLedgerModalState,
     clearAccountState
 } from '../../../redux/actions/account';
 import { clearLocalAlert } from '../../../redux/actions/status';
 import { selectAccountSlice } from '../../../redux/slices/account';
-import { selectLedgerSignInWithLedger, selectLedgerSignInWithLedgerStatus, selectLedgerTxSigned } from '../../../redux/slices/ledger';
+import { LEDGER_MODAL_STATUS, selectLedgerSignInWithLedger, selectLedgerSignInWithLedgerStatus, selectLedgerTxSigned } from '../../../redux/slices/ledger';
+import { actions as ledgerActions } from '../../../redux/slices/ledger';
```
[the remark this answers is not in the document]
marcinbodnar: Done.

packages/frontend/src/redux/reducers/status/index.js
marcinbodnar: Added a more precise description to the file.

Generally `alertReducer` was designed to relay on `ready` param that is adding to action meta - this way we can recognize if the async action is finished. With redux toolkit and `createAsyncThunk` we have separate actions (`rejected` and `fulfilled`). So for now we need to use both `rejected` and `fulfilled` actions and `ready` param to recognize the action status, but finally, when we will be using redux-toolkit for all reducers, we will be able to rely on `rejected` and `fulfilled` types of actions only and we will refactor this whole reducer. There is also an option to rebuild the whole `globalAlert` related state, but it's on hold for now.

packages/frontend/src/redux/slices/ledger/index.js
marcinbodnar: Yes, I mean, these are the first async actions that are using `globalAlert` that are refactored. I started refactoring from the simplest reducers, so there ware no case like that before.

packages/frontend/src/redux/slices/ledger/index.js
marcinbodnar: Yes, that's because `readyStatePromise` async actions will have ready and error param and `createAsyncThunk` actions will use `rejected` and `fulfilled` types of actions only. For now, we need to check both scenarios.

This might be also simplified but only after we use `redux-toolkit` reducers only.

andy-haynes (approved):
Nice work, appreciate the explanations as always @marcinbodnar

PR #2305, opened 2021-12-11
Switch to multistage build for Wallet Dockerfile, shedding 700MB of image size

mieubrisse:
…mage size

Drops image size from 1.1GB -> 323MB (and it's still so big just because the Phusion base image is 183MB)

### DIFF
--- a/packages/frontend/.dockerignore
+++ b/packages/frontend/.dockerignore
@@ -2,3 +2,6 @@
 **/node_modules
 dist
 .cache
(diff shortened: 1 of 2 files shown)


mieubrisse:
@MaximusHaximus

MaximusHaximus (approved):
:+1: This is awesome! Thanks so much for the submission, we appreciate the help! ❇️ 💖

PR #2308, opened 2021-12-14
Fix redirect after LackBalanceForState error

marcinbodnar:
Account for testing:
accountId: `tryagain.near`
seed phrase: `(redacted)`


MaximusHaximus (approved):
:+1: rock on, nice work folks <3 :)

PR #2309, opened 2021-12-14
docs: add documentation on NEAR_WALLET_ENV in the Getting Started section

andy-haynes:
This PR adds a blurb explaining the need for a valid `NEAR_WALLET_ENV` environment variable. Now that we have feature flags in the repo this variable is required.

### DIFF
--- a/packages/frontend/README.md
+++ b/packages/frontend/README.md
@@ -10,6 +10,9 @@ WARNING: This wallet shouldn't be used as the only signer for high-value account
 Getting started
 ===
 
+First ensure there is a value for the `NEAR_WALLET_ENV` environment variable. This can be set in `packages/frontend/.env` prior to bundling.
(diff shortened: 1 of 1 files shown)


andy-haynes:
Would it make sense to have a `.env.sample` file in `frontend` the way we do in `near-contract-helper`?

Patrick1904 (approved):

MaximusHaximus (approved):
:+1: <3

MaximusHaximus:
> 

I don't think it's critical, but you raise a good point about discoverability of supported env vars.  

What do you think about us updating this note in the readme in #2299 to also describe that you that you can use a `.env` file to set _any_ of the environment variables that are referenced in the new `configFromEnvironment.js` files that @esaminu has added there?  Then we won't have to keep a sample .env file updated as we add/remove env vars.

Issue #2311, opened 2021-12-15
`yarn install` should use the lockfile


andy-haynes:
**Problem**
Currently running `yarn build` modifies the `yarn.lock` file unexpectedly.

**Expected Behavior**
`yarn build` should not modify the lock file, packages should be installed according to the versions specified in `yarn.lock`.

**Steps to reproduce**

1. Run `yarn build`
2. `yarn.lock` is modified

PR #2312, opened 2021-12-15
bug: use --frozen-lockfile to install packages according to yarn.lock versions

andy-haynes:
This PR updates the `scripts` in `package.json` to replace instances of `yarn` (implicit `install`) with: `yarn install --frozen-lockfile`. Addresses #2311

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,9 +5,9 @@
     "packages/*"
   ],
   "scripts": {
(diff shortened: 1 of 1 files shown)


MaximusHaximus (approved):
:+1: legit -- good catch <3

Issue #2284, opened 2021-12-01
Implicit addresses as default for first-time account creation


corwinharrell:
## Problem
Because of the necessity for named accounts to occupy a namespace and storage, a prerequisite for account creation in the wallet has always been to fund `.near` addresses before they can be recovered. Since many new users do not have NEAR to begin with, this has created a scenario where we have had to rely on exploitable faucets and non-anonymous verification methods in order to fund new accounts on the users' behalf. This is unsustainable, and because we still have to deal with and display implicit account addresses as a part of the funding process, funds are often send to the wrong address even after onboarding is complete.

Additionally, 3rd party wallets don't currently support NEAR's account model with personalized addresses. As a result, any NEAR account created with a 3rd party will be assigned an implicit address without the option to register a personalized address. When this account is imported into the wallet, it will use the implicit address which it was originally assigned.

#### How did we discover this problem?
- 3rd party wallet support for `.near` addresses has never been enforced
- Because we've required users to send funds to their implicit address as a part of onboarding (for the manual deposit funding method), they often continue to use this address even after acquiring a personalized address, since they have already sent funds to that address once
- The coin-op faucet is regularly and unavoidably exploited. We need to remove our dependency on the faucet for account funding / initialization.
- Attempts to obscure and/or abstract away the concept of implicit accounts has ultimately resulted in more confusion and frustration on the part of the user.

#### What are we planning to do about it?
- Moving forward, implicit accounts will be treated as the default for new users of the wallet.
- Personalized addresses will be framed as an "enhanced" account option that allows users to obtain a personalized address once they have enough NEAR in an existing account to fund them.
- Because users are given a seed phrase / private key _before_ they complete the funding step, implicit accounts will enable them to recover these accounts without the concern of occupying a valuable namespace on the network
- With this work, we will be eliminating the use of our mainnet faucet: "coin-op". The responsibility of funding new accounts will fall squarely on the user.
- To retain a level of consistency with how new accounts are created across all security methods, new addresses will use the same private key as their "parent" account by default, with the option for the user to generate a new and unique private key if they so desire. This unlocks a degree of flexibility in key management.

#### What are we _not_ planning to do about it?
- This issues is specifically scoped around the ability to create and fund implicit accounts by default, followed by the ability to register personalized addresses using an existing account as a funding source. We will not be adding or making sweeping changes to security methods or our use of local storage. These changes will follow in a separate work stream.

## Measuring Success
We will be paying close attention to and measuring the following:
[...]

PR #2313, opened 2021-12-16
Implicit addresses as default for first-time account creation

Patrick1904:
This PR enables accounts creation with implicit account by default.

Exceptions include:
1. User is already logged in to at least one account
2. User is creating an account with linkdrop

The feature is currently only enabled in a `mainnet_STAGING` environment using the feature flag: `CREATE_IMPLICIT_ACCOUNT`.

More information about the feature: https://github.com/near/near-wallet/issues/2284

The flow is available for testing on the PR staging link: https://deploy-preview-2313--near-wallet-staging.netlify.app/


packages/frontend/src/redux/actions/account.js
```diff
@@ -538,27 +535,15 @@ export const handleCreateAccountWithSeedPhrase = (accountId, recoveryKeyPair, fu
...
 export const finishAccountSetup = () => async (dispatch, getState) => {
     await dispatch(refreshAccount());
-    await dispatch(getBalance());
     await dispatch(clearAccountState());
 
-    const balance = selectBalance(getState());
     const redirectUrl = selectAccountUrlRedirectUrl(getState());
     const accountId = selectAccountId(getState());
 
-    let promptTwoFactor = await TwoFactor.checkCanEnableTwoFactor(balance);
```
Patrick1904, line 548: Removed the conditional prompt of 2FA in the account creation flow in order to remove any bias towards 2FA.

packages/frontend/src/components/accounts/create/implicit_account/CreateImplicitAccount.js
```diff
@@ -0,0 +1,92 @@
...
+        disabled={!moonpayIsAvailable}
+        color='light-blue'
+        onClick={() => onClickBuyButton(amountUSD)}
+    >
+        {amountUSD ? (
+            `$${amountUSD}`
+        ) : (
+            <Translate id='account.createImplicitAccount.customAmount' />
+        )
+        }
```
MaximusHaximus, line 52: Formatting here is pretty sub-optimal and not scannable.

How about moving this into a helper function? e.g.

```javascript
function getBuyButtonLabel(amountUSD) {
  if(amountUSD) {
    return `$${amountUSD}`
  }

  return ( <Translate id='account.createImplicitAccount.customAmount' /> );
}
```

packages/frontend/src/components/accounts/create/implicit_account/CreateImplicitAccount.js
```diff
@@ -0,0 +1,92 @@
...
+}) => {
+    const [showWhereToBuyModal, setShowWhereToBuyModal] = useState(false);
+    return (
+        <>
+            <StyledContainer className='border small-centered'>
+                <h3><Translate id='account.createImplicitAccount.title' /></h3>
+                <div className='flex-center-center'>
+                    {['30', '50', '100', ''].map((amount, i) => (
+                        <BuyButton
+                            key={`${i}-${amount}`}
```
MaximusHaximus, line 70: :nit: Seems like the index isn't really relevant in any way, and `amount` is unique, so we could make the `key` property directly evaluate to `amount` (no need for the index at all)

packages/frontend/src/components/accounts/recovery_setup/new_account/SetupPassphraseNewAccount.js
```diff
@@ -0,0 +1,76 @@
...
+    const [passPhrase, setPassPhrase] = useState('');
+    const [recoveryKeyPair, setRecoveryKeyPair] = useState();
+    const [implicitAccountId, setImplicitAccountId] = useState('');
+
+    const [confirmPassphrase, setConfirmPassphrase] = useState(false);
+    const [wordId, setWordId] = useState(null);
+    const [userInputValue, setUserInputValue] = useState('');
+    const [userInputValueWrongWord, setUserInputValueWrongWord] = useState(false);
+
+    const generateAndSetPhrase = () => {
```
MaximusHaximus, line 19: Not a blocker for this PR, just thinking out loud -- I think that everything in this function and the associated state that it mutates should be a custom hook so we can re-use it in the `SetupPassphraseExistingAccount` component when we create it :)

packages/frontend/src/components/accounts/recovery_setup/new_account/SetupPassphraseNewAccount.js
```diff
@@ -0,0 +1,76 @@
...
+        generateAndSetPhrase();
+    }, []);
+
+    if (confirmPassphrase) {
+        return (
+            <ConfirmPassphrase
+                wordId={wordId}
+                userInputValue={userInputValue}
+                userInputValueWrongWord={userInputValueWrongWord}
+                handleChangeWord={(userInputValue) => {
```
MaximusHaximus, line 40: <3 just a note that I really like the choice of in-lining most event handlers here rather than making them functions in the closure scope :). It made it very apparent that `generateAndSetPhrase` was a special case that is re-used

packages/frontend/src/components/accounts/recovery_setup/new_account/SetupPassphraseNewAccount.js
```diff
@@ -0,0 +1,76 @@
...
+import ConfirmPassphrase from './ConfirmPassphrase';
+import SavePassphrase from './SavePassphrase';
+
+export default ({ handleConfirmPassphrase }) => {
+    const [passPhrase, setPassPhrase] = useState('');
+    const [recoveryKeyPair, setRecoveryKeyPair] = useState();
+    const [implicitAccountId, setImplicitAccountId] = useState('');
+
+    const [confirmPassphrase, setConfirmPassphrase] = useState(false);
+    const [wordId, setWordId] = useState(null);
```
MaximusHaximus, line 15: 1. :nit: `wordIndex` would be a bit clearer (`id` usually means something like a UUID)
2. Rather than track `wordId`, why not track the word itself -- e.g. `verificationWord` or something like that? It would make the component easier to follow and eliminate the entire concept of tracking the index just to look up the word from this component, since we never use the index for anything else

packages/frontend/src/components/common/FormButton.js
```diff
@@ -96,9 +96,9 @@ const CustomButton = styled.button`
         }
 
         &.light-blue {
-            background-color: #D6EDFF !important;
+            background-color: #D6EDFF;
```
MaximusHaximus, line 99: 💖 another one bites the dust!

packages/frontend/src/components/svg/GenerateNewIcon.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+export default () => {
```
MaximusHaximus, line 3: Can we make this a raw `.svg` file instead of a JS component? It'll be cached that way and it's not being customized in any way through JS here

packages/frontend/src/components/wallet/Sidebar.js
```diff
@@ -0,0 +1,58 @@
...
+    ];
+    return (
+        <StyledContainer>
+            {components.map((component, i) => {
+                if (i === index) {
+                    return (
+                        <div key={i}>{components[i]}</div>
+                    );
+                }
+            })}
```
MaximusHaximus, line 52: This is a bit cryptic and I think it could be simplified by not mapping across the available components at all.


We only have two choices here, and we select one of the two based on whether or not there are any non-implicit accounts available -- so I think that this can be an explicit choice  that eliminates a lot of cognitive overhead by removing the array indexes and the meaning behind them.

e.g. something like:

```javascript
export default ({ availableAccounts }) => {
    const [activeComponent, setActiveComponent] = useState(CreateCustomName);

    useEffect(() => {
        if (availableAccounts.length > 0) {
            const numNonImplicitAccounts = availableAccounts.filter(a => a.length < 64).length
            setActiveComponent(numNonImplicitAccounts === 0 ? CreateCustomName : ExploreApps);
        }
    }, [availableAccounts]);

    return (
        <StyledContainer>
            {hasNonImplicitAccounts ? <ExploreApps /> : <CreateCustomName />}
            <div className='dots'>
                <div className={`dot ${activeComponent === CreateCustomName ? 'active' : ''}`} onClick={() => setActiveComponent(CreateCustomName)}></div>
                <div className={`dot ${activeComponent === ExploreApps ? 'active' : ''}`} onClick={() => setActiveComponent(ExploreApps)}></div>
            </div>
        </StyledContainer>
    );
};
```

You might consider it slightly less DRY but overall I think it would be a lot easier to read if we didn't use an array and the index of a component...plus you could explicitly give each div a `key` that matches the type of component being rendered if that's necessary.

What do you think?

packages/frontend/src/components/wallet/Wallet.js
```diff
@@ -351,6 +350,13 @@ export function Wallet({ tab, setTab }) {
                     accountId={accountId}
                 />
             }
+            {createCustomName &&
+                <CreateCustomNameModal
+                    onClose={() => dispatch(setCreateCustomName(false))}
+                    isOpen={createCustomName}
+                    accountId='bob.near'
```
MaximusHaximus, line 357: Should this be hard coded to `bob.near`?

packages/frontend/src/redux/slices/createFromImplicit/index.js
```diff
@@ -13,6 +14,9 @@ const createFromImplicitSlice = createSlice({
     reducers: {
         setCreateFromImplicitSuccess: (state, action) => {
             state.createFromImplicitSuccess = action.payload;
+        },
```
MaximusHaximus, line 17: Not a blocker for this PR, but in general we should avoid putting the entire payload of an action into state, as this is generally discouraged; ideally the payload would contain a property with a name indicating what the value actually is, and we would destructure that value from the payload and set it in state. (Ref: https://redux.js.org/style-guide/style-guide#reducers-should-own-the-state-shape)

packages/frontend/src/routes/CreateImplicitAccountWrapper.js
```diff
@@ -0,0 +1,131 @@
...
+    const location = useSelector(getLocation);
+    const URLParams = new URLSearchParams(location.search);
+    const implicitAccountId = URLParams.get('implicitAccountId');
+    const recoveryMethod = URLParams.get('recoveryMethod');
+
+    const formattedMinDeposit = formatNearAmount(MIN_BALANCE_TO_CREATE);
+
+    // Check that the initial deposit was at least 0.2N, otherwise the users 'available balance'
+    // won't be enough to create a named account.
+    const NAMED_ACCOUNT_MIN = utils.format.parseNearAmount('0.2');
```
MaximusHaximus, line 39: Please move to module scope for consistency and so that we don't re-declare it on every render :)

packages/frontend/src/routes/CreateImplicitAccountWrapper.js
```diff
@@ -0,0 +1,131 @@
...
+                async () => {
+                    const moonpayAvailable = await isMoonpayAvailable();
+                    if (moonpayAvailable) {
+                        const moonpaySignedUrl = await getSignedUrl(implicitAccountId, window.location.href, 30);
+                        setMoonpaySignedUrl(moonpaySignedUrl);
+                    }
+                },
+                (e) => {
+                    throw e;
+                }
```
MaximusHaximus, line 59: 1. :hocho: this is a no-op (if you don't add an error handler function, the error bubbles up to the caller anyway)
2. Nothing will actually catch this error; it will bubble all the way up to the browser and be sent as a sentry exception.  Should this silently fail (wrap this in a try/catch to squelch errors), or do we want this behaviour?

packages/frontend/src/routes/CreateImplicitAccountWrapper.js
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }, [accountId, implicitAccountId, recoveryMethod]);
+
+    useEffect(() => {
+        const checkIfMoonPayIsAvailable = async () => {
+            await Mixpanel.withTracking("CA Check Moonpay available",
+                async () => {
+                    const moonpayAvailable = await isMoonpayAvailable();
+                    if (moonpayAvailable) {
+                        const moonpaySignedUrl = await getSignedUrl(implicitAccountId, window.location.href, 30);
```
MaximusHaximus, line 53: What is the last argument to this function call meant to do? It doesn't appear to be defined in the signature for `getSignedUrl`

packages/frontend/src/routes/CreateImplicitAccountWrapper.js
```diff
@@ -0,0 +1,131 @@
...
+                        throw e;
+                    } finally {
+                        if (recoveryMethod === 'ledger') {
+                            dispatch(checkAndHideLedgerModal());
+                        }
+                    }
+                },
+                (e) => {
+                    throw e;
+                }
```
MaximusHaximus, line 113: As above, this is a no-op 🔪

packages/frontend/src/routes/SetupLedgerNewAccountWrapper.js
```diff
@@ -0,0 +1,45 @@
...
+                    }
+                    await setKeyMeta(ledgerPublicKey, { type: 'ledger' });
+                    dispatch(redirectTo(`/create-implicit-account?implicitAccountId=${implicitAccountId}&recoveryMethod=ledger`));
+                } catch (e) {
+                    dispatch(showCustomAlert({
+                        errorMessage: e.message,
+                        success: false,
+                        messageCodeHeader: 'error'
+                    }));
+                    throw e;
```
MaximusHaximus, line 40: Is it intentional that this re-throw will cause a sentry alert for this branch?

packages/frontend/src/routes/SetupRecoveryImplicitAccountWrapper.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { useDispatch } from 'react-redux';
+
+import SetupRecoveryImplicitAccount from '../components/accounts/create/implicit_account/SetupRecoveryImplicitAccount';
+import { redirectTo } from '../redux/actions/account';
+
+export function SetupRecoveryImplicitAccountWrapper() {
+    const dispatch = useDispatch();
+    return (
+        <SetupRecoveryImplicitAccount
```
MaximusHaximus, line 10: I think we'd like to land this before we have tor.us support implemented, right?  If so, don't we need a case for `email` here?

packages/frontend/src/utils/moonpay.js
```diff
@@ -40,7 +40,12 @@ export const isMoonpayAvailable = async () => {
 };
 
 export const getSignedUrl = async (accountId, redirectUrl) => {
-    const widgetUrl = `${MOONPAY_BUY_URL_PREFIX}&walletAddress=${encodeURIComponent(accountId)}&currencyCode=NEAR&redirectURL=${encodeURIComponent(redirectUrl)}`;
+    const widgetUrl = `${MOONPAY_BUY_URL_PREFIX}`
+        + `&walletAddress=${encodeURIComponent(accountId)}`
+        + `&currencyCode=NEAR`
+        + `&baseCurrencyCode=usd`
```
MaximusHaximus, line 46: Did we have to add `baseCurrencyCode` here for some reason? What is the impact of this additional URL param?

packages/frontend/src/components/svg/GenerateNewIcon.js
Patrick1904: I'd prefer to keep it a component so that there's no initial flicker on the 'generate new' button on the seed phrase page. Even if it's fast, it doesn't look great with that initial load. I agree about the caching etc and do that with larger files or graphics, but since this icon is so tiny, the performance vs perceived performance here seems user will perceive it as suboptimal UI if we don't show it instantly

packages/frontend/src/components/wallet/Wallet.js
Patrick1904: Yes - though might be nice with a fading carousel style that show many difference options cc @corwinharrell

packages/frontend/src/utils/moonpay.js
Patrick1904: I added `baseCurrencyCode` here to make sure that Moonpay shows the amount in USD since that's what we show on our buttons in the wallet

packages/frontend/src/components/wallet/Wallet.js
Patrick1904: Once we implement Torus it'd be nice to pull the social login info and construct a recommended account id based on name, email, etc

packages/frontend/src/utils/moonpay.js
MaximusHaximus: ah -- gotcha :+1:

corwinharrell (changes requested):
Looking good! Here are a few things I picked up on or have questions about:

- Let's make sure copy across the board is updated to match the new designs. Right now "Protect Your Account" should read "Choose a Security Method".

- I noticed the addition of a "Start Over" button on the seed phrase confirmation screen. I don't recall if we talked about this but could you give some context on the addition?

- Once the user has complete the funding step, if they have enough funds in order to pay for the txn to reserve a custom address, they should see a modal once logged in for "Add a custom address". This more actively encourages them to do this once they have sufficient funding, vs the more passive banner that sits on the side of the page atm.

- It would be great to begin concatenating implicit address at this point (aw34f....afw34). Makes it easier to confirm an address if the user can see the first and last few characters, and this has quickly become a standard crypto UX pattern.

Great job! Excited to get this in as well as the follow up work to enable the logged in state.

Patrick1904:
> Looking good! Here are a few things I picked up on or have questions about:
> 
> * Let's make sure copy across the board is updated to match the new designs. Right now "Protect Your Account" should read "Choose a Security Method".

@corwinharrell will do 👍 

> * I noticed the addition of a "Start Over" button on the seed phrase confirmation screen. I don't recall if we talked about this but could you give some context on the addition?

It would be best to add back the 'start over' button because the verify seed phrase step is on the same route (state/view), so if the user clicks on the browser back button, the user will actually go back two steps instead of the beginning of the seed phrase flow. There are ways around it but thought this would be the best solution, also personally like being able to clearly start over. Can figure out a different way if you really don't like the button though.

> * Once the user has complete the funding step, if they have enough funds in order to pay for the txn to reserve a custom address, they should see a modal once logged in for "Add a custom address". This more actively encourages them to do this once they have sufficient funding, vs the more passive banner that sits on the side of the page atm.

It's currently set to only show the 'Add a custom address' modal if the user has funded their account with at least 0.2N - to ensure there's enough balance to create additional accounts, since the first account will be funded with 0.1N.

> * It would be great to begin concatenating implicit address at this point (aw34f....afw34). Makes it easier to confirm an address if the user can see the first and last few characters, and this has quickly become a standard crypto UX pattern.

Yep I agree 👍. Let's do this in a separate PR since it'll affect a lot of places.

corwinharrell:
> It's currently set to only show the 'Add a custom address' modal if the user has funded their account with at least 0.2N - to ensure there's enough balance to create additional accounts, since the first account will be funded with 0.1N.

Shouldn't the minimum be 0.1N + the estimated cost of the single TXN necessary to create an additional account? Does it need to be 0.2N or can it be less?

Patrick1904:
> > It's currently set to only show the 'Add a custom address' modal if the user has funded their account with at least 0.2N - to ensure there's enough balance to create additional accounts, since the first account will be funded with 0.1N.
> 
> Shouldn't the minimum be 0.1N + the estimated cost of the single TXN necessary to create an additional account? Does it need to be 0.2N or can it be less?

Yes I can lower it to 0.17N:

Initial funding: 0.2N
Reserved for storage: -0.003N
Reserved for transactions: -0.05N
Gas cost to create first account: -0.002N (likely much less)

Available balance: **0.145N** - which is enough to create a named account.

We'd want to give it a bit of padding to be on the safe side, but lowering the amount required to show the 'add custom address' modal to 0.17N should be safe as the user will end up with ~0.115N in available balance. I'll update it to that amount. We could also check the available balance after the first account has been created, but doing those calculations can sometimes be a bit slow, and the modal may not show right away when the user lands on the dashboard, which is suboptimal UX.

MaximusHaximus (approved):
:+1: this is looking great!  Since this is feature-flagged and will only be live on Staging for now, let's get this merged and create a follow-up PR with any further UI tweaks based on testing in staging

PR #2314, opened 2021-12-16
Set the previous account as active if LackBalanceForState error appears

marcinbodnar:


Patrick1904 (approved):
Thanks @marcinbodnar - works great! 💯

Issue #2296, opened 2021-12-08
Improper ordering in tx log in the wallet


marcinbodnar:
<img width="353" alt="Screen_Shot_2021-12-08_at_11 49 45_PM" src="https://user-images.githubusercontent.com/36210360/145307196-1ecf137c-77d9-46a2-b9cb-969526e13339.png">

PR #2318, opened 2021-12-20
Fix updating transactions list (#2296)

marcinbodnar:


packages/frontend/src/redux/slices/transactions/index.js
```diff
@@ -64,14 +64,17 @@ const transactionsSlice = createSlice({
             const hash = transactionsState.map((t) => t.hash_with_index);
 
             // when updating the transaction, we do not want to replace the entire array, because for some entries the tx status may already be fetched
-            transactions.forEach((t) => {
-                if (!hash.includes(t.hash_with_index)) {
-                    transactionsState.unshift(t);
-                    if (transactionsState.length > 10) {
-                        transactionsState.pop();
+            transactions
+                .reverse()
```
andy-haynes, line 68: Is there a guarantee that `transactions` will always be an ordered collection? The screenshot on the ticket appears to be unordered (2 min, 1 min, 3 min).

Would it be safer instead to sort `transactions` in the desired order before calling `.forEach()`? Something like
```
const orderedTransactions = [...transactions].sort(...);
orderedTransactions.forEach(...);
```

andy-haynes (changes requested):

packages/frontend/src/redux/slices/transactions/index.js
marcinbodnar: Yes, transactions are already sorted - `.reverse()` fixes the ordering issue after the update, which is the problem on the screenshot.

marcinbodnar: The problem was that the list was not updated correctly if there were more than one new transaction. So for example we had:
`transactionsState = [10,9,8,7,6,5,4,3,2,1]` (old list)
`transactions = [12,11,10,9,8,7,6,5,4,3]` (new list)
After first transaction update, the state was:
`[12,10,9,8,7,6,5,4,3,2]`
After second new tx update:
`[11,12,10,9,8,7,6,5,4,3]`

With updating in the correct reverse order it will be:
After first transaction update, the state was:
`[11,10,9,8,7,6,5,4,3,2]`
and 
After second new tx update:
`[12,11,10,9,8,7,6,5,4,3]`

andy-haynes: Got it, I didn't realize the top two transactions in the screenshot were the new ones. Thanks for the clarification!

andy-haynes (approved):

PR #2320, opened 2021-12-20
Show 'total' number of NFTs per contract

marcinbodnar:


marcinbodnar:
@corwinharrell @Patrick1904 Maybe a good idea would be to show the loaded number of tokens and the total number, something like "4 of 8"? With this change, it's a bit weird when the number near the contract name shows for example 10, and the user sees only 4 nfts. But definitely not a blocker.

Patrick1904 (approved):
Nice one Marcin 👍

PR #2328, opened 2021-12-22
FIX broken import of MIN_BALANCE_FOR_GAS

esaminu:
Hotfix broken import of `MIN_BALANCE_FOR_GAS`. This is causing `reservedForTransactions` to be zero.


marcinbodnar (approved):
Thanks @esaminu

PR #2331, opened 2021-12-23
Add richer media support via webm.

june07:
Currently token metadata that includes .webm content results in "Failed to load".

This change detects .webm files and renders the appropriate <video> tag in place of the  tag.  Doesn't affect image loading.

### DIFF
--- a/packages/frontend/src/components/wallet/NFTBox.js
+++ b/packages/frontend/src/components/wallet/NFTBox.js
@@ -97,7 +97,7 @@ const StyledContainer = styled.div`
         }
     }
 
-    .nft img {
+    .nft img, .nft video {
         width: 100%;
         margin-bottom: 10px;
     }
@@ -130,12 +130,22 @@ const NFTBox = ({ tokenDetails }) => {
             {
                 ownedTokensMetadata &&
                 <div className='tokens'>
-                    {ownedTokensMetadata.map(({ token_id, metadata: { mediaUrl, title } }) => {
+                    {ownedTokensMetadata.map(({ token_id, metadata: { mediaUrl, title } }, index) => {
+                        const videoProps = index === 0 ? { autoplay: "true" } : {};
                         return <div className='nft' key={token_id}>
-                            <img src={mediaUrl} alt='NFT' onError={(e) => {
-                                e.target.onerror = null;
-                                e.target.src = FailedToLoad;
-                            }}/>
+                            {
+                                mediaUrl.match(/\.webm$/i)
+                                    ? <video muted="true" loop controls { ...videoProps }>
+                                        <source src={mediaUrl} type="video/webm" onError={(e) => {
+                                        e.target.onerror = null;
+                                        e.target.src = FailedToLoad;
+                                    }}/>
+                                    </video>
+                                    : <img src={mediaUrl} alt='NFT' onError={(e) => {
+                                        e.target.onerror = null;
+                                        e.target.src = FailedToLoad;
+                                    }}/>
+                            }
                             <b>{title}</b>
                         </div>;
                     })}


june07:
Just wanted to check on this one as our last PR was already merged https://github.com/near/near-wallet/pull/2337, and this was created before...  

If there's anything else required, it would be great to get some feedback.  Thanks.

packages/frontend/src/components/wallet/NFTBox.js
```diff
@@ -130,12 +130,22 @@ const NFTBox = ({ tokenDetails }) => {
...
-                            <img src={mediaUrl} alt='NFT' onError={(e) => {
-                                e.target.onerror = null;
-                                e.target.src = FailedToLoad;
-                            }}/>
+                            {
+                                mediaUrl.match(/\.webm$/i)
+                                    ? <video muted="true" loop controls { ...videoProps }>
+                                        <source src={mediaUrl} type="video/webm" onError={(e) => {
+                                        e.target.onerror = null;
+                                        e.target.src = FailedToLoad;
```
MaximusHaximus, line 141: I noticed that `FailedToLoad` is an SVG image, but this is a `<video>` tag -- have you tested that this works and correctly displays our `FailedToLoad` SVG if the .webm file cannot be loaded by the browser?

MaximusHaximus (changes requested):
Thanks so much for the contributions, we really appreciate the help!

Just one question about error handling in case the users' browser is unable to load the .webm file -- if this error handling is working, then we can land this PR :)

packages/frontend/src/components/wallet/NFTBox.js
june07: So yes, you are correct in that a "Loading..." svg would be much better.

Initially I did just copy the logic from the img tag using the video error handler, however after receiving your feedback and testing that condition, I discovered that the video tag will not render images...

I do feel that if the nft creator's intent is to show a video, the error condition should relate to that intent?  What do you think?  So, in that sense, a video fallback ("Loading..." is probably fine.

MaximusHaximus: So I think we have two different issues here -- one is what to display while the video is being loaded, and the other is what to display if the video fails to load.

**I think that the behaviour here should be:**
1. When the page loads, the user sees 'Loading' image for the video until it is ready to be played.
2. If the video fails to load, the user sees the 'FailToLoad' image.
3. If the video loads successfully ('canplay' event fires), then the 'loading' image changes to show the default preview of the video (probably the first frame of the video in most browsers)

I don't think setting `poster` to the loading _or_ the 'loading failed' image will work for us, because it will show. the poster image before the user actually taps 'play' on the the video to download it, and they will likely think they must wait until the `loading` image changes to show something else before tapping 'play', or if we show the error image, they will assume it has failed to load even though it is still pending.

I did some reading about handling this, and I think the pattern described here is what we want, but adapted to be a react-based solution so we don't touch the DOM directly of course: https://stackoverflow.com/a/24470253

**Implementation-wise, I think we can get the behaviour pattern I described above by:**
1. Set the `poster` property to a transparent gif (it can be in-line)
2. Set an initial class on the video tag that sets the background to 'Loading...'
3. Attach a `canplay` event handler to the appropriate (video or source nested in the video) tag so that it removes the `loading` class _and_ sets the `poster` to empty; this way the user will see a proper preview of the video as soon as it is in a state that it can be played.
4. Attach an `error` event handler to the `<source>` tag inside the `<video>` tag (see the post that I referenced in my first comment, above in this thread).  If that handler fires, we set a class on the `video` tag that sets the background of the video tag to our `FailedToLoad` SVG _and_ re-configures the `poster` to be the blank gif.

What do you think?

MaximusHaximus:
@corwinharrell I think that Ideally, we'd have a `Loading` SVG image that matches the stylistic choices that were made for the `FailedToLoad` image so we can display that image in cases where the video failed to load.  Is that something you can create?

june07:
Actually, the Loading svg is likely overkill as the <video> element natively has its own loading animation which we are all very familiar with.  The key was adding the transparent gif as per your suggestion and the post you provided.  Now the failed to load svg only shows in the poster area when that condition exists, otherwise the spinning loader is shown while loading, or on success the video.

packages/frontend/src/components/wallet/NFTBox.js
```diff
@@ -130,12 +130,22 @@ const NFTBox = ({ tokenDetails }) => {
             {
                 ownedTokensMetadata &&
                 <div className='tokens'>
-                    {ownedTokensMetadata.map(({ token_id, metadata: { mediaUrl, title } }) => {
+                    {ownedTokensMetadata.map(({ token_id, metadata: { mediaUrl, title } }, index) => {
+                        const videoProps = index === 0 ? { autoPlay: true } : {};
```
MaximusHaximus, line 134: I think `autoPlay` should be all lowercase for consistency with other video html tag props

packages/frontend/src/components/wallet/NFTBox.js
```diff
@@ -130,12 +130,22 @@ const NFTBox = ({ tokenDetails }) => {
...
                         return <div className='nft' key={token_id}>
-                            <img src={mediaUrl} alt='NFT' onError={(e) => {
-                                e.target.onerror = null;
-                                e.target.src = FailedToLoad;
-                            }}/>
+                            {
+                                mediaUrl.match(/\.webm$/i)
+                                    ? <video poster="data:image/gif,AAAA" muted={true} loop controls { ...videoProps } autoPlay={true}>
```
MaximusHaximus, line 138: 🔪 this appears to always set the autoplay prop to true; probably left-over from before you added the 'only set autoplay on the tag if it is the first NFT being rendered for the current contract'

MaximusHaximus (changes requested):
Thanks so much for your work on this -- I think we're just about there, just 2 comments on the autoplay property and we can get this merged 💖

packages/frontend/src/components/wallet/NFTBox.js
june07: image

That's why I changed that.

packages/frontend/src/components/wallet/NFTBox.js
june07: And yes, you are correct.  Fixed it.

june07:
It's funny how complex such a simple thing can be sometimes and how our "understanding" of a thing often changes... my own understanding being what I'm referring to.

When making the fix above and removing the extra autoplay so that only the first video in each group is set to autoplay, I noticed while testing this change that the other video's now all showed the clear gif image as the poster (not pretty) instead of the first frame of the video.  So... digging back into that post and WHY the gif was being used in the first place, led me to understand it's because they were trying to display an image WHILE loading (which going back to my last post, I think is a bit of overkill) as everyone is familiar with 
image

which again is only shown when actual loading is happening, NOT in an error state.  Who is on slow 3G/dialup now days?  Do you concur?

june07:
Is this waiting on me for something?

MaximusHaximus (approved):
💖 Thanks so much for your help!

PR #2333, opened 2021-12-23
Sign transaction doc

esaminu:
Add docs for signing a transaction via the wallet by calling `requestSignTransactions` on `near-api-js`


worldclassdev (approved):
Looks good to me! 🙌🏾

MaximusHaximus (approved):
@esaminu Merging this since the current documentation is usable -- please create a follow-up PR with the changes we discussed earlier this week to describe the URL API specifically as its own section :)

PR #2336, opened 2021-12-24
Docs: Login with NEAR Wallet

gutsyphilip:
Added docs for login with NEAR wallet by calling requestSignIn on near-api-js


esaminu:
Should we add some docs about retrieving currently logged in user and doing stuff with it? e.g. `WalletConnection.getAccountId` seems useful to add. LGTM aside from this 😄

worldclassdev:
> Should we add some docs about retrieving currently logged in user and doing stuff with it? e.g. `WalletConnection.getAccountId` seems useful to add. LGTM aside from this 😄

Yes, I think it's a useful addition. I'll do that now.

esaminu (approved):

packages/frontend/package.json
```diff
@@ -80,6 +80,7 @@
     "babel-eslint": "^10.1.0",
     "babel-jest": "^27.0.1",
     "babel-loader": "^8.1.0",
+    "concurrently": "^6.5.1",
```
MaximusHaximus, line 83: 🔪 Was this accidental? It doesn't appear to be used in this PR

MaximusHaximus (approved):
:+1: looks good -- just one question about the yarn.lock and package.json changes which appear to be unnecessary.

While this documentation can be landed as-is for now, it is closer to the type of documentation we'd expect to see in `near-api-js` than specific to the wallet -- `near-api-js` owns the API for the `SignInOptions` and other objects like `WalletConnection` -- those aren't really 'our' APIs.  The wallet owns the URL structure and URL parameters that NAJ uses to interface with the wallet.

My instinct here is that a lot of the content in this file currently could moved to live in the `near-api-js` repo alongside the code itself, and we could link to that document from here.  

@gutsyphilip please create an issue to follow up on this; at the very least we should have a section in this file that describes the route that this uses and the URL parameters that are valid (and their formats), with a strong recommendation to use `near-api-js` to interact with the wallet as a disclaimer :)

MaximusHaximus:
Feel free to merge this once you've removed the yarn.lock/package.json changes <3 @gutsyphilip

packages/frontend/package.json
gutsyphilip: The prepush git hook script in `package.json` makes use of it. Hence, my adding it as a dev dependency.

gutsyphilip:
> 👍 looks good -- just one question about the yarn.lock and package.json changes which appear to be unnecessary.
> 
> While this documentation can be landed as-is for now, it is closer to the type of documentation we'd expect to see in `near-api-js` than specific to the wallet -- `near-api-js` owns the API for the `SignInOptions` and other objects like `WalletConnection` -- those aren't really 'our' APIs. The wallet owns the URL structure and URL parameters that NAJ uses to interface with the wallet.
> 
> My instinct here is that a lot of the content in this file currently could moved to live in the `near-api-js` repo alongside the code itself, and we could link to that document from here.
> 
> @gutsyphilip please create an issue to follow up on this; at the very least we should have a section in this file that describes the route that this uses and the URL parameters that are valid (and their formats), with a strong recommendation to use `near-api-js` to interact with the wallet as a disclaimer :)

I actually agree that much of this is `near-api-js` documentation. It had me confused for a while. But it seemed like the direction we were going for now. I'd merge this and create the follow up issue.

Issue #2330, opened 2021-12-22
Hide NFT contracts with zero NFTs


Patrick1904:
**Problem**
The wallet currently shows all NFT contracts that the user has interacted with, not taking into consideration whether the user currently has any NFTs with a specific contract.

**Solution**
Hide NFT contracts with zero NFTs

**Example**
<img width="673" alt="Screen Shot 2021-12-22 at 2 28 35 PM" src="https://user-images.githubusercontent.com/24921205/147162192-f8575f94-32a7-4977-88c0-4f6b5ff59215.png">

PR #2337, opened 2021-12-24
https://github.com/near/near-wallet/issues/2330

june07gh:
Fixed.

### DIFF
--- a/packages/frontend/src/components/wallet/NFTs.js
+++ b/packages/frontend/src/components/wallet/NFTs.js
@@ -59,6 +59,7 @@ const StyledContainer = styled.div`
 
 const NFTs = ({ tokens }) => {
     if (tokens.length) {
+        tokens = tokens.filter(tokenDetails => tokenDetails.ownedTokensMetadata && tokenDetails.ownedTokensMetadata.length)
         return (
             <StyledContainer>
                 {tokens.map((tokenDetails) => (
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/wallet/NFTs.js
```diff
@@ -59,6 +59,7 @@ const StyledContainer = styled.div`
 
 const NFTs = ({ tokens }) => {
     if (tokens.length) {
+        tokens = tokens.filter(tokenDetails => tokenDetails.ownedTokensMetadata && tokenDetails.ownedTokensMetadata.length)
         return (
             <StyledContainer>
                 {tokens.map((tokenDetails) => (
```
Patrick1904, line 65:
```suggestion
const NFTs = ({ tokens }) => {
    if (tokens.length) {
        return (
            <StyledContainer>
                {tokens.filter(tokenDetails => tokenDetails.ownedTokensMetadata?.length).map((tokenDetails) => (
```

Patrick1904, line 65: Thanks for making the change @june07gh! 🙏  Can we just move the filter down to `L64`?

june07gh: 👍

Patrick1904 (approved):
Thanks! 👍

PR #2342, opened 2022-01-04
fix: compute liquid amount locally instead of fetching

### LINKED ISSUE #2349: Lockups with outdated `are_transfers_enabled` value should display up to date "Available to transfer"
**Problem**
When a lockup has an outdated `are_transfers_enabled` value, the wallet displays 0 as "Available to transfer" even though transfers have been enabled in the `transfer_poll_account_id` contract. When "Available to transfer" is 0, the user cannot withdraw from the lockup via the wallet and so the wallet does not update the lockup `are_transfers_enabled`.

**Expected Behavior**
The wallet should call the `transfer_poll_account_id` directly and display the most up to date information in case transfers have been enabled by computing the value locally. The wallet will then update the lockup's `are_transfers_enabled` value when the user withdraws from the lockup by calling `check_transfers_vote` (currently works as expected). 

**Steps to reproduce**

1. Navigate to an account with a lockup where transfers have been enabled but the lockup has not updated (e.g. [suli.near](https://wallet.near.org/profile/suli.near))
2. Observe that the unlocked balance is not "Available to transfer"

image

esaminu:
Compute liquid amount locally instead of calling `get_liquid_owners_balance`. This is because `get_liquid_owners_balance` will return 0 when transfers are disabled on the lockup contract. Locally, we can call the transfer voting contract and compute the up to date liquid amount without spending gas to update the lockup contract. Gas will then be spent to update the lockup contract state when the user withdraws from the lockup via the wallet.

Closes #2349

### DIFF
--- a/packages/frontend/src/utils/account-with-lockup.js
+++ b/packages/frontend/src/utils/account-with-lockup.js
@@ -271,7 +271,7 @@ async function getAccountBalance(limitedAccountData = false) {
         const isAccDeletable = lockedAmount.isZero() && stakedBalanceLockup.isZero();
         const liquidOwnersBalanceTransfersEnabled = isAccDeletable
             ? new BN(lockupBalance.total)
-            : BN.min(ownersBalance, new BN(lockupBalance.total).sub(new BN(MIN_BALANCE_FOR_GAS)))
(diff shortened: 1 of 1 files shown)


MaximusHaximus:
@esaminu Looks good, but is there an issue that this should be linked to (e.g. are we fixing a reported defect?)

esaminu:
@MaximusHaximus Created #2349 and linked to this PR 😄

PR #2347, opened 2022-01-06
docs: add docs for fungible token discovery and display

### LINKED ISSUE #2346: Wallet displaying USD values from `ft_metadata` of fungible tokens
**Problem**
Wallet fungible token display will render a USD value returned by the `usd` property of `ft_metadata` payload

**Expected Behavior**
Wallet fungible token display should not render fiat data retrieved from the `ft_metadata` function call and only render fiat data from `/fiat` endpoint of contract helper.

**Steps to reproduce**

1. Own a fungible token balance with a contract that implements `ft_metadata` with a `usd` value. 
2. Observe the wallet token display:

image

esaminu:
- Add docs for fungible token discovery on NEAR wallet and how FTs are displayed.
- Fix issue where keys from `ft_metadata` that conflict with token display prop maybe be rendered by refactoring token object to be keyed by data source. (Closes #2346)


packages/frontend/src/hooks/fungibleTokensIncludingNEAR.js
```diff
@@ -1,35 +1,26 @@
...
-        },
-        ...Object.values(tokens)
-    ];
+const useNEARAsTokenWithMetadata = () => {
+    const nearBalance = useSelector(selectBalance);
+    const nearTokenFiatValueUSD = useSelector(selectNearTokenFiatValueUSD);
+
+    return {
+        balance: nearBalance?.balanceAvailable || "",
+        onChainFTMetadata: { symbol: "NEAR" },
```
MaximusHaximus, line 13: Not a blocker, but for consistency, should there be an `icon` here, I wonder....

MaximusHaximus (approved):
:+1: lookin' good! nice work <3

PR #2352, opened 2022-01-07
feat: add prompts to add and remove environments

andy-haynes:
This PR updates the `feature-flags` package to include new prompts for adding and removing environments.

When adding a new environment, the user will be prompted for the new environment name and the existing source environment whose flag state will be used for the new environment. For example, given the new environment `betanet`, choosing `testnet` as the source environment will add `betanet` with the same set of enabled and disabled flags as `testnet`.

Any modification made to `environments.json` as a result of the new commands ensures that every flag in `flags.json` has exactly one entry for each environment in `environments.json`.


packages/feature-flags/src/prompts/selectExistingEnvironment.js
```diff
@@ -0,0 +1,29 @@
+const inquirer = require('inquirer');
+
+module.exports = async function selectExistingEnvironment(environments) {
+    const { proceed } = await inquirer.prompt({
+        type: 'confirm',
+        name: 'proceed',
+        message: `
+💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥💥
+Removing or modifying existing environments WILL break any production environments using this configuration.
```
MaximusHaximus, line 9: :nit: Given we only use this prompt when deleting the environment, would it make more sense for this to be `selectEnvironmentForDeletion` and have the wording here changed to remove the reference to 'modifying existing environments' in this description?

packages/feature-flags/src/prompts/enterNewEnvironmentName.js
```diff
@@ -0,0 +1,24 @@
+const inquirer = require('inquirer');
+
+module.exports = async function newFlagName(existingEnvironments) {
```
MaximusHaximus, line 3: :nit: Since this both asks for the new environment's name _and_ gathers which env should be the sourceEnvironment, what do you think about calling this `setupNewEnvironment.js`?

Also :nit:, function name here is misnamed (probably copy/pasta'd)

packages/feature-flags/src/FlagEditor.js
```diff
@@ -27,45 +29,68 @@ class FlagEditor {
...
+                break;
             }
         }
 
+        await this.saveEnvironments();
         await this.saveFlags();
         await this.writeTypeDefinitions();
     }
 
+    async setFlagState(flagName, userEditing) {
```
MaximusHaximus, line 81: 💖

MaximusHaximus (approved):
<3 :+1:

Just a couple of nits, otherwise LGTM :shipit: 🚢

packages/feature-flags/src/prompts/enterNewEnvironmentName.js
andy-haynes: Good catch ❤️

PR #2356, opened 2022-01-11
docs: add docs for non-fungible token discovery and display

esaminu:
Add docs for non-fungible token discovery on NEAR wallet and how NFTs are displayed.


MaximusHaximus (approved):
:+1: nice work, great explanations :) ✋

PR #2368, opened 2022-01-13
Refactor: remove store usage in alerts helpers

marcinbodnar:
With this PR we are removing all imports of the `store` and using thunks instead for `alerts.js` file.

 - `dispatchWithAlert` is now `withAlert` thunk and is moved to `status` slice file.
 - `actionsPending` is now `selectActionsPending` selector.
 - `actionsPendingByPrefix ` is now `selectActionsPendingByPrefix` selector.
 - `handleClearAlert` is now a thunk and is moved to `status` slice file.


packages/frontend/src/redux/slices/status/index.js
```diff
@@ -10,3 +16,18 @@ export const selectStatusMainLoader = createSelector(selectStatusSlice, (status)
 export const selectStatusLocalAlert = createSelector(selectStatusSlice, (status) => status.localAlert || {});
 
 export const selectStatusActionStatus = createSelector(selectStatusSlice, (status) => status.actionStatus || []);
```
andy-haynes, line 18: I know this isn't new to this PR but shouldn't the fallback here be a `{}`? It looks like all usages of this selector rely have property access on them.

andy-haynes (changes requested):
Just a few comments inline and some rebase conflicts, otherwise looking good

packages/frontend/src/redux/slices/status/index.js
```diff
@@ -10,3 +16,18 @@ export const selectStatusMainLoader = createSelector(selectStatusSlice, (status)
...
+
+export const selectActionsPending = createSelector(
+    [getTypes, selectStatusActionStatus],
+    (types, actionStatus) => (typeof types === 'string' ? [types] : types).some((type) => actionStatus[type]?.pending)
+);
+
+export const selectActionsPendingByPrefix = createSelector(
```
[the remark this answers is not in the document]
marcinbodnar: Done.

packages/frontend/src/redux/slices/status/index.js
marcinbodnar: You're right. Thanks. Done.

andy-haynes (approved):
LGTM 👍

PR #2369, opened 2022-01-13
Disallow foxtrot commits

MaximusHaximus:
Adds hooks to warn you when you have just created a foxtrot commit, and to block pushing branches to the remote that contain those commits.  

Apart from the issue of bundling up changes that were not part of the feature branch's changeset into that branch, making it difficult or impossible to revert a merged branch without possibly breaking other code, this is going to be especially critical as we start generating changelog output using conventional commit contents, as not being able to follow the 'first parent' path can create serious problems with tooling that tries to generate changelog diffs automatically.

Further reading at:
https://git-blame.blogspot.com/2012/03/fun-with-first-parent.html
http://bit-booster.blogspot.cz/2016/02/no-foxtrots-allowed.html
https://stackoverflow.com/questions/35962754/how-can-i-prevent-foxtrot-merges-in-my-master-branch

The one major downside to rebasing is when you have many commits in a branch that have the _same_ conflict to be resolved while resolving the rebase; it can be tedious to re-apply the same merge resolution to multiple commits.  You can handle this by doing an interactive rebase where you squash multiple commits together so you don't need to resolve the conflict multiple times, but that is not always a good idea.  In cases like this, take advantage of the `rerere` git option to have git automatically apply a previous conflict resolution for you in cases where it detects the same conflict has been encountered in multiple commits during the rebase.
Reference: https://stackoverflow.com/questions/49500943/what-is-git-rerere-and-how-does-it-work

### DIFF
--- a/.git-hooks/helpers/detect-foxtrot-merge
+++ b/.git-hooks/helpers/detect-foxtrot-merge
@@ -1,0 +1,39 @@
+#!/bin/sh
+
+#usage:
+#   foxtrot-merge-detector [<branch>]
+#
+# If foxtrot merge detected for branch (current branch if no branch),
+# exit with 1.
+
+# foxtrot merges:
+# See http://bit-booster.blogspot.cz/2016/02/no-foxtrots-allowed.html
+# https://stackoverflow.com/questions/35962754/git-how-can-i-prevent-foxtrot-merges-in-my-master-branch
+
+remoteBranch=$(git rev-parse --abbrev-ref "$1"@{u} 2>/dev/null)
+# no remote tracking branch, exit
+if [[ -z "$remoteBranch" ]]; then
+    echo "no remote tracking branch, no foxtrot merge detected"
+    exit 0
+fi
+branch=$(git rev-parse --abbrev-ref "${1-@}" 2>/dev/null)
+# branch commit does not cover remote branch commit, exit
+if ! $(git merge-base --is-ancestor $remoteBranch $branch); then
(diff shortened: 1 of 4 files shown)


gutsyphilip (approved):
LGTM! 🙌🏾

.git-hooks/helpers/detect-foxtrot-merge
```diff
@@ -0,0 +1,41 @@
...
+# foxtrot merges:
+# See http://bit-booster.blogspot.cz/2016/02/no-foxtrots-allowed.html
+# https://stackoverflow.com/questions/35962754/git-how-can-i-prevent-foxtrot-merges-in-my-master-branch
+# https://git-blame.blogspot.com/2012/03/fun-with-first-parent.html
+# https://stackoverflow.com/questions/35962754/how-can-i-prevent-foxtrot-merges-in-my-master-branch
+
+remoteBranch=$(git rev-parse --abbrev-ref "$1"@{u} 2>/dev/null)
+# no remote tracking branch, exit
+if [[ -z "$remoteBranch" ]]; then
+    echo "no remote tracking branch, no foxtrot merge detected"
```
andy-haynes, line 18: Will we actually see the non-error `echo`s in practice? My preference would be to not get notified when nothing is wrong.

MaximusHaximus: Yup, we would, and I'd generally agree, but the husky pipeline is pretty verbose already so I didn't notice them --  commented these out for now :)

PR #2380, opened 2022-01-18
Remove store usage related to creating account from wallet.js

marcinbodnar:
With this PR we are removing all usage of `store` related to creating an account from `wallet.js`.

It's refactoring wallet methods into thunks:
 - `addLocalKeyAndFinishSetup`
 - `createIdentityFundedAccount`
 - `createNewAccount`
 - `createNewAccountWithCurrentActiveAccount`
 - `refreshAccount`

It's refactoring actions into `createAsyngThunk`:
 - `createAccountFromImplicit`
 - `createAccountWithSeedPhrase`

This could have been achieved in a simpler way, but I put some additional effort into it, so it will start some changes that we ware planning for a long time. The idea is to:
 - simplify the wallet.js - makes it more like a `FungibleTokens.js` and `NonFungibleTokens.js` services, that holds more basic functionalities and requests instead of complicated logic that is related to our project
 - this way the `wallet.js` will be much more readable and fulfill the task of service, that contributors might use to understand requests logic
 - more complicated logic will be extracted from wallet.js into thunks and from components to thunks - making also the components code more readable
 - additionally, logic related directly to creating an account will be in the same file, which is handy because there is a lot of logic related to all ways of creating an account.

There is more that might be moved into thunks, but for now, it's more like an introduction of an idea, and a first step.


marcinbodnar:
@Patrick1904 could you please check if all functionality related to creating an account is working properly with this PR? It's been a while since I was working on things related to creating an account, and I can see that a lot has changed regarding the code. I tested it, but I'm not sure If I tested all possible scenarios.

packages/frontend/src/components/accounts/CreateAccount.js
```diff
@@ -10,7 +10,6 @@ import { Mixpanel } from '../../mixpanel/index';
 import {
     checkNearDropBalance,
     checkNewAccount,
-    createNewAccount,
```
andy-haynes, line 13: ❤️

packages/frontend/src/components/accounts/SetupImplicit.js
```diff
@@ -96,7 +97,7 @@ class SetupImplicit extends Component {
         this.setState({ creatingAccount: true });
         await Mixpanel.withTracking("CA Create account from implicit",
             async () => {
-                await dispatch(createAccountFromImplicit(newAccountId, implicitAccountId, recoveryMethod));
+                await dispatch(createAccountFromImplicit({ accountId: newAccountId, implicitAccountId, recoveryMethod }));
```
andy-haynes, line 100: This might be a dumb question (my redux is a little rusty) but why do some async thunks get wrapped in `dispatch` and some are just `await`'d directly? If calling `dispatch` on them is a no-op can we always just `await` them directly?

packages/frontend/src/redux/slices/account/createAccountThunks.js
```diff
@@ -0,0 +1,176 @@
...
+        publicKey,
+        previousAccountId,
+        recaptchaToken
+    }, { dispatch }) => {
+        await wallet.checkNewAccount(accountId);
+
+        const { fundingContract, fundingKey, fundingAccountId } = fundingOptions || {};
+        if (fundingContract && fundingKey) {
+            await wallet.createNewAccountLinkdrop(accountId, fundingContract, fundingKey, publicKey);
+            await wallet.keyStore.removeKey(NETWORK_ID, fundingContract);
```
andy-haynes, line 101: I see this is a pattern we've been using elsewhere for a while but it would be nice at some point to wrap this behind a method in `wallet` (e.g. `await wallet.removeNetworkKey(...)` or something) to avoid exposing too much and protect against future implementation changes in `wallet`. Just thinking out loud, not something that needs to be addressed here 🙂

andy-haynes (changes requested):
Just a question on consistency with calling async thunks (`dispatch(...)` vs `await ...`). Looks good otherwise, just needs a rebase 🙂

packages/frontend/src/components/accounts/SetupImplicit.js
Patrick1904: @andy-haynes class components use `mapDispatchToProps` and therefore don't need the explicit `dispatch` when called. We're generally moving away from class components in favor of functional components using explicit `dispatch`:)

Patrick1904:
> @Patrick1904 could you please check if all functionality related to creating an account is working properly with this PR? It's been a while since I was working on things related to creating an account, and I can see that a lot has changed regarding the code. I tested it, but I'm not sure If I tested all possible scenarios.

Yep can we please just rebase this PR first since we recently shipped quite a lot of create account related changes

marcinbodnar:
> Yep can we please just rebase this PR first since we recently shipped quite a lot of create account related changes

@Patrick1904 Done.

packages/frontend/src/redux/slices/account/createAccountThunks.js
marcinbodnar: That is a good idea.

Patrick1904:
@marcinbodnar I did a run through and everything checks out except when creating a **second account** with email link. It gets stuck at the 'enter verification code' step. 

To reproduce:
1. First create an implicit account (or simply recover/select an existing implicit account)
2. Try creating a second account using email as recovery method
3. Observe getting stuck at code verification step

marcinbodnar:
Thanks @Patrick1904 
Fixed it, please check.

Patrick1904 (approved):
LGTM. Tested implicit account creation and then second named account funded by the first implicit account.

PR #2384, opened 2022-01-19
Display farm token rewards

esaminu:
Patch to add farm token rewards display and functionality to claim farm tokens while #2363 is in progress. 

- Fetches farms and accrued token rewards if validator is a project validator
- Renders on validators page with FT metadata
- ~~Adds functionality to claim accrued token rewards and redirect to home~~ will be done in next iteration

screencapture-localhost-1234-staking-validator2-factory01-littlefarm-testnet-2022-01-19-04_49_53


packages/frontend/src/components/staking/components/Validator.js
```diff
@@ -21,20 +27,41 @@ export default function Validator({
...
     const showConfirmModal = confirm === 'withdraw';
 
+    useEffect(() => {
+        const getFarms = async () => {
+            const farms = await validator.contract.get_farms({from_index: 0, limit: 300});
+            setFarmList(await Promise.all(farms.map(({ token_id }, i) => {
+                dispatch(fetchToken({contractName: token_id}));
+                return validator.contract
+                    .get_unclaimed_reward({ account_id: accountId, farm_id: i })
+                    .catch(() => "0")
```
esaminu, line 46: `catch` is used because `get_unclaimed_reward` [will throw](https://github.com/referencedev/staking-farm/blob/master/staking-farm/src/views.rs#L148) when the account has not accrued any rewards with the contract yet

MaximusHaximus: Follow-up: error state; if we get a network request error (or really, anything other than `ERR_NO_ACCOUNT` or `ERR_NO_FARM` type errors), we shouldn't be showing the user a 0 value here.  This will need some thought from Corwin on design.

packages/frontend/src/redux/slices/tokens/index.js
```diff
@@ -76,6 +76,26 @@ const fetchTokens = createAsyncThunk(
     }
 );
 
+const fetchToken = createAsyncThunk(
+    `${SLICE_NAME}/fetchTokens`,
```
MaximusHaximus: This action name is incorrect :)

packages/frontend/src/redux/slices/tokens/index.js
```diff
@@ -76,6 +76,26 @@ const fetchTokens = createAsyncThunk(
...
+            }
+            if(accountId) {
+                await dispatch(fetchOwnedTokensForContract({ accountId, contractName }));
+            }
+        } catch (e) {
+            // Continue loading other likely contracts on failures
+            console.warn(`Failed to load FT for ${contractName}`, e);
+        }
+    }
+);
```
MaximusHaximus, line 97: I think that this thunk should have a `condition` argument provided to it with the same code as the `fetchOwnedTokensForContract` thunk does, so that we don't re-fetch contract metadata for a contract that we are already fetching it for.

packages/frontend/src/components/staking/components/Validators.js
```diff
@@ -38,12 +50,14 @@ export default function Validators({ validators, stakeFromAccount }) {
...
-                {validators.filter(v => v.accountId.includes(validator)).map((validator, i) => 
-                    <ValidatorBox
-                        key={i}
-                        validator={validator}
-                    />
-                )}
+                {
+                    filterValidators.map((validator, i) =>
+                        <ValidatorBox
+                            key={i}
```
MaximusHaximus, line 56: Not a blocker since this was already wrong, but the `key` property should be the validator account _name_ instead of an index into the array of validators (key as array index is a react antipattern) -- we can fix this in the next PR

packages/frontend/src/components/staking/components/Validators.js
```diff
@@ -1,15 +1,27 @@
...
-export default function Validators({ validators, stakeFromAccount }) {
 
+export default function Validators({ validators, stakeFromAccount }) {
     const [validator, setValidator] = useState('');
-
+    const networkId = wallet.connection.provider.connection.url.indexOf(MAINNET) > -1 ? MAINNET : TESTNET;
+    const validatorPrefix = getProjectValidatorPrefix(networkId);
     const validValidator = validators.map(validator => validator.accountId).includes(validator);
 
+    const filterValidators = useMemo(() =>
```
MaximusHaximus, line 17: For now, we can't modify how we sort validators without a buy-in at the product level.  If I recall correctly, we intentionally _shuffle_ the list of validators every time we fetch them to avoid favoritism or 'gaming the system' by making all the validators start with the letter 'a' or other goofy things.  Since there is only 1 pool currently (AFAIK) implementing this new pool type, this would result in us putting that one pool at the top of the list every time, which we need to avoid to maintain our existing policy of randomizing the order of validators in the list.

MaximusHaximus (changes requested):
Just a couple things, then we can get this live!

packages/frontend/src/components/staking/components/BalanceBox.js
```diff
@@ -86,7 +119,26 @@ export default function BalanceBox({
...
+                    </div>
+                    {token.onChainFTMetadata?.symbol === "NEAR" &&
+                    !token.contractName ? (
+                        <Balance
+                            amount={token.balance}
+                            data-test-id={balanceTestId}
+                            symbol={false}
+                        />
+                    ) : (
+                        <TokenAmount
```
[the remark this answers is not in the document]
esaminu: added 😄

packages/frontend/src/redux/slices/tokens/index.js
esaminu: As discussed, an issue will be created for this since it requires changes to the state tree

packages/frontend/src/components/staking/components/Validators.js
esaminu: Restored initial rendering

packages/frontend/src/components/staking/components/Validator.js
esaminu: Makes sense. For now I think this call silently failing unless we have data is fine (default balance is 0).

MaximusHaximus (approved):
💖 Nice work @esaminu! Thanks for jumping into the fire and getting this ready rapidly :)

PR #2385, opened 2022-01-19
Create Account: only show verification options if funding is available

Patrick1904:


MaximusHaximus (approved):
<3 nice work, clean fix and I like the new terms for when there's no coin-op running 🚀

Issue #2383, opened 2022-01-18
Stake Farming Feature on Wallet 


pretfood:
**Problem**

On stake farming contracts, users cannot see rewards in their wallet. At present, only $NEAR rewards are visible on the wallet. In new stake farming contracts, rewards are returned in $AURORA, but these are not visible. 

**Expected Behavior**

Staking rewards should be displayed in more than one currency. 

**Steps to reproduce**

1. Stake $NEAR tokens in aurora.pool.near
2. Watch how the rewards number never moves beyond 0, despite claimed APY of 29%.

stefanopepe:
There's work in progress, but no commitment on the timing for the release, as it needs a final review on the UX side.

esaminu:
@corwinharrell Would appreciate some design input on this. We created a first iteration on #2384 that impacts the main [staking page](https://wallet.staging.near.org/staking) to display NEAR token icons and the validator page to display NEAR balance with icon and other tokens if any e.g. the [aurora validator](https://wallet.staging.near.org/staking/aurora.pool.near):

<img src="https://user-images.githubusercontent.com/5849204/150241837-ba74ca75-28ad-4005-a3a8-34bfac0212cb.png" width="500">

corwinharrell:
<img width="563" alt="Screen Shot 2022-01-20 at 6 47 43 PM" src="https://user-images.githubusercontent.com/5934188/150440558-28ad0f05-f58b-4c8e-befe-addf1cc5684c.png">

Until we have a new UI, I think it is actually clearest to group the rewards under a single section and label. This way, it simply becomes a list of `Rewards earned`, and the icon and symbol denote the particular token, without the need for separate labels and sections.

I think the only remaining necessary change then would be to change the tooltip icon to read something like:

Rewards earned but not withdrawn. NEAR token rewards are compounding and are automatically re-staked.

esaminu:
UI and copy for v1 (token rewards display) updated in #2388

PR #2388, opened 2022-01-20
chore: Update staking farm related copy

MaximusHaximus:
1. Alphabetizes the JSON files now that we have automatic tooling added to handle this whenever they are edited
2. Updates copy for token farm validators

### DIFF
--- a/packages/frontend/src/translations/en.global.json
+++ b/packages/frontend/src/translations/en.global.json
@@ -1227,7 +1227,7 @@
             },
             "farm": {
                 "button": "Claim",
-                "info": "This is project validator. You’ll get the farming reward while staking on it.",
+                "info": "This validator grants staking rewards in a token other than NEAR."
             },
             "pending": {
                 "info": "These tokens have been unstaked, but are not yet ready to withdraw. Tokens are ready to withdraw 52 to 65 hours after unstaking.",
--- a/packages/frontend/src/translations/en.global.json
+++ b/packages/frontend/src/translations/en.global.json
@@ -1228,7 +1228,6 @@
             "farm": {
                 "button": "Claim",
                 "info": "This is project validator. You’ll get the farming reward while staking on it.",
-                "title": "Total token earned"
             },
             "pending": {
                 "info": "These tokens have been unstaked, but are not yet ready to withdraw. Tokens are ready to withdraw 52 to 65 hours after unstaking.",
@@ -1240,7 +1239,7 @@
                 "title": "Total amount staked"
             },
             "unclaimed": {
-                "info": "Rewards that have been earned, but not withdrawn. Rewards earned are automatically restaked, which means your Total Amount Staked automatically increases over time, and your rewards are <a target='_blank' href='https://www.investopedia.com/terms/c/compoundinterest.asp'>compounding</a>.",
+                "info": "Rewards that have been earned, but not withdrawn. NEAR token rewards are  <a target='_blank' href='https://www.investopedia.com/terms/c/compoundinterest.asp'>compounding</a> and are automatically re-staked.",
                 "title": "Rewards earned",
                 "unavailable": {
                     "cta": "Learn More",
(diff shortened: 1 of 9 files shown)


MaximusHaximus:
@corwinharrell Would like your input on the new tooltip text and the new balance of tokens earned via the staking farm.

```
                "info": "This is a 'staking farm'. You will receive tokens other than NEAR as rewards while staking on it.",
                "title": "Un-claimed farm tokens earned"
```

corwinharrell:
@MaximusHaximus Could you add screenshots of the affected UI in context? Just helps to review things like this without going through setup locally. Would love to get into a habit of this any time there are changes that affect the front-end/UI

MaximusHaximus:
@corwinharrell You don't need to set anything up locally to see how changes look in context, since we automatically deploy previews for every PR with both testnet and mainnet configurations.  

- You can see link for the mainnet staging preview under the 'checks' area of the PR at the bottom of this screen -- click 'Details' next to the row labelled `netlify/near-wallet-staging/deploy-preview — Deploy Preview ready!`.
- You can see the testnet preview build by using the link in the comment that Render added to this PR as the first response comment (https://github.com/near/near-wallet/pull/2388#issuecomment-1017004487)

Here's a picture of the screen with the new terms being rendered for your convenience, but it would probably be a good idea for you to get in the habit of using the preview builds in general, since screenshots are pretty limited (especially for changes that can impact multiple screens)
image

corwinharrell:
Sorry for the miscommunication @MaximusHaximus. I'm totally comfortable using the deploy links. My comment about adding an image was purely an aside and for convenience sake for future PRs. Didn't mean to suggest it was a blocker. Had to step away for a bit this afternoon but looking at this now.

packages/frontend/src/translations/en.global.json
```diff
@@ -1226,8 +1226,8 @@
             },
             "farm": {
                 "button": "Claim",
-                "info": "This is project validator. You’ll get the farming reward while staking on it.",
-                "title": "Total token earned"
+                "info": "This is a 'staking farm'. You will receive tokens other than NEAR as rewards while staking on it.",
+                "title": "Total tokens earned"
```
corwinharrell, line 1230: How would we feel about:

"Additional tokens earned"

packages/frontend/src/translations/en.global.json
```diff
@@ -1226,8 +1226,8 @@
             },
             "farm": {
                 "button": "Claim",
-                "info": "This is project validator. You’ll get the farming reward while staking on it.",
-                "title": "Total token earned"
+                "info": "This is a 'staking farm'. You will receive tokens other than NEAR as rewards while staking on it.",
```
corwinharrell, line 1229: How about:

"This pool awards delegator stake with tokens other than NEAR."

corwinharrell:
Screen Shot 2022-01-20 at 6 47 43 PM

Until we have a new UI, I think it is actually clearest to group the rewards under a single section and label. This way, it simply becomes a list of rewards earned, and the icon and symbol denote the token, without the need for separate labels and sections.

I think the only remaining necessary change then would be to change the tooltip icon to read something like:

"Rewards earned but not withdrawn. NEAR token rewards are compounding and are automatically re-staked."

corwinharrell:
@MaximusHaximus See this comment: https://github.com/near/near-wallet/issues/2383#issuecomment-1018028825 since I think it has a direct impact on the changes proposed in this PR.

MaximusHaximus:
@corwinharrell I modified the UI to look like the picture you've created, and modified the translation terms accordingly -- thanks for the feedback/direction!

MaximusHaximus:
@Patrick1904 I'm pretty sure my usage of dynamic props for the container here isn't what you'd want given I haven't seen this pattern in other code -- I'm shipping with that setup in the interest of time but am keen to have a chat with you when you're back about the right way to handle the border formatting in the list 'the right way, and to understand the benefits of avoiding dynamic props usage for our style components if that's what we're generally trying to do as a team. :)

PR #2294, opened 2021-12-03
Chore: Improve linting & auto-alphabetize translations

worldclassdev:
This PR is a proposal and feature of some sort that contains a series of efforts to help ensure consistency across the codebase. See details below:

# Linting and Formatting
For consistency and predictable organization of code, this project uses [Eslint](https://eslint.org/) for linting and formatting. 

## The Setup
The configuration for Eslint can be found at the root of this project:
- Eslint - `.eslintrc.js`

VSCode is configured to automatically fix fixable linting errors on save. See `.vscode/settings.json`. You can configure Webstorm or other editors to do the same!

With in`package.json`, there are scripts that you can run to lint, autoFix and format the entire codebase.

```
"lint": "eslint --ext .js --ext .jsx .",
"fix": "eslint --ext .js --ext .jsx . --fix",
```

##  Git Commit hooks
Git commit hooks are a provision that allows us run custom scripts when specific events occur in the git workflow. [See more here](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks).

For this project, we are using the clientside commit hooks to automatically lint and alphabetize staged files before they get committed.

## Auto-alphabetization of translation files
This is meant to help ensure consistency and ease of reference within the translation files. For this, we are making use of git hooks via a package called [husky](https://typicode.github.io/husky/#/). Husky allows us to configure a defined pre-commit command which triggers the `precommit` script within modified packages. The `precommit` script in `packages/frontend/package.json` is configured to run [lint-staged](https://github.com/okonet/lint-staged). Lint-staged allows us run linters on files that have been staged through git. 

Within `package.json`, the block below defines what linters and formatters get applied to what filetype.

```
 "lint-staged": {
    "*.{js,ts,tsx, jsx}": [
      "eslint --quiet --fix",
    ],
    "src/translations/*.json": "sort-json --ignore-case true"
  }
 ```
 
Two main things are ensured through this configuration:
- Staged translation files `src/translations/*.json"` get sorted in alphabetical order automatically using a package called [sort-json](https://github.com/kesla/sort-json)
- Staged `.js` and `.jsx` files get linted for any major linting errors and committing will fail until linting errors are resolved e.g unused vars. We can make the rules as tight or loose as we want through the configuration in `.eslintrc.js`.

## Proposal: Commit messages
Within `.husky/commit-msg`, there's a script that ensures consistent formatting with commit messages using the `commit-msg` git hook. 

The proposed format is:
```
git commit -m "{commit type goes here}: {commit information}"
```

[...]

PR #2274, opened 2021-11-30
Update Portuguese translation and missing terms

VGasparini:
### Hello :wave: 

This is a PR to update the NEAR Wallet Portuguese translation.

Were an initiative made on Hacktoberfest with supervision and review by guild [NEAR Portuguese](https://www.notion.so/makersnetwork/Comunidade-NEAR-Portugu-s-f385ce229c5444468b9b9c6073e6e5db) 

### Demo
Running locally

image

### Tests
image

PR #2389, opened 2022-01-20
2022-01-19 Release

MaximusHaximus:
- #2294
- #2320 
- #2216 
- #2337 
- #2350 
- #2342 
- #2333 
- #2347 
- #2352 
- #2299 
- #2215 
- #2361 
- #2362 
- #2174 
- #2367 
- #2336 
- #2359 
- #2274 
- #2384 
- #2313
- #2356 
- #2264 
- #2385 
- #2390 
- #2388


corwinharrell (changes requested):
1. Looks like there is still some copy that was not updated in the new flow: Once the implicit account is created, the module on the dashboard reads: "Add a Custom Address..." Once the user clicks through to perform this action, the next screen reads "Reserve Account ID". Instead it should read "Add a Custom Address". We need to ensure that all of the copy is consistent with the new designs, and that this screen in particular is updated, since it contains important contextual information. cc @Patrick1904 

2. The following step is still present in the new flow:

Screen Shot 2022-01-20 at 12 49 42 PM

What we had agreed on was that this step would be eliminated in favor of defaulting to funding the new custom address with the active account. Persisting the manual deposit option here means that the user has the option to fund their new custom address via sending funds to yet _another_ temporary implicit account, which defeats the purpose of this work entirely. These options should not persist, since this flow assumed the active account has enough funds to create a new named account, and will be the one to fund it.

PR #2390, opened 2022-01-21
Fund account with implicit flow

esaminu:
Adds conditional logic to routes on the create account flow if the currently active account is an implicit account:
* Display “Add a Custom Address” copy on `/create`
* Skip `/verify-account-id screen`
* Hide cancel button on `/fund-with-existing-account`


MaximusHaximus (approved):
<3 :+1: nice work, let's get this sucker landed!

PR #2394, opened 2022-01-21
fix: disable implicit acc creation flow on testnet

esaminu:
hotfix to disable implicit account creation flow on testnet and always use named account creation flow.


MaximusHaximus (approved):
:+1: good catch @esaminu <3

Issue #2392, opened 2022-01-21
Farming validator token claiming


Solniechniy:
## Story

As a USER I see the farming token, I want to claim so I can get a reward from staking in farming validator.

## Acceptance Criteria

- [x] Claim button is available for each token on the validator page. Should open the modal for claiming.
- [x] Token prices should be taken from ref.finance and displayed below the token name. 
- [x] Since token prices are taken, we can show the APY on the validator page and for farming validators row among all validators page.
- [x] Get whitelisted token from ref.finance and display that farming token is not whitelisted by ref.
- [ ] Works with legacy Lockup accounts
- [ ] Total token rewards earned are displayed on the [staking landing page](https://wallet.testnet.near.org/staking) for both lockup and owner accounts
- [ ] Total token rewards earned are displayed on the [profile page](https://wallet.testnet.near.org/profile) for both lockup and owner accounts

## Unhappy Path
Can't claim token

## Assets

![telegram-cloud-photo-size-2-5429190957292698673-x](https://user-images.githubusercontent.com/26657914/150552758-06a8f1ac-1df4-4c38-bb69-ca75121cc94f.jpg)
![telegram-cloud-photo-size-2-5426960864243726572-y](https://user-images.githubusercontent.com/26657914/150551973-a181c59e-3184-4a93-ac8b-8d123c345482.jpg)
![image](https://user-images.githubusercontent.com/26657914/150553413-76ddc3db-f076-49b1-9183-13abef8c29b0.png)

esaminu:
@Solniechniy we need to get design and flow approval for these first so assigning to @corwinharrell and @stefanopepe for review. Could you also include all UI including ref whitelisting for their reference?

Solniechniy:
ref whitelisting looks like that 
![telegram-cloud-photo-size-2-5438138392077515063-y](https://user-images.githubusercontent.com/26657914/150820873-ad85c82c-4a0b-47ae-aacc-cfefdb167d36.jpg)
![telegram-cloud-photo-size-2-5438138392077515064-x](https://user-images.githubusercontent.com/26657914/150820891-a6079525-c6fb-463a-90c5-08d21153977f.jpg)

Solniechniy:
But there is an issue with claim modal and APY, it looks like that 
![image](https://user-images.githubusercontent.com/26657914/150821362-cb6406ac-8901-4a71-8af0-af8a21d1ec93.png)

stefanopepe:
[...]

PR #2398, opened 2022-01-24
WAL-75 Added full UI for farming validator without claim functionality

Solniechniy:
Changes according to issue [about claiming in farming validators](https://github.com/near/near-wallet/issues/2392).
.env file: NEAR_TOKEN_ID, FARMING_CLAIM_GAS,FARMING_CLAIM_YOCTO should be provided

### DIFF
--- a/packages/frontend/src/components/staking/components/Validator.js
+++ b/packages/frontend/src/components/staking/components/Validator.js
@@ -6,7 +6,6 @@ import { Mixpanel } from '../../../mixpanel';
 import { redirectTo } from '../../../redux/actions/account';
 import { claimFarmRewards, getValidatorFarmData } from '../../../redux/actions/staking';
 import { showCustomAlert } from '../../../redux/actions/status';
-import { selectAccountId } from '../../../redux/slices/account';
 import { selectValidatorsFarmData, selectFarmValidatorAPY } from '../../../redux/slices/staking';
 import { selectActionsPending } from '../../../redux/slices/status';
 import { selectTokensFiatValueUSD, selectTokenWhiteList } from '../../../redux/slices/tokenFiatValues';
@@ -75,7 +74,6 @@ export default function Validator({
     const [confirm, setConfirm] = useState(null);
     
     const nearAsFT = useSelector(selectNEARAsTokenWithMetadata);
-    const accountId = useSelector(selectAccountId);
 
     const contractMetadataByContractId = useSelector(selectAllContractMetadata);
     const tokenFiatValues = useSelector(selectTokensFiatValueUSD);
@@ -109,7 +107,7 @@ export default function Validator({
 
         try {
             setClaimingProceed(true);
-            await dispatch(claimFarmRewards(validator.accountId, accountId, token_id));
+            await dispatch(claimFarmRewards(validator.accountId, token_id));
             setClaimingProceed(false);
             return dispatch(redirectTo(`/staking/${match.params.validator}/claim`));
         } catch (e) {
@@ -129,7 +127,7 @@ export default function Validator({
     useEffect(() => {
         if (!isFarmingValidator || !validator?.accountId) return;
 
-        dispatch(getValidatorFarmData(validator.accountId, accountId));
+        dispatch(getValidatorFarmData(validator.accountId));
     }, [validator?.accountId, isFarmingValidator]);
 
     const farmList = validatorFarmData?.farmRewards || [];
--- a/packages/frontend/src/components/staking/components/Validator.js
+++ b/packages/frontend/src/components/staking/components/Validator.js
@@ -5,6 +5,7 @@ import { useDispatch, useSelector } from 'react-redux';
 import { Mixpanel } from '../../../mixpanel';
 import { redirectTo } from '../../../redux/actions/account';
 import { claimFarmRewards, getValidatorFarmData } from '../../../redux/actions/staking';
+import { showCustomAlert } from '../../../redux/actions/status';
 import { selectAccountId } from '../../../redux/slices/account';
 import { selectValidatorsFarmData, selectFarmValidatorAPY } from '../../../redux/slices/staking';
 import { selectActionsPending } from '../../../redux/slices/status';
@@ -88,6 +89,8 @@ export default function Validator({
     const [showClaimConfirmModal, setShowClaimConfirmModal] = useState(false);
     const [selectedFarm, setSelectedFarm] = useState(null);
 
+    const [claimingProceed, setClaimingProceed] = useState(false);
+
     const openModal = (farm) => {
         setSelectedFarm(farm);
         setShowClaimConfirmModal(true);
@@ -104,9 +107,20 @@ export default function Validator({
     const handleClaimAction = async (token_id) => {
         if (!validator || !isFarmingValidator || !token_id) return null;
 
-        await dispatch(claimFarmRewards(validator.accountId, accountId, token_id));
-        // TODO: handle modal close, success and fail cases;
-        return dispatch(redirectTo('/'));
+        try {
+            setClaimingProceed(true);
+            await dispatch(claimFarmRewards(validator.accountId, accountId, token_id));
+            setClaimingProceed(false);
+            return dispatch(redirectTo(`/staking/${match.params.validator}/claim`));
+        } catch (e) {
+            setClaimingProceed(false);
+            dispatch(showCustomAlert({
+                success: false,
+                messageCodeHeader: 'error',
+                messageCode: 'staking.validator.errorClaimRewards',
+            }));
+        }
+        
     };
 
     const validatorsFarmData = useSelector(selectValidatorsFarmData);
@@ -172,7 +186,7 @@ export default function Validator({
                         token={{ ...nearAsFT, balance: validator.unclaimed || '0' }}
                         hideBorder={isFarmingValidator && farmList.length > 0}
                     />
-                    {renderFarmUi({ farmList, contractMetadataByContractId, openModal, tokenPriceMetadata })}
+                    {isFarmingValidator && renderFarmUi({ farmList, contractMetadataByContractId, openModal, tokenPriceMetadata })}
                     <BalanceBox
                         title='staking.balanceBox.pending.title'
                         info='staking.balanceBox.pending.info'
@@ -211,7 +225,7 @@ export default function Validator({
                             open={showClaimConfirmModal}
                             onConfirm={handleClaimAction}
                             onClose={() => setShowClaimConfirmModal(false)}
-                            loading={loading}
+                            loading={claimingProceed}
                             farm={selectedFarm}
                         />}
                 </>
--- a/packages/frontend/src/components/staking/components/Validator.js
+++ b/packages/frontend/src/components/staking/components/Validator.js
@@ -4,47 +4,59 @@ import { useDispatch, useSelector } from 'react-redux';
 
 import { Mixpanel } from '../../../mixpanel';
 import { redirectTo } from '../../../redux/actions/account';
+import { claimFarmRewards, getValidatorFarmData } from '../../../redux/actions/staking';
 import { selectAccountId } from '../../../redux/slices/account';
+import { selectValidatorsFarmData, selectFarmValidatorAPY } from '../../../redux/slices/staking';
 import { selectActionsPending } from '../../../redux/slices/status';
-import { actions as tokensActions, selectAllContractMetadata, selectNEARAsTokenWithMetadata } from '../../../redux/slices/tokens';
-import { PROJECT_VALIDATOR_VERSION, ValidatorVersion } from '../../../utils/constants';
+import { selectTokensFiatValueUSD, selectTokenWhiteList } from '../../../redux/slices/tokenFiatValues';
+import { selectAllContractMetadata, selectNEARAsTokenWithMetadata } from '../../../redux/slices/tokens';
+import { FARMING_VALIDATOR_VERSION } from '../../../utils/constants';
 import FormButton from '../../common/FormButton';
 import SafeTranslate from '../../SafeTranslate';
 import AlertBanner from './AlertBanner';
 import BalanceBox from './BalanceBox';
+import ClaimConfirmModal from './ClaimConfirmModal';
+import { FarmingAPY } from './FarmingAPY';
 import StakeConfirmModal from './StakeConfirmModal';
 import StakingFee from './StakingFee';
 
-const { fetchToken } = tokensActions;
-
-const renderFarmUi = ({ farmList, contractMetadataByContractId, isFarmListLoading }) => {
-    if(isFarmListLoading) {
+const renderFarmUi = ({ farmList, contractMetadataByContractId, openModal, tokenPriceMetadata }) => {
+    if (!farmList.length) {
         // eslint-disable-next-line jsx-a11y/heading-has-content
-        return <h1 className="animated-dots"/>;
+        return <h1 className="animated-dots" />;
     }
 
-    return farmList.map(({ token_id, balance }, i) => {
+    return farmList.map((farm, i) => {
+        const { token_id, balance } = farm;
         const currentTokenContractMetadata = contractMetadataByContractId[token_id];
 
         if (!currentTokenContractMetadata) {
             return;
         }
+        const fiatValueMetadata = tokenPriceMetadata.tokenFiatValues[token_id];
+        const isWhiteListed = tokenPriceMetadata.tokenWhitelist.includes(token_id);
 
         return (
             <BalanceBox
                 key={token_id}
                 token={{
                     onChainFTMetadata: currentTokenContractMetadata,
-                    coingeckoMetadata: {},
+                    fiatValueMetadata,
                     balance,
                     contractName: token_id,
+                    isWhiteListed,
                 }}
-                // onClick={() => {
-                //     // TODO claim accrued rewards and redirect home where tokens will be fetched
-                //     return validator.contract.claim({token_id}).then(() => dispatch(redirectTo('/')));
-                // }}
+                onClick={() => {
+                    openModal({
+                        onChainFTMetadata: currentTokenContractMetadata,
+                        fiatValueMetadata,
+                        balance,
+                        contractName: token_id,
+                        isWhiteListed,
+                    });
+
+                }}
                 button="staking.balanceBox.farm.button"
-                buttonColor='gray-red'
                 hideBorder={farmList.length > 1 && i < farmList.length}
             />
         );
@@ -60,40 +72,26 @@ export default function Validator({
     currentValidators,
 }) {
     const [confirm, setConfirm] = useState(null);
-    const [farmList, setFarmList] = useState([]);
-    const [isFarmListLoading, setIsFarmListLoading] = useState(false);
+    
     const nearAsFT = useSelector(selectNEARAsTokenWithMetadata);
     const accountId = useSelector(selectAccountId);
+
     const contractMetadataByContractId = useSelector(selectAllContractMetadata);
+    const tokenFiatValues = useSelector(selectTokensFiatValueUSD);
+    const tokenWhitelist = useSelector(selectTokenWhiteList);
 
     const dispatch = useDispatch();
     const stakeNotAllowed = !!selectedValidator && selectedValidator !== match.params.validator && !!currentValidators.length;
     const showConfirmModal = confirm === 'withdraw';
-    const stakingPoolHasFarms = validator && validator.version === ValidatorVersion[PROJECT_VALIDATOR_VERSION];
     const pendingUpdateStaking = useSelector((state) => selectActionsPending(state, { types: ['UPDATE_STAKING'] }));
 
-    useEffect(() => {
-        const getFarms = async () => {
-            setIsFarmListLoading(true);
+    const [showClaimConfirmModal, setShowClaimConfirmModal] = useState(false);
+    const [selectedFarm, setSelectedFarm] = useState(null);
 
-            try {
-                const farms = await validator.contract.get_farms({ from_index: 0, limit: 300 });
-
-                const list = await Promise.all(farms.map(({ token_id }, i) => {
-                    dispatch(fetchToken({ contractName: token_id }));
-                    return validator.contract
-                        .get_unclaimed_reward({ account_id: accountId, farm_id: i })
-                        .catch(() => "0")
-                        .then((balance) => ({ token_id, balance, farm_id: i }));
-                }));
-
-                setFarmList(list);
-            } finally {
-                setIsFarmListLoading(false);
-            }
-        };
-        if (stakingPoolHasFarms) { getFarms(); }
-    }, [validator, stakingPoolHasFarms, accountId]);
+    const openModal = (farm) => {
+        setSelectedFarm(farm);
+        setShowClaimConfirmModal(true);
+    };
 
     const handleStakeAction = async () => {
         if (showConfirmModal && !loading) {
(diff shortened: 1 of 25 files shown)


packages/frontend/src/components/wallet/TokenAmount.js
```diff
@@ -28,21 +31,37 @@ const showFullAmount = (amount, decimals, symbol) =>
...
-                : <span className='dots' />
+const TokenAmount = ({ 
+    token: { balance, onChainFTMetadata, priceInUSD, isWhiteListed = true }, 
+    withSymbol = false, 
+    className, 
+    showFiatAmount = true, 
+    "data-test-id": testId 
+}) => {
+    const tokenBalance = balance && formatToken(balance, onChainFTMetadata?.decimals);
+    const tokenPrice = (tokenBalance && priceInUSD) && big(+tokenBalance * +priceInUSD).toFixed(2);
```
MaximusHaximus, line 42: 1. No need to use `big` here; these numbers aren't likely to exceed JS's native number type's boundaries.
2. Since we _always_ display these values to 2 decimal places, it would be nice to either store them formatted that way in our redux store, or format them for display when selecting the values from the `tokenFiatValues` reducer so that we don't have to sprinkle `toFixed(2)` throughout our components everywhere that we consume the values and if we later want to change the decimal accuracy we're displaying to, we don't need to find multiple places where we are formatting the value for display and update them all.

packages/frontend/src/components/staking/components/ValidatorBox.js
```diff
@@ -190,8 +199,20 @@ export default function ValidatorBox({
             {amount &&
                 <div className='right'>
                     {staking && <div><Translate id='staking.validatorBox.staking' /></div>}
+                    {farming && <div><Translate id='staking.validatorBox.farming' /></div>}
                     <div className='amount'>
-                        <Balance amount={amount} showBalanceInUSD={showBalanceInUSD} />
+                        {!token ? (
```
MaximusHaximus, line 204: Ideally, whatever code is rendering the box component would make a decision to render the correct component with the correct properties rather than make our `ValidatorBox` component have an implicit fall-through based on whether a property is undefined or not.  

For example, in this case _everywhere_ other than the new Modal, `ValidatorBox` receives a balance of NEAR tokens and displays them; _only_ the new modal needs to render the TokenAmount instead, and we know _in the context of the modal_ that we will want to render the `TokenAmount`  component instead of the `Balance` component, because we know what type of data we are working with in that modal context.

We try to avoid optional arguments like this in general so that our components have very specific roles and we avoid default/fallback behaviour based on falsy checks wherever we can.

In this case, rather than make `ValidatorBox` show either NEAR tokens _or_ fungible tokens based on a falsy check of one property, please create a separate component for `ValidatorBoxFarmReward` and rename `ValidatorBox` to e.g. `ValidatorBoxNEARReward` :)

Also, this would seem to be redundant as well; you are already rendering the `TokenAmount` directly in the modal itself, so making the `ValidatorBox` component also render it is redundant.  Maybe this box shouldn't even be rendered in the modal if you are going to render a `TokenAmount` component explicitly in the modal itself.

packages/frontend/src/components/wallet/TokenAmount.js
```diff
@@ -28,21 +31,37 @@ const showFullAmount = (amount, decimals, symbol) =>
...
+                    ? tokenBalance
+                    : <span className='dots' />
+                }
+                <span className='currency'>{withSymbol ? ` ${onChainFTMetadata?.symbol}` : null}</span>
+            </div>
+            {showFiatAmount &&
+                <div className='fiat-amount' style={{color: isWhiteListed ? '' : '#FF585D', display: 'flex', whiteSpace: 'normal'}}>
+                { tokenPrice 
+                    ? `≈ $${tokenPrice} USD`
+                    : <Translate id='tokenBox.priceUnavailable'/>
```
MaximusHaximus, line 56: The `TokenBox` component renders this component, and it already displays `tokenBox.priceUnavailable`; this would lead to us repeating the `priceUnavailable` text multiple times on our main fungible token component display list

packages/frontend/src/config/configFromEnvironment.js
```diff
@@ -72,4 +74,7 @@ module.exports = {
     WHITELISTED_CONTRACTS: parseCommaSeperatedStringAsArrayFromShell(
         process.env.TOKEN_CONTRACTS
     ),
+    NEAR_TOKEN_ID: process.env.NEAR_TOKEN_ID || 'wrap.testnet',
```
MaximusHaximus, line 77: 1. Because `wrap.testnet` only applies to the testnet environment; we shouldn't have a default like this here -- instead, you will need to add an entry for `NEAR_TOKEN_ID` to the `config/environmentDefaults/testnet` file
2. To ensure that we display the correct value on mainnet, you will also have to add the appropriate contract id to the `config/environmentDefaults/mainnet` and `config/environmentDefaults/mainnet_STAGING` files

packages/frontend/src/components/staking/components/StakingFee.js
```diff
@@ -46,4 +46,15 @@ export default function StakingFee({ fee }) {
             <span className='fee'>{fee}%</span>
         </Container>
     );
+}
+
+export function StakingAPY({ apy }) {
```
MaximusHaximus, line 51: By convention, we export one component per file, as the default export, where the filename matches the class/function of the component we are exporting.  Please extract the `Container` into its own file and create a new file titled `StakingAPY.js` rather than export both a `default` and a `named` component from this file.

MaximusHaximus (changes requested):
Thanks so much for the contribution! I've added a review here although i understand the design itself is not finalized, calling out some issues around state management, component structure and dependencies (primarily: we're using bn.js and migrating away from big.js :))

packages/frontend/src/utils/staking.js
```diff
@@ -95,3 +98,33 @@ export function shuffle(sourceArray) {
     }
     return sourceArray;
 }
+
+const SECONDS_IN_YEAR = 3600 * 24 * 365;
+
+export const calculateAPY = async (validator, tokenPrices) => {
```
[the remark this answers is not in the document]
askulikov: For now, I can't guarantee that this will always work as expected with the regular numbers. If you think so, we can omit library usage in this case. 

Also, is there a reason, why you prefer `bn.js` rather than `big.js`?

packages/frontend/src/components/staking/components/Validator.js
```diff
@@ -61,47 +74,59 @@ export default function Validator({
...
+        setShowClaimConfirmModal(true);
+    };
 
     const handleStakeAction = async () => {
         if (showConfirmModal && !loading) {
             await onWithdraw('withdraw', selectedValidator || validator.accountId);
             setConfirm('done');
         }
     };
+    const { currentValidator, isLoading: isFarmListLoading } = useSelector(state => state.staking.currentValidator);
```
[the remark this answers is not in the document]
askulikov: 1. Got it, will fix it, thank you.
2. At least on the Validator page `selectedValidator` always returns `null` for me, so I decided to create a separate part of the store for handling data and actions that refer to the farm. In this way, we can omit extra logic for some calculations and API calls for the regular validators.

packages/frontend/src/redux/actions/staking.js
```diff
@@ -533,3 +536,49 @@ export const handleUpdateCurrent = (accountId) => async (dispatch, getState) =>
     let currentAccount = selectStakingCurrentAccountbyAccountId(getState(), { accountId });
     dispatch(staking.updateCurrent({ currentAccount }));
 };
+
+export const selectCurrentValidator = (validatorId, accountId) => async (dispatch, getState) => {
```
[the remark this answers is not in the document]
askulikov: It is also refers to the "part of the store, that responsible only for farming". GET_VALIATORS is a huge call, that fetches all validators. By this new part, we can add some extra logic just only for those, who have farming available. Fetch up-to-date token fiat values and check the value available for claim with up-to-date APY. Without the need for updating all another.

packages/frontend/src/components/staking/components/Validator.js
```diff
@@ -61,47 +74,59 @@ export default function Validator({
...
+    const isFarmingValidator = validator?.version === FARMING_VALIDATOR_VERSION;
+
+    const handleClaimAction = async (token_id) => {
+        if (!currentValidator || !isFarmingValidator || !token_id) return null;
```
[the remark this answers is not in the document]
askulikov: Success and failure parts of this execution are still not written for the UI and UX. We should modify it when all use-cases will be confirmed. But yes, it is not ok )

askulikov:
@MaximusHaximus Thank you for the review! Most of the comments are reasonable and I will fix them soon. Especially, the part about mutating the redux store. 
But I want to clarify a few points. This is a small summary of my replies to your comments 
1. About using `big.js/bn.js`:
 - We can omit this converting to BN. But if there is a chance, that somehow these calculations will be unsafe - I prefer to calculate them in BN. 
 - Could you please explain, why do you prefer `BN.js` rather than `big.js`
 2. About token prices and whitelist:
  - For now, `tokenFiatValues` works just for NEAR and fetches the values from `https://helper.mainnet.near.org/fiat`. Token prices from ref.finance are used just for farming validators and only in a few components. I think, that for now, it is enough to just fetch them as a function call when we select the farming validator. In this case, we will trigger this function only when we need this and all data will be up-to-date (prices and APY). Also, we may easily change the provider if we need and even change the logic. 
  - On the other side, putting them to the redux will force refactoring through the project and we also will have to plan at least the final interface and the updating logic. 
  3. About the extra redux store for farming validators:
  - Now all validators are fetched on the bootstrap. It is a request that is huge for the application and even long for users. For marking the Farming Validators we have to provide some extra logic and the extra contract calls for calculating APY.
  - The decision is to make an extra store slice, which will work just for Farming Validators. When the user will enter the Validator page, the application will detect the type of validator and if needed - will fetch extra data (available farms, unclaimed rewards, recent prices, etc) and store it in this `currentValidator` part of the `staking` reducer. It won't affect the other regular validators and won't do any extra logic if we don't need this. And it will be also cleaned if a new validator will be selected. 
  - We might put this extra farming logic in a component, but as all application logic is stored in redux-actions - the best way will be to add a new part of the store.

packages/frontend/src/utils/staking.js
```diff
@@ -95,3 +98,33 @@ export function shuffle(sourceArray) {
...
+            const rewardsPerSecondInUSD = rewardsPerSecond.mul(tokenPriceInUSD);
+            const totalStakedBalanceInUSD = totalStakedBalance.mul(nearPriceInUSD);
+            const farmAPY = rewardsPerSecondInUSD.mul(SECONDS_IN_YEAR).div(totalStakedBalanceInUSD).mul(100);
+            return acc.add(farmAPY);
+        }, big(0));
+
+        return summaryAPY.round(3).toFixed(2);
+    }
+    catch (e) {
+        console.error('Error during calculating APY', e);
```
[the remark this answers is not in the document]
askulikov: Will fix add the `0`

packages/frontend/src/utils/ref-finance.js
```diff
@@ -0,0 +1,26 @@
...
+export const fetchTokenPrices = async () => {
+  try {
+    return fetch(REF_FINANCE_API_ENDPOINT + '/list-token-price').then(r => r.json());
+    
+  } catch (error) {
+    console.error(`Failed to fetch token prices: ${error}`);
+  }
+};
+
+export const fetchTokenWhiteList = async (accountId) => {
```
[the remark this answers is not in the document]
askulikov: Will add the empty array.

packages/frontend/src/components/wallet/TokenAmount.js
askulikov: Storing the formatted values in the store might do some issues in future calculations. If you want, I will also change this. But I prefer to work with original data and format it when we need this. All values in this case will be a little bit pure and close to the truth. Anyway, we have to decide how to modify `tokenFiatValues`.

MaximusHaximus (commented):
> @MaximusHaximus Thank you for the review! Most of the comments are reasonable and I will fix them soon. Especially, the part about mutating the redux store. But I want to clarify a few points. This is a small summary of my replies to your comments
>
> 1. About using `big.js/bn.js`:
> * We can omit this converting to BN. But if there is a chance, that somehow these calculations will be unsafe - I prefer to calculate them in BN.
> * Could you please explain, why do you prefer `BN.js` rather than `big.js`

Added explanation about this in-line to your original comment -- if you've identified that there are cases where the numbers involved will exceed 9 quadrillion, let's keep using the big number implementation there :). We just want to standardize on ONE big number implementation, which is `bn.js` for consistency reasons :)

> 2. About token prices and whitelist:
> * For now, `tokenFiatValues` works just for NEAR and fetches the values from `https://helper.mainnet.near.org/fiat`. Token prices from ref.finance are used just for farming validators and only in a few components. I think, that for now, it is enough to just fetch them as a function call when we select the farming validator. In this case, we will trigger this function only when we need this and all data will be up-to-date (prices and APY). Also, we may easily change the provider if we need and even change the logic.
> * On the other side, putting them to the redux will force refactoring through the project and we also will have to plan at least the final interface and the updating logic.

The `tokenFiatValues` reducer is currently only storing a single entry in its hashmap, for `NEAR` tokens, but it is intended to store _all_ fiat values for _all FTs_ in the app -- this is why the near price is stored under `tokenFiatValues['near']` instead of being stored directly as the only value in the slice. When we implemented the slice initially (for NEAR tokens only), we did not yet have a place to look up prices for other FTs; ref.finance API was not yet available.

Regarding refactoring across the codebase: code changes for this are probably able to be encapsulated inside of the `fetchTokenFiatValues` async thunk almost entirely -- any code inside of that thunk must return an object with the same shape as the current `/fiat` contract helper returns. This ensures that consumers of the slice do not need to be updated.  It would be nice if we renamed all occurrences of `coingeckoMetadata` to `fiatValueMetadata` for clarity in components that use the too-specific name, but that should be a simple global string find/replace.

My suggestion at a high level is to add a parameter to the `fetchTokenFiatValues` thunk that will allow the caller to specify an optional array of FT contract IDs to fetch fiat balances for. 

If no argument is provided, then our existing code that uses the `/fiat` endpoint of contract helper and returns the value as it was provided will still run, so you won't need to refactor any existing code that relies on the existing state slice or the existing thunk behaviour.

However, if an array of token ids _is_ provided to the `fetchTokenFiatValues` thunk, then you can use the new ref-finance helper method(s) to fetch their fiat values from ref-finance, transform the returned payload into the same shape as the `/fiat` endpoint returns, and then return those from the thunk so that they are merged into the `tokenFiatValues` state slice just like the NEAR token price is. Now any components that need to show this value will select it from the same place as we select any other tokens fiat values.

> 3. About the extra redux store for farming validators:
> * Now all validators are fetched on the bootstrap. It is a request that is huge for the application and even long for users. For marking the Farming Validators we have to provide some extra logic and the extra contract calls for calculating APY.
> * The decision is to make an extra store slice, which will work just for Farming Validators. When the user will enter the Validator page, the application will detect the type of validator and if needed - will fetch extra data (available farms, unclaimed rewards, recent prices, etc) and store it in this `currentValidator` part of the `staking` reducer. It won't affect the other regular validators and won't do any extra logic if we don't need this. And it will be also cleaned if a new validator will be selected.
> * We might put this extra farming logic in a component, but as all application logic is stored in redux-actions - the best way will be to add a new part of the store.

Added a detailed response about this as a response to your in-line comment... in shorter form, we must not duplicate chunks of our redux state if at all possible -- it's fine (preferable, even!) to create a separate slice of state to track farming specific validator data if its lifecycle is different than that of the main `validators` data, but if we are going to do that, I do have 3 requests:
(1) rather than limit the store to only one validator's data at a time, make the slice an object that is keyed on the validator account Id.  This allows users who are switching back and forth between validators to avoid their UI blocking while it re-fetches data that they just loaded a moment before.  It's fine to still dispatch a refresh of that data, but we want to avoid blocking the UI while we are refreshing it if the store already contains values that we can render the UI on (we would show an indicator next to any data that is being refreshed in the UI while the refresh is pending).
(2) Rather than only adding _some_ of the new farm specific metadata as a sub-key under the existing validator redux state (poolSummary) but _other_ farm specific data under the new slice (farmRewards), please add **all** of the farm specific data to the new slice, and leave our existing validator data slice as-is, with only non-farming-pool related data there.  (3) Finally, the farm data reducer slice should should track _only_ the farm-specific data -- do not create a reference or a copy of the validator object from the main `allValidators` slice but with some farm-specific data added to it :)

**Some thoughts about the performance and lifecycle of fetching the new farm-specific data:**
The newly added code in `GET_VALIDATORS` means we that are _already_ putting some farm specific data into the main `validators` state slice, and the 'Validator' component already relies on `validators` already being loaded, since it receives its `validator` prop as a result of a selector that selects from the `allValidators` redux slice.  This means that displaying _any_ validators is slower than before, because now our `GET_VALIDATORS` thunk is also responsible for querying _every validator contract to see if it can support `get_pool_summary`.  The thunk now fetches that farm data for _all_ validators before returning _any_ data about _any_ validators.  Ideally, the `allValidators` slice would be populated as soon as possible and any non-essential data would be asynchronously loaded without blocking the UI from rendering; in the case of any validator that is a staking farm, we have added an extra sequential call to the fetching logic here.  I noticed that the `get_pool_summary()` method appears to return the current fee fraction, which we are currently doing as a separate call on L399.  It seems that if the 'version' of a validator is a staking pool, we could call `get_pool_summary` for that pool on L399 _instead_ of `get_reward_fee_fraction`, thus removing the second sequential RPC call that the current implementation has added for each farming pool type validator.  I'm basing my understanding on this code: https://github.com/referencedev/staking-farm/blob/07c60f04d69847723ab0629acdc9379a264d21fb/staking-farm/src/views.rs#L81

Currently,  this change means:

(1) Because `Validator.js` will not render a usable screen _before_ the loading of the full validators list finishes, and your code that fetches farm-specific data requires the `validator` prop be populated before it fetches farming pool data, but the `validator` prop won't be populated until the full list of validators is loaded, the code to re-fetch the pool summary and get further farm details will not begin to execute until the full list of validators finishes loading, _including the pool summaries for every farming type of validator._

(2) Furthermore, because the `GET_VALIDATORS` thunk now _also_ fetches farm-specific data, if someone was to directly open a validator route (e.g. from a bookmark), the new code that computes `currentValidator` will be unnecessarily re-fetching the exact same `poolSummary` data that was _just_ populated when `GET_VALIDATORS` ran a moment earlier...so the user will see an unnecessary loading state while the new `selectCurrentValidator` thunk is running and re-fetching the same pool_summary data it already fetched moments before. We handle this by dispatching actions to load this type of data without `await`ing them, and using the `condition` property of `createAsyncThunk` to ensure that if two parts of the UI dispatch the same thunk in parallel, only the first one is dispatched and the rest of them abort ([reference](https://github.com/near/near-wallet/blob/ced351e456f5d0127e953c0963af07d9fc7b6508/packages/frontend/src/redux/slices/tokens/index.js#L36))

askulikov:
@MaximusHaximus Thank you for the so good and detailed review with annotations. I have finally understood the full high-level overview of app architecture and how you are going to build it further. I will come back to you with fixes.

askulikov:
@MaximusHaximus , could you please review my changes? 
Few points to discuss:
* I put token prices and token whitelist in the store. The whitelist uploaded on the bootstrap (as we suppose that this data changes rarely). But token prices uploaded with the fiat request. We talk about we are going to request them by demand. But as we anyway will request ALL tokens - there is no sense to create something more complex. At least in the context of this task, but during the future refactoring. 
* as the bonus, now we will have token prices for all available tokens from ref :)

* also I decided to not change the modal and ValidatorBox for now. I thought that maybe we will receive some design or maybe user flow for this task. But as it is unavailable - will be much better to do small refactoring with the clear task rather than trying to overwrite existing code later )

* Your suggestion about handling claim actions in some StakingService is very good. But as now unstacking functionality is handled by redux actions - I decided to create just new action for keeping code consistent.

askulikov:
Also, I will probably create a new PR, or update the current with rebase after we fix all issues. Had some problems with merging, so going to spend a little bit more time on it with fresh head

packages/frontend/src/components/staking/components/Validator.js
```diff
@@ -5,47 +5,59 @@ import { useDispatch, useSelector } from 'react-redux';
...
-    return farmList.map(({ token_id, balance }, i) => {
+    return farmList.map((farm, i) => {
+        const { token_id, balance } = farm;
         const currentTokenContractMetadata = contractMetadataByContractId[token_id];
 
         if (!currentTokenContractMetadata) {
             return;
         }
+        const fiatValueMetadata = tokenPriceMetadata.tokenFiatValues[token_id];
+        const isWhiteListed = tokenPriceMetadata.tokenWhitelist.includes(token_id);
```
Solniechniy, line 38: Is storing tokenWhitelist as a map/set instead of array suitable for this case? Getting data from a set/map will be faster than searching.

packages/frontend/src/components/staking/components/Validator.js
```diff
@@ -61,47 +73,54 @@ export default function Validator({
...
+        setShowClaimConfirmModal(true);
+    };
 
     const handleStakeAction = async () => {
         if (showConfirmModal && !loading) {
             await onWithdraw('withdraw', selectedValidator || validator.accountId);
             setConfirm('done');
         }
     };
+    const isFarmingValidator = validator?.version === FARMING_VALIDATOR_VERSION;
```
Solniechniy, line 101: Do we suppose that validator?.version is not optional param?

packages/frontend/src/components/staking/components/ValidatorBox.js
```diff
@@ -128,13 +130,16 @@ export default function ValidatorBox({
...
+    farming = false,
     clickable = true,
     style,
     label = false,
     stakeAction,
     showBalanceInUSD,
+    token = null
 }) {
     const dispatch = useDispatch();
+    const farmAPY = useSelector(state => selectFarmValidatorAPY(state, {validatorId: validator?.accountId}));
```
Solniechniy, line 142: could we move this to selectors file?

packages/frontend/src/components/staking/components/ValidatorBox.js
```diff
@@ -172,11 +177,17 @@ export default function ValidatorBox({
...
-                    {isProjectValidator && <Tooltip translate='staking.balanceBox.farm.info' />}
+                    {isFarmingValidator && <Tooltip translate='staking.balanceBox.farm.info' />}
                 </div>
                 {typeof fee === 'number' &&
                     <div className="text-left">
-                        <span>{fee}% <Translate id='staking.validatorBox.fee' /> - </span>
+                        { 
+                            isFarmingValidator && <>
+                                <span>APY&nbsp;</span>
+                                <span>{farmAPY}%&nbsp;-&nbsp;</span>
```
Solniechniy, line 187: We can use locales for this purposes with param

packages/frontend/src/redux/actions/staking.js
```diff
@@ -533,3 +525,61 @@ export const handleUpdateCurrent = (accountId) => async (dispatch, getState) =>
...
+            return validator.contract
+                .get_unclaimed_reward({ account_id: accountId, farm_id })
+                .catch(() => "0")
+                .then((balance) => ({ 
+                    token_id,
+                    balance,
+                    farm_id,
+                }));
+        } catch (error) {
+            console.error(error);
```
Solniechniy, line 552: Handling errors, in this case, should be done with modal or smth more suitable.
If any request failed - the promise will be rejected. @MaximusHaximus provides information on how it should be done properly.

packages/frontend/src/redux/actions/staking.js
```diff
@@ -533,3 +525,61 @@ export const handleUpdateCurrent = (accountId) => async (dispatch, getState) =>
...
+    }));
+
+    const farmData = {
+        poolSummary: {...poolSummary},
+        farmRewards: list,
+    };
+    await dispatch(staking.setValidatorFarmData(validator.accountId, farmData));
+};
+
+export const claimFarmRewards = (validatorId, accountId, token_id) => async (dispatch, getState) => {
```
Solniechniy, line 569: Handling errors with claiming also should be done.

packages/frontend/src/utils/ref-finance.js
```diff
@@ -0,0 +1,29 @@
+import { Contract } from "near-api-js";
+
+import sendJson from '../tmp_fetch_send_json';
+import { wallet } from "./wallet";
+
+const REF_FINANCE_API_ENDPOINT = 'https://indexer.ref-finance.net';
```
Solniechniy, line 6: Possibly we should add also endpoints for testnet indexer by checking isMainnet

askulikov:
Just updated the branch. There were some problems during rebase, so I decided to create it from scratch. Just to be sure, that I haven't missed any fixes.

packages/frontend/src/components/staking/components/Validator.js
askulikov: I'm not sure that this has any sense. As we receive just an array of token ids (`["wrap.near", "aurora", ...]`).

packages/frontend/src/components/staking/components/Validator.js
askulikov: There are a few cases when during the bootstrap of the component `validator` prop is still `undefined`. This way we can omit some errors and extra checks.

askulikov:
In the last commit were fixed some minor bugs with prices and also added error handling for claiming action. In success user will be redirected to the Success page, otherwise - the popup with an error message will be shown.

MaximusHaximus:
@stefanopepe I got a staging preview environment up for this PR for you to check out:
https://deploy-preview-2398--near-wallet-staging.netlify.app/

MaximusHaximus:
Closing as this was resolved in #2384 and claim is added in #2507

PR #2409, opened 2022-01-25
fix: replace MoonPay pre-specified amount buttons w/ generic 'Buy with MoonPay'

Patrick1904:
Screen Shot 2022-01-25 at 2 19 30 PM


MaximusHaximus (approved):
💖

PR #2412, opened 2022-01-26
feat: Jenkins integration

andy-haynes:
This PR adds a Jenkinsfile configured to deploy to the new AWS environments. The Jenkins deployment has a few key features:
- conditional building of packages (currently hardcoded for just `frontend` but can be quickly adapted for `nx`)
- parallel bundling of packages and frontend bundles, any of which failing causes the entire pipeline to fail
- deployment to S3 buckets for both merges to `master`/`stable` and PR previews

The current implementation only supports deploying to `testnet` but the `mainnet` deployment logic can be derived from the `testnet` logic once the environment is provisioned.

This PR also adds two new environments, `testnet_AWS` and `testnet_AWS_STAGING`, which will be used as a test environment before we start deploying to both the current infrastructure and the new AWS platform.


Jenkinsfile
```diff
@@ -0,0 +1,182 @@
...
+                        expression { env.BUILD_FRONTEND == 'true' }
+                    }
+                    environment {
+                        NEAR_WALLET_ENV = 'testnet_AWS'
+                        TOKEN_CONTRACTS = 'meta.pool.testnet'
+                    }
+                    steps {
+                        dir("$WORKSPACE/packages/frontend") {
+                            sh "rm -rf $FRONTEND_TESTNET_BUNDLE_PATH"
+                            sh "yarn bundle --outDir=$FRONTEND_TESTNET_BUNDLE_PATH"
```
andy-haynes, line 85: I've modified the `runBundler` code to accept arguments that can override the default values passed to the constructor. This allows us to create bundles for each environment in parallel by specifying different destination paths.

MaximusHaximus (approved):
:+1: Let's Get This Merged :D

PR #2417, opened 2022-01-27
New translations into Russian

denis-skripnik:
Translated from English into Russian what was not translated earlier.

Perhaps somewhere the context misunderstood, which is why it is incorrectly written. Please check


Patrick1904 (approved):
Thank you @denis-skripnik! 🙏

artem-p:
My 2 cents:

- "Almost there" is rather "Почти готово" than "Почти на месте"
- also typo in "логиноном"

artem-p:
Also, if I'm not mistake, "Застейканно" should be with one "н" letter, "Застейкано".

denis-skripnik:
Thank you. Fixed the mistakes.

Issue #2371, opened 2022-01-13
Tooling: Bootstrap Component Library


gutsyphilip:
## Problem
All frontend components live within the frontend package and are currently not re-usable outside of it.

#### What are we planning to do about it?
We will be creating a standalone component library that will be deployed to NPM and can be installed and utilised as an NPM package.

- [ ] Setup component library using TS(composite TS setup)
- [ ] Publish initial package to NPM

#### What are we _not_ planning to do about it?
- We will not be migrating existing components just yet. That will come after.

PR #2420, opened 2022-01-28
Feature: UI component library

gutsyphilip:
#2371 

This PR Bootstraps our UI component library with the following:

- Typescript
- Mircrobundle for bundling
- Storybook for documentation
- Jest and React testing library for testing


MaximusHaximus (changes requested):
:+1: nice work!  Just one thing, in that we have some translation changes that ended up in this PR that shouldn't be here. Once those are removed, we can get this landed!

PR #2421, opened 2022-01-28
fix: loading on receiver screen when use max is used

esaminu:
Fix loading state on receiver screen when use max is used:

- Make dependency for clearing user input more specific (`contractName` || `symbol`)
- Pass `token.balance` to `EnterReceiver` component when `isMaxAmount` is `true`
- Replace `useNEARAsTokenWithMetadata` with memoized selector `selectNEARAsTokenWithMetadata` to avoid new identity unless the state changed

### DIFF
--- a/packages/frontend/src/components/send/SendContainerV2.js
+++ b/packages/frontend/src/components/send/SendContainerV2.js
@@ -14,7 +14,7 @@ import Review from './components/views/Review';
 import SelectToken from './components/views/SelectToken';
 import Success from './components/views/Success';
 
-const { getFormattedTokenAmount, getParsedTokenAmount } = FungibleTokens;
+const { getFormattedTokenAmount, getParsedTokenAmount, getUniqueTokenIdentity } = FungibleTokens;
 
 export const VIEWS = {
     ENTER_AMOUNT: 'enterAmount',
@@ -118,7 +118,7 @@ const SendContainerV2 = ({
             setIsMaxAmount(false);
             setUserInputAmount('');
         }
-    }, [selectedToken]);
+    }, [getUniqueTokenIdentity(selectedToken)]);
 
     useEffect(() => window.scrollTo(0, 0), [activeView]);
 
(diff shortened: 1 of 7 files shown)


Patrick1904 (approved):
LGTM if we're confident about the `getRawAmount()` change:)

esaminu:
Adding my comment from slack with some context for reference:

> hmm. passing in the `getRawAmount()` directly instead of `selectedToken.balance` when `isMaxAmount` is `true` [here](https://github.com/near/near-wallet/pull/2421/files#diff-0a4e573ba5cd135e5104fc13e05bc89da8b7aa6996df8355a344f5cb489ad30cR197)  would still work because the user input is not emptied anymore so it still exists and can be parsed and displayed. The effect would be a loss of precision on the EnterReceiver screen header because we are parsing and formatting the balance a second time on that screen. The behavior in the PR is for the same parsed (full) amount to be passed to each component and they handle their formatting independently. currently, the input on the EnterAmount screen is using [getFormattedTokenAmount](https://github.com/near/near-wallet/blob/master/packages/frontend/src/services/FungibleTokens.js#L56) and the EnterReceiver screen is using [formatNearAmount](https://github.com/near/near-wallet/blob/master/packages/frontend/src/components/common/balance/helpers.js#L7). Both currently use a fracDigits value of 5

@Patrick1904 @MaximusHaximus would appreciate a follow up on this so we can get this merged

esaminu:
Thanks! Adding my comment from slack with some context for reference

>  hmm. Passing in the `getRawAmount()` instead of `selectedToken.balance` when `isMaxAmount` is `true` [here](https://github.com/near/near-wallet/pull/2421/files#diff-0a4e573ba5cd135e5104fc13e05bc89da8b7aa6996df8355a344f5cb489ad30cR197) would still work because the user input is not emptied anymore so it still exists and can be parsed and displayed. The effect would be a loss of precision on the EnterReceiver screen header because we are parsing and formatting the balance a second time on that screen. The behavior in the PR is for the same parsed (full) amount to be passed to each component and they handle their formatting independently. currently, the input on the EnterAmount screen is using [getFormattedTokenAmount](https://github.com/near/near-wallet/blob/master/packages/frontend/src/services/FungibleTokens.js#L56) and the EnterReceiver screen is using [formatNearAmount](https://github.com/near/near-wallet/blob/master/packages/frontend/src/components/common/balance/helpers.js#L7). Both currently use a fracDigits value of 5

MaximusHaximus (approved):
:+1: looks good, nice work on digging through this stuff

PR #2427, opened 2022-01-31
Login sign url docs

### LINKED ISSUE #2441: Add enforcement against usage of `URL.searchParams` or `URLSearchParams` and non selectors in `useSelector`
## Problem
1. Using `URL.searchParams` or `URLSearchParams` native objects could result in incorrect parsing and setting of query params in cases where our interface is an append type interface e.g.: `?methodNames=method1&methodNames=method2`. This is because in these cases, `URLSearchParams` instances return only the value of the first arg when you call `.get`

2. We want to enforce usage of selectors inside `useSelector`

#### What are we planning to do about it?
- Add eslint enforcement in favor of [query-string](https://github.com/sindresorhus/query-string) as opposed to  `URL.searchParams` or `URLSearchParams`.
- Add eslint enforcement to stop destructure of first argument within `useSelector` in function signature.
- Add eslint enforcement to make it illegal for the function expression passed into `useSelector` to reference properties of the first argument passed into the function (which is the top-level state).

## Resources
* [ESLint selectors doc](https://eslint.org/docs/developer-guide/selectors)
* [AST explorer](https://astexplorer.net/)
* [esquery](https://github.com/estools/esquery)
* [esquery tester](http://estools.github.io/esquery/)

esaminu:
- Add additional docs for login and sign to address the specific url params and wallet handling of them
- Add eslint enforcement against usage of `URLSearchParams`
- Add eslint enforcement against usage of `URL.searchParams`
- Add eslint enforcement against property access of state variable inside `useSelector`
- Add eslint enforcement against first argument destructure inside `useSelector`

closes #2441


MaximusHaximus (approved):
:+1: nice work :).

We should have an issue linked to this PR that describes _why_ we enforce using `query-string` and the possible API breaking change it would incur if we used `URLSearchParams' object in the login flow, just in case someone is doing some investigation later and needs to know why we use `query-string` vs. the native object.

PR #2433, opened 2022-02-02
fix: Save keypair only for recovery msg key

Patrick1904:
**Problem**
While creating a new named account with email recovery link, the user verifies the security code and upon confirmation is sent to the "Fund with an existing account." screen (shown below). Because we call `saveAccount` (key necessary for some contract-helper calls) as part of this process, the wallet adds the `not_yet_created.near` account to the list of accounts before the account has been created.

**Fix**
Call `saveAccountKeyPair` that simply adds the `not_yet_created.near` keypair to `localStorage` but not to the list of active accounts

### DIFF
--- a/packages/frontend/src/components/accounts/recovery_setup/SetupRecoveryMethod.js
+++ b/packages/frontend/src/components/accounts/recovery_setup/SetupRecoveryMethod.js
@@ -227,7 +227,7 @@ class SetupRecoveryMethod extends Component {
             const { secretKey } = parseSeedPhrase(recoverySeedPhrase);
             const recoveryKeyPair = KeyPair.fromString(secretKey);
             await validateSecurityCode(accountId, method, securityCode, validateSecurityCodeEnterpriseRecaptchaToken, 'setupRecoveryMethodNewAccount');
-            await saveAccount(accountId, recoveryKeyPair);
+            await wallet.saveAccountKeyPair({ implicitAccountId: accountId, recoveryKeyPair });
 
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/accounts/recovery_setup/SetupRecoveryMethod.js
```diff
@@ -227,7 +227,7 @@ class SetupRecoveryMethod extends Component {
             const { secretKey } = parseSeedPhrase(recoverySeedPhrase);
             const recoveryKeyPair = KeyPair.fromString(secretKey);
             await validateSecurityCode(accountId, method, securityCode, validateSecurityCodeEnterpriseRecaptchaToken, 'setupRecoveryMethodNewAccount');
-            await saveAccount(accountId, recoveryKeyPair);
+            await wallet.saveAccountKeyPair({ implicitAccountId: accountId, recoveryKeyPair });
```
marcinbodnar, line 230: Seems like it should be:
`await wallet.saveAccountKeyPair({ accountId, recoveryKeyPair });`

marcinbodnar (changes requested):

packages/frontend/src/components/accounts/recovery_setup/SetupRecoveryMethod.js
Patrick1904: Great catch! Fixed

marcinbodnar (approved):
LGTM

Issue #2348, opened 2022-01-06
Use current account as funding method by default when creating a new account & re-use recovery key


Patrick1904:
Original issue: https://github.com/near/near-wallet/issues/2284

We implemented phase one of the implicit account onboarding in https://github.com/near/near-wallet/pull/2313. Phase two includes:
1. Using current account as funding method by default when creating a new account (while logged in)
2. Enabling users to re-use the same seed phrase / key when creating a new account, essentially making it possible for users to add the same key to multiple accounts.

Figma design: https://www.figma.com/file/yR4qz7oAjAiVAJ0hJSnQQj/Onboarding-Exploration?node-id=1%3A606

Patrick1904:
@MaximusHaximus and @corwinharrell - the UI for this feature is done, but I know @MaximusHaximus has some concerns with shipping this 'shared seed phrases' feature, at least without any ability for users to manage multiple accounts with a single phrase outside of the initial setup flow. @corwinharrell any thoughts on this?

corwinharrell:
I think it's a valid approach to hold off on the shared phrase stuff until we can also design and plan for better key management, so that a user has the following options:

1. See which accounts are secured by which keys
2. See which accounts share keys
3. Easily rotate keys, change keys, etc...
4. Have messaging around actions that ensure the user understands the implications of changes to keys

PR #2434, opened 2022-02-03
Add UI for shared passphrase when creating a new account

Patrick1904:
This PR adds the necessary UI for enabling 'shared passphrase' while creating a new account. This feature will be finalized and launched once we add necessary UX to manage shared passphrases outside of the account creation flow, as outlined here: https://github.com/near/near-wallet/issues/2348#issuecomment-1034110613.

Screen Shot 2022-02-10 at 4 39 50 PM

Figma design: https://www.figma.com/file/yR4qz7oAjAiVAJ0hJSnQQj/Onboarding-Exploration?node-id=1%3A606

### DIFF
--- a/packages/frontend/src/components/accounts/recovery_setup/RecoveryOption.js
+++ b/packages/frontend/src/components/accounts/recovery_setup/RecoveryOption.js
@@ -193,7 +193,7 @@ const Icon = ({option, color}) => {
         case 'ledger':
             return <HardwareWalletIcon color={color}/>;
         default:
-            return;
+            return '';
     }
 };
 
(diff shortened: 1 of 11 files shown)


packages/frontend/src/components/accounts/recovery_setup/RecoveryOption.js
```diff
@@ -193,7 +193,7 @@ const Icon = ({option, color}) => {
         case 'ledger':
             return <HardwareWalletIcon color={color}/>;
         default:
-            return;
+            return '';
```
MaximusHaximus, line 196: Just curious about why we had to change this to return a string, and are we sure that this doesn't have any unintended impact on existing UI flows?

Patrick1904: Yep it does not have any impact on existing UI and was only necessary to make `<RecoveryOption/>` support a version without a provided icon:

Screen Shot 2022-02-10 at 4 58 58 PM

PR #2438, opened 2022-02-04
fix: misplaced usage of MIN_BALANCE_FOR_GAS

### LINKED ISSUE #2436: Lockup 'available to transfer' banner persists after transferring from lockup
**Problem**
1. Lockup 'available to transfer' banner persists after transferring from lockup
2. Lockup storage amount is included in the 'available to transfer' amount displayed on the banner

Screen Shot 2022-02-01 at 8 49 55 AM

**Expected Behavior**
1. Lockup banner should disappear after transferring from lockup
2. Lockup storage amount should only be included in the 'available to transfer' amount if the lockup is deletable and there are no funds in staking pools, etc. 

**Steps to reproduce**
1. Using an account with lockup - try transferring from lockup by clicking 'Transfer to wallet'
2. Observe that the lockup still persists after transferring, and that the amount left is the lockup storage amount

esaminu:
Fix liquid amount calculation to use `MIN_BALANCE_FOR_STORAGE` instead of `MIN_BALANCE_FOR_GAS`

closes #2436

### DIFF
--- a/packages/frontend/src/utils/account-with-lockup.js
+++ b/packages/frontend/src/utils/account-with-lockup.js
@@ -269,9 +269,10 @@ async function getAccountBalance(limitedAccountData = false) {
 
         // if acc is deletable (nothing locked && nothing stake) you can transfer the whole amount ohterwise get_liquid_owners_balance
         const isAccDeletable = lockedAmount.isZero() && stakedBalanceLockup.isZero();
+        const MIN_BALANCE_FOR_STORAGE = getLockupMinBalanceForStorage(lockupContractCodeHash);
         const liquidOwnersBalanceTransfersEnabled = isAccDeletable
             ? new BN(lockupBalance.total)
-            : BN.min(ownersBalance, new BN(lockupBalance.total).sub(new BN(MIN_BALANCE_FOR_GAS)));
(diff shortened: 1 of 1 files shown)


Patrick1904:
@esaminu Not a blocker for this PR, but I noticed while testing this fix that we throw an error when the lockup locked amount is zero and there's some tokens still in a staking pool. We still transfer the amount 'available to transfer' - in this case 0.10039 NEAR. I don't think we should show any error on the UI for this case since the transfer was successful and I expected  0.10039 NEAR to be transferred. Nothing actually went wrong here.

If we want to encourage the user to withdraw remaining funds from the lockup, then I think that should be communicated somewhere closer to the lockup section of the page with a message saying something like 'All funds in your lockup account are unlocked and can be transferred to your main account once you withdraw all funds from staking pools'. 

@corwinharrell any thoughts on this UX?

Screen Shot 2022-02-04 at 12 37 59 PM

esaminu:
@Patrick1904 I agree but I think we should get this fix merged before the next release since it affects all partially vested lockups' "available to transfer" display. We should create an issue for the UX improvement since this behavior is currently expected.

Patrick1904 (approved):

MaximusHaximus (approved):
:+1: Legit, let's get this merged so we can ship it on Monday's release

PR #2440, opened 2022-02-04
chore: Forces Yarn to be used on a project

marcinbodnar:


MaximusHaximus (approved):
Yay for no more ghost `package-lock.json` files :) Thanks for fixing a broken window!

Issue #2431, opened 2022-02-01
Web redirects from apps does not set up the correct wallet and has no ability to switch the wallet


alexauroradev:
**Problem**
Web redirects from apps does not set up the correct wallet and has no ability to switch the wallet

**Expected Behavior**
The redirect should set up the correct wallet AND user should be able to switch the signing wallet.

**Steps to reproduce**
1. The user has multiple wallets, the last wallet used is _wallet_1_
2. He logs in the app with _wallet_1_ (for example, https://ecosystem.pixeldapps.co/cryptoheroes)
3. He switches in the wallet interface to the _wallet_2_
4. During the interaction with the app and when it is requested to sign a transaction, the user is redirected to the wallet
5. The wallet displays as an active wallet _wallet_2_ instead of _wallet_1_ with inability to switch the wallet

Here's the screenshot:
<img width="1098" alt="Screenshot 2022-02-01 at 15 49 47" src="https://user-images.githubusercontent.com/5417224/151980670-d5f2f31e-e524-43f4-969d-238b271f09c4.png">

The URL is the following: https://wallet.near.org/sign?transactions=CAAAAGFzMi5uZWFyAI1rFJM6rTTaPFh4BKcB%2Bw7DGa6KT4fiFa3IODd%2BaiJYVml%2B6g01AAAPAAAAcGl4ZWx0b2tlbi5uZWFySiTUi1PV9b7fCAsDShAAi8hurggsGIJTgrP0QDO40sgBAAAAAgsAAABjaF9idXlfaXRlbVAAAAB7InRva2VuX2lkIjoiMTIzNjk4Iiwib3duZXIiOiJyb3pib3IubmVhciIsImJ1eWVyIjoiYXMyLm5lYXIiLCJwcmljZSI6IjEwMDAwMDAifQBgt5hsiAAAAQAAAAAAAAAAAAAAAAAAAA%3D%3D&referrer=Pixelpets+Dapp%3A+Pixelpets+Dapp&callbackUrl=https%3A%2F%2Fecosystem.pixeldapps.co%2Fcallback%3Fpage%3Ditem_bought

paouvrard:
+1
Some dapps like rainbowbridge expect the signer to be the connected account (wallet_1)

Patrick1904:
@esaminu is working on this here: https://github.com/near/near-wallet/issues/2402 :)

Issue #2402, opened 2022-01-24
Select transaction `signerId` on wallet sign interface


esaminu:
## Story

As a dapp user, I want the correct signer account to be selected on the wallet sign interface so I can sign a transaction successfully.

Relates to: #2396 , #2399 

## Acceptance Criteria

- [ ] Given I navigate to `/sign` with a transaction(s) the account on the transaction's `signerId` should be selected
- [ ] Given there is more than one `signerId` in a transaction batch I should see an error
- [ ] Given the account on the transaction's `signerId` does not exist on the wallet I should see an error

## Resources

- [Figma Document](https://www.figma.com/file/9ImHeER6uE2i2oof41QCu6/signerId-Errors?node-id=2%3A2)

<img width="971" alt="Screen Shot 2022-01-26 at 9 06 17 AM" src="https://user-images.githubusercontent.com/5934188/151179186-7a0d9c51-a52b-46c6-a11c-c078bc38d7d3.png">

corwinharrell:
Considering error states. My assumption is that to even _initiate_ a transaction, the user would need to be connected to the Ðapp via the wallet. Under this assumption, what is the event that a transaction's `signerId` might not exist in the wallet? Similarly, how might `signerId`s be mismatched?

Similarly,

esaminu:
Requesting a transaction signature is an independent operation to signing in (requesting an access key). A Ðapp could send us a transaction to sign that has a `signerId` that is not logged in to the wallet. That would be the first error state in the story.

The second error state is quite edge case-y. It's when the list of transactions sent via the `transactions` url param does not have the same `signerId` for all transactions. After some thought and discussion I think we can remove explicit detection and handling for this case and simply display the error returned by rpc on any transaction that was incorrectly signed. I.e. we will always handle the batch of transactions using the `signerId` of the first transaction on the list.

MaximusHaximus:
@corwinharrell

To clarify the /sign endpoint and its (lack of) relationship with the /login endpoint and the LAK that is created after a successful login transaction:

Signing in to a dapp is **not** a requirement for using of the wallet `/sign` API.

In fact, due to its very nature, because the /sign endpoint exists _specifically_ to support signing of payloads that use a FAK (thus has no `contractId` target), we can't even know if someone _is_ 'signed in' to a dapp that is relevant to the incoming transactions that are in the /sign request.

[...]

PR #2401, opened 2022-01-24
feat: add account selection functionality to /sign

esaminu:
Adds account selection functionality to `/sign` wallet integration point. Closes #2399, #2396, #2431, #2402:

Screen Shot 2022-01-25 at 12 51 29 AMScreen Shot 2022-01-25 at 12 52 04 AM

PR #2447, opened 2022-02-07
Sign tx autoselect and errors

### LINKED ISSUE #2396: Cannot select account on transaction confirmation screen
**Problem**
No matter which account is active on an app, wallet will always use the current wallet active account. Despite the account that made the transaction on the app is also logged in on wallet.near.org. 

**Expected Behavior**
As I remember, wallet uses the account that signed the transaction from the app, rather than the currently selected account on the wallet.

**Steps to reproduce**

Accounts: a.near, b.near
1. Login on any app (paras.id, app.ref.finance) as a.near
2. Ensure that a.near and b.near logged in on wallet.near.org
3. Select active account on wallet as b.near
4. Make transaction on the app with a.near

esaminu:
Adds the following functionality based on discussion on #2402 
- [x]  Switch automatically to the user on the first transaction’s `signerId` on mount of the `/sign` component
- [ ]  Build new error screens for invalid transaction batch and Account not found
- [ ]  Add checks and display error screens

Closes #2396, #2431 and #2402  

Replaces #2401

### DIFF
--- a/packages/frontend/src/components/sign/SignTransferAccountNotFound.js
+++ b/packages/frontend/src/components/sign/SignTransferAccountNotFound.js
@@ -1,0 +1,127 @@
+import React from "react";
+import { Translate } from "react-localize-redux";
+import styled from "styled-components";
+
+import AccountNotFound from "../../images/icon-account-not-found.svg";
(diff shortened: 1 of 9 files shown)


MaximusHaximus (approved):
:+1: nice work @esaminu

PR #2448, opened 2022-02-07
fix: fix logic to append additional query params

### LINKED ISSUE #2442: Success URL appended with `?` despite it has already query params in it
**Problem**
Unable to log in to dApps. The success URL after request sign-in method is appended with `?` despite it has already query params in it. 
Just found this on testnet. However, by the time I write this issue, it still working normally on mainnet

**Expected Behavior**
Wallet returns correct URL which results in successful login on dApps

**Steps to reproduce**
1. Wallet.requestSignIn from current URL appended with params ex: https://paras.id/search?q=query
2. Connect the near account to dApps
3. The URL changes to: https://paras.id/search?q=query?account_id=...&public_key=...&all_keys=... with double `?`, which resulted in failed login

If I'm not mistaken a few days ago was still normal, perhaps this may be the cause: https://github.com/near/near-wallet/blob/master/packages/frontend/src/redux/actions/account.js#L210

esaminu:
Fixes logic that appends additional query params to urls with existing query params.

Closes #2442

### DIFF
--- a/packages/frontend/src/redux/actions/account.js
+++ b/packages/frontend/src/redux/actions/account.js
@@ -207,9 +207,20 @@ export const allowLogin = () => async (dispatch, getState) => {
         const availableKeys = await wallet.getAvailableKeys();
         
         const allKeys = availableKeys.map(key => key.toString());
-        window.location = `${successUrl}?${stringify(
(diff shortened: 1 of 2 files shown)


MaximusHaximus (commented):
:+1:

PR #2449, opened 2022-02-07
handleAsyncThunkStatus helper function to handle repeated patterns for the status state object

marcinbodnar:
Creating `handleAsyncThunkStatus` helper function that simplifies and allows to automatically handle status objects for reducers. 

It's prepared using `addMatcher` because we need to handle situations when we would like to use `.pending`/`.fulfilled`/`.rejected` for additional functionality. And it's not allowed to use more than one `addCase` for the same action status.

### DIFF
--- a/packages/frontend/src/redux/slices/availableAccounts/index.js
+++ b/packages/frontend/src/redux/slices/availableAccounts/index.js
@@ -24,7 +24,7 @@ const availableAccountsSlice = createSlice({
             set(state, ['items'], Object.keys((action.payload && action.payload.accounts) || {}).sort());
         });
         handleAsyncThunkStatus({
-            asyncThunk: `refreshAccountOwner`,
+            asyncThunk: refreshAccountOwner,
             buildStatusPath: () => ['status'],
             builder
         });
--- a/packages/frontend/src/redux/slices/availableAccounts/index.js
+++ b/packages/frontend/src/redux/slices/availableAccounts/index.js
@@ -25,7 +25,7 @@ const availableAccountsSlice = createSlice({
         });
         handleAsyncThunkStatus({
             asyncThunk: `refreshAccountOwner`,
-            buildStatusPath: ({ meta: { arg: { accountId }}}) => ['status'],
+            buildStatusPath: () => ['status'],
             builder
         });
     })
--- a/packages/frontend/src/redux/slices/availableAccounts/index.js
+++ b/packages/frontend/src/redux/slices/availableAccounts/index.js
@@ -3,6 +3,7 @@ import set from 'lodash.set';
 import { createSelector } from 'reselect';
 
 import refreshAccountOwner from '../../sharedThunks/refreshAccountOwner';
+import handleAsyncThunkStatus from '../handleAsyncThunkStatus';
 import initialErrorState from '../initialErrorState';
 
 const SLICE_NAME = 'availableAccounts';
@@ -19,22 +20,13 @@ const availableAccountsSlice = createSlice({
     name: SLICE_NAME,
     initialState,
     extraReducers: ((builder) => {
-        builder.addCase(refreshAccountOwner.pending, (state) => {
-            set(state, ['status', 'loading'], true);
-            set(state, ['status', 'error'], initialErrorState);
-        });
         builder.addCase(refreshAccountOwner.fulfilled, (state, action) => {
-            set(state, ['status', 'loading'], false);
-            set(state, ['status', 'error'], initialErrorState);
-
             set(state, ['items'], Object.keys((action.payload && action.payload.accounts) || {}).sort());
         });
-        builder.addCase(refreshAccountOwner.rejected, (state, { error }) => {
-            set(state, ['status', 'loading'], false);
-            set(state, ['status', 'error'], {
-                message: error?.message || 'An error was encountered.',
-                code: error?.code
-            });
+        handleAsyncThunkStatus({
+            asyncThunk: `refreshAccountOwner`,
+            buildStatusPath: ({ meta: { arg: { accountId }}}) => ['status'],
+            builder
         });
     })
 });
--- a/packages/frontend/src/redux/slices/handleAsyncThunkStatus.js
+++ b/packages/frontend/src/redux/slices/handleAsyncThunkStatus.js
@@ -1,0 +1,41 @@
+import set from 'lodash.set';
+
+import initialErrorState from './initialErrorState';
+
+/**
+ * Automatically handle status part of reducer based on initialErrorState
+ *
+ * @param asyncThunk string representing async thunk type
+ * @param buildStatusPath array of strings representing the path to status state, needed because in a few cases we want to place the status object in more than one place
+ * @param builder objct provides addCase, addMatcher, addDefaultCase functions
+ */
+export default ({
+    asyncThunk,
+    buildStatusPath,
+    builder
+}) => builder
+    // TODO: use the same status object to all reducers, which will allow simplifying buildStatusPath 
+    .addMatcher(
+        (action) => action.type === `${asyncThunk}/pending`,
+        (state, action) => {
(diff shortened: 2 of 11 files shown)


packages/frontend/src/redux/slices/availableAccounts/index.js
```diff
@@ -19,22 +20,13 @@ const availableAccountsSlice = createSlice({
...
         });
-        builder.addCase(refreshAccountOwner.rejected, (state, { error }) => {
-            set(state, ['status', 'loading'], false);
-            set(state, ['status', 'error'], {
-                message: error?.message || 'An error was encountered.',
-                code: error?.code
-            });
+        handleAsyncThunkStatus({
+            asyncThunk: `refreshAccountOwner`,
+            buildStatusPath: ({ meta: { arg: { accountId }}}) => ['status'],
```
andy-haynes, line 28: Would it make sense to omit the arrow function parameters to make it more clear that the path is static?

packages/frontend/src/redux/slices/handleAsyncThunkStatus.js
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Automatically handle status part of reducer based on initialErrorState
+ *
+ * @param asyncThunk string representing async thunk type
+ * @param buildStatusPath array of strings representing the path to status state, needed because in a few cases we want to place the status object in more than one place
+ * @param builder objct provides addCase, addMatcher, addDefaultCase functions
+ */
+export default ({
+    asyncThunk,
```
andy-haynes, line 13: Will the async thunk object will always have properties for `pending`, `fulfilled`, and `rejected` like it does in the thunks being refactored in this PR? 

If so I think it makes more sense to pass the async thunk here rather than the type string. Then you don't need to repeat the action type when invoking `handleAsyncThunkStatus` and the typing becomes more explicit - i.e. an object with properties for `pending`, et al. rather than a string that can't be validated against known action types.

andy-haynes (changes requested):
Nice refactor, just curious if it would make more sense to retain the logic that relies on the async thunk object rather than the action type.

packages/frontend/src/redux/slices/availableAccounts/index.js
andy-haynes: Yup it needs to stay a function but since it doesn't actually need anything from the action it can take zero parameters. It makes typing a little wonky but it prevents `accountId` from being an unreferenced variable.

packages/frontend/src/redux/slices/handleAsyncThunkStatus.js
marcinbodnar: It's not possible, unfortunately. I was trying to find a better solution, but it seems that there is none. 
`pending/fulfilled/rejected` are action creators, that are being created while creating the thunk. And matcher function in `addMatched` has `action` param only - so in this case, the only usable param to match is `action.type`, and we can't get that from the thunk function.

What you describe would be possible with `builder.addCase` but it's not allowed to use two of these in one slice for the same action creator, so we can only add one `addCase` for a particular action creator (pending/fulfilled/rejected). And in some cases, we are using `addCase` for different reasons than `status` object.

packages/frontend/src/redux/slices/availableAccounts/index.js
marcinbodnar: Ahh I see, I misunderstood. Done.

packages/frontend/src/redux/slices/tokenFiatValues/index.js
```diff
@@ -23,31 +24,18 @@ const tokenFiatValuesSlice = createSlice({
...
-                state.loading = false;
-
-                // TODO: Localize this?
-                state.error = {
-                    message: error?.message || 'An error was encountered.',
-                    code: error?.code
-                };
+            handleAsyncThunkStatus({
+                asyncThunk: `${SLICE_NAME}/fetchTokenFiatValues`,
+                buildStatusPath: () => '',
```
andy-haynes, line 37: Should this return `[]` so that it doesn't add an empty string to the array when `buildStatusPath` gets called?

packages/frontend/src/redux/slices/handleAsyncThunkStatus.js
andy-haynes: Interesting, I saw your note on `addCase` in the PR description but hadn't made that connection.

packages/frontend/src/redux/slices/tokens/index.js
```diff
@@ -110,26 +111,10 @@ const tokensSlice = createSlice({
...
-        builder.addCase(fetchOwnedTokensForContract.rejected, (state, { meta, error }) => {
-            const { accountId, contractName } = meta.arg;
-
-            set(state, ['ownedTokens', 'byAccountId', accountId, contractName, 'loading'], false);
-            set(state, ['ownedTokens', 'byAccountId', accountId, contractName, 'error'], {
-                message: error?.message || 'An error was encountered.',
-                code: error?.code
-            });
+        handleAsyncThunkStatus({
+            asyncThunk: `${SLICE_NAME}/fetchOwnedTokensForContract`,
```
andy-haynes, line 115: Per feedback above would prefer to use `fetchOwnedTokensForContract.typePrefix` here instead.

packages/frontend/src/redux/slices/transactions/index.js
```diff
@@ -88,26 +89,10 @@ const transactionsSlice = createSlice({
...
-        builder.addCase(fetchTransactions.rejected, (state, { meta, error }) => {
-            const { accountId } = meta.arg;
-            
-            set(state, ['byAccountId', accountId, 'status', 'loading'], false);
-            set(state, ['byAccountId', accountId, 'status', 'error'], {
-                message: error?.message || 'An error was encountered.',
-                code: error?.code
-            });
+        handleAsyncThunkStatus({
+            asyncThunk: `${SLICE_NAME}/fetchTransactions`,
```
andy-haynes, line 93: Per feedback above would prefer to use `fetchTransactions.typePrefix` here instead.

andy-haynes (changes requested):
I just have some comments around reusing thunk properties if possible. Otherwise looking really good, I appreciate the clarifying comments around using the thunks directly.

packages/frontend/src/redux/slices/tokenFiatValues/index.js
marcinbodnar: The result will be the same, but an empty array has much more sense. Done.

packages/frontend/src/redux/slices/availableAccounts/index.js
```diff
@@ -19,22 +20,13 @@ const availableAccountsSlice = createSlice({
...
             set(state, ['items'], Object.keys((action.payload && action.payload.accounts) || {}).sort());
         });
-        builder.addCase(refreshAccountOwner.rejected, (state, { error }) => {
-            set(state, ['status', 'loading'], false);
-            set(state, ['status', 'error'], {
-                message: error?.message || 'An error was encountered.',
-                code: error?.code
-            });
+        handleAsyncThunkStatus({
+            asyncThunk: `refreshAccountOwner`,
```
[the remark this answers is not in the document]
marcinbodnar: Definitely better. Done.

packages/frontend/src/redux/slices/handleAsyncThunkStatus.js
```diff
@@ -0,0 +1,41 @@
+import set from 'lodash.set';
+
+import initialErrorState from './initialErrorState';
+
+/**
+ * Automatically handle status part of reducer based on initialErrorState
+ *
+ * @param asyncThunk actionCreator that is used to get the action type for matcher purpose
```
andy-haynes, line 8: ❤️

andy-haynes (approved):
Looks great, nice work!

marcinbodnar:
Thank you for the review @andy-haynes

PR #2478, opened 2022-02-14
fix: update old Dapp logos with new ones

picturepan2:
- [ ] Update Mintbase and Flux old logos with new ones. 

- [ ] Replace ZED with Ref Finance Logo.

image


Patrick1904 (approved):
Thanks for updating this @picturepan2! 💯 🙏

PR #2480, opened 2022-02-14
Mieubrisse/fix dockerfiles

mieubrisse:


andy-haynes (commented):
Looks great, thanks for the fix @mieubrisse!

PR #2487, opened 2022-02-15
feat: add script to disable 2fa using seedphrase

esaminu:
Add script allows the user to disable NEAR Wallet 2fa using the provided seedphrase


agileurbanite:
@esaminu @MaximusHaximus is this not something useful to also add to https://github.com/near/near-cli ?

PR #2490, opened 2022-02-16
Refactor status handle functions

marcinbodnar:
Extract `status` object into a separate file, so it will be easy to use in `initialState` for every slice. It will also standardize what the status looks like on every slice.

Simplify `handleAsyncThunkStatus` helper function by using `isFulfilled/isPending/isRejected` from redux toolkit and a set of helper functions. Thanks to that the function is more readable, also we have an easy way to change `status` object properties.

Moved all files in the proper folders structure.


andy-haynes (approved):
LGTM

PR #2492, opened 2022-02-18
Fix choosing token when two tokens have the same symbol

marcinbodnar:


Patrick1904 (approved):
Thank you @marcinbodnar! 🙏 💯

PR #2493, opened 2022-02-19
fix: Broken styling from Semantic UI removal

gutsyphilip:
Fixes some unnoticed UI quirks from the removal of Semantic UI.

- fix: /recover-with-link container


Patrick1904 (approved):
Thanks @gutsyphilip! 🙏

PR #2498, opened 2022-02-23
WAL-126: Import failure show public key

### LINKED ISSUE #2386: Specify PubKey on failed account recovery
## Story

As a USER who is failing to recover an account via [recover account](https://wallet.near.org/recover-account) functionality, I want to see the _PubKey_ that I generated with my recovery method, so I can use it when I open a Zendesk Incident

## Acceptance Criteria

- [ ] Show an error that is similar to "PubKey ed2559:xyz not found" instead of "No matching keypair found" (where xyz is the generated pubkey during the account recovery process)
- [ ] Mobile ok

## Unhappy Path
- The PubKey is not visible in the screenshot (useful when opening a Zendesk incdent)

## Assets
(it's an example, feel free to improve the messaging)

Screen Shot 2022-01-19 at 17 22 58

## Notes
This issue will vastly improve the capability of the support team to help users who can't recover their account

andy-haynes:
This PR adds the seed phrase public key to the error message when seed phrase recovery fails. I've also fixed several typos in adjacent code.

Closes #2386

### DIFF
--- a/packages/frontend/src/components/common/GlobalAlert.js
+++ b/packages/frontend/src/components/common/GlobalAlert.js
@@ -224,7 +224,6 @@ const GlobalAlertNew = ({ globalAlert, actionStatus, clearGlobalAlert, closeIcon
                                                     {noTranslationFound
                                                         ? <Translate id={`reduxActions.default.${alert.success ? 'success' : 'error'}`} />
                                                         : <SafeTranslate id={alert.messageCode} data={alert.data} />
-                                                        // : <Translate id={alert.messageCode} data={escapeHtml(alert.data)} />
                                                     }
                                                     {!alert.success &&
                                                         <a
--- a/packages/frontend/src/components/common/GlobalAlert.js
+++ b/packages/frontend/src/components/common/GlobalAlert.js
@@ -228,7 +228,7 @@ const GlobalAlertNew = ({ globalAlert, actionStatus, clearGlobalAlert, closeIcon
                                                     }
                                                     {!alert.success &&
                                                         <a
-                                                            href={(noTranslationFound || msgCode.includes('Sorry an error has occured')) 
+                                                            href={(noTranslationFound || msgCode.includes('Sorry an error has occurred'))
                                                                 ? `${zendeskBaseURL}${alert.errorMessage ? `search?query=${encodeURIComponent(alert.errorMessage.substring(0, 500))}` : ''}` 
                                                                 : `${zendeskBaseURL}search?query=${encodeURIComponent(msgCode.substring(0, 500))}`
                                                             }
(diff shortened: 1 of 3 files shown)


packages/frontend/src/utils/wallet.js
```diff
@@ -860,7 +860,7 @@ class Wallet {
         accountIds = [...accountsSet];
 
         if (!accountIds.length) {
-            throw new WalletError('Cannot find matching public key', 'recoverAccountSeedPhrase.errorInvalidSeedPhrase', { publicKey });
```
andy-haynes, line 863: Originally I looked at why the error parsing/alert box wasn't displaying this but after revisiting the original issue I noticed that the first parameter to `WalletError` is what appears in the code-styled box on the alert surfaced to the user.

packages/frontend/src/components/common/GlobalAlert.js
```diff
@@ -224,11 +224,10 @@ const GlobalAlertNew = ({ globalAlert, actionStatus, clearGlobalAlert, closeIcon
                                                     {noTranslationFound
                                                         ? <Translate id={`reduxActions.default.${alert.success ? 'success' : 'error'}`} />
                                                         : <SafeTranslate id={alert.messageCode} data={alert.data} />
```
andy-haynes, line 226: In debugging why the `publicKey` could not be interpolated into the translation string, I noticed that `SaveTranslate` appears to be receiving the wrong argument here. When I replaced `alert.data` with `alert.data.data`, the interpolation worked.

I assume this is affecting other alerts that are potentially failing to display interpolated values in translation text, but given that:
- it wasn't the ideal approach for this issue
- it's out of scope for this PR
- it could potentially break other errors across the wallet
I decided against making that change as part of this fix. However if it is a known issue affecting other parts of the app then we should address it separately.

packages/frontend/src/components/common/GlobalAlert.js
```diff
@@ -224,11 +224,10 @@ const GlobalAlertNew = ({ globalAlert, actionStatus, clearGlobalAlert, closeIcon
...
-                                                        // : <Translate id={alert.messageCode} data={escapeHtml(alert.data)} />
                                                     }
                                                     {!alert.success &&
                                                         <a
-                                                            href={(noTranslationFound || msgCode.includes('Sorry an error has occured')) 
+                                                            href={(noTranslationFound || msgCode.includes('Sorry an error has occurred'))
```
andy-haynes, line 230: This should be safe, `Sorry an error has occured` doesn't show up in any translation files with the typo.

Patrick1904 (approved):
Thanks @andy-haynes! Works well 👍 

Screen Shot 2022-02-23 at 1 07 29 PM

packages/frontend/src/components/common/GlobalAlert.js
Patrick1904: Can you please create a ticket so that we don't forget to follow up on this?

cc @marcinbodnar

andy-haynes: Done #2504

PR #2499, opened 2022-02-23
fix: prevent transaction approval until account matches signer

### LINKED ISSUE #2482: Able to sign with wrong signerId on /sign route (esp. slow connection)
**Problem**
Able to sign a TX using wrong `signerId` on `/sign` before the correct account has been selected / made active in wallet

**Expected Behavior**
Not able to sign a TX using wrong `signerId` on `/sign` before the correct account has been selected / made active wallet

**Steps to reproduce**

1. Log in on https://berryclub.io/ with account A
2. In a separate tab, make account B active on wallet.near.org
3. Refresh page on https://berryclub.io/
4. Initiate buying something on https://berryclub.io/
5. Observe that the 'approve' button is still enabled before the `signerId` matches the active wallet account

andy-haynes:
This PR adds a flag prop to the transaction signing components to prevent approving transactions when the active account has not yet switched over to the signing account. This will always be temporary since the account is guaranteed to be valid to render the button and the wrapper dispatches the action to switch accounts if the active account does not match the signing account.

Closes #2482

### DIFF
--- a/packages/frontend/src/components/sign/v2/SignTransactionSummary.js
+++ b/packages/frontend/src/components/sign/v2/SignTransactionSummary.js
@@ -54,6 +54,7 @@ export default ({
     onClickMoreInformation,
     accountUrlReferrer,
     submittingTransaction,
+    isSignerValid,
     isValidCallbackUrl
 }) => {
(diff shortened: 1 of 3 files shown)


Patrick1904 (approved):
Thanks @andy-haynes! 🙏 💯

PR #2501, opened 2022-02-23
fix: radio button

gutsyphilip:
Before:
image

After:
image


Patrick1904 (approved):
Thanks @gutsyphilip - looks great!

PR #2503, opened 2022-02-24
fix(recovery): bypass key creation during recovery when account is active or available

### LINKED ISSUE #2494: Non draining NEAR recovery \ login link
## Story

As a developer, I want generate & pre-fund implicit accounts for users and send them a link they can use to log-in into their account with near wallet. I found this recovery deep-link (for example): 

```
https://wallet.testnet.near.org/recover-with-link/<accountId>/<seed-phrase>
```

And it works great. But there is one problem with such link - even if user already have keys for account in their browser storage (meaning they already "recovered" their account via link once) the link will still drain some near and make new access key. 

Is it possible to have a link which first checks the keys availability and only if they are not found, do the recover. Something like recover \ login link?

## Acceptance Criteria

- [ ] User can recover their account by going through link the first time
- [ ] User don't spend additional near going through the same link over and over again if browser storage has needed keys

andy-haynes:
This PR modifies the link recovery component to check on update whether the account being recovered is already in the account state (i.e. already has keys for in the current browser's localStorage). If it is already the active account, the user is redirected to the root page. If it is not currently active but is in the set of available accounts, then the current account is switched to that the account being recovered.

Recovering accounts in browsers without a key in localStorage will still require the creation of an access key from the seed phrase in the link.

To test using the preview link:
- Visit https://near-wallet-pr-2503.onrender.com/recover-with-link/pr2503_test.testnet/virtual%20arrive%20cliff%20hunt%20stand%20cupboard%20syrup%20bachelor%20safe%20basic%20dawn%20inside.
- Verify that the `Continue` button allows you to recover the account.
- Visit the link again, this time verifying that you are redirected to the account page because you are logged into the account already.

Closes #2494

### DIFF
--- a/packages/frontend/src/components/accounts/RecoverWithLink.js
+++ b/packages/frontend/src/components/accounts/RecoverWithLink.js
@@ -10,9 +10,11 @@ import {
     recoverAccountSeedPhrase,
     refreshAccount,
     redirectTo,
-    clearAccountState
+    clearAccountState,
+    makeAccountActive,
 } from '../../redux/actions/account';
-import { selectAccountSlice } from '../../redux/slices/account';
+import { selectAccountId, selectAccountSlice } from '../../redux/slices/account';
+import { selectAvailableAccounts } from '../../redux/slices/availableAccounts';
 import { selectActionsPending, selectStatusMainLoader } from '../../redux/slices/status';
 import copyText from '../../utils/copyText';
 import isMobile from '../../utils/isMobile';
@@ -134,6 +136,7 @@ class RecoverWithLink extends Component {
 
         this.state = {
             accountId: this.props.accountId,
+            isSwitchingAccount: false,
             seedPhrase: this.props.seedPhrase,
             successSnackbar: false,
             successView: true
@@ -176,8 +179,23 @@ class RecoverWithLink extends Component {
         );
     }
 
-    render() {
+    componentDidUpdate() {
+        const { accountId, isSwitchingAccount } = this.state;
+        const { isAccountActive, isAccountAvailable } = this.props;
 
+        // redirect user to home page if active account is being recovered
+        // switch to account being recovered if already in the set of available accounts
+        if (isAccountActive) {
+            this.props.redirectTo('/');
+        } else if (isAccountAvailable && !isSwitchingAccount) {
+            // reset flag to ensure account switch is only dispatched once
+            this.setState({ isSwitchingAccount: true });
+            this.props.makeAccountActive(accountId);
+            this.props.refreshAccount();
+        }
+    }
+
+    render() {
         const { accountId, successSnackbar, successView } = this.state;
         const { mainLoader, history, continueSending } = this.props;
 
@@ -241,10 +259,11 @@ class RecoverWithLink extends Component {
 }
 
 const mapDispatchToProps = {
-    recoverAccountSeedPhrase, 
+    recoverAccountSeedPhrase,
     refreshAccount,
     redirectTo,
-    clearAccountState
+    clearAccountState,
+    makeAccountActive,
 };
 
 const mapStateToProps = (state, { match }) => ({
@@ -252,7 +271,9 @@ const mapStateToProps = (state, { match }) => ({
     accountId: match.params.accountId,
     seedPhrase: match.params.seedPhrase,
     mainLoader: selectStatusMainLoader(state),
-    continueSending: selectActionsPending(state, { types: ['RECOVER_ACCOUNT_SEED_PHRASE'] })
+    continueSending: selectActionsPending(state, { types: ['RECOVER_ACCOUNT_SEED_PHRASE'] }),
+    isAccountActive: selectAccountId(state) === match.params.accountId,
+    isAccountAvailable: selectAvailableAccounts(state).some((accountId) => accountId === match.params.accountId),
 });
 
 export const RecoverWithLinkWithRouter = connect(
--- a/packages/frontend/src/components/accounts/RecoverWithLink.js
+++ b/packages/frontend/src/components/accounts/RecoverWithLink.js
@@ -84,6 +84,8 @@ const UserName = styled.span`
 const ButtonWrapper = styled.div`
     display: flex;
     flex-direction: column;
+    align-items: flex-end;
+    justify-content: space-evenly;
     margin-top: 30px;
 
     @media (min-width: 768px) {
@@ -96,17 +98,12 @@ const ButtonWrapper = styled.div`
             max-width: 300px;
         }
 
-        &:first-of-type {
-            margin-top: 0 !important;
-        }
-
         &:last-of-type {
             margin-top: 25px;
             position: relative;
             overflow: hidden;
 
             @media (min-width: 768px) {
-                margin-left: 25px;
                 margin-top: 0;
             }
 
@@ -137,6 +134,7 @@ class RecoverWithLink extends Component {
         this.state = {
             accountId: this.props.accountId,
             isSwitchingAccount: false,
+            isTargetAccountSelected: false,
             seedPhrase: this.props.seedPhrase,
             successSnackbar: false,
             successView: true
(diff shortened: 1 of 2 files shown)


Patrick1904:
This looks great to me. There are some corner cases that I think we may want to address:

1. Some users, especially users that used a temporary funding address, may want to recover both their implicit and named account. For example, the user may send funds at a later point to the funding address by mistake, and a simple way to 'recover' that address is by simply importing the account again, since we check and recover **all** accounts corresponding to the recovery key. If we add this behavior, the only way for the user to actually import their account again would be to empty `localStorage`. We could call `getAccountIds` to make sure there aren't additional accounts, before deciding that the user is already logged in and redirect.

2. I do see from time to time users reporting having a corrupted key in `localStorage`, I believe the error is something like `cannot find matching key...`. This is usually resolved by simply importing the account again, since a new key is created and added to localStorage. I guess we could mitigate this by grabbing the key from localStorage and checking to make sure it's active on chain, before deciding that the user is already properly logged in.

Performing some of the checks may take a sec, so might want to disable the continue button while checking, assuming the `accountId` is part of the wallet accounts list already.

What do you think?

andy-haynes:
@Patrick1904 those are really interesting edge cases, but I think the scope of the changes may be small enough to not cause problems when they arise.

1. Since the route requires an `accountId` parameter that gets passed through to `recoverAccountSecretKey`, it looks like the component can ever recover exactly one account. It looks like all you'd need to do to recover an implicit account is use that account ID in a different URL from the one used to recover the named account (i.e. different accounts, same phrase). Does that sound right or am I maybe missing something?
2. In this scenario would it be more likely that they would use the `Import Account` button to fix the corrupted localStorage? This new behavior would only come into play if they had followed the `/recover-with-link` URL and had the account active/available, in which case they should still be able to use the `Import Account` button.


> Performing some of the checks may take a sec, so might want to disable the continue button while checking, assuming the `accountId` is part of the wallet accounts list already.

Nice catch! I'll add a disabled condition for the continue button when the account is being switched.

andy-haynes:
@Patrick1904 and I discussed this offline further and he's completely right about potential unintended consequences of this change for the edge cases he outlined. While these are not cases we need to fix with this PR, we agreed that the best solution would leave it to the user to decide whether they want to either:
- skip recovery because their account is already logged in on the current browser
- continue on with recovery (despite being logged in) to allow the recovery link to continue working as a way to recover implicit or corrupted accounts

So the page will automatically check whether the account being recovered is already logged in:
- **if the account is not logged in**, then there are no changes to the current flow - the user is still presented with the `Continue` and `Copy URL` buttons
- **if the account is logged in**, we present the user with a dialog indicating that they may skip recovery if they wish, otherwise they may continue on with account recovery

In the case that the user is logged in, it would be a good idea to indicate that continuing on with recovery for a named account would also recover the implicit account, as this is a side effect of recovering via link that may not be well understood by users.

@corwinharrell do you have thoughts on how a prompt like this should look and behave?

agileurbanite:
@andy-haynes do you have any recommendations of how a prompt like this should look for @corwinharrell ?  I'm wondering if you can propose a PR that can get @corwinharrell to sign off on.  As long as we use the design system or existing components, I don't believe that this will introduce any new UI to the user that deviates from the existing prompts that we already have.  Would love to know your thoughts.

andy-haynes:
> @andy-haynes do you have any recommendations of how a prompt like this should look for @corwinharrell ? I'm wondering if you can propose a PR that can get @corwinharrell to sign off on. As long as we use the design system or existing components, I don't believe that this will introduce any new UI to the user that deviates from the existing prompts that we already have. Would love to know your thoughts.

I'm mostly curious to what extent the team thinks this requires a new prompt, whether it's simply another confirmation button, additional copy to convey the situation to the user, a modal alert, or some combination. Being unfamiliar with the typical frontend dev process for the wallet I thought it prudent to seek additional UX insight. I do believe that the implementation as is would address most scenarios but if there were users relying on this flow to address the edge cases mentioned then the new behavior could potentially exacerbate things.

corwinharrell:
This may actually be slightly more complex than has been accounted for, simply because of the fact that NEAR allows for a single private key to be shared across even more than just 2 accounts. In this scenario, and in the future, the optimal solution would actually allow the user to select from a list the accounts that they wish to import. This is similar to the behavior we see with Ledger, where a single Ledger device can be used to secure any number of accounts.

It may make sense to take into account the work being done on [Multiple Ledger Transports](https://nearinc.atlassian.net/jira/software/projects/WAL/boards/7?selectedIssue=WAL-45), which addresses a very similar scenario in the context of Ledger recovery.

To demonstrate possible implications on the UI, here is an example from the Ledger work, where 2/3 accounts have already been imported, but one account has been detected that is yet to be imported:

Screen Shot 2022-03-01 at 12 34 55 PM

I don't see any reason why we couldn't leverage a similar approach for other key recovery methods (eg "We found 3 accounts secured by this passphrase"), where we detect the number of accounts secured by a given key, display them, and allow the user to either skip recovery, or recover the remaining accounts secured by that key.

Thoughts @andy-haynes @Patrick1904?

corwinharrell:
I believe at the moment, recovering with passphrase just imports all accounts secured by that phrase automatically, with no option to see or select the accounts until after all of them have been imported (This should change to match what we will have going on with ledger for example though).

I totally understand the limitations of the current route. I think what I'm suggesting is that the decision to explicitly call out a single wallet address within the UI may have actually been a design flaw originally implemented without sufficient foresight, and now we're having to iterate around it.

Calling out a single address in the UI when the private key provided is actually used to secure multiple, and given that you can import these remaining accounts, is pretty confusing. I don't think we need to change the address referenced in the URL here necessarily, but I do think it would be ideal to treat the UI conditionally, and to be upfront about the accounts available to import as well as their state:

If < 2 accounts & account **has not** been imported:

Desktop - 1

If < 2 accounts & account **has already** been imported:

Desktop - 2

If > 1 accounts & 1/2 **has already** been imported:

Desktop - 3

agileurbanite:
I think the idea here was to help folks who are also using this when they had pre-funded implicit accounts to bypass this screen.

Patrick1904:
Just want to summarize what the exact problems are because I think some of it got lost in our discussions above.

Every time a user clicks on their email/sms recovery link, they land on the [link recovery page](https://near-wallet-pr-2503.onrender.com/recover-with-link/pr2503_test.testnet/virtual%20arrive%20cliff%20hunt%20stand%20cupboard%20syrup%20bachelor%20safe%20basic%20dawn%20inside) and is presented with the 'Recover Account' button. Once the user clicks the button, we add a brand new key to the users `localStorage`. This is fine, however, **we never check if the user is already logged in with that account**.

If the user is already logged in to the specific account, we'd like to **avoid adding a new key**, and simply 1. make the account active in the wallet and 2. redirect to dashboard.

There are two caveats to this approach that we'd like to address with some potential UX:

1. The user might be a legacy user that accidentally sent funds to the original 'funding address', and may wish to import both their implicit/funding address account, as well as their named account. We automatically import all accounts associated with the recovery link seed phrase, but as outlined above, if the user is already logged in to the named account, we'd skip this 'import action' all together, unless we can display a message to the user saying something like 'Your account corwin.near is already imported in this wallet, do you want to import it again?', and then show buttons 'import again' or 'skip' (skip would just make the named account active and redirect to dashboard).
2. Another caveat is that a user could somehow corrupt their `localStorage` and may want to re-import the account in order to clean up their `localStorage` with a fresh key. It would therefore be a good idea to allow the user to import their account again, despite the account already existing in the wallet.

We'd basically like to offer the user the option to re-import the account even if the user is already logged in to the account as a way to override the existing key and import all accounts associated with the corresponding seed phrase/key.

corwinharrell:
After additional discussion, @andy-haynes @Patrick1904 and I have concluded the following:

1. This route will not be concerned with or used as a solution for the issue we are seeing where users' localStorage keys become corrupt. This issue will instead be solely focused on offering users who are importing their account a way to bypass the process if the account already exists. We'll used the proposed designs above to enable this, additionally offering a way for users to import any other accounts that may be secured by the URL (ie implicit)

2. The issue of corrupted localStorage keys still needs to be addressed. We should take some time to look deeper into what is causing users to encounter this error in order to determine if there is more that can be done to mitigate it.

3. In the meantime, it may make sense to prioritize the long-awaited "Remove Account" feature as a means to solve for errors such as these. In general, these issues can be resolved by removing the affected account's existing access key and re-importing it with a new one. Rather than have users fiddle with their cache/localStorage manually, offering a simple and straightforward way to remove one's account is preferred. As a follow-up to this discussion, we'll be putting together designs for the "Remove Account" feature, and will likely want to prioritize this soon as A) It helps users to troubleshoot in scenarios like these, and B) Is one of the most commonly requested features in our backlog.

Patrick1904:
Closing this PR since the scope of this project has changed and now addressed in https://github.com/near/near-wallet/pull/2521

PR #2505, opened 2022-02-24
chore: add new project validator factory to regex

esaminu:
This PR adds support for validators using wallet factory `factory.colorpalette.testnet` as well as aurora's `factory01.littlefarm.testnet`


MaximusHaximus (approved):
:+1:

PR #2507, opened 2022-02-25
Feat farming validator UI

esaminu:
Adopted branch for the new stake farming feature as implemented on #2398


esaminu:
Updated `LOCKUP_ACCOUNT_ID_SUFFIX` on render preview to `lockup.devnet`. We can now use https://testnet.lockup.tech/ to deploy lockups for testing. Just update `wallet.testnet.near.org` to `near-wallet-pr-2507.onrender.com` when logging in and signing transactions.

esaminu:
Validators for testing:

- [zentriav2.factory.colorpalette.testnet](https://near-wallet-pr-2507.onrender.com/staking/zentriav2.factory.colorpalette.testnet) 
- [domanodes.factory.colorpalette.testnet](https://near-wallet-pr-2507.onrender.com/staking/domanodes.factory.colorpalette.testnet)

esaminu:
[testnet preview url](https://near-wallet-pr-2507.onrender.com)

To test, the following farm validators can be used:

* [zentriav2.factory.colorpalette.testnet](https://near-wallet-pr-2512.onrender.com/staking/zentriav2.factory.colorpalette.testnet)
* [domanodes.factory.colorpalette.testnet](https://near-wallet-pr-2512.onrender.com/staking/domanodes.factory.colorpalette.testnet)

To test lockup staking, use https://testnet.lockup.tech/ to add a lockup to your account and update `wallet.testnet.near.org` to `near-wallet-pr-2512.onrender.com` when logging in and signing transactions.

packages/frontend/src/hooks/fungibleTokensIncludingNEAR.js
```diff
@@ -11,5 +12,11 @@ export const useFungibleTokensIncludingNEAR = function () {
         selectTokensWithMetadataForAccountId(state, { accountId })
     );
 
-    return [NEARAsTokenWithMetadata, ...fungibleTokens];
+    const fungibleTokenPrices = useSelector(selectTokensFiatValueUSD);
+    const fungibleTokensWithPrices = fungibleTokens.map((ft) => ({
```
MaximusHaximus, line 16: **Not a blocker for this PR**, but this should be memoized so that we don't create an entirely new value for `fungibleTokensWithPrices` here _every time this function runs_.

packages/frontend/src/config/environmentDefaults/development.js
```diff
@@ -42,4 +43,9 @@ export default {
     SMS_BLACKLIST: [],
     STAKING_GAS_BASE: '25000000000000', // 25 Tgas
     WHITELISTED_CONTRACTS: ['meta.pool.testnet'],
+    NEAR_TOKEN_ID: 'wrap.testnet',
+    FARMING_CLAIM_GAS: parseNearAmount('0.00000000015'),
+    FARMING_CLAIM_YOCTO: '1',
+    REF_FINANCE_API_ENDPOINT: 'https://dev-indexer.ref-finance.com/',
+    REF_FINANCE_CONTRACT: 'ref-finance-101.testnet'
```
MaximusHaximus, line 50: 💖 nice work finding the correct configuration for dev :D

packages/frontend/src/redux/actions/staking.js
```diff
@@ -6,28 +6,20 @@ import {
     ACCOUNT_HELPER_URL,
     REACT_APP_USE_TESTINGLOCKUP,
     STAKING_GAS_BASE,
+    FARMING_CLAIM_GAS,
```
MaximusHaximus, line 9: :hocho: Opportunistic reformatting in this file that changed a bunch of imports from one-per-line to multiple-per-line.  Since we prefer 1-per-line for large objects, this will result in 'flapping' next time someone else touches the file.

packages/frontend/src/redux/actions/staking.js
```diff
@@ -533,3 +524,85 @@ export const handleUpdateCurrent = (accountId) => async (dispatch, getState) =>
...
+export const getValidatorFarmData = (validator, accountId) => async (dispatch, getState) => {
+
+    if (validator?.version !== FARMING_VALIDATOR_VERSION || !accountId) return;
+
+    const poolSummary = await validator.contract.get_pool_summary();
+    const farms = await validator.contract.get_farms({ from_index: 0, limit: 300 });
+
+    const list = await Promise.all(farms.map(({ token_id, farm_id, active }) => {
+        try {
+            dispatch(fetchToken({ contractName: token_id }));
```
MaximusHaximus, line 537: `await` should be here so that if this call throws an error, we fall into the `catch()` block before trying to call `get_unclaimed_reward` (if the token cannot be loaded, we won't know how to _format_ the unclaimed reward)

packages/frontend/src/redux/actions/staking.js
```diff
@@ -533,3 +524,85 @@ export const handleUpdateCurrent = (accountId) => async (dispatch, getState) =>
...
+        }
+        
+    }));
+
+    const farmData = {
+        poolSummary: {...poolSummary},
+        farmRewards: list,
+    };
+    await dispatch(staking.setValidatorFarmData(validator.accountId, farmData));
+    return farmData;
```
MaximusHaximus, line 564: Since this puts data into state where it should be selected using a selector, it would be an anti-pattern for the caller to also wait for the returned result.  🔪  this return so that the method doesn't encourage someone to `await` this thunk rather than using selectors to load the data that this thunk populates.

packages/frontend/src/redux/actions/staking.js
```diff
@@ -533,3 +524,85 @@ export const handleUpdateCurrent = (accountId) => async (dispatch, getState) =>
     let currentAccount = selectStakingCurrentAccountbyAccountId(getState(), { accountId });
     dispatch(staking.updateCurrent({ currentAccount }));
 };
+
+export const getValidatorFarmData = (validator, accountId) => async (dispatch, getState) => {
```
MaximusHaximus, line 528: **Not a blocker for this PR**, but...

1. This thunk should be setting loading state and error states so that the UI doesn't show a 0 balance in case
2. Directly related to the lack of loading and error states managed here, we should also not be blindly swallowing _all_ errors encountered from `get_unclaimed_reward` and returning a `0` value for balance in those cases, which would be misleading to the user.

@esaminu please make sure we've got this issue documented so we can folllow up and fix this up so there is user feedback when something has gone wrong instead of them seeing a 0 balance.

packages/frontend/src/redux/actions/staking.js
```diff
@@ -533,3 +524,85 @@ export const handleUpdateCurrent = (accountId) => async (dispatch, getState) =>
...
+
+    const farmData = {
+        poolSummary: {...poolSummary},
+        farmRewards: list,
+    };
+    await dispatch(staking.setValidatorFarmData(validator.accountId, farmData));
+    return farmData;
+};
+
+export const claimFarmRewards = (validatorId, token_id) => async (dispatch, getState) => {
```
MaximusHaximus, line 567: I am curious-- why doesn't this code use a `farm_id` instead of `token_id`? Is `token_id` guaranteed to be unique in the same way that `farm_id` is?

packages/frontend/src/redux/actions/staking.js
```diff
@@ -533,3 +524,85 @@ export const handleUpdateCurrent = (accountId) => async (dispatch, getState) =>
...
+    try {
+        const accountId = selectStakingMainAccountId(getState());
+        const currentAccountId = selectStakingCurrentAccountAccountId(getState());
+        const isLockup = currentAccountId !== accountId;
+
+        const validators = selectStakingAllValidators(getState());
+        const validator = { ...validators.find((validator) => validator?.accountId === validatorId)};
```
MaximusHaximus, line 574: **Not a blocker for this PR** , but this should be using `selectStakingFindContractByValidatorId` selector instead AFAICT.

packages/frontend/src/redux/actions/staking.js
```diff
@@ -533,3 +524,85 @@ export const handleUpdateCurrent = (accountId) => async (dispatch, getState) =>
...
+                    const account = await wallet.getAccount(accountId);
+                    await fungibleTokensService.transferStorageDeposit({
+                         account,
+                         contractName: token_id,
+                         receiverId: accountId,
+                         storageDepositAmount: FT_MINIMUM_STORAGE_BALANCE_LARGE
+                        });
+                }
+                catch (e) {
+                    console.warn(e);
```
MaximusHaximus, line 591: **Not a blocker for this PR**, but -- :hocho: this catch block -- if we are re-throwing the error, we should not `console.warn()` it here. This kind of pattern tends to lead to errors being logged multiple times as they bubble up the stack, which can be confusing in many ways.

packages/frontend/src/redux/actions/staking.js
```diff
@@ -533,3 +524,85 @@ export const handleUpdateCurrent = (accountId) => async (dispatch, getState) =>
...
+    };
+    await dispatch(staking.setValidatorFarmData(validator.accountId, farmData));
+    return farmData;
+};
+
+export const claimFarmRewards = (validatorId, token_id) => async (dispatch, getState) => {
+    try {
+        const accountId = selectStakingMainAccountId(getState());
+        const currentAccountId = selectStakingCurrentAccountAccountId(getState());
+        const isLockup = currentAccountId !== accountId;
```
MaximusHaximus, line 571: This feels like a dangerous way to 'infer' if this is a lockup account. Don't we use a regex match to make this determination generally?

packages/frontend/src/redux/actions/staking.js
```diff
@@ -533,3 +524,85 @@ export const handleUpdateCurrent = (accountId) => async (dispatch, getState) =>
...
+        const args = { token_id };
+
+        if (isLockup) {
+            const lockupId = selectStakingLockupId(getState());
+            args.delegator_id = lockupId;
+        }
+
+        return validator.contract.claim(args, FARMING_CLAIM_GAS, FARMING_CLAIM_YOCTO);
+    } catch (error) {
+        throw error;
```
MaximusHaximus, line 605: **Not a blocker for this PR**, but  🔪 this catch block does nothing, so the entire try/catch should be removed unless we add actual error handling here.

packages/frontend/src/redux/slices/tokenFiatValues/index.js
```diff
@@ -4,14 +4,38 @@ import { createSelector } from 'reselect';
...
 const SLICE_NAME = 'tokenFiatValues';
 
 const fetchTokenFiatValues = createAsyncThunk(
     `${SLICE_NAME}/fetchTokenFiatValues`,
-    () => sendJson('GET', ACCOUNT_HELPER_URL + '/fiat')
+    async () => {
+        const nearFiatValue = await sendJson('GET', ACCOUNT_HELPER_URL + '/fiat');
+        // TODO: Rewrite this function call on by demand or even create separate action.
+
+        const tokenPrices = await fetchTokenPrices();
```
MaximusHaximus, line 19: **Not a blocker for this PR**, but these can run in parallel.

packages/frontend/src/redux/slices/tokenFiatValues/index.js
```diff
@@ -4,14 +4,38 @@ import { createSelector } from 'reselect';
...
+        const last_updated_at = Date.now() / 1000; 
+
+        const tokenFiatValues = Object.keys(tokenPrices).reduce((acc, curr) => {
```
MaximusHaximus, line 22: **Not a blocker for this PR**, but this can just be replaced with `_. mapValues` for clarity & brevity here. Using `Object.keys()` here makes our reducer function impure which is undesirable; if we're going to reference variables in the parent function scope, then doing a for loop or using forEach would be more idiomatic.

MaximusHaximus (changes requested):
Just a few minor things to be fixed before this can land -- let's get this landed today!

Added notes about things we should address but that don't block this landing; we should address those in a follow-up PR.

packages/frontend/src/redux/actions/staking.js
esaminu: I don't think `token_id` guarantees the same uniqueness but the smart contract does use it for [claiming](https://github.com/referencedev/staking-farm/blob/master/staking-farm/src/farm.rs#L272)

MaximusHaximus: I think we've got a problem here :(

Looking at the implementation of `get_unclaimed_rewards`, it looks like we're going to have a problem with our displayed balances and a confusing/frustrating behaviour for users who are claiming balances when there is more than one farm for the same `token_id` defined.

**At a high-level:**
1. Rewards are accrued 'per `session`' (epoch/time period).
2. The _currently in-progress `session`_ reward amounts need to be tracked _per `farm_id`_, because each farm may have a different return/APY, even though multiple farms in the same staking contract **may** accrue rewards for the same `token_id` (FT contract).
3. At the end of each `session` (epoch), rewards are 'distributed' to each person who has staked tokens.  At this time, reward balances _per `farm_id`_ are combined/added into the balance of tokens that have accrued in _prior_ `session`s (epochs) , which are _not_ tracked per `farm_id`, they are tracked _per `token_id`_ (FT contract)
4. `get_unclaimed_rewards()` takes a `farm_id` as an argument, because it returns an unclaimed balance that includes tokens that are accruing _during the current `session` (epoch)_.  This method looks up the rewards that are _still being accrued in the current epoch_, then it _adds_ the amount of rewards that were accrued in _prior_ epochs to that amount, and returns the result.

**(4) above causes multiple problems for us, because it returns _one_ balance for unclaimed rewards, but it is a _sum_ of _farm-specific_ rewards plus _token-specific_ rewards, and there is no way for us as the caller to know which portion of the balance is which.**

To illustrate the problem, imagine we have a staking pool with 2 farms, but _both farms generate rewards of the same token types_

**Farm 1:** AURORA tokens @ 10%
**Farm 2:** AURORA tokens @ 20%

If we have, across the last few months, accrued a total of **10 AURORA tokens**, then that balance is tracked as a single value for _prior_ epochs, even though that balance is actually a combination of rewards that were distributed by _two different `farm_id`s_
If, in the current epoch, I have accrued 0.01 AURORA tokens for Farm 1 and 0.02 AURORA tokens for Farm 2, those balances are stored _per farm_id_ for the current epoch.

**Imagine that I want to show the claimable balance for a user _per farm_ (since the user must claim their balance _per farm_ and I can only retrieve the unclaimed balance _per farm_.**
If I want to show total claimable balance for a user, naively I would map across each farm and call `get_unclaimed_rewards(farm_id)`.  I can only _claim_ tokens _per farm_id_, so it is natural for me to render a separate `claim` button _per farm_, along with a separate claimable balance next to each button.  However, if I show the amount claimable _per farm_, it will be very confusing to the user because both called to `get_unclaimed_rewards` included both the per `token_id` balance _and_ the per `farm_id` balance!
With the naive implementation we would see something misleading like this:
**Farm 1:**: 10.01 <CLAIM>
**Farm 2:**: 10.02 <CLAIM>

When the user sees this UI, they will think they can claim a total of **20.03 tokens**.

But when the user clicks <CLAIM> for _either_ farm, they will receive either **10.01** or **10.02** tokens, and the claimable balances will refresh to show 0.00 for the farm they clicked <CLAIM> next to, and _just the balance accrued in the current epoch_ for the farm that they _didn't_ click <CLAIM> next to.

**Now imagine that we _don't_ want to show a separate balance next to _each farm_, and would like to just show a total claimable balance per _token_id_ instead, with a single `claim` button next to the token id.**
Unfortunately, we have no way of telling _what the total claimable balance is; if we sum up the balances returned by calling `get_unclaimed_rewards()` for each `farm_id`, the sum will be 20.02, but if we choose any one farm, the balance will not be accurate because it will not include the rewards accrued in the current epoch for any of the _other_ farms for a given token.

**Without being able to tell what portion of the claimable rewards balance is tracked _per farm id_ vs. what amount is being tracked _per token_id_, our balances will be inaccurate, and claiming tokens will be a confusing experience for users who have accrued tokens for multiple farms that farmed the same token_id**

I also have a suspicion that this mismatch will make it impossible to accurately tell what the APY has been per-farm when multiple farms exist for the same token, since the balance of previously rewarded tokens isn't tracked per-farm...

cc: @agileurbanite @askulikov @esaminu

packages/frontend/src/services/StakingFarmContracts.js
```diff
@@ -0,0 +1,49 @@
...
+            { account_id, farm_id }
+        );
+    }
+
+    static getFarmListWithUnclaimedRewards = async ({ contractName, account_id, from_index, limit }) => {
+        const farms = await StakingFarmContracts.getFarms({ contractName, from_index, limit });
+        return Promise.all(
+            farms.map(({ token_id, farm_id, active }) =>
+                StakingFarmContracts.getUnclaimedRewards({ contractName, account_id, farm_id })
+                .catch(() => '0')
```
MaximusHaximus, line 39: Re-adding comment on this line since the last few refactors hide the last one -- we shouldn't be silently returning 0 balance for failed calls to `getUnclaimedRewards` here since it'll be misleading to the user and could cause some strange/inconsistent UI behaviors.

MaximusHaximus (approved):
:+1: nice work @esaminu @askulikov ! I'm merging this in so it'll be live on [testnet](https://wallet.testnet.near.org/) and [staging](https://wallet.staging.near.org/) in a few minutes!



This comment is something we have to solve:
https://github.com/near/near-wallet/pull/2507#discussion_r833881765

packages/frontend/src/redux/actions/staking.js
agileurbanite: @MaximusHaximus qq what if we hid the reward amount but just unblocked users from claiming the reward?  Showing the claim amount might be something we can address in a separate PR that might require some changes on upstream dependencies.  Thoughts?

askulikov: @MaximusHaximus I haven't learnt rust yet, but I can't ignore this. So sorry, If this question is a dummy. I checked the source code of the contract. It seems like rewards are calculated by farm and I wasn't able to find anything, that takes the rewards by token_id in the contract. Just for double check. Maybe I'm looking into the wrong place.
[get_unclaimed_rewards](https://github.com/referencedev/staking-farm/blob/a39240d34fe97b21453a012a0abfb1d1afbc8775/staking-farm/src/views.rs#L144)
[internal_unclaimed_balance](https://github.com/referencedev/staking-farm/blob/a39240d34fe97b21453a012a0abfb1d1afbc8775/staking-farm/src/farm.rs#L121)

By the way, claiming by `token_id` looks like a problem, I agree.

MaximusHaximus: Updated my comment to correct the behaviour the user will see right now if we render a separate `claim` button and balance _per farm_.

Since this issue only impacts staking farms contracts where more than one farm is distributing rewards for the same token, we can land this and get the claim button live, then iterate on fixing the underlying issue that keeps us from displaying accurate claimable token balances, so I've merged this PR for now.

**Brainstorming, I came up with a few options that might work here:**
#### 1. We could refactor the contract's state storage to track rewards accrued in prior epochs by `farm_id` rather than by `token_id`. **I believe this would be the 'most correct' solution.**
- This may be _necessary_ if we want to show accurate APY _per farm_ rather than _per token ID_, or if we want to show a UI where each farm shows the amount of a particular token that were rewarded _by that farm_.  IMO this is the only 'correct' solution _if we need to know how much of the current claimable tokens for a particular `token_id` came from each farm that farms that `token_id`_.  If we don't need to show per-farm rows in the UI with accurate data, but instead want to show per-_token_ rewards, then this seems like an unnecessary change.
#### 2. Since `claim()` takes `token_id` as an argument, we could modify `get_unclaimed_rewards` to take the `token_id` as an argument also, instead of `farm_id`. **I think this is a decent compromise/middle-ground**
- If we did this, the method could be modified to return the _sum_ of the accrued rewards from _prior_ epochs/sessions for the passed `token_id` _plus_ the sum of the rewards being accrued in the _current_ epoch for every farm that yields that `token_id`.  This would give us a number that correctly matches what the user would receive if they clicked a `claim` button  for that token.  We probably would also want to stop rendering a separate row _per farm_ in the UI at all at this point, and just show rewards _per token_.
#### 3. We could make `claim()` _only_ transfer tokens that have been accrued in _prior sessions_, and stop including the amounts being accrued in the current session in the result provided by `get_unclaimed_rewards`. **This feels like a sub-optimal work-around.**
- I _think_ this breaks one of the more attractive properties of a staking farm, which is the ability to withdraw accrued FT rewards without needing to be tied to epoch/sessions starting/ending.  Also, `get_unclaimed_rewards` at that point should absolutely be modified to take `token_id` as an argument since there would be no farm-specific balances included in the value we returned.

cc: @ilblackdragon

PR #2508, opened 2022-02-28
fix: bug when viewing earned token rewards on refresh

esaminu:
- Fixes a bug on #2398 where token rewards are always 0 when navigating directly to the validator page or refreshing
- Use `farm_id` instead of `token_id` as key when rendering farm list


MaximusHaximus (approved):
:+1: LGTM -- let's get this merged so we can do a final review on #2507 :)

PR #2509, opened 2022-02-28
WAL-115 Remove option for enabling 2FA with SMS

morgsmccauley:
This PR removes the option for enabling 2FA with SMS and it's associated code. As we will most likely expand out these options again in future the concept of "options" still remains in code.

Removing SMS leaves us with a single option making the radio options a bit redundant but we've decided to push this out as is to stop new SMS 2FA accounts ASAP. An updated UI will be addressed in a follow up PR.

image

### DIFF
--- a/packages/frontend/src/translations/en.global.json
+++ b/packages/frontend/src/translations/en.global.json
@@ -1411,6 +1411,7 @@
         "desc": "Protect your account by requiring confirmation via SMS or email when authorizing transactions.",
         "disable": {
             "desc": "Keep in mind that transactions won't have to be confirmed with 2FA once it's disabled.",
+            "phoneDesc": "The SMS option for two-factor authentication is being deprecated, and cannot be re-enbaled.",
             "disable": "Disable 2FA",
             "keep": "No, keep 2FA",
             "title": "Are you sure you want to disable 2FA?"
--- a/packages/frontend/src/components/accounts/two_factor/EnableTwoFactor.js
+++ b/packages/frontend/src/components/accounts/two_factor/EnableTwoFactor.js
@@ -1,7 +1,6 @@
 import { utils } from 'near-api-js';
 import React, { useState, useEffect } from 'react';
 import { Translate } from 'react-localize-redux';
-import PhoneInput, { isValidPhoneNumber } from 'react-phone-number-input';
 import { useSelector, useDispatch } from 'react-redux';
 import styled from 'styled-components';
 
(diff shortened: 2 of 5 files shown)


Patrick1904:
Looks great @morgsmccauley! Thanks for doing a proper cleanup too.

One nit: What do you think about just removing the radio button (`:before`) and extra margin (`35px`) on the left side? We do a similar treatment on the `/staking` page when there's no additional accounts to choose from. 

Screen Shot 2022-02-28 at 2 46 45 PM

Patrick1904 (approved):
This is awesome! 💯  Thanks again for the proper cleanup, it helps keep the code clean 💎

Patrick1904 (commented):
@morgsmccauley, just taking a second look, and realizing that `<EnterVerificationCode/>` component could be used outside of 2FA verification, so it may be best to keep it as-is since we might use phone/SMS as an option to verify an account, recovery method, or something else at some point.

For example, the `<SetupRecoverymethod/>` component has the phone option in there, but it's currently only disabled with a flag. However, if we merge this, phone wouldn't work should we enable the flag again.

Any thoughts on above?

morgsmccauley:
Good catch @Patrick1904, definitely worth keeping that round 😅. I'll revert the changes made to `<EnterVerificationCode />`.

morgsmccauley:
@Patrick1904 Also found that the `twoFactor.phone` translation is still required in [`<TwoFactorAuth />`](https://github.com/near/near-wallet/blob/master/packages/frontend/src/components/profile/two_factor/TwoFactorAuth.js#L88) so reverted the translation changes too 🤦🏽‍♂️.

andy-haynes:
@MaximusHaximus @agileurbanite did we decide whether we wanted to include a warning when disabling existing SMS 2FA? Should that be part of this work?

agileurbanite:
> @MaximusHaximus @agileurbanite did we decide whether we wanted to include a warning when disabling existing SMS 2FA? Should that be part of this work?

Yes, @morgsmccauley can you also include an option so that if a user had enabled 2FA SMS in the past and then disabled it, they would be provided a warning / alert dialog that would state something like: "Once disabled, you will not be able to enable 2FA with SMS again.  You will be able to enable 2FA with email again but SMS as an option for 2FA will no longer be available."  Something like that, maybe @corwinharrell has a more eloquent and concise way he can recommend.

morgsmccauley:
Updated disable UI as per discussion with @corwinharrell

image

packages/frontend/src/translations/en.global.json
```diff
@@ -1411,6 +1411,7 @@
         "desc": "Protect your account by requiring confirmation via SMS or email when authorizing transactions.",
         "disable": {
             "desc": "Keep in mind that transactions won't have to be confirmed with 2FA once it's disabled.",
+            "phoneDesc": "The SMS option for two-factor authentication is being deprecated, and cannot be re-enbaled.",
```
morgsmccauley, line 1414: How to go about updating other languages?

Patrick1904: The UI will display the string in English while there's no corresponding translation available, which I think is reasonable here since we want to prioritize shipping this change

Patrick1904 (approved):
Looks good! Thanks for adding the deprecation disclaimer

agileurbanite:
@morgsmccauley @Patrick1904 will this impact our other users who already have 2FA with SMS enabled?

morgsmccauley:
@agileurbanite only difference for existing users with 2FA with SMS will be the updated UI when attempting to disable as displayed here https://github.com/near/near-wallet/pull/2509#issuecomment-1055931755

agileurbanite (approved):
LGTM

PR #2510, opened 2022-02-28
fix: Invalid transfer url

andy-haynes:
This PR applies the fix from `stable` in which callback URLs used in signing transactions are considered invalid when they contain JavaScript code.


MaximusHaximus (approved):
:+1: thx for following up on this!

PR #2511, opened 2022-03-01
feat: add new claim confirm modal

esaminu:
This PR:
* Adds a new modal to claim token rewards for #2392 based on the design in https://github.com/near/near-wallet/issues/2392#issuecomment-1044901440 
* Hides farms if they are inactive + have no unclaimed rewards


corwinharrell:
@esaminu can you add instructions to test?

esaminu:
@corwinharrell @Patrick1904 I added instructions to test on https://github.com/near/near-wallet/pull/2512#issue-1155994650. You should start accruing rewards pretty quickly after which the "Claim" button will be enabled and you'll see the modal 😄

Patrick1904:
Thanks @esaminu - this is looking great! Tested the flow and noticed a few minor UI things that would be nice if we addressed:

The text in the tooltip is gray but I believe it should be white:ish. Also, the validator name cuts off a bit weird? Should normally be ellipsis (if needed).

Screen Shot 2022-03-02 at 5 16 22 PM

I tried hovering on the tooltip on the stake confirm screen but it doesn't show up for some reason

Screen Shot 2022-03-02 at 5 16 34 PM

The mix and inconsistency with info button and exclamation mark button size/color and rest of color scheme looks a bit weird here for the custom tokens, is this intended @corwinharrell? I couldn't find a Figma for this part of the flow to verify

Should we add the 'price unavailable' here instead - the red feels like something is wrong?
Screen Shot 2022-03-02 at 5 35 39 PM

Screen Shot 2022-03-02 at 5 25 48 PM

stefanopepe:
@esaminu a few reviews, following what Patrick shared:
1. open a separate bug issue for the validator name ellipsis
2. specify in the tooltip `This token isn't whitelisted by ref-finance. Make sure you research this token!` - specifying the source of this information
3. automatic claim tokens on withdraw event (tbd the copy in the explanation)

corwinharrell (changes requested):
- [x] Remove APY. This was just tacked on and it wasn't considered how to elegantly add this to the UI without breaking it. This should be a separate issue.
- [x] The USD amount doesn't need a color change. Having this displayed in red doesn't communicate anything to the user other than that something is wrong, which is untrue in this case.
- [x] Hiding warnings behind tooltips is contradictory and poor UX. If there is something worth warning the user about, it should be displayed clearly at the point at which it is relevant, not buried behind a tooltip. Let's remove these tooltips altogether. Instead, the user should see the warning message displayed in a banner within the claim dialogue. If we determine that there is risk involved _before_ the point at which they will claim their tokens, we should use a similar treatment there instead.
- [ ] Rewards are not displayed on the main staking dashboard. These should be displayed here, but _only_ if there are rewards to claim.
- [ ] There is extra padding at the bottom of the validator card within the dialogue.
- [ ] The validator name within the dialogue does not need to be a link to the explorer, especially in this context. If we are going to link to the explorer, we should do that from the validator profile, but this is outside of the scope of this work.
- [ ] The button loading state when claiming reads "Sending..."
- [ ] There is a missing divider below the section for Rewards Earned
- [ ] The success state is a divergence from the regular success states, and has an awkward translation. We should make these as consistent as possible and make a pass at copy (Just use your best judgement here)
- [ ] After claiming, the tokens appear to still be available to claim from the same validator (ie no change in state)

One other note is that it feels odd to have a claim button for these farming tokens but not for the NEAR rewards elsewhere in the UI. I anticipate users asking about this, but perhaps this is something we can address in a redesign.

esaminu:
@corwinharrell @Patrick1904 @stefanopepe Made UI changes after discussion with @corwinharrell including removal of APY and rendering the whitelist per validator. I opened #2520 to fix the UI issues with validator box in a follow up release as discussed. I also added the auto claim behavior on withdrawal of NEAR from a validator.

I rebased #2512 so please give it a whirl on the updated preview

@MaximusHaximus would also appreciate if you look this and #2508 over so we can merge these branches into each other into master.

MaximusHaximus (approved):
:+1: LGTM, let's get this merged :)

PR #2513, opened 2022-03-02
fix(ledger): use the new action reference

andy-haynes:
This PR fixes the reference to `checkAndHideLedgerModal`


MaximusHaximus (approved):
<3 good catch -- LGTM! :shipit:

PR #2514, opened 2022-03-04
feat: NFT transfer

andy-haynes:
This PR is an extension of #2205 since I haven't figured out how to incorporate my changes into the original PR.


packages/frontend/src/components/nft/NFTDetailWrapper.js
```diff
@@ -0,0 +1,49 @@
...
+    const { contractId: contractName, tokenId } = match.params;
+    const accountId = useSelector(selectAccountId);
+    const { balanceAvailable: nearBalance} = useSelector(selectBalance);
+    const nft = useSelector((state) => selectTokenForAccountForContractForTokenId(state, {
+        accountId,
+        contractName,
+        tokenId,
+    }));
+
+    const transferredNft = useSelector((state) => selectTransferredTokenForContractForTokenId(state, {
```
andy-haynes, line 26: This is the biggest change from the original PR - rather than fetching the NFT metadata by `tokenId` (without regards to `accountId`) I've added two new selectors to fetch the NFT state:
- `selectTokenForAccountForContractForTokenId` gets the NFT metadata from state initialized when `fetchNFTs` was dispatched.
- `selectTransferredTokenForContractForTokenId` gets NFT metadata from a new key in the redux state that holds the set of tokens the user has transferred. Once transferred, the NFT no longer exists in state as an owned token since it is no longer owned by the user. Rather than refetch the NFT after transferring I've opted to simply add the transferred token to the new state. Refetching the token is still an option however and only requires a small change to the new `transferNFT` action/reducer.

packages/frontend/src/components/nft/NFTDetailWrapper.js
```diff
@@ -0,0 +1,49 @@
...
+        contractName,
+        tokenId,
+    }));
+
+    const dispatch = useDispatch();
+    const { fetchNFTs } = nftActions;
+
+    useEffect(() => {
+        if (accountId && !nft) {
+            dispatch(fetchNFTs({ accountId, contractName }));
```
andy-haynes, line 36: This should only happen if the user refreshes the page but it's still probably worth adding an action to fetch just the token with `tokenId` rather than fetching all for this `accountId` + `contractName` pair.

packages/frontend/src/redux/slices/nft/index.js
```diff
@@ -74,6 +78,17 @@ const fetchOwnedNFTsForContract = createAsyncThunk(
     }
 );
 
+const transferNFT = createAsyncThunk(
+    `${SLICE_NAME}/transferNFT`,
+    async ({ accountId, contractName, nft }, { dispatch, getState }) => {
+        debugLog('THUNK/transferNFT');
+
+        const { actions: { transferToken } } = nftSlice;
+        dispatch(transferToken({ contractName, nft }));
```
andy-haynes, line 87: We can refetch the token here rather than cache the transferred token object, but there isn't much interaction after the transfer so it may not be strictly necessary.

packages/frontend/src/components/nft/NFTDetailWrapper.js
andy-haynes: Added new action

Patrick1904 (approved):
LGTM! Thank you for the last miles effort @andy-haynes 🙏

Patrick1904:
Seems there might be something off with the Mintbase NFTs, because they don't render on the NFT details page at all. I just bought this one with our test account and it's a blank page: https://deploy-preview-2514--near-wallet-staging.netlify.app/nft-detail/gambiarra.mintbase1.near/6556

Suspecting mintbase img format/links might be different and we're not taking that into consideration on the NFT details page?

andy-haynes:
> Seems there might be something off with the Mintbase NFTs, because they don't render on the NFT details page at all. I just bought this one with our test account and it's a blank page: https://deploy-preview-2514--near-wallet-staging.netlify.app/nft-detail/gambiarra.mintbase1.near/6556
> 
> Suspecting mintbase img format/links might be different and we're not taking that into consideration on the NFT details page?

Still looking into this, but the problem seems to be a disparity in the shape and completeness of data returned from different method calls. Regardless of the root cause we'll need a way to handle this disparity in the new detail component as Mintbase NFTs are currently breaking navigation, requiring a reload after trying to view the token detail.

Patrick1904:
> > Seems there might be something off with the Mintbase NFTs, because they don't render on the NFT details page at all. I just bought this one with our test account and it's a blank page: https://deploy-preview-2514--near-wallet-staging.netlify.app/nft-detail/gambiarra.mintbase1.near/6556
> > Suspecting mintbase img format/links might be different and we're not taking that into consideration on the NFT details page?
> 
> Still looking into this, but the problem seems to be a disparity in the shape and completeness of data returned from different method calls. Regardless of the root cause we'll need a way to handle this disparity in the new detail component as Mintbase NFTs are currently breaking navigation, requiring a reload after trying to view the token detail.

I see, it seems like the [`selectTokenForAccountForContractForTokenId`](https://github.com/near/near-wallet/pull/2514/files#diff-3a6ae3f7eedde449bfed180fce29c98a7f9b17bcfb0cfcaf5848ac1ff6b26db1R19-R23) selector is returning an empty object for Mintbase NFTs, and that's why nothing renders on the page. For some reason the navigation is still intact for me, no reason to refresh the page

Patrick1904 (approved):
Thank you for working through the edge cases 🙏 🥇

PR #2517, opened 2022-03-08
style(lint): nonexistent named imports error

andy-haynes:
This PR adds a new eslint dependency and rules to verify the existence of named imports, ensuring that we no longer reference named imports not exported from the target module.

### DIFF
--- a/packages/frontend/.eslintrc.js
+++ b/packages/frontend/.eslintrc.js
@@ -18,6 +18,7 @@ module.exports = {
         'no-extra-boolean-cast':'off',
         'no-extra-semi':'off',
         'no-irregular-whitespace':'off',
+        'import/named': ['error', 'always'],
         'import/order': [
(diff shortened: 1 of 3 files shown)


packages/frontend/.eslintrc.js
```diff
@@ -53,6 +54,9 @@ module.exports = {
             'message': 'Please use a selector for any state accesses within useSelector'
         }]
     },
+    settings: {
+        'import/ignore': ['src/config/*'],
```
andy-haynes, line 58: `config.js` produced lots of false positives, I'm assuming because individual keys aren't exported directly but do get destructured in our `import`s. I didn't see a way to get around this besides whitelisting the module but if there's a better way to do this I think it's worth checking out.

Patrick1904 (approved):
LGTM 👍

PR #2523, opened 2022-03-15
WAL-177 Render pending 2FA actions in confirmation modal

morgsmccauley:
With the migration to Twilio Verify 2FA SMS messages will no longer show the transaction details. This PR adds UI to the 2FA verification modal to render the same details which would be present in the SMS.

This is essentially a 1:1 replacement of the existing messaging _except_ for setting up SMS 2FA. I've left that out as we have disabled that flow.

158480330-ac9cad39-93b0-4a9e-b6b6-8f61c639e772
158478845-8dd09ae8-6be8-45fd-80da-a432a89bc5f2
158478597-d0cfa5df-f71f-4141-8c8c-23b4bb807611

### DIFF
--- a/packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
+++ b/packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
@@ -22,6 +22,39 @@ const Form = styled.form`
     width: 100%;
 `;
 
+const StyledBannerContainer = styled.div`
+    &&& {
+        div:first-child {
+            margin: 0px;
+            width: 100%;
+            @media (max-width: 450px) {
+                 border-radius: 4px;
+            }
+        }
+
+        pre {
+            margin: 0px;
+        }
+    }
+`;
+
+const ActionDetailsBanner = ({ multisigRequest }) => {
+    const isAddingFullAccessKey = multisigRequest.actions.some(({ enum: type, permission }) => type === 'addKey' && !permission);
+
+    return  (
+        <StyledBannerContainer>
+            <AlertBanner theme={isAddingFullAccessKey ? 'warning' : 'light-blue'}>
+                {getTranslationsFromMultisigRequest(multisigRequest).map(({ id, data }, index, arr) => (
+                    <React.Fragment key={JSON.stringify(data)}>
+                        <Translate id={id} data={data} />
+                        {index !== arr.length - 1 && <br />}
+                    </React.Fragment>
+                ))}
+            </AlertBanner>
+        </StyledBannerContainer>
+    );
+};
+
 const TwoFactorVerifyModal = ({ open, onClose }) => {
 
     const [method, setMethod] = useState();
@@ -89,15 +122,7 @@ const TwoFactorVerifyModal = ({ open, onClose }) => {
             <h2 className='title'><Translate id='twoFactor.verify.title'/></h2>
             <p className='font-bw'><Translate id='twoFactor.verify.desc'/></p>
             <p className='color-black font-bw' style={{ marginTop: '-10px', fontWeight: '500', height: '19px'}}>{method && method.detail}</p>
-            {multisigRequest && (
-                <AlertBanner theme='alert'>
-                        <>
-                            <Translate id={id} data={data} />
-                            {index !== arr.length - 1 && <br />}
-                        </>
-                    ))}
-                </AlertBanner>
-            )}
+            {multisigRequest && <ActionDetailsBanner multisigRequest={multisigRequest} />}
             <Form onSubmit={(e) => {handleVerifyCode(); e.preventDefault();}}>
                 <TwoFactorVerifyInput
                     code={code}
--- a/packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
+++ b/packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
@@ -91,10 +91,9 @@ const TwoFactorVerifyModal = ({ open, onClose }) => {
             <p className='color-black font-bw' style={{ marginTop: '-10px', fontWeight: '500', height: '19px'}}>{method && method.detail}</p>
             {multisigRequest && (
                 <AlertBanner theme='alert'>
-                    {getTranslationsFromMultisigRequest(multisigRequest).map(({ id, data }) => (
                         <>
                             <Translate id={id} data={data} />
-                            <br />
+                            {index !== arr.length - 1 && <br />}
                         </>
                     ))}
                 </AlertBanner>
--- a/packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
+++ b/packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
@@ -180,7 +180,7 @@ const TwoFactorVerifyModal = ({ open, onClose }) => {
             <p className='font-bw'><Translate id='twoFactor.verify.desc'/></p>
             <p className='color-black font-bw' style={{ marginTop: '-10px', fontWeight: '500', height: '19px'}}>{method && method.detail}</p>
             {multisigRequest && (
-                <AlertBanner theme="alert">
+                <AlertBanner theme='alert'>
                     {getTranslationsFromMultisigRequest(multisigRequest).map(({ id, data }) => (
                         <>
                             <Translate id={id} data={data} />
--- a/packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
+++ b/packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
@@ -39,7 +39,7 @@ const StyledBannerContainer = styled.div`
 `;
 
 const ActionDetailsBanner = ({ multisigRequest }) => {
-    const isAddingFullAccessKey = multisigRequest.actions.some(({ enum: type, permission }) => type === 'addKey' && !permission);
+    const isAddingFullAccessKey = multisigRequest.actions.some(({ enum: type, [type]: action }) => type === 'addKey' && !action.accessKey.permission);
 
     return  (
         <StyledBannerContainer>
--- a/packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
+++ b/packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
@@ -27,6 +27,7 @@ const StyledBannerContainer = styled.div`
         div:first-child {
             margin: 0px;
             width: 100%;
+            word-break: break-word;
             @media (max-width: 450px) {
                  border-radius: 4px;
             }
--- a/packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
+++ b/packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
@@ -1,5 +1,3 @@
-import BN from 'bn.js';
-import { utils } from 'near-api-js';
 import React, { useEffect, useState } from 'react';
 import { Translate } from 'react-localize-redux';
 import { useDispatch, useSelector } from 'react-redux';
@@ -9,6 +7,7 @@ import { Mixpanel } from '../../../mixpanel/index';
 import { resendTwoFactor, get2faMethod, getMultisigRequest } from '../../../redux/actions/account';
 import { selectAccountSlice, selectAccountMultisigRequest } from '../../../redux/slices/account';
 import { selectActionsPending, selectStatusSlice } from '../../../redux/slices/status';
+import getTranslationsFromMultisigRequest from '../../../utils/getTranslationsFromMultisigRequest';
 import { WalletError } from '../../../utils/walletError';
 import AlertBanner from '../../common/AlertBanner';
 import FormButton from '../../common/FormButton';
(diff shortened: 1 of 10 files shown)


packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
```diff
@@ -1,32 +1,125 @@
...
+                            publicKey: new PublicKey({ keyType: KeyType.ED25519, data: action.publicKey.data }).toString(),
+                        }
+                    };
+                case 'functionCall':
+                    return {
+                        id: 'twoFactor.action.functionCall',
+                        data: {
+                            receiverId,
+                            methodName: action.methodName,
+                            deposit: formatNear(action.deposit, 4),
```
Patrick1904, line 89: Think we can cut the `4` here since the second param isn't used in the `formatNear` function

packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
```diff
@@ -85,6 +179,16 @@ const TwoFactorVerifyModal = ({ open, onClose }) => {
             <h2 className='title'><Translate id='twoFactor.verify.title'/></h2>
             <p className='font-bw'><Translate id='twoFactor.verify.desc'/></p>
             <p className='color-black font-bw' style={{ marginTop: '-10px', fontWeight: '500', height: '19px'}}>{method && method.detail}</p>
+            {multisigRequest && (
+                <AlertBanner theme="alert">
```
Patrick1904, line 183: nit: single quote for consistency :)

packages/frontend/src/translations/en.global.json
```diff
@@ -1439,6 +1439,16 @@
...
+        },
+        "action": {
+            "addKey": {
+                "limited": "Adding key ${publicKey} limited to call ${methodNames} on ${receiverId} and spend up to ${allowance} on gas",
+                "full": "WARNING: Entering the code below will authorize full access to your NEAR account: ${accountId}. If you did not initiate this action, please DO NOT continue.<br /><br />This should only be done if you are adding a new seed phrase to your account. In all other cases, this is very dangerous.<br /><br /> The public key you are adding is: ${publicKey} ",
+            },
+            "deleteKey": "Deleting key ${publicKey}",
+            "functionCall": "Calling method: ${methodName} in contract: ${receiverId} with amount ${deposit} and with args ${args}",
+            "stake": "Staking: ${deposit} to validator: ${receiverId}",
+            "transfer": "Transferring: ${deposit}Ⓝ to ${receiverId}"
```
Patrick1904, line 1451: nit: whenever you see the `Ⓝ` symbol in legacy copy for NEAR currency, let's switch it out for just `NEAR` instead :) We used to have inconsistent usage of the currency symbol but have settled on just using `NEAR`

packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
```diff
@@ -1,32 +1,125 @@
...
     align-items: center;
     width: 100%;
 `;
 
+const formatNear = (amount) => {
+    try {
+        return formatNearAmount(amount, 4);
+    } catch {
+        // Near amount may be stored as hex rather than decimal
+        return formatNearAmount(new BN(amount, 16).toString(), 4);
```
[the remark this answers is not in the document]
morgsmccauley: No I wasn't able to figure out why 😢 . I _suspect_ it has something to do with `near-api-js` because when I query the contract directly with `near-cli` the values come back in the right format. This also seems specific to `functionCall` as I didn't encounter this with any of the other actions.

I'm planning to do some more digging to hopefully get to the bottom of this 😄 

```js
// from `near-api-js`
{
    "functionCall": {
        "methodName": "nft_buy",
        "args": [/* buffer */],
        "gas": "5af3107a4000",
        "deposit": "01a9e7b50fe5dae8400000"
    },
    "enum": "functionCall"
}
```
```js
// from `near-cli`
{
  actions: [
    {
      type: 'FunctionCall',
      method_name: 'nft_buy',
      args: 'eyJ0b2tlbl9zZXJpZXNfaWQiOiIyNjYiLCJyZWNlaXZlcl9pZCI6InNwdWQudGVzdG5ldCJ9',
      deposit: '2011280000000000000000000',
      gas: '100000000000000'
    }
  ]
}
```

packages/frontend/src/translations/en.global.json
morgsmccauley: Sounds good! 👍🏽

packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
morgsmccauley: Woops, left behind from when this was `formatNearAmount` 😅

packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
```diff
@@ -1,32 +1,125 @@
...
+    }
+};
+
+const getTranslationsFromMultisigRequest = ({ actions, receiverId, accountId }) => {
```
Patrick1904, line 45: This is a great function and feels like it might be used in other places outside of the two factor modal. Should we put it in a helper, etc?

packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
```diff
@@ -1,32 +1,125 @@
...
+                        id: 'twoFactor.action.transfer',
+                        data:{
+                            receiverId,
+                            deposit: formatNearAmount(action.deposit, 4),
+                        }
+                    };
+                case 'stake':
+                    return {
+                        id: 'twoFactor.action.stake',
+                        data: {
```
morgsmccauley, line 104: This matches what we have in `near-contract-helper` but doesn't line up with what i've found in the [docs](https://nomicon.io/RuntimeSpec/Actions#stakeaction). Staking is actually achieved through calling a function on the validator contract (`functionCall`), it seems this is more related to validators themselves, is it actually possible to trigger this action from the wallet?

packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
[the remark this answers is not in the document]
Patrick1904: @MaximusHaximus do you have any idea why we'd see hex for `functionCall` from `near-api-js` ❓

packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
morgsmccauley: Yep definitely makes sense as a helper! Do you think we should keep the function as is, i.e. still returning the `<Translation />` data, or should it perhaps be more generic by returning the `multisigRequest` in a more usable format? If it was more generic we'd most likely need to map it again here which could still bloat this component file 🤔

packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
Patrick1904: Hmm good catch. I think `stake` might be referring to actual staking, not delegating to a validator that then stakes on behalf of the user. Actual staking is not possible in the wallet at the moment.

packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
morgsmccauley: It's also worth noting that `near-contract-helper` also covers messaging around setting up 2FA for the first time (`requestId === -1`) but since we've disabled that flow for SMS i've left that out.

Also, initial verification is done within `<EnterVerificationCode />` rather than this modal so it wouldn't be needed in this component, but may be needed within this new helper if we chose to eventually add something similar to `<EnterVerificationCode />` 🤔

Patrick1904: I like it the way it is because it'd be easy to just plug it into any UI and yes additional bloat seems not worth it IMO 👍

morgsmccauley:
> There's some built in mobile screen styling in the `AlertBanner` that causes issues inside this modal but we can fix that with the style added below
> 
> Screen Shot 2022-03-15 at 12 11 34 PM
> 
> Screen Shot 2022-03-15 at 12 12 00 PM

Good catch @Patrick1904! I'll update :)

packages/frontend/src/utils/twoFactor.js
```diff
@@ -48,6 +48,20 @@ export class TwoFactor extends Account2FA {
         });
     }
 
+    async getMultisigRequest() {
+        const request = this.getRequest();
+        const { receiver_id } = await this.viewFunction(
```
morgsmccauley, line 53: ideally we could just use the result of `getRequest()` but it doesn't contain the `receiver_id` so we need to fetch this from the contract.

packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
```diff
@@ -20,13 +22,45 @@ const Form = styled.form`
     width: 100%;
 `;
 
+const StyledBannerContainer = styled.div`
+    &&& {
+        div:first-child {
+            margin: 0px;
+            border-radius: 0;
+            width: 100%;
```
Patrick1904, line 30: nit: we want to actually override the border-radius change on 450px screen width so maybe something like this:

```suggestion
            margin: 0px;
            width: 100%;
            @media (max-width: 450px) {
                 border-radius: 4px;
            }
```

packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
```diff
@@ -20,13 +22,45 @@ const Form = styled.form`
...
+        }
+    }
+`;
+
+const ActionDetailsBanner = ({ multisigRequest }) => {
+    const isAddingFullAccessKey = multisigRequest.actions.some(({ enum: type, permission }) => type === 'addKey' && !permission);
+
+    return  (
+        <StyledBannerContainer>
+            <AlertBanner theme={isAddingFullAccessKey ? 'warning' : 'light-blue'}>
```
Patrick1904, line 44: 👍 nice!

Patrick1904:
Any idea why there would be such an inconsistency between stake and unstake? Seems to think that I'm unstaking 0N but actually 0.001N

Screen Shot 2022-03-15 at 3 38 58 PM

Screen Shot 2022-03-15 at 3 37 36 PM

packages/frontend/src/translations/en.global.json
```diff
@@ -1439,6 +1439,16 @@
             "resending": "Sending...",
             "resent": "Code sent!",
             "title": "Enter Two-Factor Verification Code"
+        },
+        "action": {
+            "addKey": {
+                "limited": "Adding key ${publicKey} limited to call ${methodNames} on ${receiverId} and spend up to ${allowance} NEAR on gas",
+                "full": "WARNING: Entering the code below will authorize full access to your NEAR account: ${accountId}. If you did not initiate this action, please DO NOT continue.<br /><br />This should only be done if you are adding a new seed phrase to your account. In all other cases, this is very dangerous.<br /><br /> The public key you are adding is: ${publicKey} ",
```
Patrick1904, line 1446: nit: Just to avoid confusion here, I think we want to say `if you are adding a new full access key to your account` instead of `if you are adding a new seed phrase to your account` since passphrase is just one way to add a full access key.

packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
Patrick1904: Yea probs worth removing since it won't be used here if we add support anyway

morgsmccauley:
> Any idea why there would be such an inconsistency between stake and unstake? Seems to think that I'm unstaking 0N but actually 0.001N
> 
> I can see that the 2FA email I'm getting has the same issue so almost certainly an existing bug?
> 
> Screen Shot 2022-03-15 at 3 38 58 PM
> 
> Screen Shot 2022-03-15 at 3 37 36 PM

@Patrick1904 Not entirely sure.. looks to be an implementation detail of the staking contract itself. I'd guess, and am probably wrong, that staking requires a `deposit` so the funds can be transferred and unstaking only requires we specify the `amount` (in args) to be released with nothing physically being transferred (hence the `0` deposit).

You have highlighted another issue with this PR though - i'm not parsing out `amount` or `deposit` from the arguments and formatting as Near 😄 .

packages/frontend/src/translations/en.global.json
morgsmccauley: Good call, i'll update 👍🏽 . I also found this confusing since this message will pop up when disabling 2FA 🤔

packages/frontend/src/utils/getTranslationsFromMultisigRequest.js
```diff
@@ -0,0 +1,93 @@
...
+        ...args,
+        ...(args.amount && { amount: formatNear(args.amount) }),
+        ...(args.deposit && { deposit: formatNear(args.deposit )}),
+    };
+
+    return JSON.stringify(parsedArgs, null, 2);
+};
+
+export default function getTranslationsFromMultisigRequest({ actions, receiverId, accountId }) {
+    const fullAccessKeyAction = actions.find(({ enum: type, permission }) => type === 'addKey' && !permission);
```
Patrick1904, line 38: I'm getting full access key warning modal when recovering my account secured with email link + 2FA despite no full access key being added. The 2FA email that I received from [contract-helper](https://github.com/near/near-contract-helper/blob/436762c54a519b400692c17d2ead910a1a589fdf/middleware/2fa.js#L143) is showing the correct information though.

Looks like it's `accessKey.permission`

Screen Shot 2022-03-16 at 12 50 42 PM

Patrick1904:
> You have highlighted another issue with this PR though - i'm not parsing out amount or deposit from the arguments and formatting as NEAR like we are currently 😄 .

Parsing this improved the UX quite a lot - nice one 👍

Since contract-helper is not parsing the NEAR amount, can you please make the same parsing in a follow-up PR for contract-helper too? That way there is alignment between what the user sees in the email and on the modal.

packages/frontend/src/utils/getTranslationsFromMultisigRequest.js
morgsmccauley: Good catch,  I wasn't even parsing out the action itself either. Fixed 👍🏽

packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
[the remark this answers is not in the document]
morgsmccauley: I'm still looking in to this, but for now i've opted to fetch and use the multisig request from the contract (`this.viewFunction()`) directly rather than request stored by `near-api-js` (`this.getRequest()`) as it doesn't store the Near values in hex.

https://github.com/near/near-wallet/pull/2523/files#diff-0592ce7ed5478b8cd8fd764e47824321d8a8bae9b77b2616da331efc69a02a94R51

packages/frontend/src/utils/getTranslationsFromMultisigRequest.js
```diff
@@ -0,0 +1,76 @@
+import { formatNearAmount } from 'near-api-js/lib/utils/format';
+
+const NEAR_FRACTIONAL_DIGITS = 4;
+
+const formatNear = (amount) => formatNearAmount(amount, NEAR_FRACTIONAL_DIGITS);
```
Patrick1904, line 5: I think we can re-use the existing NEAR formatting function in the balance [helper](https://github.com/near/near-wallet/blob/master/packages/frontend/src/components/common/balance/helpers.js) file too

morgsmccauley: Sounds good 👍🏽

Patrick1904 (approved):
This is great! 💯  Thank you 🙏

PR #2524, opened 2022-03-16
chore: Remove all code related to coin-op 'active account' state

Patrick1904:
https://nearinc.atlassian.net/browse/WAL-195

The coin-op ‘active account’ feature that gated parts of the wallet until the user had deposited funds is no longer relevant and should be removed from the wallet codebase. This cleanup helps organize the code for the ‘zero balance’ wallet support.


andy-haynes (approved):
Nice work, loving all the 🟥

PR #2527, opened 2022-03-16
WAL-110  - Lock / Unlock components and workflows for encrypting accessing local storage private key

gutsyphilip:
In this PR:

- Implements idle timer for wallet
- Implements set /change password UI
- Implements timeout modal
- Implements Browser password settings
- Implements wallet unlock UI
- Implements password reset UI
- Implements new seedphrase input UI

image

### DIFF
--- a/packages/frontend/src/components/accounts/SetupSeedPhrase.js
+++ b/packages/frontend/src/components/accounts/SetupSeedPhrase.js
@@ -265,7 +265,6 @@ class SetupSeedPhrase extends Component {
                                             onSubmit={this.handleOnSubmit}
                                             isLinkDrop={parseFundingOptions(this.props.location.search) !== null}
                                             hasSeedPhraseRecovery={hasSeedPhraseRecovery}
-                                            seedPhrase={seedPhrase}
                                         />
                                     </form>
                                 </Container>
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/common/Button.js
```diff
@@ -2,6 +2,26 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import styled from 'styled-components';
 
+const backgroundColorMap = {
```
Patrick1904, line 5: Nice mapping of colors! I'm curious why you decided to use this button component vs the regular `FormButton`? Seems it would be good if we used the same button as much as possible for consistency sake, and less components to maintain.

Patrick1904:
@gutsyphilip is this PR ready for review? If not, can you please mark it as in progress / draft?

packages/frontend/src/components/common/Button.js
gutsyphilip: Definitely. Will switch to the FormButton.

Patrick1904:
> > @gutsyphilip is this PR ready for review? If not, can you please mark it as in progress / draft?
> 
> It was ready for review @Patrick1904. But I did the UI for the passphrase task (WAL-37) in here because it was part of the design. I'm now working to revert that and then create a separate PR for that(tied to that ticket particularly).

Thanks for clarifying. I converted this to draft for now so please convert it back when it's ready and request a review :)

PR #2528, opened 2022-03-16
Created field for showing collected farm rewards

askulikov:


marcinbodnar:
@askulikov Hey, do you have maybe a testnet account that I can use for testing this particular feature?

askulikov:
@marcinbodnar Hi! Sorry for the late reply. Somehow missed the notification.
Yes, you can take this one. It already has staked tokens in needed validators.
`basket cement cross nasty tornado pill weapon exact bullet casual nerve cattle`

packages/frontend/src/redux/crossStateSelectors/selectCollectedAvailableForClaimData.js
```diff
@@ -0,0 +1,81 @@
...
+                return [...acc, ...validatorsFarmData[validatorId].farmRewards];
+            }, [])
+            .filter((farm) => farm.active && +farm.balance > 0);
+
+        const collectedBalance = filteredFarms.reduce((acc, farm) => {
+            const tokenId = farm.token_id;
+            if (acc[tokenId]) {
+                return {
+                    ...acc,
+                    [tokenId]: new BN(acc[tokenId].balance)
```
marcinbodnar, line 33: The `collectedBalance` is not well calculated here. The bug is whit this line, it should be `acc[tokenId]`.

Also, it seems like the `reduce` can be simplified here, for example, something like this:
```js
const collectedBalance = filteredFarms.reduce((acc, farm) => ({
    ...acc,
    [farm.token_id]: new BN(acc[farm.token_id])
        .add(new BN(farm.balance))
        .toString()
}), {});
```

packages/frontend/src/redux/crossStateSelectors/selectCollectedAvailableForClaimData.js
```diff
@@ -0,0 +1,81 @@
...
+    selectTokenWhiteList,
+} from '../slices/tokenFiatValues';
+import { selectAllContractMetadata } from '../slices/tokens';
+
+const collectFarmingData = (...args) => {
+    try {
+        if (args.some((el) => isEmpty(el))) return [];
+
+        const [
+            validatorsFarmData,
```
marcinbodnar, line 17: Seems like there is a problem with `validatorsFarmData`, if you check with your account, you would see that when you first enter the Staking page there is only one farm - when you enter the third validator and get back the second farm is there.

marcinbodnar (changes requested):
Thank you for this PR and good job @askulikov 
Few comments below.

packages/frontend/src/components/staking/StakingContainer.js
```diff
@@ -203,6 +205,14 @@ export function StakingContainer({ history, match }) {
         dispatch(handleUpdateCurrent(accountId));
     };
 
+    useEffect(() => {
+        if (!accountId || !validators.length) return;
+        
+        validators
+            .filter((validator) => validator.version === FARMING_VALIDATOR_VERSION && validator.active)
+            .forEach((validator) => dispatch(getValidatorFarmData(validator, accountId)));
+    }, [accountId, validators]);
```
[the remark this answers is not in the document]
askulikov: I prefer this solution for now. It is not very beautiful, I don't like it this way. But batching all actions in one dispatch may cause problems in the future. Now we can inspect each fetch action and we don't really care if some of them fail for some reason (as it fails now for incorrect validators). I thought a lot about this part of the code. And finally decided to let it as it is. 
It might be that perfect is the enemy of the good in this case.

packages/frontend/src/components/staking/components/Staking.js
```diff
@@ -3,6 +3,7 @@ import React from 'react';
 import { Translate } from 'react-localize-redux';
```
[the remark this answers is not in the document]
askulikov: Found a reason for it. the forEach-dispatch above calculates APY just for active validators. 
The component ValidatorBox is checking, and its initial APY is null - it returns loading dots. I will add the check `isActive` in this component. As the validator is inactive - his APY should be 0. So nice catch!

packages/frontend/src/redux/crossStateSelectors/selectCollectedAvailableForClaimData.js
```diff
@@ -0,0 +1,81 @@
...
+
+import { selectValidatorsFarmData } from '../slices/staking';
+import {
+    selectTokensFiatValueUSD,
+    selectTokenWhiteList,
+} from '../slices/tokenFiatValues';
+import { selectAllContractMetadata } from '../slices/tokens';
+
+const collectFarmingData = (...args) => {
+    try {
```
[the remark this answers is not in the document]
askulikov: Not really. But the catch block will return `[]` if something will go wrong. Just in case. Probably, if the app somewhere in the future will ride to TS - this can be removed.

packages/frontend/src/redux/crossStateSelectors/selectCollectedAvailableForClaimData.js
```diff
@@ -0,0 +1,81 @@
...
+            validatorsFarmData,
+            contractMetadataByContractId,
+            tokenFiatValues,
+            tokenWhitelist,
+        ] = args;
+        const filteredFarms = Object.keys(validatorsFarmData)
+            .reduce((acc, validatorId) => {
+                return [...acc, ...validatorsFarmData[validatorId].farmRewards];
+            }, [])
+            .filter((farm) => farm.active && +farm.balance > 0);
```
askulikov, line 26:
```suggestion
            .filter((farm) => +farm.balance > 0);
```

@marcinbodnar Could you check this please?

packages/frontend/src/redux/crossStateSelectors/selectCollectedAvailableForClaimData.js
askulikov: Yes, you are right. I changed the code and use your suggestion. But there is a problem with filtering farms. When I worked with it, the second farm was active. And now all tokens, that have been framed from it, do not sum to an active farm. Maybe
we have to delete this filtering on active farms. Check the suggestion, please https://github.com/near/near-wallet/pull/2528#discussion_r842957198

packages/frontend/src/redux/crossStateSelectors/selectCollectedAvailableForClaimData.js
askulikov: I'm not sure about the `business` part of this solution. Don't the validators must withdraw farmed tokens when become inactive?

packages/frontend/src/components/staking/StakingContainer.js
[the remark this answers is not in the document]
marcinbodnar: What I mean is to create a thunk that will handle dispatching all actions one by one like it is now. Thunk will:
 - select validators and filter validators using a selector
 - select `accountId` using selector
 - then it will dispatch `getValidatorFarmData` actions one by one like it is now.
It's just that all of it will happen in a separate thunk instead of the component hook.

Generally, we are trying to extract this type of functionality into redux thunks.

Also, we don't need to check if `!validators.length` is true because if so the forEach will do nothing.

packages/frontend/src/redux/crossStateSelectors/selectCollectedAvailableForClaimData.js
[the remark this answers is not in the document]
marcinbodnar: I see, so it can be removed. Using try/catch makes sense in situations when you are expecting that error might appear, for example, the request will throw an error, and you need to handle this error.

Here, you are basically filtering data, so if you expect that there might be a problem with particular data, you should handle it there explicitly.

packages/frontend/src/components/staking/StakingContainer.js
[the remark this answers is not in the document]
marcinbodnar: But that's not a blocker. Just a suggestion.

packages/frontend/src/redux/crossStateSelectors/selectCollectedAvailableForClaimData.js
marcinbodnar: It seems like the problem is with `validatorsFarmData` - if you enter the staking page there is only one validator, if you click on the third validator and get back, the second validator appears in `validatorsFarmData` and the total balance is calculated properly.

marcinbodnar:
@askulikov I will pass this review to @esaminu . Thank you for all the changes.

packages/frontend/src/components/staking/StakingContainer.js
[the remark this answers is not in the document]
esaminu: @askulikov The `validator.active` condition here should be removed. This is currently not accumulating unclaimed rewards from inactive validators due to this. To repro you can confirm that ILLU rewards from domanodes validator are currently not added in askulikov.testnet

packages/frontend/src/components/staking/components/Staking.js
```diff
@@ -105,6 +106,24 @@ export default function Staking({
                     />
                 </>
                 : null}
+            {!loading && collectedFarmData.map((tokenData, i) => {
+                return (
```
esaminu, line 110: I think we should move the filtering for zero balance rewards here for reusability purposes and in keeping with the other `BalanceBox` logic on this component

packages/frontend/src/redux/crossStateSelectors/selectCollectedAvailableForClaimData.js
esaminu: We should not be filtering out inactive farms or validators. Users may still have unclaimed rewards on expired farms for example.

packages/frontend/src/redux/crossStateSelectors/selectCollectedAvailableForClaimData.js
```diff
@@ -0,0 +1,71 @@
...
+            validatorsFarmData,
+            contractMetadataByContractId,
+            tokenFiatValues,
+            tokenWhitelist,
+        } = args;
+        const filteredFarms = Object.keys(validatorsFarmData)
+            .reduce((acc, validatorId) => {
+                return [...acc, ...validatorsFarmData[validatorId].farmRewards];
+            }, [])
+            .filter((farm) => farm.active && +farm.balance > 0);
```
esaminu, line 26: If we move the filtering logic to the component, this whole block can simply be reduced to:
```suggestion
           const filteredFarms = Object.values(validatorsFarmData).reduce((acc, {farmRewards}) => [...acc, ...farmRewards], [])
```

esaminu (changes requested):
Looking good on first pass. Left minor comments on unclaimed rewards for inactive farms and validators and token filtering logic 👍

esaminu:
@askulikov Please resolve completed points 😄

packages/frontend/src/redux/crossStateSelectors/selectCollectedAvailableForClaimData.js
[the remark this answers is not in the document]
askulikov: Resolve this as during my last commits catch blocks were mostly required for returning empty data.
I can't assume that something will crash here, but I think that this may save us at least from some invalid data on devnet and testned.

packages/frontend/src/redux/crossStateSelectors/selectCollectedAvailableForClaimData.js
askulikov: Resolved this thread. The fix will be in other threads

packages/frontend/src/components/staking/components/Staking.js
askulikov: Done

packages/frontend/src/redux/crossStateSelectors/selectCollectedAvailableForClaimData.js
askulikov: Done. Nice suggestion, I like it. Thank you!

packages/frontend/src/redux/crossStateSelectors/selectCollectedAvailableForClaimData.js
askulikov: Done

esaminu (approved):

esaminu:
Thanks @askulikov !

PR #2529, opened 2022-03-16
WAL-160 fix(recovery): provide publicKey to security code validation endpoints

andy-haynes:
This PR updates calls to the recovery method endpoints responsible for validating security codes to include the public key associated with the recovery method being validated. The public key will be used to uniquely identify the recovery method document in DynamoDB and so should be provided when calling any endpoint responsible for querying or mutating recovery methods.

### DIFF
--- a/packages/frontend/src/routes/SetupRecoveryImplicitAccountWrapper.js
+++ b/packages/frontend/src/routes/SetupRecoveryImplicitAccountWrapper.js
@@ -20,6 +20,7 @@ export function SetupRecoveryImplicitAccountWrapper() {
     const [verifyingEmailCode, setVerifyingEmailCode] = useState(false);
     const [resendingEmailCode, setResendingEmailCode] = useState(false);
     const [seedPhrasePublicKey, setSeedPhrasePublicKey] = useState(null);
+    const [isInitializingRecoveryLink, setIsInitializingRecoveryLink] = useState(false);
 
     const handleInititalizeEmailRecoveryLink = async () => {
         const passPhrase = await wallet.initializeRecoveryMethodNewImplicitAccount({ kind: 'email', detail: email });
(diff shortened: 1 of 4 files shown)


packages/frontend/src/routes/SetupRecoveryImplicitAccountWrapper.js
```diff
@@ -60,7 +62,7 @@ export function SetupRecoveryImplicitAccountWrapper() {
                 Mixpanel.track('SR Verify email code');
                 try {
                     setVerifyingEmailCode(true);
-                    await wallet.validateSecurityCodeNewImplicitAccount(implicitAccountId, { kind: 'email', detail: email }, securityCode);
+                    await wallet.validateSecurityCodeNewImplicitAccount(implicitAccountId, { kind: 'email', detail: email }, securityCode, seedPhrasePublicKey);
```
andy-haynes, line 65: Since `handleInititalizeEmailRecoveryLink` does not get `await`'d and `showVerifyEmailCode` is set to `true` before it's called, I believe there's a race condition in which this can be called without a value for `publicKey`. Would it be safe to start `await`ing that method call and only setting `showVerifyEmailCode` once it's completed its API call?

Patrick1904: Yea I don't see any issues with awaiting the API call before setting `showVerifyEmailCode` to `true` - I suspect it'll be quick too. The only thing to consider is potentially adding a disabled/loading state for the main button in case the call takes some time to finish since right now we send the user to the 'verify' step right away.

andy-haynes: @Patrick1904 good call, I've added a flag to the component to disable the button during link initialization. Let me know if you think there's a more appropriate way to convey the loading.

Patrick1904: Sweet! nit: can we also please add `sending={isInitializingRecoveryLink}` as a prop to the `FormButton` so that we get that `Sending...` copy for the button?

Patrick1904 (approved):
Thank you! 🙏

PR #2530, opened 2022-03-17
Use createEntityAdapter for transactions reducer

marcinbodnar:
Introducing solution: Using entity adapter to perform CRUD operations on the transactions array.


marcinbodnar:
Thank you @judith-Near

PR #2531, opened 2022-03-18
fix: Disallow switching account on onboarding/recovery pages

Patrick1904:
Fixes:
1. Disallow switching account on onboarding/recovery pages.
2. Makes sure to pull wallet recovery methods and disable the 'continue' button for already enabled recovery methods (regression).

To test:
1. While logged in to an account, navigate to any of the onboarding or/and recovery setup pages:
```
    /create
    /set-recovery
    /setup-seed-phrase
    /recover-account
    /recover-seed-phrase
    /sign-in-ledger
    /fund-create-account
    /verify-account
    /initial-deposit
    /setup-ledger
```
3. Expect the main navigation dropdown to disable option to switch account
4. Navigate to a different page and expect option to switch account to be enabled again

### DIFF
--- a/packages/frontend/src/components/Routing.js
+++ b/packages/frontend/src/components/Routing.js
@@ -179,41 +179,36 @@ class Routing extends Component {
 
         this.props.setActiveLanguage(activeLang);
         // this.addTranslationsForActiveLanguage(defaultLanguage)
-
-        const {
-            refreshAccount,
-            history,
-            handleRedirectUrl,
-            handleClearUrl,
-            handleClearAlert,
-            handleFlowLimitation,
-            router
-        } = this.props;
-
-        history.listen(async (location) => {
-            handleRedirectUrl(router.location);
-            handleClearUrl();
-            handleFlowLimitation();
-            if (!WALLET_CREATE_NEW_ACCOUNT_FLOW_URLS.find((path) => router.location.pathname.indexOf(path) > -1)) {
-                await refreshAccount(true);
-            }
-
-            handleClearAlert();
-        });
     }
 
     componentDidMount = async () => {
         const {
             refreshAccount,
             handleRefreshUrl,
+            history,
+            handleRedirectUrl,
+            handleClearUrl,
             router,
             fetchTokenFiatValues,
+            handleClearAlert,
+            handleFlowLimitation
         } = this.props;
 
         fetchTokenFiatValues();
         this.startPollingTokenFiatValue();
         handleRefreshUrl(router);
         refreshAccount();
+
+        history.listen(async () => {
+            handleRedirectUrl(this.props.router.location);
+            handleClearUrl();
+            if (!WALLET_CREATE_NEW_ACCOUNT_FLOW_URLS.find((path) => this.props.router.location.pathname.indexOf(path) > -1)) {
+                await refreshAccount(true);
+            }
+
+            handleClearAlert();
+            handleFlowLimitation();
+        });
     }
 
     componentDidUpdate(prevProps) {
--- a/packages/frontend/src/components/Routing.js
+++ b/packages/frontend/src/components/Routing.js
@@ -18,6 +18,7 @@ import { Mixpanel } from '../mixpanel/index';
 import * as accountActions from '../redux/actions/account';
 import { handleClearAlert } from '../redux/reducers/status';
 import { selectAccountSlice } from '../redux/slices/account';
+import { actions as flowLimitationActions } from '../redux/slices/flowLimitation';
 import { actions as tokenFiatValueActions } from '../redux/slices/tokenFiatValues';
 import { CreateImplicitAccountWrapper } from '../routes/CreateImplicitAccountWrapper';
 import { ImportAccountWithLinkWrapper } from '../routes/ImportAccountWithLinkWrapper';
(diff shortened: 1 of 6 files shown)


andy-haynes (changes requested):
Looks good, question inline around `isNewAccount`.

I did notice some minor odd behavior when switching accounts while adding a phrase to an existing account:
1. click `Enable` on Passphrase for target account
2. switch accounts to an account that already has a phrase (form button is now disabled)
3. switch back to the original target account

After switching back to the original account, the form button is rendered as enabled but briefly becomes disabled before being re-enabled. Not a huge deal if it always ends up enabled but it may be indicative of a weird intermediate state that could cause problems. If you only switch between accounts _without_ phrases then the button correctly remains enabled.

corwinharrell:
@andy-haynes @Patrick1904 saw this and wanted to ask:

What is the benefit of allowing users to switch accounts during the recovery flow? Why not just treat recovery as an action that occurs "outside" of the context of the account UI? If this was the case, a user could exit the recovery flow via a "Cancel" action at any point. Exiting would return them to the dashboard of their active account or the homepage if they don't have one.

andy-haynes:
> Why not just treat recovery as an action that occurs "outside" of the context of the account UI?

I think this would be the ideal approach, but I see it as part of a more fundamental change. With the account selector being part of the top-level SPA component, I'd imagine that every screen in the wallet needs to gracefully handle account switching regardless of whether it makes sense for the actions taken on that screen.

To me it would make more sense to restrict account switching to a handful of screens in which no account-specific actions are being taken, but I don't know how difficult it would be to make this distinction.

corwinharrell:
@andy-haynes gotcha agree that perhaps it's a larger change so we're on the same page 👍 

> To me it would make more sense to restrict account switching to a handful of screens in which no account-specific actions are being taken, but I don't know how difficult it would be to make this distinction.

I agree with this and think that things like the wallet nav and account switcher don't need to persist during things like onboarding and recovery. Only when you are actively using the wallet portion of the app are these things relevant.

packages/frontend/src/components/Routing.js
```diff
@@ -176,36 +179,43 @@ class Routing extends Component {
...
         } = this.props;
 
-        fetchTokenFiatValues();
-        this.startPollingTokenFiatValue();
-        handleRefreshUrl(router);
-        refreshAccount();
-
-        history.listen(async () => {
-            handleRedirectUrl(this.props.router.location);
+        history.listen(async (location) => {
```
Patrick1904, line 193: Moved `history.listen` into the constructor so that it works as intended

Patrick1904:
@andy-haynes @corwinharrell I changed the behavior as we agreed upon and updated the [PR description ](https://github.com/near/near-wallet/pull/2531#issue-1173084457)

packages/frontend/src/redux/slices/flowLimitation/index.js
```diff
@@ -21,11 +22,18 @@ const initialState = {
 const handleFlowLimitation = createAsyncThunk(
     `${SLICE_NAME}/handleFlowLimitation`,
     async (_, thunkAPI) => {
+        const { actions: { setFlowLimitation } } = flowLimitationSlice;
         const { dispatch, getState } = thunkAPI;
         const { pathname } = getState().router.location;
-        const { redirect_url } = getState().account.url;
-        const { actions: { setFlowLimitation } } = flowLimitationSlice;
 
+        if (WALLET_CREATE_NEW_ACCOUNT_FLOW_URLS.some((url) => pathname.includes(url))) {
```
andy-haynes, line 29: This covers 10 URLs, are we pretty confident that there are no valid use cases for switching accounts in any of them? In particular I'm wondering about those involved with funding an account, where a user might want to switch to the funding account to send funds to the new account. Would this affect that flow at all?

andy-haynes (changes requested):
This is looking great, just have a couple questions around some potential impact.

packages/frontend/src/redux/slices/flowLimitation/index.js
```diff
@@ -21,11 +22,18 @@ const initialState = {
...
+            // Disallow account switching on account creation/recovery pages
+            dispatch(setFlowLimitation({ subMenu: true }));
+        } else {
+            dispatch(setFlowLimitation({ subMenu: false }));
+        }
```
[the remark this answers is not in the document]
Patrick1904: Love it

packages/frontend/src/components/accounts/SetupSeedPhrase.js
```diff
@@ -51,32 +50,30 @@ class SetupSeedPhrase extends Component {
     }
 
     componentDidMount = async () => {
-        const { accountId, fetchRecoveryMethods } = this.props;
         this.refreshData();
-
-        if (accountId === wallet.accountId) {
```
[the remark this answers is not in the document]
Patrick1904: We're fetching recovery methods here now instead whenever the account from URL exists :) : https://github.com/near/near-wallet/pull/2531/files#diff-c583b238dac7e863a5b4e152206ee3f69b72513aa9efbaab221c05ba201957f8R64-R66

packages/frontend/src/redux/slices/flowLimitation/index.js
Patrick1904: Yep any pages where switching account would be relevant, like funding pages, have an account selector on the page 👍

andy-haynes (approved):
Great work! This will address a lot of unintuitive behavior around recovery method routes ❤️

PR #2537, opened 2022-03-21
fix(nft): restore data URI scheme rendering for NFTs

### LINKED ISSUE #2534: NFT images from URI is broken in testnet
Happened after fcf8564806e4d0321910cfa9c5a98e4c5bb3fc7b because of https://github.com/near/near-wallet/commit/fcf8564806e4d0321910cfa9c5a98e4c5bb3fc7b#r69059531.

Feature https://github.com/near/near-wallet/pull/2227 is currently disabled in testnet.

andy-haynes:
This PR adds a `buildMediaUrl` method to the NFT class to simplify the logic around extracting URI data from NFT metadata. In particular it restores the logic responsible for rendering `data:image` URIs.

Closes #2534

### DIFF
--- a/packages/frontend/src/services/NonFungibleTokens.js
+++ b/packages/frontend/src/services/NonFungibleTokens.js
@@ -93,24 +93,26 @@ export default class NonFungibleTokens {
             .map((token) => this.mapTokenMediaUrl(token, base_uri));
     }
 
-    static mapTokenMediaUrl = ({ metadata, ...token }, base_uri) => {
-        const { media } = metadata;
-        let mediaUrl;
-        if (media && !media.includes('://')) {
-            if (base_uri) {
-                mediaUrl = `${base_uri}/${media}`;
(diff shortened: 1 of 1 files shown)


willemneal:
Out of curiosity why are the static methods defined as arrow functions?

MaximusHaximus:
> > Out of curiosity why are the static methods defined as arrow functions?
> 
> Not sure what the original intention was, @MaximusHaximus might have more context.

No particular reason here; it's probable that we just didn't have any references to `this` so it was a non-issue either way, but the safest thing would be to make them all standard function definitions instead of arrow functions, so if we need to add a reference to another static method we don't at the same time also need to modify the function definition from an arrow function to a standard function.

morgsmccauley (approved):

Patrick1904 (approved):
Thank you for addressing this! 🙏

PR #2538, opened 2022-03-22
Fixing calculation of APY for farming validators

askulikov:
This branch was created for discussing and fixing the APY for farming validators.

Acceptance criteria: 

- [ ] Issue found
- [ ] Issue solved

@stefanopepe @esaminu

### DIFF
--- a/features/flags.json
+++ b/features/flags.json
@@ -57,6 +57,35 @@
       "lastEditedAt": "2022-01-31T21:13:09.948Z"
     }
   },
+  "FARMING_VALIDATOR_APY_DISPLAY": {
+    "createdBy": "esaminu",
+    "createdAt": "2022-03-09T21:26:58.971Z",
+    "development": {
+      "enabled": true,
+      "lastEditedBy": "Arsen Kulikov",
+      "lastEditedAt": "2022-03-22T20:36:21.751Z"
+    },
+    "testnet": {
+      "enabled": true,
+      "lastEditedBy": "Arsen Kulikov",
+      "lastEditedAt": "2022-03-22T20:36:21.751Z"
+    },
+    "mainnet": {
+      "enabled": true,
+      "lastEditedBy": "Arsen Kulikov",
+      "lastEditedAt": "2022-03-22T20:36:21.751Z"
+    },
+    "mainnet_STAGING": {
+      "enabled": true,
+      "lastEditedBy": "Arsen Kulikov",
+      "lastEditedAt": "2022-03-22T20:36:21.751Z"
+    },
+    "testnet_STAGING": {
+      "enabled": true,
+      "lastEditedBy": "Arsen Kulikov",
+      "lastEditedAt": "2022-03-22T20:36:21.751Z"
+    }
+  },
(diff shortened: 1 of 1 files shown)


stefanopepe:
@askulikov please open a github ticket, so we keep this thread only for code review and implementation discussion. cc @agileurbanite for the allocation of this task

askulikov:
@stefanopepe @esaminu, could you please describe the problem with APY? All calculations were done by the formulas provided earlier. Could you please check this, before we start? 

`
export const calculateAPY = (poolSummary, tokenPrices) => {
    **// Handle if there are no farms:
    const activeFarms = poolSummary?.farms;
    if (!activeFarms) return 0;**

    try {

        **if (activeFarms.some((farm) => !+tokenPrices[farm.token_id]?.usd)) return 0;**
        const totalStakedBalance = nearTo(poolSummary.total_staked_balance);

        const summaryAPY = activeFarms.reduce((acc, farm) => {
            const tokenPriceInUSD = +tokenPrices[farm.token_id].usd;
            const nearPriceInUSD = +tokenPrices[NEAR_TOKEN_ID].usd;

            const rewardsPerSecond = farm.amount / ((farm.end_date - farm.start_date) * 1e9);
            const rewardsPerSecondInUSD = rewardsPerSecond * tokenPriceInUSD;
            const totalStakedBalanceInUSD = totalStakedBalance * nearPriceInUSD;
            const farmAPY = rewardsPerSecondInUSD * SECONDS_IN_YEAR / totalStakedBalanceInUSD * 100;
            return acc + farmAPY;
        }, 0);

        return summaryAPY.toFixed(2);
    }
    catch (e) {
        console.error('Error during calculating APY', e);
        return '-';
    }
};
`

The only thing, that I see now, is that the checks and rules might be incorrect (bold code). They were written by me at the beginning and I guess we have to review and fix them:
* instead of checking the amount of farms, we have to check the amount of active farms.
* instead of checking that some of farms haven't prices, we have to filter and calculated only for those, which have prices.

Could you please review this?

MaximusHaximus:
@askulikov could you take a look [at the comment here](https://github.com/near/near-wallet/pull/2507#discussion_r833881765) and see if the issue I described there will cause problems with APY calculations per-farm? I think it will.

It seems like tracking some rewards by `token_id` and others by `farm_id` causes some problems, because the value returned by the `get_unclaimed_reward` method includes a sum of both, but rewards accrued from previous sessions could be from multiple different farms that happen to farm the same token.

askulikov:
@MaximusHaximus It seems that without changes on the contract side, there is nothing to change on the frontend side. Because we can't calculate the correct values because of the lack of the correct data. And also it will be much more clear after clarifying the rules of calculation, which is described by @stefanopepe in this [thread](https://github.com/near/near-wallet/issues/2539#issuecomment-1076798365). Correct me if I'm wrong, please. I carefully re-read your comments a few times, and it looks like there aren't any options to fix this in the current implementation. Also, I let a small comment about calculating rewards just for clarifying in the [thread](https://github.com/near/near-wallet/pull/2507#discussion_r834277397)

askulikov:
The last commit just fixes the initial conditions for selecting farms.

PR #2541, opened 2022-03-23
fix: passing data from WalletError to actionStatus state

marcinbodnar:
It one-line fix, but a few hours of tinkering and checking.

I analyzed several possible solutions. We cannot use `alert.data.data` (like is mentioned in the issue description) because it will interfere with other usages of global alerts that are proper. We cannot change `walletError` implementation because it's also used in two other cases, and it seems reasonable as it is. The only proper solution is to destructure `payload.data` into `data` prop in `actionStatus`. It's solving the problem properly.

Generally, we are planning to slowly get rid of global alerts in favor of custom error UI implemented in all functionalities. So basically every UI will have its own way to show error information.

### DIFF
--- a/packages/frontend/src/redux/reducers/status/index.js
+++ b/packages/frontend/src/redux/reducers/status/index.js
@@ -73,6 +73,7 @@ const alertReducer = (state, { error, ready, payload, meta, type }) => {
                 errorMessage: (error && payload?.message) || (type === 'SHOW_CUSTOM_ALERT' && payload.errorMessage) || undefined,
                 data: {
                     ...meta?.data,
+                    ...payload?.data,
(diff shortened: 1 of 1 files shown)


andy-haynes:
@marcinbodnar is there any easy way to tell which error throws are affected by this?

marcinbodnar:
@andy-haynes yes, you can check the `WalletError` usage. At this moment there are two cases where `data` for `WalletError` is used, but in both cases, the data is used for a different purpose than showing info in global error, and it's working properly.

andy-haynes (approved):
Good find! Hopefully we can deprecate this soon but in the meantime it will be nice to not have to think about it.

PR #2544, opened 2022-03-25
chore: Wrappify Dashboard and Navigation components

Patrick1904:
Changes:
1. Wrappify main dashboard (`Wallet.js`) and Navigation components as part of zero balance wallet support
2. Convert from class components to function components

### DIFF
--- a/packages/frontend/src/components/navigation/Navigation.js
+++ b/packages/frontend/src/components/navigation/Navigation.js
@@ -79,10 +79,10 @@ export default ({
         }
     }, [menuOpen]);
 
-    const handleSelectAccount = (accountId) => {
+    const handleSelectAccount = useCallback((accountId) => {
         selectAccount(accountId);
         setMenuOpen(false);
-    };
+    }, []);
     
     return (
         <Container id='nav-container' open={menuOpen}>
--- a/packages/frontend/src/components/navigation/Navigation.js
+++ b/packages/frontend/src/components/navigation/Navigation.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useCallback } from 'react';
 import styled from 'styled-components';
 
 import DesktopContainer from './DesktopContainer';
@@ -46,22 +46,19 @@ export default ({
         if (menuOpen) {
             document.addEventListener('keydown', handleKeyDown);
             document.addEventListener('click', handleClick);
+        } else {
+            document.removeEventListener('keydown', handleKeyDown);
+            document.removeEventListener('click', handleClick);
         }
     }, [menuOpen]);
 
-    const handleCloseMenu = () => {
-        setMenuOpen(false);
-        document.removeEventListener('keydown', handleKeyDown);
-        document.removeEventListener('click', handleClick);
-    };
-
-    const handleKeyDown = (e) => {
+    const handleKeyDown = useCallback((e) => {
         if (e.keyCode === 27) {
-            handleCloseMenu();
+            setMenuOpen(false);
         }
-    };
+    }, []);
 
-    const handleClick = (e) => {
+    const handleClick = useCallback((e) => {
         const desktopMenu = document.getElementById('desktop-menu');
         const mobileMenu = document.getElementById('mobile-menu');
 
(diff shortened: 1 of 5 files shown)


packages/frontend/src/routes/WalletWrapper.js
```diff
@@ -0,0 +1,75 @@
...
+import { selectLinkdropAmount, actions as linkdropActions } from '../redux/slices/linkdrop';
+import { selectTokensWithMetadataForAccountId, actions as nftActions } from '../redux/slices/nft';
+import { actions as tokensActions, selectTokensLoading } from '../redux/slices/tokens';
+
+const { fetchNFTs } = nftActions;
+const { fetchTokens } = tokensActions;
+const { setLinkdropAmount } = linkdropActions;
+const { setCreateFromImplicitSuccess, setCreateCustomName } = createFromImplicitActions;
+
+export function WalletWrapper({
```
MaximusHaximus, line 19: Rock on, good stuff <3

packages/frontend/src/routes/WalletWrapper.js
```diff
@@ -0,0 +1,75 @@
...
+    tab,
+    setTab
+}) {
+
+    const accountId = useSelector((state) => selectAccountId(state));
+    const balance = useSelector((state) => selectBalance(state));
```
MaximusHaximus, line 25: :nitpick: since these aren't parameterized selectors
```suggestion
    const accountId = useSelector(selectAccountId);
    const balance = useSelector(selectBalance);
```

packages/frontend/src/routes/WalletWrapper.js
```diff
@@ -0,0 +1,75 @@
...
+    }, [accountId]);
+
+    useEffect(() => {
+        if (!accountId) {
+            return;
+        }
+
+        dispatch(fetchNFTs({ accountId }));
+        dispatch(fetchTokens({ accountId }));
+    }, [accountId]);
```
MaximusHaximus, line 50: Any reason not to combine these into a single effect?

```suggestion
    useEffect(() => {
        if (accountId) {
            Mixpanel.identify(Mixpanel.get_distinct_id());
            Mixpanel.people.set({ relogin_date: new Date().toString() });
            
            dispatch(fetchNFTs({ accountId }));
            dispatch(fetchTokens({ accountId }));
        }
    }, [accountId]);
```

packages/frontend/src/routes/WalletWrapper.js
```diff
@@ -0,0 +1,75 @@
...
+            accountId={accountId}
+            balance={balance}
+            linkdropAmount={linkdropAmount}
+            createFromImplicitSuccess={createFromImplicitSuccess}
+            createCustomName={createCustomName}
+            fungibleTokensList={fungibleTokensList}
+            tokensLoader={tokensLoader}
+            availableAccounts={availableAccounts}
+            sortedNFTs={sortedNFTs}
+            handleCloseLinkdropModal={
```
MaximusHaximus, line 65: Wrap in `useCallback()` for a stable function reference here so we don't pass a new function identity to Wallet component on every render

MaximusHaximus (changes requested):
Nice work! One step closer to hooks-errywhere and container components :+1: Just a few things added in-line

packages/frontend/src/components/navigation/Navigation.js
```diff
@@ -36,114 +29,85 @@ const Container = styled.div`
...
-    flowLimitationSubMenu: selectFlowLimitationSubMenu(state)
-});
-
-const mapDispatchToProps = {
-    switchAccount,
-    getAvailableAccountsBalance,
-    getAccountBalance
+    }, [menuOpen]);
+
+    const handleSelectAccount = (accountId) => {
```
MaximusHaximus, line 82: :nit: this should also be wrapped in `useCallback()`

MaximusHaximus (approved):
:+1: lookin' good -- just one nitpicky thing that isn't a blocker, but would be good for completeness.

packages/frontend/src/components/navigation/Navigation.js
Patrick1904: Added :)

PR #2549, opened 2022-03-25
fix: add conditional claim all on withdraw

### LINKED ISSUE #2548: Unable to withdraw funds from staking (after release period)
**Problem**
When trying to withdraw the funds from a validator (stacking) I'm unable to go on with the tx (there's no error feedback)

**Expected Behavior**
I should be able to withdraw my funds after the the pending release period

**Steps to reproduce**

1. Go to the [Near Wallet stacking](https://wallet.near.org/staking/) page
2. Select a validator to which you've previously clicked `Unstake` and waited the 4 epoch period
3. Click `Withdraw`
4. Nothing happens... No visual feedback

```json
{
  "block_height": 62126595,
  "error": "wasm execution failed with error: FunctionCallError(MethodResolveError(MethodNotFound))",
  "logs": []
}
```

esaminu:
Only claim all tokens on withdrawal if the validator is a staking farm validator.

Closes #2548

### DIFF
--- a/packages/frontend/src/components/staking/components/Validator.js
+++ b/packages/frontend/src/components/staking/components/Validator.js
@@ -98,25 +98,28 @@ export default function Validator({
         setShowClaimTokenFarmRewardsModal(true);
     };
 
+    const isFarmingValidator = validator?.version === FARMING_VALIDATOR_VERSION;
+
     const handleStakeAction = async () => {
         if (showConfirmModal && !loading) {
-            await StakingFarmContracts.getFarmListWithUnclaimedRewards({
-                contractName: validator.contract.contractId,
-                account_id: currentAccountId,
-                from_index: 0,
-                limit: 300,
-            }).then((res) => 
-                Promise.all([
-                    (res || [])
-                    .filter(({balance}) => !new BN(balance).isZero())
-                    .map(({token_id}) => dispatch(claimFarmRewards(validator.accountId, token_id)))
-                ])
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/staking/components/Validator.js
```diff
@@ -98,25 +98,28 @@ export default function Validator({
...
-                limit: 300,
-            }).then((res) => 
-                Promise.all([
-                    (res || [])
-                    .filter(({balance}) => !new BN(balance).isZero())
-                    .map(({token_id}) => dispatch(claimFarmRewards(validator.accountId, token_id)))
-                ])
-            );
+            if (isFarmingValidator) {
+                await StakingFarmContracts.getFarmListWithUnclaimedRewards({
```
MaximusHaximus, line 106: Not a blocker for this, but we should fix this up in a follow-up -- since `claim()` takes `token_id` as an argument, we only need to call this once per `token_id` that _any_ farm on the contract yields.

MaximusHaximus (approved):
:+1: thanks for the speedy fix @esaminu  :)

packages/frontend/src/components/staking/components/Validator.js
esaminu: Added to WAL-213

zcstarr:
haven't tested this yet on the prod side but 🎉  thanks all for speedy response!

PR #2553, opened 2022-03-28
Refactor redux state to be keyed by accountId

marcinbodnar:
This is the proposition of implementing redux state three to be keyed by accountId.

The main goal was to create an additional layer of functionality that will handle keying account-related state by accountId without the additional complication of the redux store usage within the application.

The main functionalities are:
 - `injectReducer` method added to the `store` that is using `replaceReducer` to inject new reducer in case a new account is being created or recovered
 - separation of main and account reducers - main reducers are top lever reducers shared between accounts or used for the application purposes
 - `setupAccountReducer` function that is used to check existing accounts and prepare reducers
 - main selectors that are used to help simplify usage of the new state three.

Also, I reorganized the redux folder structure to make it more straightforward, and simple.


packages/frontend/src/index.js
```diff
@@ -20,6 +20,10 @@ const history = createBrowserHistory();
 
 export const store = createStore(createRootReducer(history), createMiddleware(history));
 
+store.injectReducer = () => {
+    store.replaceReducer(createRootReducer(history));
```
andy-haynes, line 24: This feels a little heavy for just to key part of the redux state by account ID. Couple questions:
1. If I change accounts and then change back, does all the state for the original account have to be re-fetched?
2. What happens if we don't call `injectReducer`? Does the account-specific state get updated?
3. Could we get around needing to do this if we had top-level redux state for the currently active account?

andy-haynes (changes requested):
This is a great start - just some questions around organization and `injectReducer` 👍

packages/frontend/src/redux/createReducers/combinedAccountReducers.js
```diff
@@ -0,0 +1,5 @@
+import tokensSlice from '../slices/tokens';
```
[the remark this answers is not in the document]
marcinbodnar: My idea was to keep the whole functionality regarding keying by `accountId` as a separate layer of functionality, so while developing you don't have to worry about it or know how it works. 

For example, if someone would like to add a new reducer, he just opens `combinedAccountReducers.js` or `combinedMainReducers.js` (like before) and he doesn't have to worry about `setupAccountReducer` at all. If we add it inline `setupAccountReducer` function it will be much less intuitive.

So It's a separate module because the code will be organized better.

packages/frontend/src/redux/createReducers/setupAccountReducer.js
```diff
@@ -0,0 +1,26 @@
...
+import combinedAccountReducers from './combinedAccountReducers';
+
+export default (history) => {
+    const accounts = Object.keys(wallet.accounts);
+    if (!accounts) {
+        return {};
+    }
+
+    return accounts.reduce((x, accountId) => {
+        const reducer = combineReducers(combinedAccountReducers(history));
```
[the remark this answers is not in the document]
marcinbodnar: Your right, we don't need it, and it probably won't change. Done.

packages/frontend/src/redux/createReducers/setupAccountReducer.js
```diff
@@ -0,0 +1,26 @@
...
+        const inicialState = reducer(store?.getState()[accountId], {});
+
+        return ({
```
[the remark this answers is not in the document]
marcinbodnar: Right. Done.

packages/frontend/src/redux/selectors/mainSelectors/selectSliceByAccountId.js
```diff
@@ -0,0 +1,5 @@
+import { createSelector } from 'reselect';
```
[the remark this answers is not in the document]
marcinbodnar: Sure, it makes sense. Added `topLevel.js` as the file name, as it's a bit more descriptive - what do you think?

packages/frontend/src/redux/selectors/crossStateSelectors/selectNEARAsTokenWithMetadata.js
```diff
@@ -1,7 +1,7 @@
 import { createSelector } from 'reselect';
```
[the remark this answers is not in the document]
marcinbodnar: We keep selectors that belong to a specific slice in the slice file, and these are the selectors that use states from different slices. So it's more `cross slice/reducer` than `cross state`. 

There will be more of them and as we moved `mainSelectors` into one direct file I would prefer to keep them in a separate folder so it will be organized a bit, and not mixed with other files. Is that ok with you?

packages/frontend/src/index.js
marcinbodnar: Ad 1. The idea is to keep the state, so the data will be instantly visible. But it's not possible for now, as we are purging `account` reducer on the account switch.
Ad 2. `injectReducer` is used only in the case of adding or recovering an account when we need to add a new reducer because it doesn't exist in the current state. Reducers for existing accounts are created right away based on the account list in local storage, but if we want to create a new reducer we need to dynamically add it using `replaceReducer`. If we add/recover a new account and don't add a new reducer it will not exist, no data will be stored.
Ad 3. Not sure what you mean :) could you please clarify?

andy-haynes: I see, so `saveAccount` is only used for adding a new account (either creation or recovery) to LocalStorage?

Looking at the `replaceReducer` documentation I see the outlined use cases are code-splitting and HMR - neither of which are conventional production runtime use cases like this usage is. It seems excessive to replace the entire root reducer just for adding one new account. If the intention here is to populate state for a new account ID key in `byAccountId`, then why do we need to replace the entire root reducer? What are the performance and UI implications of doing this?

If this is the way we need to do it then I think this needs to be feature flagged - with my limited understanding of `replaceReducer` I think I'm missing something crucial about how this new structure is supposed to work. It would be better to iterate on it if we're not yet sure what the final implementation should look like (i.e. `TODO: figure out better way to inject reducer`).

re: my comment on the top-level reducer, I am picturing this conceptually as a map of account IDs in the redux state whose value is some `AccountState` type (which, going from the diagram in the ticket, includes `tokenBalancesNFT`, `tokenBalancesFT`, etc.). Since you can only have one active account, the active account is tracked by another key at the top level of the state. So your state for a single account would be something like:
```
{
  shared: {
    fiatValues: { usd: 10.50 },
  },
  activeAccountId: 'spaghetti.near',
  byAccountId: {
    'spaghetti.near': {
      tokenBalancesNFT: { ... },
      ...
    },
  },
}
```

Then if I add a new account, it becomes:
```
{
  shared: {
    fiatValues: { usd: 10.50 },
  },
  activeAccountId: 'lasagna.near',
  byAccountId: {
    'spaghetti.near': {
      tokenBalancesNFT: { ... },
      ...
    },
    'lasagna.near': {
      tokenBalancesNFT: {},
      ...
    },
  },
}
```

So I've only added a key (with initial value) to the `byAccountId` map and updated `activeAccountId` to the new account. Other state (`shared`) is not impacted by this because it's outside the scope of what's changed by adding a new account. In my naive model here, it seems like I should be able to dispatch an `addAccount` action that:
- adds the new account ID
- initializes the reducer state under this new account ID key in `byAccountIds` to an empty state
- fires off requests to populate this new state with values from on chain + contract helper (only if recovering, new accounts probably wouldn't have any data on chain that's not already in context)

But in this implementation we're reinitializing and replacing the entire root reducer, which would include `shared`, even though non-account specific data hasn't changed. Given my understanding this seems excessive because we know exactly what account state needed to change when we added a new account - I don't see what makes this state different than any other in the store.

I see that `injectReducer` is mentioned explicitly in the diagram on the ticket, so I'm assuming that this is the plan - I'm just having trouble reconciling my understanding with this implementation. Maybe @MaximusHaximus can weigh in here with what the intention behind replacing the root reducer is.

marcinbodnar: Thank you andy for checking these changes in detail. It's always very valuable, especially with such important changes.

> I see, so saveAccount is only used for adding a new account (either creation or recovery) to LocalStorage?

`saveAccount` is the last step of both creating and recovering an account, it's adding the account to the local storage and to Wallet `accounts` array. 

Using `saceAccount` for injecting reducer is a temporary solution (that's why the comment) because we don't want to use the `store` in the wallet.js (it's causing circular reference problems), but I'm using it there because with this PR I want to propose a concept of splitting the state tree. The next PR will be about moving `injectReducer` away from `saveAccount`, and preparing a more reasonable solution. Ideally, I would like to avoid adding `injectReducer` in a few places (for every way that we create or recover an account), I have a few concepts on how to do it - generally, I would like to have a thunk that will be dispatched as the last step of creating/recovering account (for every method) so we will be able to use it for every functionality that is common for every create/recover account method - in this case, it will be `injectReducer`, but it might be handy for future changes as well.

Btw I would love to organize the code responsible for creating/recoverying accounts in a better way at some point, now it's a bit complicated regarding the code structure.

> Looking at the replaceReducer documentation I see the outlined use cases are code-splitting and HMR - neither of which are conventional production runtime use cases like this usage is. It seems excessive to replace the entire root reducer just for adding one new account. 

So we are not replacing the whole root reducer. We are just adding the set of reducers for the new account.

> If the intention here is to populate the state for a new account ID key in byAccountId, then why do we need to replace the entire root reducer? What are the performance and UI implications of doing this?

The intention is not to populate the state, but to inject the reducers dynamically. At the start of the app, we are setting reducers for all accounts that the user has in local storage. But when the user creates/recovers an additional account, we need to add a new set of reducers (account, staking, status, etc.). The result is that every account has its own set of reducers, instead of the current solution where every reducer is holding the state for each account.

> It would be better to iterate on it if we're not yet sure what the final implementation should look like (i.e. TODO: figure out better way to inject reducer).

I mean it is the final implementation of adding reducers dynamically. The next PR will be to figure out a better way than using `saveAccount` method - because it's relevant but I want to split changes into smaller PRs. I mean it's working well as it is, and we are using a similar solution in the project, but we want to get rid of using `store` in `wallet.js` - so the next step will be PR that will base on this one, and will introduce a better way to do it. 

> re: my comment on the top-level reducer, I am picturing this conceptually as a map of account IDs in the redux state whose value is some AccountState type (which, going from the diagram in the ticket, includes tokenBalancesNFT, tokenBalancesFT, etc.). Since you can only have one active account, the active account is tracked by another key at the top level of the state.

So this is how it's implemented now, the only change is that right now we are not holding the shared reducers in the `shared` object, and we don't hold accounts in the `byAccountId` object. But besides that, that is exactly how it's working now. I can easily close all shared reducers in the `shared` object, and all accounts in the `byAccountId` object, like it is on the diagram, I'm just not sure if we need an additional object level. It will look better, the state will be more organized for sure, so we can change that - easy change.

About `activeAccountId` - I'm not sure if we need it, which account is active we know from local storage (it's set in `wallet.accountId`) - so we always know which account is active. I'm not sure if we will ever use `activeAccountId` state from Redux.

Btw my solution is inspired by this comment from Dan Abramov:
https://stackoverflow.com/questions/32968016/how-to-dynamically-load-reducers-for-code-splitting-in-a-redux-application/33044701#33044701
It's a different use case, but it's showing a concept of how to add a new reducer dynamically.

Also this PR:
https://github.com/near/near-wallet/pull/2555
is showing how it will work within an example of `tokens` reducer. It's based on this PR.
We are splitting `tokens` reducer into:
 - `tokens` that is a reducer that every account will have
 - `tokensMetadata ` that is shared.
You can test it by recovering an additional account, you will see that `tokensMetadata` is not replaced or purged.

packages/frontend/src/redux/createReducers/index.js
```diff
@@ -0,0 +1,9 @@
+import { combineReducers } from 'redux';
+
+import combinedMainReducers from './combinedMainReducers';
+import setupAccountReducer from './setupAccountReducer';
+
+export default (history, accountId) => combineReducers({
+    ...combinedMainReducers(history),
+    accountsReducer: combineReducers(setupAccountReducer(accountId))
```
andy-haynes, line 8: Let's just call this `accounts`. `accountsReducer` is redundant since all keys in the object passed to `combineReducers` have values which are reducers.

Really we just need something that doesn't conflict with `allAccounts` and makes it clear that we have account data under it.

marcinbodnar: Sure. Done.

andy-haynes (changes requested):
Looking good, just one question on initialization otherwise LGTM 🎉

packages/frontend/src/index.js
```diff
@@ -10,15 +10,20 @@ import { createStore } from 'redux';
...
 import createMiddleware from './redux/middleware';
 import { initSentry } from './utils/sentry';
+import { wallet } from './utils/wallet';
 
 initSentry();
 
 const history = createBrowserHistory();
 
-export const store = createStore(createRootReducer(history), createMiddleware(history));
+export const store = createStore(createRootReducer(history, wallet.accountId), createMiddleware(history));
```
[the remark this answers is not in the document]
marcinbodnar: So If we want to pass `accountId` as a param (and don't use `wallet.accountId` inside `setupAccountReducer`) then we need to pass it also here because while creating a store we are adding an initial state for every account that is available, and also we are setting active reducer for an active account. So if we will not pass the accountId here, we will not have an active reducer for the active account, and there will be no data saved by any action.

andy-haynes (approved):

marcinbodnar:
@andy-haynes 
Was working on refactoring https://github.com/near/near-wallet/pull/2555 yesterday, to match recent changes for this PR and ran into a few issues that might appear while refactoring other low-level reducers. 

Generally, we cannot pass `accountId` as a param to `setupAccountReducer` - `accountId` inside `setupAccountReducer` must be a dynamic value, it must handle situations where active accountId is changing for example while switching accounts. 

We removed the usage of `wallet.accountId` inside `setupAccountReducer` which was a good decision. So if we can't pass it as a param and we don't want to use `wallet.accountId` I had to figure out the way so that the low-level reducer will know the value of active accountId.

To handle it I added an additional reducer one level up, so now the `accounts` is an active reducer, that is storing `activeAccount` data as well as the all accounts data. This way low-level reducer can operate using `activeAccount` data.

It's a much more robust implementation, `accounts` reducer has all the necessary data to operate, and we don't have to use `store.getState()` anymore.

packages/frontend/src/redux/actions/account.js
```diff
@@ -544,6 +545,7 @@ export const refreshAccount = (basicData = false) => async (dispatch, getState)
     }
 
     dispatch(setLocalStorage(wallet.accountId));
+    dispatch(activeAccountActions.setAccountId({ accountId: wallet.accountId }));
```
andy-haynes, line 548: So if this now needs to be called before `setupAccountReducer`, can we just inject the reducers for the new account here instead of calling `addAccountReducer` in `wallet.saveAccount()`?

marcinbodnar: Nope, `refreshAccount` is dispatched on every route change automatically and in many other situations. `addAccountReducer` needs to be called only in a situation when a new account is added.

I will work on introducing one action, that will be dispatched when the account creation/recovery is over, and then I will move `addAccountReducer` there. But it's not that easy as we have many methods for recovery or creating accounts, and it will be a part of separate PR.

packages/frontend/src/utils/wallet.js
```diff
@@ -460,6 +460,9 @@ class Wallet {
         this.getAccountsLocalStorage();
         await this.setKey(accountId, keyPair);
         this.accounts[accountId] = true;
+
+        // TODO: figure out better way to inject reducer
+        store.addAccountReducer(accountId);
```
andy-haynes, line 465: @marcinbodnar please remove the `accountId` parameter here as it is no longer used

marcinbodnar: Thanks. Done.

PR #2555, opened 2022-03-28
Refactor tokens reducer into two reducers

marcinbodnar:
The idea is to have two separate reducers:
1. `tokens` - that will be a part of account reducers
2. `tokensMetadata` - that will be shared between accounts.

### DIFF
--- a/packages/frontend/src/components/staking/components/Validator.js
+++ b/packages/frontend/src/components/staking/components/Validator.js
@@ -12,7 +12,7 @@ import selectNEARAsTokenWithMetadata from '../../../redux/selectors/crossStateSe
 import { selectValidatorsFarmData, selectFarmValidatorAPY, selectStakingCurrentAccountAccountId } from '../../../redux/slices/staking';
 import { selectActionsPending } from '../../../redux/slices/status';
 import { selectTokensFiatValueUSD, selectTokenWhiteList } from '../../../redux/slices/tokenFiatValues';
-import { selectAllContractMetadata } from '../../../redux/slices/tokens';
+import { selectContractsMetadata } from '../../../redux/slices/tokensMetadata';
 import StakingFarmContracts from '../../../services/StakingFarmContracts';
 import { FARMING_VALIDATOR_VERSION } from '../../../utils/constants';
 import FormButton from '../../common/FormButton';
@@ -78,7 +78,7 @@ export default function Validator({
     
     const NEARAsTokenWithMetadata = useSelector(selectNEARAsTokenWithMetadata);
 
-    const contractMetadataByContractId = useSelector(selectAllContractMetadata);
+    const contractMetadataByContractId = useSelector(selectContractsMetadata);
     const tokenFiatValues = useSelector(selectTokensFiatValueUSD);
     const tokenWhitelist = useSelector(selectTokenWhiteList);
     const currentAccountId = useSelector(selectStakingCurrentAccountAccountId);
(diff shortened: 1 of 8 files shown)


packages/frontend/src/redux/slices/tokensMetadata/index.js
```diff
@@ -0,0 +1,50 @@
...
+    ...tokensMetadataSlice.actions
+};
+export const reducer = tokensMetadataSlice.reducer;
+
+// Top level selectors
+const selectTokensMetadataSlice = (state) => state[SLICE_NAME] || initialState;
+
+const getContractNameParam = createParameterSelector((params) => params.contractName);
+
+export const selectContractsMetadata = createSelector(selectTokensMetadataSlice, ({ byContractName }) => byContractName || {});
```
andy-haynes, line 45: Does this still need to fall back on the default value of `{}` since `initialState` will have a value for `byContractName`?

packages/frontend/src/components/staking/components/Validator.js
```diff
@@ -78,7 +78,7 @@ export default function Validator({
     
     const NEARAsTokenWithMetadata = useSelector(selectNEARAsTokenWithMetadata);
 
-    const contractMetadataByContractId = useSelector(selectAllContractMetadata);
+    const contractMetadataByContractId = useSelector(selectContractsMetadata);
```
andy-haynes, line 81: Off topic question: why do we use a `token_id` as a key in a map that's keyed by contract name? Are contract names and token IDs the same thing for fungible tokens?

packages/frontend/src/redux/slices/tokensMetadata/index.js
```diff
@@ -0,0 +1,50 @@
...
+        }
+    }
+});
+
+export default tokensMetadataSlice;
+
+export const actions = {
+    ...tokensMetadataSlice.actions
+};
+export const reducer = tokensMetadataSlice.reducer;
```
andy-haynes, line 38: I think we can drop the exports for `actions` and `reducers` - only `actions` is referenced but that gets destructured from `tokensMetadataSlice` directly.

andy-haynes (changes requested):
This is awesome, much clearer to break these up 🎉 

Just some minor feedback otherwise LGTM

packages/frontend/src/redux/slices/tokensMetadata/index.js
marcinbodnar: Done.

packages/frontend/src/redux/slices/tokensMetadata/index.js
marcinbodnar: Done.

packages/frontend/src/components/staking/components/Validator.js
marcinbodnar: which line of code do you mean?

andy-haynes: https://github.com/near/near-wallet/blob/WAL-216-refactor-tokens-reducer-into-two-reducers/packages/frontend/src/components/staking/components/Validator.js#L35

marcinbodnar: I see. I'm not sure, it seems like `token_id` is a part of what`get_farms` function is returning -> `StakingFarmContracts.getFarms` - I didn't participate in these changes.

andy-haynes (approved):
Nice work!

andy-haynes (approved):

PR #2558, opened 2022-03-29
WAL-215 Twilio Verify error handling

morgsmccauley:
Twilio Verify has several different rate limits which do not exist with the current SMS delivery service. This introduces new user error states which now need to be handled. These 2FA rate limits users are likely to encounter are:
* User has sent/resent the code too many times
* User has incorrectly verified the code too many times
* A pending verification was left for an extended period and expired

This PR will surface these error messages to the user via `<GlobalAlert />` and also close the 2FA modal preventing any further rate limited verifications/resends. From here, new transactions will be blocked until the rate limit has subsided (usually ~10mins).

### DIFF
--- a/packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
+++ b/packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
@@ -95,15 +95,15 @@ const TwoFactorVerifyModal = ({ open, onClose }) => {
     const handleResendCode = async () => {
         setResendCode('resending');
         await Mixpanel.withTracking('2FA Modal Resend code', 
-            async () => await dispatch(resendTwoFactor()),
+            async () => {
+                await dispatch(resendTwoFactor());
+                setResendCode('resent');
+                setTimeout(() => { setResendCode(); }, 3000);
+            },
             (e) => {
                 setResendCode();
                 throw e;
             },
-            () => {
-                setResendCode('resent');
-                setTimeout(() => { setResendCode(); }, 3000);
-            }
         );
     };
     
--- a/packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
+++ b/packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
@@ -15,6 +15,8 @@ import Modal from '../../common/modal/Modal';
 import ModalTheme from '../ledger/ModalTheme';
 import TwoFactorVerifyInput from './TwoFactorVerifyInput';
 
+const TOO_MANY_REQUESTS_STATUS = 429;
+
 const Form = styled.form`
     display: flex;
     flex-direction: column;
@@ -101,7 +103,12 @@ const TwoFactorVerifyModal = ({ open, onClose }) => {
                 setTimeout(() => { setResendCode(); }, 3000);
             },
             (e) => {
-                setResendCode();
+                if (e.status === TOO_MANY_REQUESTS_STATUS) {
+                    onClose(false, e);
+                } else {
+                    setResendCode();
+                }
+
                 throw e;
             },
         );
--- a/packages/frontend/src/components/login/v2/ConfirmLoginWrapper.js
+++ b/packages/frontend/src/components/login/v2/ConfirmLoginWrapper.js
@@ -3,7 +3,6 @@ import { useSelector, useDispatch } from 'react-redux';
 
 import { Mixpanel } from '../../../mixpanel/index';
 import { allowLogin } from '../../../redux/actions/account';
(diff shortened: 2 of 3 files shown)


packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
```diff
@@ -95,15 +97,20 @@ const TwoFactorVerifyModal = ({ open, onClose }) => {
...
+                } else {
+                    setResendCode();
+                }
+
                 throw e;
             },
-            () => {
-                setResendCode('resent');
-                setTimeout(() => { setResendCode(); }, 3000);
-            }
```
morgsmccauley, line 106: We only want to run this on success, but it also gets run on failure: https://github.com/near/near-wallet/blob/master/packages/frontend/src/mixpanel/index.js#L23-L27

I've moved this block to main function so it's only run after a successful resend.

packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyModal.js
```diff
@@ -95,15 +97,20 @@ const TwoFactorVerifyModal = ({ open, onClose }) => {
...
-            async () => await dispatch(resendTwoFactor()),
+            async () => {
+                await dispatch(resendTwoFactor());
+                setResendCode('resent');
+                setTimeout(() => { setResendCode(); }, 3000);
+            },
             (e) => {
-                setResendCode();
+                if (e.status === TOO_MANY_REQUESTS_STATUS) {
+                    onClose(false, e);
```
morgsmccauley, line 107: Close the modal when we hit a rate limit, preventing the user for verifying/resending when we know it will fail.

morgsmccauley: `e` here will also get surfaced to the pending action ensuring that the `<GlobalAlert />` is populated with the error message.

packages/frontend/src/components/login/v2/ConfirmLoginWrapper.js
```diff
@@ -33,20 +32,7 @@ export default ({
...
-                await Mixpanel.withTracking('LOGIN',
-                    async () => {
-                        await dispatch(allowLogin());
-                    },
-                    (e) => {
-                        dispatch(showCustomAlert({
-                            success: false,
-                            messageCodeHeader: 'error',
-                            errorMessage: e.message
-                        }));
```
morgsmccauley, line 46: `allowLogin` already handles populating the `alert` `meta` so this actually just produces a duplicate alert.

packages/frontend/src/components/login/v2/ConfirmLoginWrapper.js
```diff
@@ -33,20 +32,7 @@ export default ({
...
-                    }
-                );
-            }}
+            onClickConnect={() => Mixpanel.withTracking('LOGIN', () => dispatch(allowLogin()))}
```
Patrick1904, line 35: Sweet! 🙏

Patrick1904 (approved):
This is great! Thanks for doing some cleanup/optimization at the same time.

Patrick1904:
One thing I noticed while testing this, is that the ['invalid code'](https://github.com/near/near-wallet/blob/43c28d0dbead6094b27a6edf4c0b6e9f482cdb7c/packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyInput.js#L58) error string on the 2FA modal isn't displaying when entering the wrong code - I suspect it's been broken for a while. It feels a bit unclear what's actually happening because we just show the modal again, without any UI error explaining what's wrong. Do you think you could look into that too? Separate PR seems fine as it's not regression from these changes.

morgsmccauley:
> One thing I noticed while testing this, is that the ['invalid code'](https://github.com/near/near-wallet/blob/43c28d0dbead6094b27a6edf4c0b6e9f482cdb7c/packages/frontend/src/components/accounts/two_factor/TwoFactorVerifyInput.js#L58) error string on the 2FA modal isn't displaying when entering the wrong code - I suspect it's been broken for a while. It feels a bit unclear what's actually happening because we just show the modal again, without any UI error explaining what's wrong. Do you think you could look into that too? Separate PR seems fine as it's not regression from these changes.

`VERIFY_TWO_FACTOR` doesn't actually get called for 2FA requests, instead, we call `near-api-js` which then verifies the OTP by calling `near-contract-helper` directly. Because of this `VERIFY_TWO_FACTOR` doesn't get populated with the error and therefore won't prompt this alert.

The only time we do call `VERIFY_TWO_FACTOR` is the user sets up an SMS recovery method. If they enter a wrong code there the alert does display 😅 

I'll investigate whether we can have something similar for these 2FA requests :)

PR #2560, opened 2022-03-30
fix(sign): Prevent runtime error on failed sign tx

morgsmccauley:
When a user has 2FA enabled and gets rate limited by Twilio the sign request fails with no successful transactions, making `successHashes` `undefined`. This PR defaults `successHashes` to an empty array to prevent some of the down stream logic which assume it exists from failing:

image


packages/frontend/src/redux/reducers/sign/index.js
```diff
@@ -20,6 +21,7 @@ const sign = handleActions({
         const allActions = transactions.flatMap((t) => t.actions);
         
         return {
+            ...state,
```
marcinbodnar, line 24: If I understood correctly you added this line so `successHashes` will not be cleared by `parseTransactionsToSign` action.

It will work well at this point, I'm just not sure if spreading the state, in this case, is the best idea because it might spread the state that belongs to a previous sign action. I think now it's not possible in the code, but it might happen if for example at some point `parseTransactionsToSign` or `handleRefreshUrl` will be dispatched. 

I would propose to use `initialState` and spread it instead of `state`.

Overall I think that at some point we should refactor `sign` state tree and hold transaction data keyed by `transactions` prop. The current state tree is not the best idea, of course now signing transaction is always related to opening a new browser session but clearing the whole reducer state is never the best idea.

marcinbodnar (approved):
Good job @morgsmccauley

PR #2561, opened 2022-03-30
feat: add restore-account-contract command

### LINKED ISSUE #2496: Disable 2FA script question
Observing https://github.com/near/near-wallet/pull/2487/files#diff-af1599f511c250583625e7629817717180c2b3693cb7d8727d2c9250e612fd6cR22 - which contact is been deployed?

Also why is contract been deployed for disabling 2FA?
My expectation that next sequence of actions are done:
 - remove functional access key and replace them back with full access keys except for the one by NEAR Wallet
 - deploy state cleaning contract
 - call it to clean state after 2FA contract
 - deploy empty contract (to get to 0x0 hash)

Also if account had a contract before, a special circumstances are applied and probably should be done manually:
 - lookup old contract before by finding a offending transaction and it's block
 - fetch the old contract code via next command `http post http://rpc.mainnet.near.org/ method=query params:=‘{“account_id”: “pool.near”, “request_type”: “view_code”, “block_id”: [59131110](tel:59131110) }’ jsonrpc=2.0 id=123` (need to replace block_id to before deployment and account id to the specific account)
 - deserialize the resulting base64 encoded file into .wasm file
 - deploy that contract by using the key from the seed phrase (`near deploy <account id> decoded.wasm --seedPhrase='...'`)
 - replace functional access keys with full access keys except for one by NEAR Wallet (`near keys <account id>`, `near add-key <account id> <key> --seedPhrase='...'`)
 -

esaminu:
Adds a command to `utilities` package to restore an account's code to a given `blockHash`.

Together with the `disable-2fa` command added in #2533 , this closes #2496

### DIFF
--- a/packages/utilities/commands/restoreAccountContract.js
+++ b/packages/utilities/commands/restoreAccountContract.js
@@ -1,0 +1,57 @@
+const { InMemoryKeyStore } = require("near-api-js/lib/key_stores");
+const { parseSeedPhrase } = require("near-seed-phrase");
+const { Connection, KeyPair } = require("near-api-js");
+const { Account } = require("near-api-js/lib/account");
+
+module.exports = {
+    command: `restore-account-contract`,
+    builder: (yargs) =>
+        yargs
+            .option("accountId", {
+                desc: "accountId to disable the 2fa on",
+                type: "string",
+                required: true,
+            })
+            .option("seedPhrase", {
+                desc: "seedPhrase for the accountId",
+                type: "string",
+                required: true,
+            })
+            .option("nodeUrl", {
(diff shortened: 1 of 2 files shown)


Eki08:
Recovery wallet sultan.08 to 634357ff1aef531e64ba63774dac27e01656c8fe5e19814a92306053d65eec3c

esaminu:
This PR only implements code restoration to a given `blockHash`. Full code **and** state restoration is only possible once near/core-contracts#208 is merged

MaximusHaximus (approved):
:+1: The end of an era! ;)

Issue #2559, opened 2022-03-29
[Feature] Support for viewing Video NFTs in collectibles tab


roshaans:
#### What problem are we solving with? Is this epic part of the planned features [here](https://github.com/near/near-wallet-roadmap/projects/1), and which ones?

- Currently, the collectibles tab defaults to viewing collectible's metadata within a `<img>` tag. This should be dynamic based on the file type. 

#### How did this problem originate? If it was revealed through user feedback, how was that feedback received and from whom?
- My video nfts show up blank on the collectibles tab. 

#### Job Story(s)
- When I view a video file embdded NFTs/collectibles, I want to see them in the collectibles view similar to how Paras.id does it so I can enjoy my collectibels for what they are suppose to be made for.
#### What are we planning to do about it?
- What is the scope of the design and engineering work required for this project?
- Along the way, this epic will be linked with issues and pull requests
it is easier to assume that the picture from the metadata is an image. Many IPFS links in the token data do not expose what type of metadata it is. The metadata link type should be embedded as part of the NFT standard. Then, the front should decide what tag to choose `<img>` or `<video>` based on this. 

#### What are we _not_ planning to do about it?
- What items have been considered but may fall out of the scope of design and engineering work for this project?
Sometimes the scope is too large, and this section may lead to other epics, other features [here](https://github.com/near/near-wallet-roadmap/projects/1) or even other products
Do a sophisticated hacked-out approach that pings the IPFS link and awaits to see what type of data it returned. Possible, but introduces a lag for loading all collectible items, even images. Perhaps, we can load images by default and if images do not load, then only then check for other types of media. 

## Measuring Success
- Are there a set of quantifiable metrics that we can surface once the work has been completed in order to determine how successful our solution was?
Being able to view a dynamic set of media in the collectibles tab. 

### Are there any resources that are necessary or that will aid in the completion of this work?
- paras.id appraoch

roshaans:
I noticed that the webm format NFTs or any Video NFTs are not supported on the NFT detail tab at all. 

I made a PR for it if you would kindly review it. 

https://github.com/near/near-wallet/pull/2562

agileurbanite:
I believe this is  closed by #2562

PR #2562, opened 2022-03-30
feat: added video to show up on the NFT detail screen

roshaans:
Users are unable to see their video NFTs on the NFT-detail screen when they click their NFTs. Video NFTs are not supported. 

### What Changed? 
- Users are able to video and regular image NFTs in the NFT details tab. 

| Before   |      After      |
|----------|:-------------:|
| Before | After |

## Image NFTs show up as normal 

| Collectible Screen   |      NFT Detail Screen      |
|----------|:-------------:|
| image | image|

### DIFF
--- a/packages/frontend/src/components/nft/NFTDetail.js
+++ b/packages/frontend/src/components/nft/NFTDetail.js
@@ -2,6 +2,7 @@ import React, { useState } from 'react';
 import { Translate } from 'react-localize-redux';
 import styled from 'styled-components';
 
+import FailedToLoad from '../../images/failed_to_load.svg';
 import UserIconGrey from '../../images/UserIconGrey';
 import { NFT_TRANSFER_DEPOSIT, NFT_TRANSFER_GAS } from '../../services/NonFungibleTokens';
 import BackArrowButton from '../common/BackArrowButton';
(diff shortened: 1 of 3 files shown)


andy-haynes:
This is great, thank you for the contribution @roshaans!

Since the change being made here is already in place in `components/nft/NFTBox.js` it would be best to refactor this so that both components are rendering NFT media using the exact same logic. If you have time would you mind please adding a new functional component in `components/nft/NFTMedia.js` that takes a `mediaUrl` prop and renders either the `<video />` or `` tag using the same logic in your change and `NFTBox.js`? Then both `NFTBox.js` and `NFTDetail.js`  can render `<NFTMedia mediaUrl={...} />` instead of having it inline.

This will ensure that any future changes to how we render NFT media are always applied to both components. If you don't have time I am happy to make the change as well, but I may need push access to your fork. Let me know if this works for you otherwise I'm sure we can work something out to get your fix into the codebase.

agileurbanite:
@andy-haynes let's wait till 4/4/22 for @roshaans to reply before we take over and put this in the proper place.

agileurbanite:
Related to #2559

roshaans:
> 

Yes, I will get this in as soon I can!

roshaans:
@andy-haynes @agileurbanite I believe that takes care of the refactor. Let me know if you want anything else changed and happy to contribute!

andy-haynes (approved):
Great work, @roshaans! Thank you for your bug fix and refactor ❤️

PR #2563, opened 2022-03-31
feat: Make wallet usable for accounts without record on chain

Patrick1904:
This PR is part of effort to enable '[zero balance accounts](https://nearinc.atlassian.net/browse/WAL-87)'.

This PR makes the wallet usable for accounts with no record on chain, a vital step in supporting zero balance accounts.

Main UX changes for non-existent accounts:
1. User is able to visit dashboard, login, send, staking, receive, and buy
2. The wallet app does not suddenly break (blank page) when an account doesn't exist on chain or the RPC is unable to confirm the status of an account

Easiest way to test is to visit the [staging test link](https://deploy-preview-2563--near-wallet-staging.netlify.app/) and add the following key/value pairs to `localStorage`:

`_4:wallet:accounts_v2 ////// {"28f511e62ab449da82c7689d33161122312c1bb86d6fe389f78a5760cbafad76":true}`

`_4:wallet:active_account_id_v2 ////// 28f511e62ab449da82c7689d33161122312c1bb86d6fe389f78a5760cbafad76`

`wallet.account:28f511e62ab449da82c7689d33161122312c1bb86d6fe389f78a5760cbafad76:default:confirmed ////// true`

`nearlib:keystore:28f511e62ab449da82c7689d33161122312c1bb86d6fe389f78a5760cbafad76:default ////// ed25519:5mTyEaDFLXwvaQ9XNQUrMFKHPdq89J17ysr51kFvjas15TPXSAyCLipkKqcSDtEM9zqmV9jeBF3cmbADArkaDdpq`

Screen Shot 2022-03-31 at 4 13 13 PM

To test switching between a zero balance accounts and an existing account, simply import an existing account into the wallet.

Screen Shot 2022-03-31 at 3 58 34 PM

Screen Shot 2022-04-12 at 5 48 05 PM

### DIFF
--- a/packages/frontend/src/redux/sharedThunks/refreshAccountOwner.js
+++ b/packages/frontend/src/redux/sharedThunks/refreshAccountOwner.js
@@ -28,7 +28,11 @@ export default createAsyncThunk(
                 }
                 if (nextAccountId) {
                     dispatch(makeAccountActive(nextAccountId));
+                    //FIX: Automatic switching when account doesn't exist makes it impossible to switch to a zero balance account
+                    // However, redux 'availableAccounts' array becomes empty if you switch to a zero balance account and refresh the page
                 }
+
+                console.log('createAsyncThunk refreshAccountOwner wallet.accounts',wallet.accounts);
 
                 // TODO: Make sure "problem creating" only shows for actual creation
                 return {
--- a/packages/frontend/src/redux/sharedThunks/refreshAccountOwner.js
+++ b/packages/frontend/src/redux/sharedThunks/refreshAccountOwner.js
@@ -26,7 +26,9 @@ export default createAsyncThunk(
                         break;
                     }
                 }
-                dispatch(makeAccountActive(nextAccountId));
+                if (nextAccountId) {
+                    dispatch(makeAccountActive(nextAccountId));
+                }
 
                 // TODO: Make sure "problem creating" only shows for actual creation
                 return {
--- a/packages/frontend/src/redux/sharedThunks/refreshAccountOwner.js
+++ b/packages/frontend/src/redux/sharedThunks/refreshAccountOwner.js
@@ -1,5 +1,6 @@
 import { createAsyncThunk } from '@reduxjs/toolkit';
 
+import { isImplicitAccount } from '../../utils/account';
 import { getAccountConfirmed, setAccountConfirmed } from '../../utils/localStorage';
 import { wallet } from '../../utils/wallet';
 import { makeAccountActive } from '../actions/account';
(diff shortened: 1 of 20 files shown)


packages/frontend/src/routes/WalletWrapper.js
```diff
@@ -28,7 +28,7 @@ export function WalletWrapper({
     const createFromImplicitSuccess = useSelector(selectCreateFromImplicitSuccess);
     const createCustomName = useSelector(selectCreateCustomName);
     const fungibleTokensList = useFungibleTokensIncludingNEAR();
-    const tokensLoader = useSelector((state) => selectTokensLoading(state, { accountId })) || !balance?.total;
+    const tokensLoader = useSelector((state) => selectTokensLoading(state, { accountId }));
```
andy-haynes, line 31: Probably out of scope for this PR but should this be `tokensLoading` instead?

andy-haynes (changes requested):
Dropped a few observations inline, @marcinbodnar probably has some more substantive feedback 😄 

>The wallet app does not suddenly break (blank page) when an account doesn't exist on chain or the RPC is unable to confirm the status of an account

😍

marcinbodnar (commented):
I checked the functionality mainly, UI looks very slick.

Few things I noticed:
1. Account page doesn't work - just showing Global Alert. I wonder maybe it would be a bit better if we would avoid showing the alert, and show the Account page with limited and empty data and information in the UI, like with `Recent activity`?
2. Global Alert is not being closed after a route change.
3. When I choose `Send` and `Select asset` the balance for `NEAR` is loading.
4. When I added another account by recovery, I'm not able to switch back to the zero balance account - zero balance account appears in the menu, but after switching, it's switching back to the nonempty account. I'm not sure how we want to handle it though. Do we want to allow to switch to the empty account, or maybe remove the account from local storage if there is a proper account?

I'm just thinking about this implementation. Now we are implementing a solution where we are taking the accountId from local storage when there is no account in the redux state and handle the problems that might appear if there is no proper data for the account. I wonder wouldn't it be simpler if generally, we would fill the `account` reducer with empty data (accountId, etc), and instead of checking the accountId from local storage, we will handle the problems that might occur in the Wallet where the account that is in the redux is zero balance account.

I'm probably not explaining well what's on my mind :D but I wonder about the general idea, where instead of thinking about what would happen if there is no balance object (and handle it), we would just fill the balance object with all standard props but the values will be `0`.

it's just a thought, it's not needed for this implementation. But maybe we can think about something like that in the further works on the zero balance account.

Patrick1904:
> I'm just thinking about this implementation. Now we are implementing a solution where we are taking the accountId from local storage when there is no account in the redux state and handle the problems that might appear if there is no proper data for the account. I wonder wouldn't it be simpler if generally, we would fill the `account` reducer with empty data (accountId, etc), and instead of checking the accountId from local storage, we will handle the problems that might occur in the Wallet where the account that is in the redux is zero balance account.

Thanks for the advice @marcinbodnar - I think it's a reasonable approach. I'll tinker with this and update the PR.

packages/frontend/src/redux/sharedThunks/refreshAccountOwner.js
```diff
@@ -26,7 +26,13 @@ export default createAsyncThunk(
                         break;
                     }
                 }
-                dispatch(makeAccountActive(nextAccountId));
+                if (nextAccountId) {
+                    dispatch(makeAccountActive(nextAccountId));
+                    //FIX: Automatic switching when account doesn't exist makes it impossible to switch to a zero balance account
```
Patrick1904, line 31: @marcinbodnar removing this automatic account switching when the account doesn't exist works fine until you refresh the page, because for some reason that causes the `availableAccounts` array in redux to be completely empty. Do you have any idea why that is? `wallet.accounts` still has the accounts listed, they're just not getting into the `availableAccounts` array. Should we maybe change the way we're loading `availableAccounts` to not depend on `refreshAccountOwner` and instead just pull them from `localStorage` without additional checking?

Patrick1904: Fixed it here: https://github.com/near/near-wallet/pull/2563/commits/bbf75775a3ccc2faa5c5256e8d1019aea6fb418f

Please review this solution

MaximusHaximus (approved):
:+1: LGTM -- good work @Patrick1904 

Feel free to merge as long as @marcinbodnar has no blockers from his review

marcinbodnar (approved):
Looks great, works well.

packages/frontend/src/routes/WalletWrapper.js
Patrick1904: Great catch! Fixed it here: https://github.com/near/near-wallet/pull/2563/commits/71f4692d487ce9f5cd165498e2b0e2f073f997f4

PR #2565, opened 2022-04-01
fix(env): more descriptive environment error

### LINKED ISSUE #2564: how can I fix it?
image

Hi,  When I run the wallet (https://github.com/near/near-wallet/tree/master/packages/frontend), its interface is blank. I see the following error in Chrome, how can I fix it?

andy-haynes:
This PR expands the "no environment set" error message to provide more context and enumerate the set of valid environments (values for `NEAR_WALLET_ENV` env var).

Closes #2564

### DIFF
--- a/packages/feature-flags/src/index.js
+++ b/packages/feature-flags/src/index.js
@@ -4,7 +4,10 @@ function initFeatureFlags({ flagState, currentEnvironment, environments }) {
     }
 
     if (!Object.values(environments).includes(currentEnvironment)) {
-        throw Error(`invalid environment: "${currentEnvironment}"`);
+        throw Error(`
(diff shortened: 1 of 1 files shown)


yarn.lock
```diff
@@ -2027,15 +2027,6 @@
     call-me-maybe "^1.0.1"
     glob-to-regexp "^0.3.0"
 
-"@near-wallet/feature-flags@^0.0.4":
```
andy-haynes, line 2030: @MaximusHaximus not sure why this dropped the reference altogether. What is the correct way to bump the feature flag package? Do we need to be running a `lerna` command?

morgsmccauley (approved):
🚀

PR #2571, opened 2022-04-06
feat: Enable 'Remove Wallet Account'

Patrick1904:
https://nearinc.atlassian.net/browse/WAL-58

**Description**
This feature allows users to remove a single account from the wallet. The wallet automatically makes the next available account active once the user removes an account. The user is redirected to the guest landing page if no other accounts are available.

**Notes**
All changes to keys are limited to the browsers `localStorage`. There are currently no changes made to keys on-chain.

Screen Shot 2022-04-05 at 6 54 18 PM

### DIFF
--- a/packages/frontend/src/utils/wallet.js
+++ b/packages/frontend/src/utils/wallet.js
@@ -15,7 +15,7 @@ import { actions as ledgerActions } from '../redux/slices/ledger';
 import sendJson from '../tmp_fetch_send_json';
 import { decorateWithLockup } from './account-with-lockup';
 import { getAccountIds } from './helper-api';
-import { setAccountConfirmed } from './localStorage';
+import { setAccountConfirmed, setWalletAccounts, removeActiveAccount, removeAccountConfirmed } from './localStorage';
 import { TwoFactor } from './twoFactor';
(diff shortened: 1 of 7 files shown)


packages/frontend/src/utils/wallet.js
```diff
@@ -142,8 +142,18 @@ class Wallet {
         this.accountId = localStorage.getItem(KEY_ACTIVE_ACCOUNT_ID) || '';
     }
 
+    removeWalletAccount(accountId) {
+        let walletAccounts = this.getAccountsLocalStorage();
+        delete walletAccounts[accountId];
+        setWalletAccounts(KEY_WALLET_ACCOUNTS, walletAccounts);
+        this.keyStore.removeKey(NETWORK_ID, accountId);
```
MaximusHaximus, line 149: This method is `async` - we need to either `await` it or add a .catch() to it that swallows the error or logs the error, or any errors calling `removeKey()` will become unhandled promise rejections

MaximusHaximus (approved):
:+1: nice work!  Just one thing about the `removeKey` method being async but not being handled as one currently, then this is ready to rock!

PR #2572, opened 2022-04-06
feat: added additional permission to LAK login display

### LINKED ISSUE #2567: Incorrect information when authorizing an app
**Problem**
Whenever I authorize an app, like https://app.burrow.cash/ in wallet it shows me the following:
image

It claims that app can only view address and balance of the account. However, it does much more, it adds an access key that allows app to call anything from the application contract using my account as a transaction signer.

**Expected Behavior**
The wallet should state explicitly that application can call any method of the contract on behalf of my account.

**Steps to reproduce**

1. Open https://app.burrow.cash/
2. Login with some wallet account. Observe the above screenshot that states that application can only view account address and balance;
3. Go to explorer and observe that an access key that allows all method calls has been added.

**Note**
In general, it seems to be an anti-pattern and a potential security issue to add an access key every time the Logs out and then Logs in. It leads to a significant proliferation of access keys if the user is an active DeFi user on NEAR. This makes access keys unmanageable, since the user would need to review history of transactions to understand the origin of all access keys. Given that the keys are used to control the account access it creates potential for user mistake or security exploits, e.g. a user can by accident delete a wrong key.

esaminu:
This PR adresses concerns in and closes #2567 :
* Adds a new permission display regarding the ability to call methods on the contract

image

### DIFF
--- a/packages/frontend/src/components/login/v2/ConfirmLogin.js
+++ b/packages/frontend/src/components/login/v2/ConfirmLogin.js
@@ -111,6 +111,7 @@ const LimitedAccessUI = () => (
     <>
         <PermissionItem translateId='login.v2.connectConfirm.permissions.viewAddress' />
         <PermissionItem translateId='login.v2.connectConfirm.permissions.viewBalance' />
+        <PermissionItem translateId='login.v2.connectConfirm.permissions.callMethodsOnContract' />
         <PermissionItem permitted={false} translateId='login.v2.connectConfirm.permissions.notTransferTokens' />
     </>
 );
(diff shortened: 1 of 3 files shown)


Patrick1904 (changes requested):
Thanks for addressing this! I'm getting a weird UI bug in Chrome, but you can fix that by adding `min-width: 32px` in [PermissionItem.js](https://github.com/near/near-wallet/blob/master/packages/frontend/src/components/login/v2/PermissionItem.js#L13)

Screen Shot 2022-04-06 at 1 56 25 PM

corwinharrell:
I would also encourage us to always try to think of articulating these types of things in layman's terms, so that the messaging is valuable to non-developers. How might this permission be framed in such a way that doesn't depend on prerequisite knowledge about smart contracts and programing?

esaminu:
Thanks for catching @Patrick1904 ! Updated 😄

Patrick1904 (approved):
Thanks @esaminu! I'm merging this. Let's update the copy in a follow-up if needed.

PR #2574, opened 2022-04-07
fix: Create implicit: Fix MoonPay button loading state

Patrick1904:
Screen Shot 2022-04-06 at 8 31 29 PM

### DIFF
--- a/packages/frontend/src/components/accounts/create/implicit_account/CreateImplicitAccount.js
+++ b/packages/frontend/src/components/accounts/create/implicit_account/CreateImplicitAccount.js
@@ -111,6 +111,8 @@ export default ({
                 */}
                 <FormButton
                     disabled={!moonpayIsAvailable}
+                    sending={!implicitAccountId || moonpaySignedUrl === null}
+                    sendingString='button.loading'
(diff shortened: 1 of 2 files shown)


MaximusHaximus (approved):
:+1: LGTM!

I wonder if the entire `moonpayIsAvailable` prop on `CreateImplicitAccount` is redundant now, though? Do we ever drive that prop off of anything _other_ than the presence of a moonpay signed URL elsewhere in the codebase? Does it even make sense to ever render `CreateImplicitAccount` with `moonpayIsAvailable` set to `false` but `moonpaySignedUrl` set to an URL or vice-versa? Seems like the component might even render in an inconsistent state in (some of?) those cases usability-wise...

If not, it could be nice to remove the `moonpayIsAvailable` prop entirely at this point and just use the value of `moonpaySignedUrl` to drive whatever internal logic inside of `CreateImplicitAccount` that currently is based on the value of the `moonpayIsAvailable` prop. Not a big deal or a blocker, just wondering if it would make sense to simplify the `CreateImplicitAccount` component API and behaviours if there's no reason to have two props here instead of one.

PR #2488, opened 2022-02-16
Custom HD Path for account recovery

gutsyphilip:
Closes #2200 

Bonus: 
Closes #2557

PR #2581, opened 2022-04-14
Support for active ledger transports

marcinbodnar:
This PR includes supporting two additional types of transport for Ledger and refactoring UI for Sign In With Ledger functionality.

Additionally handled scenarios:
 - when a user clicks send and there is no ledger connection yet - in this case, Global Alert is shown
 - when a user for example disconnected the device in a case where the connection was previously established - in this case, Global Alert is shown

Additionally rebuilt Sign In With Ledger components into a new structure.


marcinbodnar:
@MaximusHaximus Please check this PR if you have a moment. I was able to handle a few additional situations and also refactored the Sign in with ledger components. With this implementation, we will be able to get rid of `store` usage in `wallet.js` I have an idea how exactly it will look like and I will handle it in a separate PR.

@corwinharrell
This is a fully functional version for new ledger transport and Sign in With Ledger UI. There are a few topics that I would like to discuss with you:
1. Connect Ledger button and screen.
If you check the Figma designs, we have a screen `MacBook Pro 9` and a set of modals below. I'm not sure, but my idea is that this screen should be visible after the user clicks `Connect Ledger` button, because we need him to see additional information and we want him to confirm.

If so, then there are a few things:
 - the way establishing the connection is working, we don't need any modals, because it's happening instantly and the user doesn't have to confirm it on the device
 - the only situation that we would need modal is a case when the connection was unsuccessful
 - but there is one problem with `MacBook Pro 9` screen - for example, the user is trying to send tokens, he puts the amount, chooses a token, enters the receiver id and then he's clicking Send button - now we are redirecting him to the `MacBook Pro 9` screen - we can redirect him back to the Send screen after successful connection but the user would have to enter all data again.

The way I implemented it now is that there is no additional screen, if a user clicks Connect Ledger the connection is established. And if for example ledger device is not connected user will see Global Alert.

I think that a good solution would be to add a small additional window that will appear near the Connect Ledger button (the same type of window as we use for the account submenu) where we will be displaying all additional information - so it will be used for these situations:
 - the user clicks Connect Ledger and we want him to read additional information and confirm - in this case, a small window will appear with the information and the button to confirm
 - user is trying to perform an operation (Send, Sign in, etc) but the connection is not established yet - in this case, we can show him the same window with the information that there is no connection and a button to confirm a connection
 - connection is already established but a user disconnected the device fiscally - in this case, we will show the window with the information that the device was disconnected and he needs to connect it again.

Such a small window will not interfere with wallet functionality and will be used for several situations instead of global alert.

2. Importing accounts. 
If you check `MacBook Pro 4` screen - in this case, we are showing the list of accounts as the page and there is a set of modals below. The way I understand we want to show these modals while the user is accepting the accounts. 

But the problem is that the accounts are accepted one by one, the user is accepting the first account, then we are showing the loader inside the modal, and after that, we instantly starting with next account, so in practice, these modals will be visible constantly, so the user will almost never see the list of accounts behind modals. 

The way I implemented it now is that we don't use modals at all in Sign In With Ledger flow. Please check how it looks in this PR. When it comes to the screen with a list of account, all informations are visible there - the user sees when the account is being processed, that he needs to confirm on the device, was it a success or not, and which one will be next so it seems like modals are not needed.

3. There was an idea to add a possibility that the user will be able to choose which account he wants to import - now he can decline the account, but in a case where a user has few dozens of accounts and wants to import only one, he would have to decline all of the accounts manually. The idea was to add checkboxes for each account and options "select/unselect all" so it will be much easier. I can implement it with this PR, it would just need some small additional changes to the design.

4. There is also a window that appears if we didn't find any accounts - then the user is able to enter accountId manually. I prepared it using the same design but please check if it looks ok.

agileurbanite:
I suggest we split this PR out to two parts.  We need to index on getting the functional blockers that block chrome users from using ledger.  The UX approach can be revisited in a separate PR.  Let's make it work, then make it better.

corwinharrell:
@agileurbanite that's fine. Once @marcinbodnar and I have synced up and we've aligned on the things that need to be addressed, I'll leave it up to him to determine how best to divide the work in order to get a functional implementation out the door. He can then to follow up with any improvements to the UX if there are any that are considered supplemental and not critical.

packages/frontend/src/components/accounts/ledger/LedgerConnectModal/views/Connect.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
```
andy-haynes, line 1: Do these need to be separate components? The only things different between the three is the icon and the translation strings, seems like they could be consolidated for better reuse.

If we think these will dramatically diverge in the future then it makes sense to keep them separate but if we don't anticipate many changes then it probably makes more sense to combine them into a single component that takes an `isError` prop and translation string values. What do you think?

packages/frontend/src/components/accounts/ledger/SignInLedgerWrapper.js
```diff
@@ -106,73 +111,84 @@ export function SignInLedger(props) {
...
+        if (signInWithLedgerStatus === LEDGER_MODAL_STATUS.CONFIRM_ACCOUNTS || signInWithLedgerStatus === LEDGER_MODAL_STATUS.SUCCESS) {
+            return VIEWS.IMPORT_ACCOUNTS;
+        }
+    };
+
+    const getCurrentViewComponent = (view) => {
+        switch (view) {
+            case VIEWS.AUTHORIZE:
+                return (
+                    <Authorize
```
andy-haynes, line 141: We're not passing in the `props` prop here, does that prevent the cancel button from working in the `Authorize` component?

packages/frontend/src/translations/en.global.json
```diff
@@ -302,11 +304,33 @@
...
+        "two": "Follow the instructions on your device to confirm and grant access to each account you'd like to import."
     },
     "connecting": "Connecting",
+    "connectLedger": {
+        "connectLedger": "Connect Ledger",
+        "ledgerConnected": "Ledger Connected",
+        "modal": {
+            "connect": {
+                "header": "Connect to your Ledger device.",
+                "one": "Make sure your Ledger is connected securely, and that the NEAR app is open on your device."
```
andy-haynes, line 321: Can we factor this out to be directly under the `modal` key? If we need to update it in the future it would be better to have it in one place

packages/frontend/src/redux/slices/ledger/index.js
```diff
@@ -16,14 +18,73 @@ const SLICE_NAME = 'ledger';
...
-    ENTER_ACCOUNTID: 'enter-accountId'
+    ENTER_ACCOUNTID: 'enter-accountId',
+    SUCCESS: 'success'
+};
+
+export const CONNECT_MODAL_TYPE = {
+    CONNECT: 'connect',
+    CONNECTION_ERROR: 'connection-error',
+    DISCONNECTED: 'disconnected',
+    HIDE: false
```
andy-haynes, line 29: Can we make this a string for the sake of consistency? If we don't need to rely on it being a bool then it should have the same type as the other constants. I think it would also be more clear when querying the modal's `type` to get `"hidden"` rather than just `false`

packages/frontend/src/utils/ledgerManager.js
```diff
@@ -0,0 +1,72 @@
...
+import { createClient, getSupportedTransport, setDebugLogging } from 'near-ledger-js';
+
+setDebugLogging(true);
+listen(console.log);
+
+window.Buffer = Buffer; // Exists due to `bs58` import
+
+class LedgerManager {
+    constructor() {
+        this.bs58 = bs58;
```
andy-haynes, line 12: This doesn't appear to be referenced anywhere. Do we need it?

packages/frontend/src/components/navigation/ConnectLedger.js
```diff
@@ -0,0 +1,52 @@
...
+`;
+
+export default ({
+    connectLedger,
+    ledgerConnectionAvailable,
+    showConnectLedgerButton
+}) => {
+    return (
+        <>
+            {showConnectLedgerButton &&
```
andy-haynes, line 34: Seems odd to render an empty component based on a prop. What if we removed this prop and had the parent only render this when `showConnectLedgerButton` is `true`?

packages/frontend/src/components/accounts/ledger/SignInLedgerWrapper.js
```diff
@@ -106,73 +111,84 @@ export function SignInLedger(props) {
...
         if (signInWithLedgerStatus === LEDGER_MODAL_STATUS.ENTER_ACCOUNTID) {
-            dispatch(clearSignInWithLedgerModalState());
+            return VIEWS.ENTER_ACCOUNT_ID;
+        }
+        if (signInWithLedgerStatus === LEDGER_MODAL_STATUS.CONFIRM_ACCOUNTS || signInWithLedgerStatus === LEDGER_MODAL_STATUS.SUCCESS) {
+            return VIEWS.IMPORT_ACCOUNTS;
+        }
+    };
+
+    const getCurrentViewComponent = (view) => {
```
andy-haynes, line 137: Now that the rendered component calls only this method within a `Container` component, what would you think about inlining this to get rid of the `switch` statement altogether? Just thinking it might be easier to parse the component if it renders these directly rather than through a function.

e.g.
```
<Container ...>
    {activeView === VIEWS.AUTHORIZE && (
        <Authorize ... />
    )}
</Container>
```

andy-haynes (changes requested):
This is awesome, nice work @marcinbodnar! Just a few minor questions/suggestions but otherwise looking great

andy-haynes:
@marcinbodnar checking out the latest preview link on Chrome:
- using a ledger to create an account works but doesn't dismiss modal after account is created (still shows `Processing...`)
- when adding ledger recovery method, if the user rejects when adding a key the modal doesn't dismiss
- when recovering with ledger, the option is given to change the HD path: changing the path prevents recovery from working, even if you go back and try again... it only starts working if you clear localStorage

The only other thing I noticed is that we've dramatically increased the amount of logging we do to the console. Can we put this behind a debug flag or remove the logging please? Too much noise in the console makes it harder for users to submit relevant information when they encounter errors in the wallet, which in turn makes it more difficult for us to get a handle on errors that are written to the console.

packages/frontend/src/components/accounts/ledger/SignInLedgerViews/Authorize.js
```diff
@@ -0,0 +1,54 @@
...
+import LedgerHdPaths from '../LedgerHdPaths';
+
+const Authorize = ({
+    status,
+    path,
+    setPath,
+    setConfirmedPath,
+    handleSignIn,
+    signingIn,
+    props
```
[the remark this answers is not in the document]
marcinbodnar: Right, I was planning to do it differently. Changed it to `redirectTo` action. It dispatches `push` from `connected-react-router` and makes redux devtool work properly.

packages/frontend/src/components/accounts/ledger/SignInLedgerWrapper.js
marcinbodnar: Thanks for catching it, I was planning to do it differently. Changed it to `redirectTo` action. It dispatches `push` from `connected-react-router` and makes redux devtool work properly.

packages/frontend/src/translations/en.global.json
marcinbodnar: Right, thanks for catching it. That was the thing I forgot to ask @corwinharrell  about - because we have three different modals but the same text on design. I think we should provide some additional info for every modal that might help users to understand the problem. Something like:
 - on error - your device is probably not physically connected or you need to enter a pin
 - on disconnected - your device was probably disconnected physically.
Maybe you can provide some nice text and I will change it?

packages/frontend/src/redux/slices/ledger/index.js
marcinbodnar: Your right, it looks odd :) I just removed it, and now we are setting `type` to `undefined` if we don't want to show modal. Is that ok?

packages/frontend/src/components/accounts/ledger/SignInLedgerWrapper.js
andy-haynes: Thinking about this a bit more, I think my suggestion doesn't make things any clearer 🤔 

What if we combine `activeView` and `getCurrentViewComponent` into one function that gets rendered as a component? Then we don't need to map to `VIEW` types and take advantage of the fact that `getCurrentViewComponent` renders a component.

```js
const LedgerAction = () => {
  if (!signWithLedgerStatus) {
    return (
      <Authorize ... />
    );
  }
  ...
  return null;
}

return (
  <Container>
    <LedgerAction />
  </Container>
);
```

packages/frontend/src/utils/ledgerManager.js
marcinbodnar: Right, we don't use it. Done.

packages/frontend/src/components/navigation/ConnectLedger.js
marcinbodnar: Right, refactored it to return `null` - this is the pattern that we use in a few places - the idea is that the parent component doesn't have to care about showing/not showing the component - everything related to this component is inside it. Are you ok with this solution?

packages/frontend/src/components/accounts/ledger/LedgerConnectModal/views/Connect.js
andy-haynes: I'm suggesting passing in the text as separate props - `isError` would just be to determine the icon. Though the icon could be a separate prop as well.

packages/frontend/src/components/accounts/ledger/SignInLedgerWrapper.js
marcinbodnar: I took this implementation from `SendContainerV2.js` component - figured that it's the way we want to implement such scenarios. But I'm ok with both solutions, just not sure which one we prefare for consistency.

packages/frontend/src/components/navigation/ConnectLedger.js
andy-haynes: But the parent _does_ care because it knows the value of `showConnectLedgerButton`. To me it's more clear in the parent to know whether or not a child component is rendered. Child components shouldn't take a flag that prevents the component from rendering, it should just be a dumb component that always renders and the parent decides whether or not to render it.

packages/frontend/src/components/accounts/ledger/SignInLedgerWrapper.js
andy-haynes: IMO the two big things here are
- the unnecessary abstraction that comes with mapping application state to `VIEW` types
- returning components from invoking functions with arguments rather than React components with props

By merging the functions we get a solution with less indirection that is easier to read because it just boils down to "render X component when Y application state". As it is here and `SendContainerV2` the relationships between application state (e.g. `signWithLedgerStatus`) and component (`<Authorize />`) are unnecessarily separated, forcing developers to follow the arbitrary connection to a `VIEW` value.

marcinbodnar:
@andy-haynes 
> using a ledger to create an account works but doesn't dismiss modal after account is created (still shows Processing...)

Done

> when adding ledger recovery method, if the user rejects when adding a key the modal doesn't dismiss

Done

> when recovering with ledger, the option is given to change the HD path: changing the path prevents recovery from working, even if you go back and try again... it only starts working if you clear localStorage

So I tested it and seems to work well, probably happens in specific scenario - could you please try to reproduce it?

> The only other thing I noticed is that we've dramatically increased the amount of logging we do to the console. Can we put this behind a debug flag or remove the logging please? Too much noise in the console makes it harder for users to submit relevant information when they encounter errors in the wallet, which in turn makes it more difficult for us to get a handle on errors that are written to the console.

You're right. Done.

andy-haynes:
> So I tested it and seems to work well, probably happens in specific scenario - could you please try to reproduce it?

I can still consistently reproduce it. While it's not a typical user flow, anything we break that requires clearing localStorage to fix is very bad.

andy-haynes (approved):
Let's address the rest of this feedback in the second PR since it doesn't block the implementation.

The issue with recovering accounts via the wrong path is a preexisting problem so we will address it separately.

marcinbodnar:
> > So I tested it and seems to work well, probably happens in specific scenario - could you please try to reproduce it?
> 
> I can still consistently reproduce it. While it's not a typical user flow, anything we break that requires clearing localStorage to fix is very bad.
> 
> 

Great catch. Adding a comment to #2488 about it.

packages/frontend/src/components/accounts/ledger/SignInLedgerWrapper.js
marcinbodnar: Done:
https://github.com/near/near-wallet/pull/2595/commits/dedad83c677497d6178a1f67799f883c35b12057

packages/frontend/src/components/navigation/ConnectLedger.js
marcinbodnar: I see your point. Actually, we don't need to pass these props, this component can handle all necessary data - it would be much more effective:
https://github.com/near/near-wallet/pull/2595/commits/8e3d121d6c5d1a038aa8445fc5689eb6a8fa2a85

packages/frontend/src/components/accounts/ledger/LedgerConnectModal/views/Connect.js
marcinbodnar: Implemented perfect solution. One component, no text in props, simpler wrapper:
https://github.com/near/near-wallet/pull/2595/commits/e7c209e9955ac960ad9386907d85987731e358e1

PR #2568, opened 2022-04-05
feat:add USN swap page and USN to the list of tokens (#2)

OlegYak:
This PR is part of an effort to include the USN contract.

This PR gives the wallet the right to use USN tokens, exchange them for NEAR and vice versa.

Main changes: 

      1.Added a list of tokens and their total balance.
      2.Added swap page.
      3.Added a button to go to the swap page. 
      4.Added token exchange functionality.

Screenshot from 2022-04-05 21-25-42
Screenshot from 2022-04-05 21-26-06

PR #2583, opened 2022-04-18
Feat usn rebased

esaminu:
Adopted and rebased branch of #2568 

Also:
* Enabled flag on `testnet` for preview purposes
* Fix bugs from https://github.com/near/near-wallet/pull/2568#pullrequestreview-944202624


Patrick1904 (commented):
The UI/UX is quite rough but I understand that we're going to ship a follow-up PR with UI improvements. Some obvious and minor fixes that I noted were:

- ‘Continue’ button enabled despite no NEAR or USN available to swap (results in failed TX)
- Ledger ‘processing’ modal keeps showing after swapping
- Since there’s no confirmation step before swapping (as FT flow), I think the button should say ‘Confirm’ vs ‘Continue’
- Confirmation screen copy says ‘You’ve swapped X NEAR To X USN’. Let’s align with FT flow: ‘You swapped X NEAR to X USN’
- Confirmation screen button says ‘Back to Swap’ - let’s say ‘Continue’ (align with FT flow) that goes to the dashboard (vs to swap page)
- Confirmation screen button says ‘To the main screen’ - let’s say ‘Swap More’

I think it's at least worth fixing the ledger issue before merging the PR

andy-haynes:
I made some quality of life style/formatting changes after reviewing it last night, I think they'll help with long-term maintenance. We can roll them back if they break anything but I didn't notice any new problems when testing. Other than that the PR looks pretty good, the only thing I noticed was the leading `0` that persists even after entering a swap value, i.e. entering a `1` gets rendered in the input as `01`.

Once Patrick's feedback is addressed we should get this into testnet 🎉

packages/frontend/src/components/Routing.js
```diff
@@ -199,7 +200,6 @@ class Routing extends Component {
         } = this.props;
 
         fetchTokenFiatValues();
-        this.startPollingTokenFiatValue();
```
andy-haynes, line 202: Are we no longer polling the `/fiat` endpoint? Should this be feature flagged?

andy-haynes: After talking with @Patrick1904 and @MaximusHaximus we've decided to revert this change and keep fiat polling in `Routing`

Patrick1904 (approved):
LGTM (for testnet)

PR #2590, opened 2022-04-21
fix: fiat selector and text align

esaminu:
- Fix the fallback logic on the fiat value selector to only fallback when fiat data actually exists
- Fix the conditional text alignment on the token list component


andy-haynes (approved):
LGTM 🚀

PR #2593, opened 2022-04-22
feat: added functionality "Donate to Ukraine"

OlegYak:
This is a PR to add the ability to donate to Ukraine
Screenshot from 2022-04-22 14-16-20
.


packages/frontend/src/components/DonateToUkraine/DonateContainer.js
```diff
@@ -0,0 +1,96 @@
...
+            parseNearAmount={parseNearAmount}
+            formatNearAmount={formatNearAmount}
+            fungibleTokens={fungibleTokensList}
+            isMobile={isMobile()}
+            explorerUrl={EXPLORER_URL}
+            showNetworkBanner={SHOW_NETWORK_BANNER}
+            activeView={activeView}
+            setActiveView={(view) => setActiveView(view)}
+            nearTokenFiatValueUSD={nearTokenFiatValueUSD}
+            handleSendToken={async (rawAmount, receiverId, contractName) => {
```
agileurbanite, line 58: nit inline anon functions can be abstracted so we can see common patterns and create reusable abstractions.  This encourages modular code that is easier to test.

packages/frontend/src/components/DonateToUkraine/DonateContainerV2.js
```diff
@@ -0,0 +1,207 @@
...
+    transactionHash,
+    nearTokenFiatValueUSD
+}) => {
+    const [userInputAmount, setUserInputAmount] = useState('');
+    const [isMaxAmount, setIsMaxAmount] = useState(false);
+
+    const [selectedToken, setSelectedToken] = useState(fungibleTokens[0]);
+
+    useEffect(() => {
+        let targetToken = fungibleTokens.find(({ contractName }) =>
```
agileurbanite, line 95: nit: this is hard to read imo might be worth it to break out into a common function as both sides of this Maybe are 80% the same.

packages/frontend/src/components/svg/SwapIconTwoArrows copy.js
```diff
@@ -1,10 +1,26 @@
...
+=======
+>>>>>>> 95d188d3c359b03201b06fb38769a606bf1c7c75
 import React from 'react';
 
 const SwapIconTwoArrows = ({ color, width, height, margin }) => {
     const stroke = color || '#000';
     const currentWidth = width || '20';
     const currentHeight = height || '20';
     // const currentMargin = margin || '0px';
+<<<<<<< HEAD
```
agileurbanite, line 20: remove these merge conflict commits

packages/frontend/src/components/svg/SwapIconTwoArrows copy.js
```diff
@@ -1,10 +1,26 @@
...
+<<<<<<< HEAD
+import React from "react";
+
+const SwapIconTwoArrows = ({ color, width, height, margin }) => {
+    const stroke = color || "#000";
+    const currentWidth = width || "20";
+    const currentHeight = height || "20";
+    const currentMargin = margin || "0px";
+=======
```
agileurbanite, line 10: remove these merge conflict commits

packages/frontend/src/components/svg/SwapIconTwoArrows copy.js
```diff
@@ -1,10 +1,26 @@
+<<<<<<< HEAD
```
agileurbanite, line 1: remove these merge conflict commits

packages/frontend/src/components/wallet/TokenBox.js
```diff
@@ -10,6 +10,56 @@ import Balance from '../common/balance/Balance';
...
+    if (formattedAmount === `0.${'0'.repeat(FRAC_DIGITS)}`) {
+        return `< ${
+            !FRAC_DIGITS ? '0' : `0.${'0'.repeat((FRAC_DIGITS || 1) - 1)}1`
+        }`;
+    }
+    return amountWithCommas(removeTrailingZeros(formattedAmount));
+};
+
+
+// const FRAC_DIGITS = 5;
```
agileurbanite, line 41: why comment, just remove it.

agileurbanite (changes requested):
Overall, I am curious why this PR is so verbose?  Can't we just have a button that sets the state of the component for sender input to the ukraine address?  I feel like this PR introduces unnecessary complexity.

morgsmccauley:
Agree with @agileurbanite here

I don't think we need to add an entire new set of components and route for something that's already very similar to what we have. We have the capability for deep-linking to send tokens to a specific account, let's just utilise that.

I'd suggest we keep `NavLink.js` as it is, i.e. show the "Donate to Ukraine" link in the header, but instead route to `/send-money/ukraine` (where `ukraine` is the accountId). Then we can remove `/donate-to-ukraine` and related code.

This does add an additional step compared to this flow - confirming the receiver ID, but this will be pre-populated with `ukraine` as it's pulled from the URL. This is a bit more work for users but is probably more familiar since it's what they're used to.

agileurbanite:
Closing as duplicate of #2639

PR #2595, opened 2022-04-22
Refactor Active Ledger Transport components

marcinbodnar:
These are the changes requested in #2581 review.


packages/frontend/src/components/navigation/ConnectLedger.js
```diff
@@ -24,11 +29,17 @@ const ConnectLedgerButton = styled.div`
...
+
+    const { pathname } = useSelector(getLocation);
+    const ledgerConnectionAvailable = useSelector(selectLedgerConnectionAvailable);
+    const hasLedger = useSelector(selectLedgerHasLedger);
+
+    const connectLedger =  () => dispatch(handleShowConnectModal());
+    
+    const showConnectLedgerButton = hasLedger || ['sign-in-ledger', 'setup-ledger'].includes(pathname.split('/')[1]);
+
     return showConnectLedgerButton
```
andy-haynes, line 43: 💄 Could we short-circuit the return here instead of returning a ternary?

i.e.
```js
if (!showConnectionLedgerButton) {
  return null;
}

return (
  ...
);
```

packages/frontend/src/translations/en.global.json
```diff
@@ -320,16 +320,22 @@
...
+                "cancelButton": "Cancel",
+                "connectButton": "Connect",
                 "header": "Connect to your Ledger device.",
                 "one": "Make sure your Ledger is connected securely, and that the NEAR app is open on your device."
             },
+            "connection-error": {
+                "cancelButton": "Cancel",
+                "connectButton": "Retry",
+                "header": "Connection unsuccessful.",
+                "one": "Make sure your ledger is connected securely, and that the NEAR app is open on your device."
```
andy-haynes, line 332: Can we please rename these `"one"` keys? `one` is not a descriptive name and although we use it in other places I think it's a habit we should break in order to produce code that is easier to read and debug.

Something like `connectionPrompt` would work well here I think, but really anything that describes what this text indicates is viable.

packages/frontend/src/utils/wallet.js
```diff
@@ -295,7 +295,7 @@ class Wallet {
         }
 
         const { data: recoveryMethods } = await this.getRecoveryMethods();
-        const methodsToRemove = recoveryMethods.filter((method) => method.kind !== 'ledger');
+        const methodsToRemove = (recoveryMethods || []).filter((method) => method.kind !== 'ledger');
```
andy-haynes, line 298: Is this related to your previous changes?

It looks like the destructure on L297 is broken - `this.getRecoveryMethods()` returns an array but it gets destructured like an object with a `data` property. The fallback you're adding will always be required so long as the wrong data type is returned.

The correct fix would appear to be to change L297 to:
```js
const recoveryMethods = await this.getRecoveryMethods();
```

If you want to make the changes to that line then I think it's fine either in this PR or a new one. But let's please remove this fallback and address the root of the issue instead. Once addressed, the fallback should never be required.

andy-haynes (changes requested):
Thanks for cross-referencing feedback from the previous PR - just a couple minor comments inline

packages/frontend/src/components/navigation/ConnectLedger.js
marcinbodnar: Done.

packages/frontend/src/translations/en.global.json
marcinbodnar: Done.

packages/frontend/src/utils/wallet.js
marcinbodnar: Right. It's not related to ledger transport changes. Done.

andy-haynes (approved):
Awesome work!

PR #2591, opened 2022-04-21
update zh translation for swap

think-in-universe:
update Chinese translation for swap feature on behalf of @RubyJiao

PR #2596, opened 2022-04-22
fix: missing zh translation

think-in-universe:
add the missing translation as mentioned here: https://github.com/near/near-wallet/pull/2591#issuecomment-1105913862


think-in-universe:
> Your [Render](https://render.com) PR Server URL is https://near-wallet-pr-2596.onrender.com.
> 
> Follow its progress at https://dashboard.render.com/static/srv-c9hbs6f6d9kgmrqcnutg.

The render preview seems to be broken?

Patrick1904 (approved):

Patrick1904:
> 

I think it was because the tests failed, though not sure why. Checked out locally and works fine.

Issue #2597, opened 2022-04-23
Cannot confirm `FunctionCall` for `aurora` contract with 2FA account


think-in-universe:
**Problem**

In most of the cases, the args of FunctionCall can be deserialized into JSON successfully. However, this is not the case for Aurora Engine. In the `call()` function of Aurora Engine, the parameters are [deserialized as Borsh](https://github.com/aurora-is-near/aurora-engine/blob/5c8691ea6ea5f1b309ef227f7f5c719ffea45d28/engine/src/lib.rs#L226) instead of JSON. 

This caused issue when parsing the FunctionCall args for 2FA account in NEAR Wallet, when the contract ID is `aurora`. 

Example:

1. One 2FA user Alice wanted to make a call to `aurora` contract, she sent the request successfully: https://explorer.testnet.near.org/transactions/GWuEAKJ3LLJ9tGEYkQJLZJtfbVEg42U2e957qii3LxWU
2. Then she wanted to confirm the request with verification code sent to the email / phone. However, the step failed because NEAR wallet cannot parse the arguments into JSON, and the screen became be blank, with the below error message in developer console:

<img width="716" alt="image" src="https://user-images.githubusercontent.com/46699230/164866318-007af966-e497-4cfb-98dd-58f4386b144d.png">

The error was thrown from this line of code: 

https://github.com/near/near-wallet/blob/43cf19a18c2d1f050a6c22fcf1aa6ab914dc0d05/packages/frontend/src/utils/getTranslationsFromMultisigRequest.js#L5

The error only happens to 2FA users, and other users can make the call successfully. 

**Expected Behavior**

2FA users should be able to make a call to `aurora` contract as other contracts. 

**Steps to reproduce**

1. Create a 2FA account on testnet
2. Make a call to `aurora` contract with the 2FA account via Ref Finance test version: https://dev.ref-finance.com/#aurora|usdc.fakes.testnet
3. Swap ETH to USDC and the error will occur

PR #2598, opened 2022-04-23
fix: fallback to string if args cannot be parsed as JSON

think-in-universe:
Fixes: #2597

When the `FunctionCall` args is not JSON, display the raw base64 string, so it won't break the user workflow when the contract is `aurora`.

Please give this high priority if possible, so it won't stop the release of some important features of Ref Finance.


morgsmccauley (approved):
Thanks @think-in-universe!

Also found that we weren't handling long text (such as encoded args) very well so updated the styles to allow scrolling:

before:
image
after:
Screen Shot 2022-04-26 at 9 56 29 AM

PR #2600, opened 2022-04-25
Top up page redesign

andrii4ire:


packages/frontend/src/components/buy/BuyNear.js
```diff
@@ -170,6 +196,23 @@ export function BuyNear({ history }) {
         checkMoonPay();
     }, [accountId]);
 
+    const PayMethods = useMemo(() => {
+        return {
+            moonPay: {
```
agileurbanite, line 201: nit we should abstract this out to a config

agileurbanite (changes requested):
@andrii4ire can you make sure that utorg is included as an option in this PR?

andy-haynes:
Addressed by #2651

PR #2604, opened 2022-04-25
Improvement available wallet balance

andrii4ire:


agileurbanite:
@andrii4ire can you provide more details of what you were trying to accomplish with this PR?  would love to understand the problem you were hoping it can solve.

SerDorosh:
This is a task where “Available Balance” should reflect the fiat sum of all tokens in their wallet that have a fiat conversion available.

PR #2605, opened 2022-04-25
feat(cookie-policy): add cookie policy banner

PinkiNice:
Added cookie policy banner with "Accept" button. 
It shows if user have not accepted cookies yet. 

If he presses "Accept" or "Close" it will set a cookie named "cookie-policy" to a value of "true".


packages/frontend/src/components/cookie-policy/CookiePolicy.js
```diff
@@ -0,0 +1,140 @@
+import React from 'react';
+import styled from 'styled-components';
+
+import cookiePolicyPdf from '../../documents/cookie-policy.pdf';
+import { Cookie } from '../../utils/cookies';
+import { StyledButton } from '../common/Button';
+
+// Copypasted close icon from modal/CloseButton.js
+const CloseIcon = () => (<svg width="35" height="35" viewBox="0 0 35 35" fill="none" xmlns="http://www.w3.org/2000/svg">
```
roshaans, line 9: I think it would be cleaner and more future proof to create a `CloseIcon.js` file inside `packages/frontend/src/components/svg/` and import from that folder.

roshaans (commented):
Were there any design specs for this? I suppose it would be nice to have this be looked by the design team before merging because for mobile users, it takes up more than half the screen when the website is first loaded. 

Also, I verified that changes work as intended. Just some minor nits that would clean this up as well as improve performance a tad bit:)

PR #2607, opened 2022-04-25
fix: support ipfs:// base_uri

kalloc:
I don't really understand, is it useful or not.
I think base_uri may be equal to ipfs://

### DIFF
--- a/packages/frontend/src/services/NonFungibleTokens.js
+++ b/packages/frontend/src/services/NonFungibleTokens.js
@@ -99,7 +99,7 @@ export default class NonFungibleTokens {
             return media;
         }
 
-        if (base_uri) {
+        if (base_uri && base_uri != 'ipfs://') {
             return `${base_uri}/${media}`;
         }
 


morgsmccauley:
Thanks for the contribution @kalloc!

Do you have examples of dApps using the `ipfs://` scheme? From what i've seen, most will specify an IPFS public gateway. It would be good to learn more about this use case + test this out before landing :)

andy-haynes:
Closing for inactivity. Please feel free to re-open to resume discussion or file an issue instead if there is a question around functionality.

kalloc:
> Thanks for the contribution @kalloc!
> 
> Do you have examples of dApps using the `ipfs://` scheme? From what i've seen, most will specify an IPFS public gateway. It would be good to learn more about this use case + test this out before landing :)

hi.

You are totally right, but you miss that every developers follow wallet/paras as the main UI.
So, even wallet don't support ipfs:// schema, nobody will support baseUri with ipfs://
Hard-coding gateway, may follow to making that NFTs unreachable in future (when gateway died)

PR #2614, opened 2022-04-26
BUGFIX: error when loading profile page on wallet see screenshot

gutsyphilip:
In this PR:
- Fixes error with missing method call when accessing lockup profiles

image


agileurbanite (approved):
LGTM

PR #2616, opened 2022-04-26
fix: key farmRewards by accountId

esaminu:
Bug fix to correctly display accumulated farm rewards for lockup accounts as well as main account when the radio toggle is visible. Also refetch unclaimed rewards when the selected account changes to a lockup account on the staking dashboard.

This can be tested by deploying a lockup using https://testnet.lockup.tech/ on the deploy preview

cc: @askulikov

### DIFF
--- a/packages/frontend/src/components/staking/StakingContainer.js
+++ b/packages/frontend/src/components/staking/StakingContainer.js
@@ -214,7 +214,7 @@ export function StakingContainer({ history, match }) {
         
         validators
             .filter((validator) => validator.version === FARMING_VALIDATOR_VERSION)
-            .forEach((validator) => dispatch(getValidatorFarmData(validator, currentAccount.accountId)));
+            .forEach((validator) => dispatch(getValidatorFarmData({ validator, accountId: currentAccount.accountId })));
     }, [currentAccount.accountId, validators]);
 
     const handleAction = async (action, validator, amount) => {
--- a/packages/frontend/src/components/staking/StakingContainer.js
+++ b/packages/frontend/src/components/staking/StakingContainer.js
@@ -210,7 +210,7 @@ export function StakingContainer({ history, match }) {
     };
 
     useEffect(() => {
-        if (!accountId || !validators.length) return;
+        if (!currentAccount.accountId || !validators.length) return;
         
         validators
             .filter((validator) => validator.version === FARMING_VALIDATOR_VERSION)
--- a/packages/frontend/src/components/staking/StakingContainer.js
+++ b/packages/frontend/src/components/staking/StakingContainer.js
@@ -214,8 +214,8 @@ export function StakingContainer({ history, match }) {
         
         validators
             .filter((validator) => validator.version === FARMING_VALIDATOR_VERSION)
-            .forEach((validator) => dispatch(getValidatorFarmData(validator, accountId)));
-    }, [accountId, validators]);
+            .forEach((validator) => dispatch(getValidatorFarmData(validator, currentAccount.accountId)));
+    }, [currentAccount, validators]);
 
     const handleAction = async (action, validator, amount) => {
         let id = Mixpanel.get_distinct_id();
--- a/packages/frontend/src/components/staking/StakingContainer.js
+++ b/packages/frontend/src/components/staking/StakingContainer.js
@@ -215,7 +215,7 @@ export function StakingContainer({ history, match }) {
         validators
             .filter((validator) => validator.version === FARMING_VALIDATOR_VERSION)
             .forEach((validator) => dispatch(getValidatorFarmData(validator, currentAccount.accountId)));
-    }, [currentAccount, validators]);
+    }, [currentAccount.accountId, validators]);
 
     const handleAction = async (action, validator, amount) => {
         let id = Mixpanel.get_distinct_id();
--- a/packages/frontend/src/components/staking/components/BalanceBox.js
+++ b/packages/frontend/src/components/staking/components/BalanceBox.js
@@ -124,7 +124,7 @@ export default function BalanceBox({
                     <div className='title'>
                         {title && <Translate id={title}/>}
                         {info && <Tooltip translate={info}/>}
-                        {loading && <span className="animated-dots" style={{width: 16}}/>}
+                        {loading && <span className="animated-dots" />}
                     </div>
                 }
                 <div className='token-balance'>
--- a/packages/frontend/src/components/staking/components/BalanceBox.js
+++ b/packages/frontend/src/components/staking/components/BalanceBox.js
@@ -124,6 +124,7 @@ export default function BalanceBox({
                     <div className='title'>
                         {title && <Translate id={title}/>}
                         {info && <Tooltip translate={info}/>}
+                        {loading && <span className="animated-dots" style={{width: 16}}/>}
                     </div>
                 }
                 <div className='token-balance'>
--- a/packages/frontend/src/redux/actions/staking.js
+++ b/packages/frontend/src/redux/actions/staking.js
@@ -574,7 +574,7 @@ export const getValidatorFarmData = (validator, accountId) => async (dispatch, g
 
     const farmData = {
         poolSummary: {...poolSummary},
-        farmRewards: farmList,
+        farmRewards: {[accountId]: farmList},
     };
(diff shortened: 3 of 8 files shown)


packages/frontend/src/components/staking/StakingContainer.js
```diff
@@ -214,8 +214,8 @@ export function StakingContainer({ history, match }) {
         
         validators
             .filter((validator) => validator.version === FARMING_VALIDATOR_VERSION)
-            .forEach((validator) => dispatch(getValidatorFarmData(validator, accountId)));
-    }, [accountId, validators]);
+            .forEach((validator) => dispatch(getValidatorFarmData(validator, currentAccount.accountId)));
```
morgsmccauley, line 217: This is the real issue here - we were fetching the unclaimed rewards with the wallet account, which doesn't change when selecting a different staking account, and therefore doesn't fetch the relevant validator info.

Using `currentAccount` is really the only change necessary to fix this issue, I don't think we need to go as far as updating redux to key `farmRewards` by `accountId`. Every time we switch staking accounts we fetch fresh validator info and replace the old (account) state, so there isn't must value in keeping it around. Instead, we can just fire this action with the selected staking account (`currentAccount.accountId`) and expect that `farmRewards` has been updated with selected accounts info.

I'd suggest we remove the redux and related changes and just land the changes in this file, what do you think?

packages/frontend/src/components/staking/StakingContainer.js
```diff
@@ -214,8 +214,8 @@ export function StakingContainer({ history, match }) {
...
+    }, [currentAccount, validators]);
```
morgsmccauley, line 218:
```suggestion
    }, [currentAccount.accountId, validators]);
```
It's good to specify individual values rather than passing whole objects so we avoid running this effect when other values within the object we don't care about change.

packages/frontend/src/components/staking/StakingContainer.js
morgsmccauley: We should also update the `if` condition above to check `currentAccount.accountId` rather than `accountId` since we're no longer using that.

morgsmccauley (changes requested):
Good job on this and thanks for walking me through it 😄. I think we may be doing more than we need here, added a comment inline explaining my reasoning behind this.

packages/frontend/src/components/staking/StakingContainer.js
esaminu: Thanks for catching the `if` condition update.

I kept the `farmRewards` for the other account around for UX reasons primarily. When you toggle between the 2 it would display the value for the other account or remove the whole "Available to Claim" node if we clear it which seemed pretty glitchy to me. The alternative is to render a placeholder on a loading state but I think the last loaded state is even better. We could do both i.e. use the placeholder for when we don't have a last loaded value for the first fetch.

morgsmccauley: > remove the whole "Available to Claim" node

I think this is actually related to having specific `farmRewards` state for each account. Since we only load the rewards for the currently selected account (`currentAccount.accountId`), when we switch account, we are attempting to index the `farmRewards` for the newly selected account, which doesn't exist yet and therefore causes the rewards UI to not render. Additionally, when `farmRewards` for both accounts exist (both have been selected and therefore state exists for both), when switching again the rewards will change twice, first to the existing state of the rewards, and second when the fetch to get up to date rewards finishes. This gif displays both issues:

<details>
  <summary>Current PR implementation</summary>

  
</details>

Without the additional redux logic we don't have either of these issues. `farmRewards` is always populated (after first load) so it will never be removed from the UI, and the value only changes once on select. Though, the issue with this  is that it is a little delayed, ideally we could have a placeholder/loading UI for the rewards value but that's probably out of scope for this PR. You can see what I mean in this gif:

<details>
  <summary>Proposed implementation</summary>


</details>

Overall, I think we end up with a better UX without the redux logic, thoughts?

esaminu: Both implementations have the same issue on first load. The difference is whether we display the last loaded for the currently selected account or the previously selected account. The values for both accounts are likely to be highly out of sync and keep in mind there would be multiple tokens the user has available to claim.

#### Currently it's:

**first load:** empty **&rarr;** fetched current account's tokens
**toggle:** other account's tokens **&rarr;** fetched current account's tokens

#### In this PR it's:

**first load:** empty **&rarr;** current account's tokens
**toggle:** other account's tokens **&rarr;** last loaded current account's tokens **&rarr;**  current account's tokens

#### I think we should do:

**first load:** empty **&rarr;**  placeholder **&rarr;**  current account's tokens
**toggle:** other account's tokens **&rarr;**  last loaded current account's tokens **&rarr;**  current account's tokens

The best we can do without the redux changes is

**first load:** empty **&rarr;**  placeholder **&rarr;**  current account's tokens
**toggle:** placeholder **&rarr;**  current account's tokens

Lmk what you think

esaminu: Updated the PR with a loading state per discussion @morgsmccauley . It's worth highlighting a point we uncovered on this thread as well:
We cannot render a loading state before all farm validators have been queried because it is possible that the user does not have any unclaimed rewards in which case the UI will be collapsed which is not ideal.

packages/frontend/src/redux/actions/staking.js
```diff
@@ -553,8 +551,7 @@ export const handleUpdateCurrent = (accountId) => async (dispatch, getState) =>
     dispatch(staking.updateCurrent({ currentAccount }));
 };
 
-export const getValidatorFarmData = (validator, accountId) => async (dispatch, getState) => {
-
+export const getValidatorFarmData = createAsyncThunk('staking/setValidatorFarmData', async ({ validator, accountId }, { dispatch }) => {
```
morgsmccauley, line 554:
```suggestion
export const getValidatorFarmData = createAsyncThunk('staking/getValidatorFarmData', async ({ validator, accountId }, { dispatch }) => {
```
typo?

morgsmccauley, line 554:
```suggestion
export const getValidatorFarmData = createAsyncThunk('staking/getValidatorFarmData', async ({ validator, accountId }, { dispatch }) => {
```
typo?
```suggestion
export const getValidatorFarmData = createAsyncThunk('staking/setValidatorFarmData', async ({ validator, accountId }, { dispatch }) => {
```

packages/frontend/src/components/staking/components/BalanceBox.js
```diff
@@ -124,6 +124,7 @@ export default function BalanceBox({
                     <div className='title'>
                         {title && <Translate id={title}/>}
                         {info && <Tooltip translate={info}/>}
+                        {loading && <span className="animated-dots" style={{width: 16}}/>}
```
morgsmccauley, line 127: Just curious why we need to set the `width`? It doesn't seem to make any noticeable difference when removing it.

morgsmccauley, line 127: Just curious why we need to set the `width`? It doesn't seem to make any noticeable difference when removing it.

morgsmccauley (commented):
Couple small questions otherwise looking good!

packages/frontend/src/components/staking/StakingContainer.js
morgsmccauley: Nice job 👍🏽 the loading looks good, the initial loading state isn't ideal but like you said, it doesn't make sense for users who don't have rewards :)

packages/frontend/src/redux/actions/staking.js
esaminu: Either works but was intended to replace the removed action of the same name 😄 . I guess since there's a fetch this time `getValidatorFarmData` makes more sense 👍

packages/frontend/src/components/staking/components/BalanceBox.js
esaminu: Removed 👍

esaminu: Removed and tested 👍

morgsmccauley (approved):
LGTM, good job :)

PR #2623, opened 2022-04-28
Wal 277 withdrawing lockup funds issue

esaminu:
This PR adds an additional condition in `isAccDeletable` for lockup accounts to check for unclaimed token rewards. `isAccDeletable` will affect whether the "Available to transfer" for the lockup includes the storage amount (3.5 / 35 Ⓝ) and whether we delete the lockup when withdrawing from it.

### DIFF
--- a/packages/frontend/src/redux/crossStateSelectors/selectCollectedAvailableForClaimData.js
+++ b/packages/frontend/src/redux/crossStateSelectors/selectCollectedAvailableForClaimData.js
@@ -71,7 +71,7 @@ const selectCollectedAvailableForClaimData = createSelector(
     }
 );
 
-export const selectCollectedAvailableForClaimDataForAccountId = createSelector(
+export const selectCollectedAvailableForClaimDataByAccountId = createSelector(
     [
         selectValidatorsFarmData,
         selectAllContractMetadata,
--- a/packages/frontend/src/redux/crossStateSelectors/selectCollectedAvailableForClaimData.js
+++ b/packages/frontend/src/redux/crossStateSelectors/selectCollectedAvailableForClaimData.js
@@ -46,7 +46,7 @@ const collectFarmingData = (args) => {
     }
 };
 
-export default createSelector(
+const selectCollectedAvailableForClaimData = createSelector(
     [
         selectValidatorsFarmData,
         selectAllContractMetadata,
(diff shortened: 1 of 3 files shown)


packages/frontend/src/services/StakingFarmContracts.js
```diff
@@ -46,4 +60,34 @@ export default class StakingFarmContracts {
...
+        limit,
+    }) => {
+        return (
+            StakingFarmContracts.isFarmingValidator(contractName) &&
+            (await StakingFarmContracts.getFarmListWithUnclaimedRewards({
+                contractName,
+                account_id,
+                from_index,
+                limit,
+            }).then(
```
morgsmccauley, line 86:
```suggestion
            StakingFarmContracts.getFarmListWithUnclaimedRewards({
                contractName,
                account_id,
                from_index,
                limit,
            }.then(
```
`await` is unnecessary here

packages/frontend/src/services/StakingFarmContracts.js
```diff
@@ -46,4 +60,34 @@ export default class StakingFarmContracts {
...
+                accountId
+            ) === FARMING_VALIDATOR_VERSION
+        );
+    }
+
+    static hasUnclaimedRewards = async ({
+        contractName,
+        account_id,
+        from_index,
+        limit,
```
morgsmccauley, line 77:
```suggestion
        from_index = 0,
        limit = 300,
```
Are these likely to change per use-case? Should we default them here instead of passing them at the call-site?

packages/frontend/src/services/StakingFarmContracts.js
```diff
@@ -1,12 +1,26 @@
...
     // View functions are not signed, so do not require a real account!
-    static viewFunctionAccount = wallet.getAccountBasic('dontcare');
+    static viewFunctionAccount = new Account(
+        Connection.fromConfig({
+            networkId: NETWORK_ID,
+            provider: {
+                type: 'JsonRpcProvider',
+                args: { url: NODE_URL + '/' },
+            },
+            signer: {}
```
morgsmccauley, line 20: Why did this change? to remove the `signer`?

packages/frontend/src/utils/account-with-lockup.js
```diff
@@ -266,6 +275,12 @@ async function getAccountBalance(limitedAccountData = false) {
         let totalBalance = new BN(lockupBalance.total);
         let stakedBalanceLockup = new BN(0);
         const stakingPoolLockupAccountId = await this.wrappedAccount.viewFunction(lockupAccountId, 'get_staking_pool_account_id');
+        let hasUnclaimedTokenBalance = stakingPoolLockupAccountId && await StakingFarmContracts.hasUnclaimedRewards({
```
morgsmccauley, line 278:
```suggestion
        const hasUnclaimedTokenBalance = stakingPoolLockupAccountId && await StakingFarmContracts.hasUnclaimedRewards({
```
nit: `const`

packages/frontend/src/redux/crossStateSelectors/selectCollectedAvailableForClaimData.js
```diff
@@ -70,3 +70,36 @@ export default createSelector(
         });
     }
 );
+
+export const selectCollectedAvailableForClaimDataForAccountId = createSelector(
```
morgsmccauley, line 74:
```suggestion
export const selectCollectedAvailableForClaimDataByAccountId = createSelector(
```
This is a bit of a mouthful 😅  - I feel like this makes it a bit clearer, thoughts?

morgsmccauley (changes requested):
Good job on this :D just a few small comments!

packages/frontend/src/services/StakingFarmContracts.js
esaminu: It was to address some require cycles that were causing some imports to be `undefined`

packages/frontend/src/utils/account-with-lockup.js
esaminu: ✅

packages/frontend/src/redux/crossStateSelectors/selectCollectedAvailableForClaimData.js
esaminu: 😆 yeah its less repetitive

morgsmccauley (approved):
LGTM :)

PR #2626, opened 2022-04-29
add: added USN Banner

OlegYak:
add: added USN Released Banner !


gutsyphilip:
I have gone through the updates and it all looks good from a code perspective. I have nudged designers on the team for a design review.

cc: @corwinharrell @jc-near

corwinharrell:
image

After taking a look at the designs above from this work as shared by @gutsyphilip, I have a few concerns:

1. From a strategic perspective, this borders on advertising. I would question whether or not we want to be leveraging the UI in order to push users to purchase tokens other than the native token that allows them to actually use the wallet and/or dApps. There is some liability introduced in doing this which I'm not sure is something that we want to assume.

2. There is actually already a banner patterns that exists that is similar to this. Is this a new pattern or is it reusing an existing components. I realize that in general there is a lack of documentation around component usage and best practices, but this component is typically reserved for status indicators such as errors or issues related to the network. If we do end up going the route of advertising things like tokens launches, we may want to do so in a way that differs from the patterns that we rely on to communicate application and network state to the user.

PR #2632, opened 2022-05-03
refactor: disable usn on mainnet

agileurbanite:
refactor: disable usn on mainnet

### DIFF
--- a/features/flags.json
+++ b/features/flags.json
@@ -158,17 +158,17 @@
       "lastEditedAt": "2022-05-02T17:12:41.421Z"
     },
     "mainnet": {
-      "enabled": true,
+      "enabled": false,
       "lastEditedBy": "ignatKorokov",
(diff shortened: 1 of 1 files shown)


features/flags.json
```diff
@@ -158,17 +158,17 @@
...
     },
     "mainnet_STAGING": {
-      "enabled": true,
+      "enabled": false,
       "lastEditedBy": "ignatKorokov",
       "lastEditedAt": "2022-05-02T17:12:41.421Z"
     },
     "testnet_STAGING": {
-      "enabled": true,
+      "enabled": false,
```
andy-haynes, line 171: Not necessary to disable this, until it has a dedicated environment it should just be the same as `testnet`.

andy-haynes (approved):
LGTM

In the future we should avoid shipping initial feature flag PRs enabled in all environments.

gutsyphilip (approved):

PR #2635, opened 2022-05-05
Add nearorg environments

esaminu:
This PR:

1. Adds `testnet_NEARORG`, `testnet_STAGING_NEARORG`, `mainnet_NEARORG`, `mainnet_STAGING_NEARORG` environments for use on `*wallet.near.org`
2. Enables `CREATE_USN_CONTRACT` feature on `development`, `testnet`, `mainnet`, `mainnet_STAGING` and `testnet_STAGING` environments
3. Sets the `USE_CLOUDFLARE` environment variable to `false` on netlify.toml. This environment has been set to true via the netlify UI for https://wallet.near.org and https://wallet.staging.near.org 

The `NEAR_WALLET_ENV` needs to be updated to the new NEARORG environments on netlify and render once this is merged


features/flags.json
```diff
@@ -26,35 +26,75 @@
...
-      "lastEditedAt": "2022-04-26T08:16:00:UTC"
+      "enabled": true,
+      "lastEditedBy": "esaminu",
+      "lastEditedAt": "2022-05-05T18:11:03.325Z"
     },
     "mainnet": {
-      "enabled": false,
-      "lastEditedBy": "agileurbanite ",
-      "lastEditedAt": "2022-04-26T08:16:00:UTC"
+      "enabled": true,
```
andy-haynes, line 65: I don't think this PR shouldn't be modifying any existing flag states. Since this needs to be landed before we can start using the `*_NEARORG` we run the risk of inadvertently enabling flags in existing environments. Better to enable flags in the current environments once we have all `*.near.org` frontend sites using the `*_NEARORG` environment (and update the configuration repositories so that it can't be automatically reverted to their original values). We just need to be absolutely sure that this feature can't be reenabled as a side effect of this migration.

features/flags.json
```diff
@@ -26,35 +26,75 @@
...
+      "lastEditedBy": "esaminu",
+      "lastEditedAt": "2022-05-05T18:07:28.665Z"
+    },
+    "testnet_NEARORG": {
+      "enabled": true,
+      "lastEditedBy": "esaminu",
+      "lastEditedAt": "2022-05-05T18:07:50.110Z"
+    },
+    "testnet_STAGING_NEARORG": {
```
andy-haynes, line 45: `testnet_STAGING` doesn't really need to be duplicated since it won't be supported by any sites under `*wallet.near.org`. It really only exists now so that I can deploy to that environment with the new CI/CD pipeline, but since the domain migration takes priority I'm not going to bother deploying under the `*.near.org` domain. It's safe to ignore as part of this work 🙂

packages/frontend/netlify.toml
```diff
@@ -23,13 +23,13 @@
     TOKEN_CONTRACTS = 'berryclub.ek.near,wrap.near,6b175474e89094c44da98b954eedeac495271d0f.factory.bridge.near'
     RECAPTCHA_ENTERPRISE_SITE_KEY = "6LcpJ3EcAAAAAFgA-nixKFNGWMo9IG9FQhH4XjSY"
     RECAPTCHA_CHALLENGE_API_KEY = "6LeRzswaAAAAAGeS7mSasZ1wDcGnMcH3D7W1gy1b"
-    USE_CLOUDFLARE = "true"
+    USE_CLOUDFLARE = "false"
```
andy-haynes, line 26: Don't we want to leave this enabled for the existing configurations and only apply it to the new Netlify environments?

andy-haynes (changes requested):
Looking good, just a few changes we'll want to make before going ahead with this

features/flags.json
esaminu: Good point, disabled 👍

features/flags.json
esaminu: deleted the environment 👍

packages/frontend/netlify.toml
esaminu: The issue is that `netlify.toml` [will override any envs we set on the UI](https://docs.netlify.com/configure-builds/environment-variables/#app) so I enabled it on the Netlify UI for the current environments and added an explicit disable on our `netlify.toml`

andy-haynes: If `netlify.toml` overrides the site configuration, then wouldn't this be disabling it for `*.near.org` where we want it to remain on?

Is it possible to specify a new configuration in `netlify.toml` for the new sites or are we limited to `[build.environment]` and `[context.production.environment]`?

esaminu: > If netlify.toml overrides the site configuration, then wouldn't this be disabling it for *.near.org where we want it to remain on?

Good point, it means I cant set the explicit `false`. I removed it from the toml

> Is it possible to specify a new configuration in netlify.toml for the new sites or are we limited to [build.environment] and [context.production.environment]

I was looking for this on [the docs](https://docs.netlify.com/configure-builds/file-based-configuration/) but I couldn't find anything

andy-haynes: I just realized that now `USE_CLOUDFLARE=true` means that we aren't using Cloudflare 😂 

Good call setting it explicitly in Netlify, it'll be nice when we can remove all references to it 🎉

andy-haynes:
@esaminu I snuck in a new feature flag (`USE_INDEXER_SERVICE`) after you created this PR - it's not showing up as a conflict but I think it will break once the `*NEARORG` environments try to reference it. Would you mind please rebasing and adding the new envs to the flag?

andy-haynes (approved):
LGTM 🚀

PR #2636, opened 2022-05-05
Seed phrase import: enable import of zero balance account

Patrick1904:
Referencing issues:
https://nearinc.atlassian.net/browse/WAL-193
https://nearinc.atlassian.net/browse/WAL-191

This PR enables import of a 'zero balance account' when importing an account with seed phrase on the `/recover-seed-phrase` route.

The wallet first tries to find all associated accounts using the provided seed phrase, and shows a modal with 'no accounts found' in case of no active accounts found. The user is then presented with a button to 'import anyway', which converts the seed phrase into an implicit account and imports it into the wallet.

The wallet runs a check against the provided seed phrase using [bip39-light](https://www.npmjs.com/package/bip39-light) to ensure that the provided seed phrase is valid. The wallet shows the error toast in case the provided seed phrase is:
1. Less than 12 words
2. Invalid as per bip-39-light standard
Screen Shot 2022-05-05 at 1 57 28 PM
Screen Shot 2022-05-05 at 1 57 47 PM
Screen Shot 2022-05-05 at 1 58 19 PM

### DIFF
--- a/packages/frontend/src/components/accounts/RecoverAccountSeedPhrase.js
+++ b/packages/frontend/src/components/accounts/RecoverAccountSeedPhrase.js
@@ -1,10 +1,13 @@
 import { getRouter } from 'connected-react-router';
+import { KeyPair } from 'near-api-js';
+import { parseSeedPhrase } from 'near-seed-phrase';
 import { parse as parseQuery, stringify } from 'query-string';
 import React, { Component } from 'react';
 import { Translate } from 'react-localize-redux';
 import { connect } from 'react-redux';
 import { withRouter } from 'react-router-dom';
 import styled from 'styled-components';
+
 
 import { Mixpanel } from '../../mixpanel/index';
 import {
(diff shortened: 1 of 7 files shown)


yarn.lock
```diff
@@ -2057,15 +2057,6 @@
     call-me-maybe "^1.0.1"
     glob-to-regexp "^0.3.0"
 
-"@near-wallet/feature-flags@^0.0.4":
```
morgsmccauley, line 2060: This lock update doesn't seem quite right 🤔  

Weird that it's trying to remove `feature-flags`, maybe `yarn` get's confused when updating `packages/frontend/package.json` since `feature-flags` actually exists in the root `package.json`?

I think we can just revert this file change, I can see that `bip38-light@1.0.7` already exists and we don't want to remove these packages.

packages/frontend/src/translations/en.global.json
```diff
@@ -1702,7 +1707,8 @@
             "error": "Import of your account failed. Please try again or contact support for assistance."
         },
         "recoverAccountSeedPhrase": {
-            "errorInvalidSeedPhrase": "No accounts were found for this passphrase."
+            "errorInvalidSeedPhrase": "No accounts were found for this passphrase.",
+            "errorSeedPhraseNotValid": "The provided seed phrase is not valid. Please check your seed phrase and try again."
```
morgsmccauley, line 1711: Should we be using "passphrase" instead of "seed phrase" here?

packages/frontend/src/components/accounts/RecoverAccountSeedPhrase.js
```diff
@@ -127,6 +152,21 @@ class RecoverAccountSeedPhrase extends Component {
                         handleChange={this.handleChange}
                     />
                 </form>
+                {showCouldNotFindAccountModal && (
+                    <CouldNotFindAccountModal
+                        onClickImport={async () => {
+                            const { secretKey } = parseSeedPhrase(seedPhrase);
+                            const recoveryKeyPair = KeyPair.fromString(secretKey);
+                            const implicitAccountId = Buffer.from(recoveryKeyPair.publicKey.data).toString('hex');
+                            await wallet.importZeroBalanceAccount(implicitAccountId, recoveryKeyPair);
```
morgsmccauley, line 161: What's the likelihood of this failing? Do we need to catch and show an alert?

morgsmccauley (changes requested):
Couple small questions :)

yarn.lock
Patrick1904: Good catch! Reverted

packages/frontend/src/components/accounts/RecoverAccountSeedPhrase.js
Patrick1904: Small but worth handling the error :)

morgsmccauley (approved):
LGTM!

PR #2637, opened 2022-05-05
feat(indexer): indexer service

andy-haynes:
This PR adds an indexer service responsible for calling out to the contract helper REST endpoints responsible for querying the indexer. Aside from consolidating this logic in one module, this change introduces a new set of service URLs pointed to the new backend indexer service which has been separated from the contract helper for performance and stability concerns.

### DIFF
--- a/packages/frontend/src/components/navigation/NavLinks.js
+++ b/packages/frontend/src/components/navigation/NavLinks.js
@@ -146,7 +146,8 @@ const NavLinks = () => (
             }
             target='_blank' 
             className="usn-button"
-            onClick={() => Mixpanel.track('Click Buy USN')} rel="noreferrer"
+            onClick={() => Mixpanel.track('Click Buy USN')}
+            rel="noreferrer"
             >
             <div>
                 <img src={USN_LOGO} alt='open-link'></img>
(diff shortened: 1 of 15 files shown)


packages/frontend/src/utils/helper-api.js
```diff
@@ -2,13 +2,18 @@
 import * as nearApiJs from 'near-api-js';
 import { parseSeedPhrase } from 'near-seed-phrase';
 
+import { USE_INDEXER_SERVICE } from '../../../../features';
 import { ACCOUNT_HELPER_URL } from '../config';
+import { listAccountsByPublicKey } from '../services/indexer';
 
 export let controller;
```
andy-haynes, line 9: As far as I can tell this doesn't get used, so I've omitted it from the new call.

packages/frontend/src/components/navigation/NavLinks.js
```diff
@@ -146,7 +146,7 @@ const NavLinks = () => (
             }
             target='_blank' 
             className="usn-button"
-            onClick={() => Mixpanel.track('Click Buy USN')}
+            onClick={() => Mixpanel.track('Click Buy USN')} rel="noreferrer"
```
morgsmccauley, line 149:
```suggestion
            onClick={() => Mixpanel.track('Click Buy USN')}
            rel="noreferrer"
```

morgsmccauley (commented):
just one nit :)

packages/frontend/src/components/navigation/NavLinks.js
andy-haynes: Heh didn't even notice, I only included it to fix the lint break 🙄

PR #2640, opened 2022-05-06
fix: Feature flag 'import zero balance account' feature for testnet

Patrick1904:


packages/frontend/src/components/accounts/RecoverAccountSeedPhrase.js
```diff
@@ -107,9 +107,13 @@ class RecoverAccountSeedPhrase extends Component {
                 await recoverAccountSeedPhrase(seedPhrase);
                 await refreshAccount();
             }, (e) => {
-                if (e.message.includes('Cannot find matching public key')) {
-                    this.setState({ showCouldNotFindAccountModal: true });
+
+                if (IMPORT_ZERO_BALANCE_ACCOUNT) {
+                    if (e.message.includes('Cannot find matching public key')) {
```
agileurbanite, line 112: should we use error codes in the future?

Patrick1904: Yep will address this in a follow-up PR for import zero balance with link/ledger

PR #2644, opened 2022-05-06
fix: add missing flags

esaminu:
Add missing feature flags for new environments to flags.json

### DIFF
--- a/features/flags.json
+++ b/features/flags.json
@@ -27,6 +27,11 @@
       "lastEditedBy": "Andy Haynes",
       "lastEditedAt": "2022-01-31T21:13:09.948Z"
     },
+    "testnet_NEARORG": {
+      "enabled": true,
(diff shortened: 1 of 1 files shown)


features/flags.json
```diff
@@ -246,6 +246,21 @@
       "enabled": true,
       "lastEditedBy": "Patrick Tajima",
       "lastEditedAt": "2022-05-06T17:30:49.430Z"
+    },
+    "testnet_NEARORG": {
+      "enabled": false,
```
andy-haynes, line 251: This should be `true` to match `testnet`

andy-haynes (changes requested):
Just one value that doesn't line up otherwise LGTM

features/flags.json
esaminu: great catch! thanks 👍

andy-haynes (approved):

PR #2645, opened 2022-05-07
fix: review and update Vietnamese translation

that83:
Some Vietnamese sentences was not translated correctly or was not translated so we have updated them. (Reviewed by CuongDCDev)


Patrick1904 (approved):
Thanks @that83!

PR #2650, opened 2022-05-10
Feat: Coingecko call directly from frontend wrapped with dataloader and node-cache

judith-Near:
Screen Shot 2022-05-06 at 1 50 57 PM

Dataloader implementation modified from [near-contract-helper](https://github.com/near/near-contract-helper/blob/b43422e2096fe0787c36e135ad11b831c9135689/FiatValueManager.js#L29)


packages/frontend/package.json
```diff
@@ -12,6 +12,7 @@
     "@sentry/browser": "^6.4.1",
     "bip39-light": "^1.0.7",
     "connected-react-router": "^6.6.0",
+    "dataloader": "^2.1.0",
```
agileurbanite, line 15: did we need to introduce this new package, how were we handling fetch requests before?

packages/frontend/src/components/wallet/Wallet.js
```diff
@@ -39,6 +39,14 @@ const StyledContainer = styled(Container)`
             margin-top: -15px;
         }
     }
+    
+    .coingecko {
+        color: #B4B4B4;
+        line-height: 150%;
```
agileurbanite, line 45: why the `line-height: 150%;`?

packages/frontend/package.json
judith-Near: this new package is specifically to cache and batch on the frontend any fetch requests that we make. this dependency was originally in near-contract-helper, because it gave us caching and batching/coalescing requests out of the box, so we're just moving it to the frontend.

packages/frontend/package.json
```diff
@@ -32,6 +33,7 @@
     "near-api-js": "^0.43.1",
     "near-ledger-js": "^0.2.1",
     "near-seed-phrase": "^0.2.0",
+    "node-cache": "^5.1.2",
```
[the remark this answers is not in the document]
judith-Near: i believe that the repo is using redux as a cache of sorts, ie functions that reference "getting from the cache" are just pulling from the redux store. node-cache is needed to specify a TTL on the dataloader cache, because dataloader out of the box is only a short lived per request cache. we want something with a little longer TTL (30 seconds). this is also a dependency pulled from near-contract-helper.

packages/frontend/src/components/wallet/Wallet.js
```diff
@@ -494,6 +502,7 @@ const FungibleTokens = ({
                             />
                         </>
                     )}
+                    <div className="coingecko"><Translate id="poweredByCoinGecko"/></div>
```
Patrick1904, line 505: nit: let's try to stick with single quotes when possible just for consistency sake

packages/frontend/src/components/wallet/Wallet.js
```diff
@@ -39,6 +39,14 @@ const StyledContainer = styled(Container)`
             margin-top: -15px;
         }
     }
+    
+    .coingecko {
+        color: #B4B4B4;
+        line-height: 150%;
+        align-self: end;
+        padding: 10px 20px;
```
Patrick1904, line 47: nit: can we please use margin instead of padding for spacing around the div here, and fix alignment on smaller screens. Something like this should be fine:

```
margin: 20px;
@media (max-width: 991px) {
    margin: -25px 0 25px 0;
}
```

packages/frontend/src/components/wallet/Wallet.js
Patrick1904: We can skip the line-height if we apply the [margin](https://github.com/near/near-wallet/pull/2650/files#r870748785)

PR #2655, opened 2022-05-10
test(flags): verify that every feature flag specifies every environment

andy-haynes:
This PR adds a test to ensure that all feature flags have a key for every valid environment.


morgsmccauley (commented):
Is this actually run in travis/jenkins?

andy-haynes:
> Is this actually run in travis/jenkins?

Only Jenkins, we'll need it to short circuit the deployment since one could merge it in this state without conflicts

morgsmccauley (approved):

PR #2661, opened 2022-05-11
Custom HD Path for account creation and recovery

gutsyphilip:
Reverts near/near-wallet#2658

Re-adds custom hd path implementation from #2488


marcinbodnar (approved):
Works well. Just one comment, not a blocker.

PR #2621, opened 2022-04-28
feat: FIFS-174 Wrap/Unwrap

SerDorosh:
Wrap / Unwrap NEAR feature

PR #2663, opened 2022-05-11
fix: enable and fix USN with copy update

esaminu:
This PR:

- Enables USN feature on non `_NEARORG` environments
- Re-adds the USN route (removed in #2621)
- Updates the USN route to `/swap-usn` from `/swap-money`
- Updates the copy on the Swap button, Swap navlink and `swap-usn` title to "Swap USN" from "Swap" to avoid duplicate


agileurbanite (approved):
LGTM

PR #2666, opened 2022-05-12
Custom hd path fix

marcinbodnar:
1. open incognito browser window
2. recover account using seed phrase
3. go to Account page and Enable Ledger
4. try for example to send near

### DIFF
--- a/packages/frontend/src/components/accounts/ledger/SetupLedger.js
+++ b/packages/frontend/src/components/accounts/ledger/SetupLedger.js
@@ -145,6 +145,9 @@ const SetupLedger = (props) => {
                     }
                 } else {
                     try {
+                        if (ledgerHdPath) {
+                            setLedgerHdPath({ accountId, path: ledgerHdPath });
+                        }
                         await dispatch(addLedgerAccessKey(ledgerHdPath));
                     } catch (error) {
                         dispatch(checkAndHideLedgerModal());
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/accounts/ledger/SetupLedger.js
```diff
@@ -85,10 +85,8 @@ const SetupLedger = (props) => {
                         await setKeyMeta(publicKey, { type: 'ledger' });
                         Mixpanel.track('SR-Ledger Set key meta');
 
-                        // Set custom path to localstorage
-                        if (ledgerHdPath) {
-                            setLedgerHdPath({ accountId, path: ledgerHdPath });
-                        }
+                        // Set path to localstorage
+                        setLedgerHdPath({ accountId, path: ledgerHdPath });
```
gutsyphilip, line 89: I think we can just call this once before the if else block?
I had this in a PR yesterday(#2659),I missed it while closing and recreating the PR severally.

marcinbodnar: We cannot because we want to add it only if adding a ledger as a recovery option was a success.

gutsyphilip: Not sure if I'm looking at this right. But aren't we setting it before we attempt adding ledger as an option? Looks that way in the try block

Patrick1904 (approved):
LGTM

PR #2669, opened 2022-05-12
fix: wallet buttons overflow

esaminu:
This PR fixes styles on mobile when 5 buttons are rendered on the wallet dashboard and adds a shadow to indicate that the content can be horizontally scrolled


Patrick1904:
image

Works well but for some reason I don't see any shadow indicating scrollable section. I'm using iPhone with Safari.

esaminu:
Closing in favor of preferred solution in #2670

PR #2670, opened 2022-05-13
feat: wrap wallet buttons on overflow

esaminu:
Wrap wallet buttons on the wallet dashboard on overflow

image


Patrick1904 (approved):
Thank you @esaminu! 💯

PR #2679, opened 2022-05-16
feat: Enable import of zero balance Ledger account (testnet)

Patrick1904:
This enables a user to import a non-existent Ledger implicit account into the wallet (on testnet)

To test:
1. Go to `https://my-near-wallet-pr-2679.onrender.com/sign-in-ledger`
2. Change Ledger HD path to any path without an existing account
3. Try importing the account
4. Click 'Import anyway' when prompted
5. Confirm key on Ledger device
6. Observe that the Ledger account has been imported with given key

### DIFF
--- a/packages/frontend/src/components/accounts/CouldNotFindAccountModalWrapper.js
+++ b/packages/frontend/src/components/accounts/CouldNotFindAccountModalWrapper.js
@@ -3,38 +3,91 @@ import { useDispatch } from 'react-redux';
 
 import {
     redirectTo,
-    refreshAccount
+    refreshAccount,
+    getLedgerPublicKey
 } from '../../redux/actions/account';
 import { showCustomAlert, clearGlobalAlert } from '../../redux/actions/status';
+import { setLedgerHdPath } from '../../utils/localStorage';
 import { getImplicitAccountIdFromSeedPhrase, getKeyPairFromSeedPhrase } from '../../utils/parseSeedPhrase';
-import { wallet } from '../../utils/wallet';
+import { wallet, setKeyMeta } from '../../utils/wallet';
 import { CouldNotFindAccountModal } from './CouldNotFindAccountModal';
 
-export function CouldNotFindAccountModalWrapper ({
+export function CouldNotFindAccountModalWrapper({
     isOpen,
     onClose,
-    seedPhrase
+    seedPhrase,
+    ledgerHdPath,
+    recoveryMethod
 }) {
     const dispatch = useDispatch();
+
+    const handleImportNonLedgerImplicitAccount = async () => {
+        const recoveryKeyPair = getKeyPairFromSeedPhrase(seedPhrase);
+        const implicitAccountId = getImplicitAccountIdFromSeedPhrase(seedPhrase);
+        try {
+            await wallet.importZeroBalanceAccount(implicitAccountId, recoveryKeyPair);
+            dispatch(refreshAccount());
+            dispatch(redirectTo('/'));
+            dispatch(clearGlobalAlert());
+        } catch (e) {
+            dispatch(showCustomAlert({
+                success: false,
+                messageCodeHeader: 'error',
+                messageCode: 'walletErrorCodes.recoverAccountSeedPhrase.errorNotAbleToImportAccount',
+                errorMessage: e.message
+            }));
+        }
+    };
+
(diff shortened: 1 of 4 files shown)


packages/frontend/src/components/accounts/CouldNotFindAccountModalWrapper.js
```diff
@@ -3,38 +3,91 @@ import { useDispatch } from 'react-redux';
...
+            }));
+        }
+    };
+
     return (
         <CouldNotFindAccountModal
             isOpen={isOpen}
             onClose={onClose}
+            recoveryMethod={recoveryMethod}
             onClickImport={async () => {
```
morgsmccauley, line 86: nit: I think the `await`s are not actually necessary here?

packages/frontend/src/components/accounts/ledger/SignInLedgerWrapper.js
```diff
@@ -39,7 +41,8 @@ export function SignInLedgerWrapper(props) {
 
     const [accountId, setAccountId] = useState('');
     const [loader, setLoader] = useState(false);
-    const [confirmedPath, setConfirmedPath] = useState(1);
+    const [confirmedPath, setConfirmedPath] = useState(10);
```
morgsmccauley, line 44: Was the change to `10` intentional?

packages/frontend/src/components/accounts/CouldNotFindAccountModalWrapper.js
```diff
@@ -3,38 +3,91 @@ import { useDispatch } from 'react-redux';
...
+        try {
+            const ledgerPublicKey = await dispatch(getLedgerPublicKey(ledgerHdPath));
+            const implicitAccountId = Buffer.from(ledgerPublicKey.data).toString('hex');
+            const account = wallet.getAccountBasic(implicitAccountId);
+            try {
+                const accountState = await account.state();
+                if (accountState) {
+                    const errorMessage = `Implicit account ID ${implicitAccountId} derived from Ledger HD path ${ledgerHdPath} already exists but is not controlled by this Ledger device.`;
+                    // This could happen if user creates an implicit account with Ledger but then switches to seed phrase recovery, etc.
+                    console.log(errorMessage);
```
morgsmccauley, line 52: 🔪 ? Or do we want these in prod?

packages/frontend/src/components/accounts/CouldNotFindAccountModalWrapper.js
```diff
@@ -3,38 +3,91 @@ import { useDispatch } from 'react-redux';
...
+                    dispatch(showCustomAlert({
+                        errorMessage: errorMessage,
+                        success: false,
+                        messageCodeHeader: 'error'
+                    }));
+                    return;
+                }
+            } catch (e) {
+                if (e.message.includes('does not exist while viewing')) {
+                    console.log('Ledger implicit Account ID does not exist on chain. Importing as zero balance account.');
```
morgsmccauley, line 62: Same comment as above ⬆️

morgsmccauley (changes requested):
Couple small Qs :)

packages/frontend/src/components/accounts/ledger/SignInLedgerWrapper.js
Patrick1904: Thanks for catching!! Leftover from testing. Definitely not intentional.

packages/frontend/src/components/accounts/CouldNotFindAccountModalWrapper.js
Patrick1904: I forgot to move down the common 'post import' actions here :)

packages/frontend/src/components/accounts/CouldNotFindAccountModalWrapper.js
Patrick1904: I think it's worth keeping this prints to console because it'll be helpful for support in case of user error and for technical users

packages/frontend/src/components/accounts/CouldNotFindAccountModalWrapper.js
Patrick1904: I think it's worth keeping this prints to console because it'll be helpful for support in case of user error and for technical users

morgsmccauley (approved):
LGTM!

Issue #2620, opened 2022-04-28
[UX Suggestion] Copy wallet address button


DRIX-dot:
It could be great if there would be a [copy wallet address] On the right-upper corner of the Near Wallet main page.

Currently, user would have to navigate to the [account] tab in order to find the forementioned button.

gutsyphilip:
Thanks for opening this issue @DRIX-dot. We have added this to our backlog and will follow up as necessary.

PR #2682, opened 2022-05-18
feat: add copy button to address list

1bitl:
Adds a "Copy to clipboard" button to the accounts list.

This required a nudge on the "Copied" text so that it doesn't cut off by the surrounding list.
Also changes the tooltip text to a description of what the button does.

Fixes https://github.com/near/near-wallet/issues/2620

image

image

### DIFF
--- a/packages/frontend/src/components/accounts/account_selector/Account.js
+++ b/packages/frontend/src/components/accounts/account_selector/Account.js
@@ -2,6 +2,8 @@ import React, { useState } from 'react';
 import styled from 'styled-components';
 
 import Balance from '../../common/balance/Balance';
+import ClickToCopy from '../../common/ClickToCopy';
+import CopyIcon from '../../svg/CopyIcon';
 import EyeIcon from './EyeIcon';
 
 const StyledContainer = styled.div`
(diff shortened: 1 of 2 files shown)


Patrick1904 (changes requested):
Thanks for the PR @1bitl! I noticed that when you click on the copy button on an account other than the current account in the dropdown, the copy button does not work and the menu closes on click. Can you please look into that?

1bitl:
@Patrick1904 Thanks for the quick review, and nice catch! Fixed and added an event to the `ClickToCopy` component.

Patrick1904:
@1bitl the issue still persists for me while testing - basically no copy and the dropdown closes.

I'm also noticing that the success toast for the `ClickToCopy` component gets cut off at the top - can we please avoid style changes to the `ClickToCopy` component?

Screen Shot 2022-05-23 at 1 33 03 PM

Patrick1904:
Great job @1bitl - it works well now! 👍

Patrick1904 (approved):
Great job @1bitl - it works well now! 👍

PR #2688, opened 2022-05-19
Swap and SwapUSN consolidation

judith-Near:
1. make functionality of swap page generic such that users can select tokens to swap
2. selecting tokens to swap reuses the "select asset" view seen in send functionality
3. tokens that you can select from are based on tokens that you have
4. tokens you can swap to are based on the token you are swapping from (must be a valid pair)
5. based on the token you have selected to swap, the token to swap to will default to a valid pair
6. refactors multiplier logic
7. when clicking the middle swap positions button, the amounts switch correctly
8. culls out editing the bottom token amount, you can only edit the top amount
9. fix translations
10. refactor review screen and success screen
11. delete "buy USN" and "Swap USN" buttons and prompts
12. create slippagePicker component

Design
[Figma Swap](https://www.figma.com/file/Wcl0YlyeasgfBFxJNx7vYS/WAL-48%3A-Wrap%2FUnwrap-NEAR?node-id=2063%3A7314)
+ Select token screen should be reused from send functionality (signed off by corwin and jacob)
+ Can only edit the top token amount (signed off by jacob)

Known issues:
"View on explorer" on the success screen is not available for USN swaps, will address in future PR (requires exporation on usn contract and how to surface most recent transaction)

### DIFF
--- a/packages/frontend/src/hooks/fungibleTokensIncludingNEAR.js
+++ b/packages/frontend/src/hooks/fungibleTokensIncludingNEAR.js
@@ -1,9 +1,12 @@
 import { useSelector } from 'react-redux';
 
+import { NEAR_TOKEN_ID, USN_CONTRACT } from '../config';
 import selectNEARAsTokenWithMetadata from '../redux/selectors/crossStateSelectors/selectNEARAsTokenWithMetadata';
 import { selectActiveAccountId } from '../redux/slices/activeAccount';
 import { selectTokensFiatValueUSD } from '../redux/slices/tokenFiatValues';
 import { selectTokensWithMetadataForAccountId } from '../redux/slices/tokens';
+import compare from '../utils/compare';
+
 
 export const useFungibleTokensIncludingNEAR = function ({ showTokensWithZeroBalance = false } = {}) {
     const NEARAsTokenWithMetadata = useSelector(selectNEARAsTokenWithMetadata);
@@ -18,5 +21,11 @@ export const useFungibleTokensIncludingNEAR = function ({ showTokensWithZeroBala
         fiatValueMetadata: ft.fiatValueMetadata?.usd ? ft.fiatValueMetadata : {...fungibleTokenPrices[ft.contractName]}
     }));
 
+    const sortingOrder = {
+        [USN_CONTRACT]: 1,
+        [NEAR_TOKEN_ID]: 2
+    };
+    fungibleTokensWithPrices.sort(compare({key: 'contractName', sortingOrder}));
+    
     return [NEARAsTokenWithMetadata, ...fungibleTokensWithPrices];
 };
--- a/packages/frontend/src/utils/compare.js
+++ b/packages/frontend/src/utils/compare.js
@@ -1,0 +1,15 @@
+export default function compare({key, order = 'asc', sortingOrder}) {
+    return function (a, b) {
+        if (!a.hasOwnProperty(key) || !b.hasOwnProperty(key)) return 0;
+		
+        const first = (a[key].toLowerCase() in sortingOrder) ? sortingOrder[a[key]] : Number.MAX_SAFE_INTEGER;
+        const second = (b[key].toLowerCase() in sortingOrder) ? sortingOrder[b[key]] : Number.MAX_SAFE_INTEGER;
+
+        let result = 0;
+        if (first < second)
+            result = -1;
+        else if (first > second)
+            result = 1;
+        return (order === 'desc') ? ~result : result;
+    };
+};


packages/frontend/src/components/send/components/RawTokenAmount.js
```diff
@@ -10,7 +10,7 @@ const RawTokenAmount = ({
     withSymbol = true,
     showFiatAmountForNonNearToken
 }) => {
-    if (decimals && symbol) {
+    if (decimals && symbol || decimals === 0 && symbol) {
```
judith-Near, line 13: decimals in this component is being used to determine how many spots to the left to move the decimal over to.
ie if the amount is 1, and decimals is 5, the resulting calculation from the components is .00005. if i don't want to do this kind of formatting, the easiest way is to just say decimals 0.

packages/frontend/src/components/swap/components/helpers.js
```diff
@@ -0,0 +1,220 @@
...
+            }
+        };
+
+        getCommission();
+    }, [debounceValue, exchangeRate]);
+
+    return { commissionFee, isLoadingCommission };
+};
+
+export const exchangeRateTranslation = ({ inputtedAmountOfToken, calculateAmountOfToken, balance, exchangeRate }) => {
```
judith-Near, line 184: exchangeRateTranslation and useInterval are the only new code added

packages/frontend/src/components/swap/components/helpers.js
```diff
@@ -0,0 +1,223 @@
+import * as nearApiJs from 'near-api-js';
```
judith-Near, line 1: findTokenSwapToList, exchangeRateTranslation, and useInterval are the only new code added

packages/frontend/src/components/swap/components/TransactionDetails.js
```diff
@@ -0,0 +1,117 @@
...
+            >
+                {isUSNSwap && 
+                    <SlippagePicker
+                        translateIdTitle={'swap.priceImpact'}
+                        translateIdInfoTooltip="swap.translateIdInfoTooltip.priceImpact"
+                        setSlippage={setSlippage}
+                    />
+                }
+                
+                {!isUSNSwap &&
```
Patrick1904, line 53: What do you think about making separate `TransactionsDetails` components for USN and wNEAR swapping to make it easier to read and avoid all the conditionals?

packages/frontend/src/components/wallet/TokenBox.js
```diff
@@ -104,15 +101,12 @@ const StyledContainer = styled.div`
     }
 
     .balance {
-        margin-left: ${({IS_USN}) =>(IS_USN ? 0 : 'auto')};
+        margin-left: auto !important;
```
Patrick1904, line 104: nit: can we please opt to use `&` for added specificity when needed vs `!important`?

if `.balance` is already getting styled up above maybe we can/should rename it here to avoid needing more specificity all together?

packages/frontend/src/components/send/components/RawTokenAmount.js
Patrick1904: Could we do something like this instead?

`decimals = null,`

`if (decimals !== null && symbol)`

Patrick1904:
Can we add the max decimal constraint/formatting to the input value too?

Screen Shot 2022-05-26 at 12 45 44 PM

Patrick1904:
nit: can we please fix the `font-weight` here:

Screen Shot 2022-05-26 at 12 49 17 PM

Patrick1904:
nit: can we please align the color w/ figma

Screen Shot 2022-05-26 at 12 52 13 PM

Patrick1904:
Can we add both amounts to the success page?

Screen Shot 2022-05-26 at 12 55 16 PM

Patrick1904:
nit: can we please add more top margin to the button to align w/ figma?

Screen Shot 2022-05-26 at 12 56 56 PM

Patrick1904:
nit: the 'view on explorer' button would be sweet

Screen Shot 2022-05-26 at 1 01 24 PM

Patrick1904:
This is looking really great @judith-Near! 💯 Big PR! USN swap UX/UI is a massively improved with these changes.

I don't want any `nit` comments to block merging this PR when it's ready. We can do that in follow up PR.

packages/frontend/src/components/swap/components/TransactionDetails.js
judith-Near: yeah that makes sense, will do 👍

packages/frontend/src/components/wallet/TokenBox.js
judith-Near: after getting rid of the swap usn button inline under your portfolio, realized this isn't even needed. thanks for the catch!

packages/frontend/src/components/swap/components/ReviewTransactionDetails.js
```diff
@@ -151,6 +159,60 @@ const ReviewTransactionDetails = ({
...
+                        selectedTokenTo={tokenTo}
+                        estimatedFeesInNear={`${
+                            amountTokenFrom > 1
+                                ? Math.trunc(amountTokenFrom).toString()
+                                : '1'
+                        }`}
+                        estimatedMinReceived={estimatedMinReceived}
+                    />  
+                );
+            default:
```
esaminu, line 200: I suggest we collapse this into just the `USN` and the `default` cases and name the component something more generic like `SwapTransactionDetails` wdyt?

esaminu (changes requested):
Awesome work on this @judith-Near ! I left some minor comments otherwise LGTM! 😄

packages/frontend/src/components/swap/Swap.js
```diff
@@ -93,21 +136,33 @@ const StyledContainer = styled(Container)`
...
     const [activeTokenTo, setActiveTokenTo] = useState();
     const [swappingToken, setSwappingToken] = useState(false);
     const [transactionHash, setTransactionHash] = useState(null);
+    const [reversePositionsJustClicked, setReversePositionsJustClicked] = useState(false);
     const dispatch = useDispatch();
+
+    useEffect(() => {
+        // setsMultiplier as multiplierResult gets populated
+        setMultiplier(multiplierResult);
+    }, [multiplierResult]);
```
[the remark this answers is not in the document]
judith-Near: yes we can just use multiplier result directly, oversight on my part!

esaminu (approved):
LGTM! Great work on this 👍

agileurbanite:
@judith-Near is this safe to merge?

Patrick1904 (approved):
Great job @judith-Near! 🙏

PR #2689, opened 2022-05-20
feat: Finalize wallet setup for funded zero balance account

Patrick1904:
Changes:
1. Finalize wallet setup by adding a new key in `localStorage` when the zero balance account is active/funded. This final setup happens when the user navigates to the user `/profile` since the main reason for a `localStorage` key is for Ledger to pull and manage recovery methods. Non-ledger users do not in theory need a new access key for `localStorage` since the initial key is already present, however, it will be added for consistency sake since that's the current behavior during onboarding and when importing an account.
2. Turn `off` the feature flag `IMPORT_ZERO_BALANCE_ACCOUNT` in all environments until we've shipped the final pieces.

Main way to test the UI/UX here is to:
1. Change the flag `IMPORT_ZERO_BALANCE_ACCOUNT` to `on` in your environment
2. Import an implicit account that does not exist
3. Navigate to the wallet `/profile`. If you have Ledger, you'll be prompted with a modal asking you to add a key. In all other cases, a key will be added automatically and a success toast notifying you that your wallet account has now been setup successfully.

The next steps after this PR is:
1. Enable ability to add a recovery method record in contract-helper DB before an account has been created. This is currently possible for email/phone, and has to be possible for Ledger and seed phrase too so that the correct recovery method displays as active/inactive once the user has funded their account.
5. Skip the funding step during onboarding and treat the account as a zero balance account

### DIFF
--- a/features/flags.json
+++ b/features/flags.json
@@ -285,7 +285,6 @@
       "enabled": false,
       "lastEditedBy": "Patrick Tajima",
       "lastEditedAt": "2022-05-06T17:30:49.430Z"
-
     },
     "testnet_STAGING": {
       "enabled": false,
--- a/features/flags.json
+++ b/features/flags.json
@@ -272,7 +272,7 @@
       "lastEditedAt": "2022-05-06T17:30:49.429Z"
     },
     "testnet": {
-      "enabled": true,
+      "enabled": false,
       "lastEditedBy": "Patrick Tajima",
       "lastEditedAt": "2022-05-06T17:30:49.430Z"
     },
@@ -285,14 +285,15 @@
       "enabled": false,
       "lastEditedBy": "Patrick Tajima",
       "lastEditedAt": "2022-05-06T17:30:49.430Z"
+
     },
     "testnet_STAGING": {
-      "enabled": true,
+      "enabled": false,
       "lastEditedBy": "Patrick Tajima",
       "lastEditedAt": "2022-05-06T17:30:49.430Z"
     },
     "testnet_NEARORG": {
-      "enabled": true,
+      "enabled": false,
       "lastEditedBy": "esaminu",
       "lastEditedAt": "2022-05-09T21:50:59.559Z"
     },
--- a/packages/frontend/src/components/profile/zero_balance/ZeroBalanceAccountWrapper.js
+++ b/packages/frontend/src/components/profile/zero_balance/ZeroBalanceAccountWrapper.js
@@ -1,0 +1,109 @@
+import React, { useEffect, useState } from 'react';
+import { useSelector, useDispatch } from 'react-redux';
+
+import { refreshAccount } from '../../../redux/actions/account';
+import { showCustomAlert } from '../../../redux/actions/status';
+import { selectAccountExists, selectAccountFullAccessKeys, selectAccountId } from '../../../redux/slices/account';
+import { finishLocalSetupForZeroBalanceAccount } from '../../../redux/slices/account/createAccountThunks';
+import { actions as ledgerActions, selectLedgerConnectionAvailable } from '../../../redux/slices/ledger';
+import { wallet } from '../../../utils/wallet';
+import { AddLedgerKeyModal } from './AddLedgerKeyModal';
+
+
+const { handleShowConnectModal } = ledgerActions;
+
+export function ZeroBalanceAccountWrapper() {
+    const dispatch = useDispatch();
+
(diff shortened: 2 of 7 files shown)


packages/frontend/src/components/profile/zero_balance/ZeroBalanceAccountWrapper.js
```diff
@@ -0,0 +1,109 @@
...
+        } catch (e) {
+            dispatch(showCustomAlert({
+                success: false,
+                messageCodeHeader: 'zeroBalance.addPhraseKey.error.header',
+                messageCode: 'zeroBalance.addPhraseKey.error.message',
+            }));
+        }
+
+        dispatch(refreshAccount());
+    };
```
morgsmccauley, line 87: Thoughts on combining these? 🤔 

```js
    const handleAddLocalAccessKey = async (recoveryMethod) => {
        const translationId = recoveryMethod === 'ledger' ? 'addLedgerKey' : 'addPhraseKey';

        try {
            await dispatch(finishLocalSetupForZeroBalanceAccount({
                implicitAccountId: accountId,
                recoveryMethod,
            }));
            dispatch(showCustomAlert({
                success: true,
                messageCodeHeader: `zeroBalance.${translationId}.success.header`,
                messageCode: `zeroBalance.${translationId}.success.message`
            }));
        } catch (e) {
            dispatch(showCustomAlert({
                success: false,
                messageCodeHeader: `zeroBalance.${translationId}.error.header`,
                messageCode: `zeroBalance.${translationId}.error.message`,
            }));
        }

        dispatch(refreshAccount());
    };
```
Then using it:
```js
handleLocalAccessKey('ledger');
```

features/flags.json
```diff
@@ -285,14 +285,15 @@
       "enabled": false,
       "lastEditedBy": "Patrick Tajima",
       "lastEditedAt": "2022-05-06T17:30:49.430Z"
+
```
morgsmccauley, line 288: Did the flag tool do this? 🤔

packages/frontend/src/translations/en.global.json
```diff
@@ -1738,5 +1741,33 @@
...
+        "ledgerModal": {
+            "desc": "Add a Ledger limited access key to the wallet to manage multiple recovery methods.",
+            "confirmOnLedger": "Confirm the action on your Ledger device.",
+            "title": "Your account is now active!",
+            "addLedgerKey": "Add Ledger Key"
+        },
+        "addLedgerKey": {
+            "success": {
+                "header": "Ledger access key added successfully!",
+                "message": "A new Ledger key was added to your account. You can now manage your accounts recovery methods using your Ledger device."
```
morgsmccauley, line 1755:
```suggestion
                "message": "A new Ledger key was added to your account. You can now manage your accounts recovery methods without using your Ledger device."
```
Might be wrong on this - is this messaging based around the limited access key we add for authenticating requests to contract-helper? If so, then it would enable users to managing recovery methods _without_ their ledger?

packages/frontend/src/redux/slices/account/createAccountThunks.js
```diff
@@ -187,3 +191,31 @@ export const finishSetupImplicitAccount = createAsyncThunk(
...
+        recoveryMethod
+    }, { dispatch }) => {
+        try {
+            if (recoveryMethod === 'ledger') {
+                const ledgerHDPath = getLedgerHDPath(implicitAccountId);
+                await dispatch(signInWithLedger({ path: ledgerHDPath })).unwrap();
+            } else {
+                const account = await wallet.getAccount(implicitAccountId);
+                const accessKeys = await account.getAccessKeys();
+                const fullAccessKeys = accessKeys.filter((it) => it.access_key && it.access_key.permission === 'FullAccess');
```
morgsmccauley, line 208:
```suggestion
                const fullAccessKeys = accessKeys.filter((it) => it.access_key?.permission === 'FullAccess');
```
nit: optional chaining 😄

morgsmccauley (changes requested):
Looks good, just couple minor comments :)

packages/frontend/src/translations/en.global.json
Patrick1904: It's just so that they can pull the information from the contract-helper, not actually manage the method, but figured it's easier for user to understand 'manage' in this context

morgsmccauley (approved):
LGTM :)

PR #2690, opened 2022-05-23
Feat: Coingecko call directly from frontend wrapped with dataloader and node-cache

judith-Near:
[CoinGecko Call Design Doc](https://docs.google.com/document/d/1mvnNVavH12IleKsr_hvZDG7GrZcQtXz21MlraT-3H-4/edit?usp=sharing)

Note: closed PR#2641 due to merging master, proper process is rebasing

"The reason for caching the data on the frontend and batching our requests is because there are API-per-minute limits for coingecko which appear to be enforced by source IP -- and we don’t want our wallet users being throttled and seeing errors from the coingecko APIs when they need to request a large number of prices for many tokens at once." Daryl Collins

1) deletes utils/ref-finance and consolidates logic into utils/fiatValueManager which makes use of dataloader and node-cache
2) simplifies logic of the thunk fetchTokenFiatValues
3) adds logic to reducer to only update state of the tokens if the token values have changed prices and are more recently updated
4) USN values were being implied via tether stablecoin values, changed to fetch usn values from coingecko in testnet and usn values from ref-finance in mainnet


packages/frontend/src/redux/slices/tokenFiatValues/index.js
```diff
@@ -85,11 +85,11 @@ export const selectNearTokenFiatValueUSD = createSelector(selectNearTokenFiatDat
 
 export const selectUSDNTokenFiatData = createSelector(
     selectAllTokenFiatValues,
-    ({ tokens }) => tokens.tether || {}
+    ({ tokens }) => tokens.usn || {}
```
judith-Near, line 88: USN values were being implied via tether stablecoin values, changed to fetch usn values from coingecko in testnet and usn values from ref-finance in mainnet

packages/frontend/src/utils/fiatValueManager.js
```diff
@@ -0,0 +1,117 @@
...
+
+    async getPrice(tokens = ['near', 'usn']) {
+        const byTokenName = {};
+        const prices = await this.fiatValueDataLoader.loadMany(tokens);
+        tokens.forEach((tokenName, ndx) => byTokenName[tokenName] = prices[ndx]);
+        return byTokenName;
+    };
+
+    async fetchTokenPrices(dummyToken = ['near']) {
+        const [prices] = await this.refFinanceDataLoader.loadMany(dummyToken);
```
judith-Near, line 91: it doesn't matter what dummyToken is, but dataloader's API dictates that loadMany needs an array of values. typically this array would be the list of token names, but this ref-finance call is dynamic and returns different values dependent on environment. dummyToken is ignored in the implementation of the batch function.

"DataLoader must be constructed with a function which accepts Array<key> and returns Promise<Array<value>> of the same length as the Array of keys"

packages/frontend/src/redux/slices/tokenFiatValues/index.js
```diff
@@ -50,10 +38,22 @@ const tokenFiatValuesSlice = createSlice({
...
+                    if (fetchedLastUpdatedAt > previousLastUpdatedAt && fetchedPrice !== previousPrice) {
+                        beenUpdatedTokens[token] = action.payload[token];
+                    }
+                });
                 // Using merge instead of `assign()` so in the future we don't blow away previously loaded token
                 // prices when we load new ones with different token names
-                merge(state.tokens, action.payload);
+                if (Object.keys(beenUpdatedTokens).length) {
+                    merge(state.tokens, beenUpdatedTokens);
+                }
```
esaminu, line 56: I think it would be cool and cleaner to import [lodash `mergeWith`](https://lodash.com/docs/4.17.15#mergeWith) here. Something like:

```suggestion
mergeWith(state.tokens, action.payload, (previous, fetched) =>
    fetched?.last_updated_at > previous?.last_updated_at &&
    fetched?.usd != previous?.usd
        ? fetched
        : previous
);
```

I'm also somewhat concerned about things aside from `usd` changing. Maybe we can address that by importing [lodash omit](https://lodash.com/docs/4.17.15#omit) and [isEqual](https://lodash.com/docs/4.17.15#isEqual):

```suggestion
mergeWith(state.tokens, action.payload, (previous, fetched) =>
    fetched?.last_updated_at > previous?.last_updated_at &&
    !isEqual(omit(fetched, 'last_updated_at'), omit(previous, 'last_updated_at'))
        ? fetched
        : previous
);
```

packages/frontend/src/redux/slices/tokenFiatValues/index.js
```diff
@@ -50,10 +38,22 @@ const tokenFiatValuesSlice = createSlice({
...
+                const tokens = cloneDeep(state).tokens;
+                Object.keys(action.payload).map((token) => {
+                    const previousLastUpdatedAt = tokens[token] ? tokens[token].last_updated_at : 0;
+                    const previousPrice = tokens[token] ? tokens[token].usd : 0;
+                    const fetchedLastUpdatedAt = action.payload[token].last_updated_at;
+                    const fetchedPrice = action.payload[token].usd;
+                    if (fetchedLastUpdatedAt > previousLastUpdatedAt && fetchedPrice !== previousPrice) {
+                        beenUpdatedTokens[token] = action.payload[token];
+                    }
+                });
```
esaminu, line 51: This can be removed if below `mergeWith` suggestion is used:
```suggestion
```

esaminu (changes requested):
Nice feature and cleanup! Added a comment regarding reducer code otherwise LGTM!! 👍

packages/frontend/src/utils/fiatValueManager.js
```diff
@@ -0,0 +1,118 @@
...
+            },
+            {
+                cacheMap: wrapNodeCacheForDataloader(new Cache({ stdTTL: 30, checkperiod: 0, useClones: false }))
+            }
+        );
+    };
+
+    async getPrice(tokens = ['near', 'usn']) {
+        const byTokenName = {};
+        const prices = await this.fiatValueDataLoader.loadMany(tokens);
```
esaminu, line 86: typo here?
```suggestion
        const prices = await this.coinGeckoFiatValueDataLoader.loadMany(tokens);
```

packages/frontend/src/utils/fiatValueManager.js
```diff
@@ -0,0 +1,118 @@
...
+                    );
+                    return tokenIds.map((id) => tokenFiatValues[id]);
+                } catch (error) {
+                    console.error(`Failed to fetch coingecko prices: ${error}`);
+                    // DataLoader must be constructed with a function which accepts 
+                    // Array<key> and returns Promise<Array<value>> of the same length
+                    // as the Array of keys
+                    return new Promise((resolve, reject) => {
+                        return [].fill({}, tokenIds.length);
+                    });
```
esaminu, line 56: `[].fill({}, tokenIds.length)` will always return an empty array AFAICT also, the promise does not resolve
```suggestion
                   return Promise.resolve(Array(tokenIds.length).fill({}));
```

packages/frontend/src/utils/fiatValueManager.js
```diff
@@ -0,0 +1,118 @@
...
+            async (dummyToken) => {
+                try {
+                    const refFinanceTokenFiatValues = await sendJson(
+                        'GET',
+                        REF_FINANCE_API_ENDPOINT + '/list-token-price'
+                    );
+                    return [refFinanceTokenFiatValues];
+                } catch (error) {
+                    console.error(`Failed to fetch ref-finance prices: ${error}`);
+                    return new Promise((resolve, reject) => [{}]);
```
esaminu, line 75:
```suggestion
                    return Promise.resolve([{}]);
```

esaminu (changes requested):
Just some things I missed on first pass. Should be good to merge after they're fixed 👍

packages/frontend/src/utils/fiatValueManager.js
```diff
@@ -0,0 +1,118 @@
...
+    };
+
+    async getPrice(tokens = ['near', 'usn']) {
+        const byTokenName = {};
+        const prices = await this.fiatValueDataLoader.loadMany(tokens);
+        tokens.forEach((tokenName, ndx) => byTokenName[tokenName] = prices[ndx]);
+        return byTokenName;
+    };
+
+    async fetchTokenPrices(dummyToken = ['near']) {
```
esaminu, line 91: nit: I think these method names could be more descriptive e.g. `fetchRefFinanceTokenPrices` or similar. That way, when we `Promise.all` or similar, it's clear why we're calling multiple methods and awaiting multiple promises

esaminu (approved):
Great stuff! 👍

PR #2694, opened 2022-05-24
Kwlt 6 ability to export private keys

esaminu:
This PR:

1. Adds [functionality to show a modal, confirm `accountId` and view private key](https://nearinc.atlassian.net/browse/KWLT-6)
2. Adds an icon to the "Remove Account from Wallet" button

[Figma link](https://www.figma.com/file/w4SQVO6kBvAP5D65TGlm7y/Wallet-Domain-Migration)

### DIFF
--- a/packages/frontend/src/components/profile/export_private_key/ExportKeyModal.js
+++ b/packages/frontend/src/components/profile/export_private_key/ExportKeyModal.js
@@ -6,6 +6,7 @@ import styled from 'styled-components';
 import CheckMarkNoBorderIcon from '../../../images/icon-check-no-border.svg';
 import { selectAccountId } from '../../../redux/slices/account';
 import isMobile from '../../../utils/isMobile';
+import ClickToCopy from '../../common/ClickToCopy';
 import FormButton from '../../common/FormButton';
 import Modal from '../../common/modal/Modal';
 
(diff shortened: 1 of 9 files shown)


packages/frontend/src/components/profile/remove_account/RemoveAccountWrapper.js
```diff
@@ -11,8 +12,28 @@ import Container from '../../common/styled/Container.css';
...
+    padding-top: 0;
+    padding-bottom: 0;
     > button {
         width: 100%;
+        display: flex;
+        justify-content: center;
+        align-items: center;
+
+        > svg {
+            width: 22px !important;
```
Patrick1904, line 25: nit: can we please opt to use `&` (if absolutely needed) for added specificity instead of `!important`

packages/frontend/src/components/profile/export_private_key/ExportKeyModal.js
```diff
@@ -0,0 +1,175 @@
...
+
+        .text-select-display {
+          display: flex;
+          flex-direction: row;
+          align-items: flex-start;
+          padding: 12px;
+          gap: 8px;
+
+          height: 96px;
+          width: 99%;
```
Patrick1904, line 73: Not sure why the gap, height, width is necessary? Can we remove that and just make padding 20px?

Patrick1904:
This is looking great @esaminu! Not a blocker but just something I noticed / thought of:
1. Should we specify that the key is the 'key on this machine' and not the key from the users recovery method? Seems one might assume it's the recovery method key otherwise
2. Would be nice with ability to copy the key. Simple thing would be to just wrap the key div in a `ClickToCopy` component.

cc @heycorwin

Patrick1904 (dismissed):
LGTM.

@esaminu not a blocker but can we please wrap it in a `form` so that user can hit enter to continue after entering account id

Patrick1904 (approved):
💯

PR #2697, opened 2022-05-24
Fix sorting actions for Sign Transaction Details page

marcinbodnar:
This PR fixes the order of `actions` on the Transactions Details page (Sign flow) by removing mutating the `transactions` array and replacing `.sort` function that results in changing the order of transactions in browsers without stable sort.


andy-haynes (approved):
LGTM 🚀

PR #2698, opened 2022-05-24
feat: Skip funding step when creating implicit account

Patrick1904:
This PR treats an implicit account as a 'zero balance account' during onboarding and therefore skips the funding step and sends the user directly to the wallet dashboard.

The feature (`IMPORT_ZERO_BALANCE_ACCOUNT`) is currently turned `OFF` in all environments, and a separate PR to turn the feature `ON` on `testnet` and `staging` will be opened once this PR is merged.

To test this PR, simply turn the flag `IMPORT_ZERO_BALANCE_ACCOUNT` `ON` in your environment and then test onboarding or/and import of a zero balance account. Note that only `mainnet` accounts need funding.

### DIFF
--- a/packages/frontend/src/redux/slices/account/createAccountThunks.js
+++ b/packages/frontend/src/redux/slices/account/createAccountThunks.js
@@ -6,10 +6,18 @@ import { KeyType } from 'near-api-js/lib/utils/key_pair';
 import * as Config from '../../../config';
 import { actions as ledgerActions } from '../../../redux/slices/ledger';
 import sendJson from '../../../tmp_fetch_send_json';
-import { setReleaseNotesClosed, getLedgerHDPath } from '../../../utils/localStorage';
-import { CONTRACT_CREATE_ACCOUNT_URL, FUNDED_ACCOUNT_CREATE_URL, IDENTITY_FUNDED_ACCOUNT_CREATE_URL, RELEASE_NOTES_MODAL_VERSION, wallet } from '../../../utils/wallet';
+import { setReleaseNotesClosed, getLedgerHDPath, setLedgerHdPath } from '../../../utils/localStorage';
+import {
+    CONTRACT_CREATE_ACCOUNT_URL,
+    FUNDED_ACCOUNT_CREATE_URL,
+    IDENTITY_FUNDED_ACCOUNT_CREATE_URL,
+    RELEASE_NOTES_MODAL_VERSION,
+    wallet,
+    setKeyMeta
+} from '../../../utils/wallet';
 import { WalletError } from '../../../utils/walletError';
 import { finishAccountSetup } from '../../actions/account';
+import { showCustomAlert } from '../../actions/status';
 import { SLICE_NAME } from './';
 
 const {
@@ -19,7 +27,8 @@ const {
 const {
     RECAPTCHA_ENTERPRISE_SITE_KEY,
     NETWORK_ID,
-    RECAPTCHA_CHALLENGE_API_KEY
+    RECAPTCHA_CHALLENGE_API_KEY,
+    ACCOUNT_HELPER_URL
 } = Config;
 
 export const addLocalKeyAndFinishSetup = createAsyncThunk(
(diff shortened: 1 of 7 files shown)


packages/frontend/src/redux/slices/account/createAccountThunks.js
```diff
@@ -219,3 +228,70 @@ export const finishLocalSetupForZeroBalanceAccount = createAsyncThunk(
...
+        recoveryKeyPair
+    }, { dispatch }) => {
+        try {
+            try {
+                await sendJson('POST', `${ACCOUNT_HELPER_URL}/account/seedPhraseAdded`, {
+                    accountId: implicitAccountId,
+                    publicKey: recoveryKeyPair.publicKey.toString()
+                });
+            } catch (e) {
+                if (e.status === 400) {
```
morgsmccauley, line 245:
```suggestion
                if (e.message.includes('ConditionalCheckFailedException')) {
```
contract-helper throws `400` by default so it may be better to check the message explicitly. Ideally contract-helper would throw a specific error relating to creating duplicates but best we can do now is check for the DynamoDB error.

Thoughts @andy-haynes ?

packages/frontend/src/redux/slices/account/createAccountThunks.js
```diff
@@ -219,3 +228,70 @@ export const finishLocalSetupForZeroBalanceAccount = createAsyncThunk(
...
+        ledgerHdPath
+    }, { dispatch }) => {
+        try {
+            try {
+                await sendJson('POST', `${ACCOUNT_HELPER_URL}/account/ledgerKeyAdded`, {
+                    accountId: implicitAccountId,
+                    publicKey: ledgerPublicKey.toString()
+                });
+            } catch (e) {
+                if (e.status === 400) {
```
morgsmccauley, line 278: Same as above ⬆️

packages/frontend/src/redux/slices/account/createAccountThunks.js
```diff
@@ -219,3 +228,70 @@ export const finishLocalSetupForZeroBalanceAccount = createAsyncThunk(
...
+                    console.log(`Ledger public key ${ledgerPublicKey.toString()} has previously been added as recovery method to account. Continueing setup...`);
+                } else {
+                    throw new WalletError(e, 'initiateSetupForZeroBalanceAccountLedger.error');
+                }
+            }
+            await setKeyMeta(ledgerPublicKey, { type: 'ledger' });
+            await setLedgerHdPath({ accountId: implicitAccountId, path: ledgerHdPath });
```
morgsmccauley, line 285: Previously we would check if the `path` exists before we call `setLedgerHdPath()`, is it safe to assume it will always have a value now?

morgsmccauley: Also this function isn't `async` 😄

morgsmccauley (changes requested):
Looking good! Just couple small questions 😄

packages/frontend/src/redux/slices/account/createAccountThunks.js
Patrick1904: Yep it has the default `ledgerHdPath` set in `SetupLedgerNewAccount` here: https://github.com/near/near-wallet/blob/master/packages/frontend/src/components/accounts/ledger/SetupLedgerNewAccount.js#L17 so it'll always have a path :)

packages/frontend/src/redux/slices/account/createAccountThunks.js
```diff
@@ -219,3 +228,70 @@ export const finishLocalSetupForZeroBalanceAccount = createAsyncThunk(
...
+    }, { dispatch }) => {
+        try {
+            try {
+                await sendJson('POST', `${ACCOUNT_HELPER_URL}/account/seedPhraseAdded`, {
+                    accountId: implicitAccountId,
+                    publicKey: recoveryKeyPair.publicKey.toString()
+                });
+            } catch (e) {
+                if (e.status === 400) {
+                    console.log(`Public key ${recoveryKeyPair.publicKey.toString()} has previously been added as recovery method to account. Continueing setup...`);
```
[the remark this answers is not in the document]
Patrick1904: Good one:)

packages/frontend/src/redux/slices/account/createAccountThunks.js
Patrick1904: Yep I'll use `ConditionalCheckFailedException` for now and we can update if we end up adding better error in contract-helper later

morgsmccauley (approved):
LGTM :)

PR #2701, opened 2022-05-25
Kwlt 12 pre migration banner

gutsyphilip:
In this PR:

- Adds Migration Banner
- Implements [Pre](https://nearinc.atlassian.net/browse/KWLT-12?atlOrigin=eyJpIjoiNjk3MjM3NDIxNDk1NGYzMGIxNDkzNWJlYzJjNWJjMGIiLCJwIjoiaiJ9) and [Post Migration](https://nearinc.atlassian.net/browse/KWLT-13?atlOrigin=eyJpIjoiNjU4YmU1MzMyYjUwNDdjNTkwYzk4OTM0NTA1NTk5NjMiLCJwIjoiaiJ9) Callouts
- Implements feature flagging(SHOW_MIGRATION_BANNER) for Migration banner

Note:
-Introduced new envs - MIGRATION_START_DATE and MIGRATION_END_DATE

Figma link:
https://www.figma.com/file/w4SQVO6kBvAP5D65TGlm7y/Wallet-Domain-Migration?node-id=2%3A70


esaminu:
@gutsyphilip [render deploys](https://dashboard.render.com/static/srv-ca6ogvf9re0l3kma6nsg/deploys/dep-ca7m6nv5f9959gfok8g0) appear to be failing with this error:
```
error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.
```
Could you check it out?

packages/frontend/src/components/common/MigrationBanner.js
```diff
@@ -0,0 +1,160 @@
...
+    height: 40px;
+
+    @media (max-width: 768px) {
+        margin-top: 16px !important;
+    }
+`;
+
+const MigrationBanner = ({ account }) => {
+    const migrationStartDate = new Date( MIGRATION_START_DATE * 1000);
+    const migrationEndDate = new Date(MIGRATION_END_DATE * 1000);
```
esaminu, line 84: I believe the default values for `MIGRATION_START_DATE` are incorrect here. We also aren't retrieving it from `process.env` so this is not currently overridable by an env variable. We need to add `MIGRATION_START_DATE` and `MIGRATION_END_DATE` to `configFromEnvironment.js`.

Also, I suggest we do this in `envParsers.js` by adding a new `parseDateFromShell` function and adding comments to that function indicating exactly what we expect from the environment variable and returning a `Date` object if it's provided and `undefined` if it's not.

packages/frontend/src/components/common/MigrationBanner.js
```diff
@@ -0,0 +1,160 @@
...
+
+    const showPostMigrationBanner = isAfter(new Date(), new Date(migrationStartDate));
```
esaminu, line 86: I believe we're calling `new Date` on a `Date` instance here. Also, I don't see why we need an external dependency to perform this comparison. I believe we can just use the > and < operators. [It seems that's what they're doing anyway](https://github.com/date-fns/date-fns/blob/master/src/isAfter/index.ts)

packages/frontend/src/components/common/MigrationBanner.js
```diff
@@ -0,0 +1,160 @@
...
+
+
+    const getInviteCalendarEventURl =()=>{
+        const calenderURL = new URL('https://calendar.google.com/calendar/u/0/r/eventedit');
+        const calendarStartDate= migrationStartDate.toISOString().replace(/-|:|\.\d\d\d/g,'');
+        const calendarEndDate= migrationEndDate.toISOString().replace(/-|:|\.\d\d\d/g,'');
+
+        calenderURL.searchParams.append('dates', `${calendarStartDate}/${calendarEndDate}`);
+        calenderURL.searchParams.append('text', 'Wallet Migration Event');
+        calenderURL.searchParams.append('details', 'This is the official start date of the near wallet migration. Please make sure to start migrating your account as soon as possible.');
```
esaminu, line 111: We need to set these up for localization

packages/frontend/src/components/common/MigrationBanner.js
```diff
@@ -0,0 +1,160 @@
...
+    }, [account]);
+
+    const preMigrationMarkup = (
+            <ContentWrapper>
+               <div className='content'>
+                    <IconAlertTriangle/>
+                    <Translate id='preMigration.message' data={{ startDate: migrationStartDate.toLocaleDateString() }} />
+               </div>
+                <CustomButton onClick={handleAddToCalendar}>
+                <IconBell/> Set a Reminder
```
esaminu, line 136: This needs to be added to localization

packages/frontend/src/components/common/MigrationBanner.js
```diff
@@ -0,0 +1,160 @@
...
+    );
+
+    const postMigrationMarkup = (
+        <ContentWrapper>
+            <div className='content'>
+                <IconAlertTriangle/>
+                <Translate id='postMigration.message' data={{ endDate: migrationEndDate.toLocaleDateString() }} />
+            </div>
+            <CustomButton onClick={()=>{}}>
+               <IconShare/>  Transfer My Accounts
```
esaminu, line 148: Needs to be added to localization

packages/frontend/src/config/environmentDefaults/mainnet.js
```diff
@@ -19,6 +19,9 @@ export default {
     INDEXER_SERVICE_URL: 'https://api.kitwallet.app',
     LINKDROP_GAS: '100000000000000',
     LOCKUP_ACCOUNT_ID_SUFFIX: 'lockup.near',
+    MIGRATION_CALENDAR_EVENT: 'https://calendar.google.com/event?action=TEMPLATE&tmeid=MGkwbjMxcDI5bWJxaHEzazQwdGs0a244YXIgcGhpbGlwLmMub2Jvc2lAbQ&tmsrc=philip.c.obosi%40gmail.com',
```
esaminu, line 22: This does not appear to be used anywhere and is set in all the `environmentDefaults` files

esaminu (changes requested):
Nice work! Left some comments regarding adding a dependency, env value setup and localization 👍

packages/frontend/src/config/environmentDefaults/mainnet.js
gutsyphilip: Yeah. Switched the implementation.

packages/frontend/src/components/common/MigrationBanner.js
gutsyphilip: Sure thing! Thanks.

packages/frontend/src/components/common/MigrationBanner.js
gutsyphilip: Sure thing! Thanks.

packages/frontend/src/components/common/MigrationBanner.js
gutsyphilip: Sure thing! Thanks.

agileurbanite:
@gutsyphilip can you address comments and conflicts?

features/features.d.ts
```diff
@@ -8,4 +8,7 @@ export type Features = {
 	FARMING_VALIDATOR_APY_DISPLAY: boolean;
 	IMPORT_ACCOUNT_WITH_LINK_V2: boolean;
 	IMPORT_ZERO_BALANCE_ACCOUNT: boolean;
+	USE_INDEXER_SERVICE: boolean;
+	USN_BUTTON: boolean;
```
esaminu, line 12: These were removed in base I believe

esaminu (changes requested):
Awesome stuff 👍 Minor comments inline

features/flags.json
```diff
@@ -262,5 +262,137 @@
...
+      "lastEditedBy": "esaminu",
+      "lastEditedAt": "2022-05-05T18:06:54.374Z"
+    },
+    "mainnet_STAGING_NEARORG": {
+      "enabled": false,
+      "lastEditedBy": "esaminu",
+      "lastEditedAt": "2022-05-05T18:07:28.665Z"
+    }
+  },
+  "SHOW_MIGRATION_BANNER": {
```
esaminu, line 354: IIRC this should be enabled in `_NEARORG` environments?

yarn.lock
```diff
@@ -6067,7 +6067,7 @@ find-root@1.1.0:
   resolved "https://registry.yarnpkg.com/find-root/-/find-root-1.1.0.tgz#abcfc8ba76f708c42a97b3d685b7e9450bfb9ce4"
   integrity sha512-NKfW6bec6GfKc0SGx1e07QZY9PE99u0Bft/0rzSD5k3sO/vwkVUpDUKVm5Gpp5Ue3YfShPFTX2070tDs5kB9Ng==
 
-find-up@^2.0.0, find-up@^2.1.0:
```
andy-haynes, line 6070: Unless you're modifying `package.json` or bumping a transitive dependency, you shouldn't have changes in the lockfile. Please make sure you're running `yarn install --frozen-lockfile` when installing and only include the lockfile in your PR if you're deliberately modifying packages as part of your work.

packages/frontend/src/config/environmentDefaults/development.js
```diff
@@ -18,6 +20,8 @@ export default {
     INDEXER_SERVICE_URL: 'https://testnet-api.kitwallet.app',
     LINKDROP_GAS: '100000000000000',
     LOCKUP_ACCOUNT_ID_SUFFIX: 'lockup.devnet',
+    MIGRATION_START_DATE: parseDateFromShell('1654977600'),
```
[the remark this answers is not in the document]
gutsyphilip: Yeah. this makes sense

esaminu (approved):
LGTM 👍 we'll configure flagging in a follow up

PR #2703, opened 2022-05-25
Refactoring folder accounts and svg in type script

SerDorosh:


andy-haynes:
This is a _massive_ change, something of this scope needs to be discussed and planned in some depth. Neither this PR nor the Jira issue have any context on them, which is really required for a change of this scope.

Please see the comment here: https://github.com/near/near-wallet/pull/2676#issuecomment-1146205856

PR #2718, opened 2022-06-03
When importing an implicit account it's showing FT data from a previous active account

marcinbodnar:
Three fixes:
1. Fix refreshing account, we need to wait till the `refreshAccount` action is resolved, even if currently it's throwing an error. That's because the active account is changed by `refreshAccount` action.
If we redirect to the dashboard before the `refreshAccount` is resolved, FTs are fetched for the precious account, because it's still an active account.
2. Fixed redirect. We should use `redirectToApp` instead of `redirectTo`. If for example the user will be redirected to Wallet to login to the app in case there is no imported account yet, and he will import a zero balance account, we should redirect the user to the `/login` page anyway, so user will be able to import different account and proceed with login.
3. Clean up the code formatting.

### DIFF
--- a/packages/frontend/src/components/accounts/CouldNotFindAccountModalWrapper.js
+++ b/packages/frontend/src/components/accounts/CouldNotFindAccountModalWrapper.js
@@ -2,9 +2,9 @@ import React from 'react';
 import { useDispatch } from 'react-redux';
 
 import {
-    redirectTo,
     refreshAccount,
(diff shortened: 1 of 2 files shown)


andy-haynes (approved):
Nice work with the quick fix 😎

marcinbodnar:
Thanks @andy-haynes it was hard to figure out, what's exactly is the problem

PR #2720, opened 2022-06-05
fix: only include NEAR contractName with optional prop

esaminu:
This PR fixes reliance on NEAR token instances not having a `contractName` prop by only including that prop optionally and setting explicitly to true in cases where `contractName` is required


esaminu:
Merging hotfix for now, post deploy review / discussion can be done

PR #2723, opened 2022-06-06
KWLT-18 Update MixPanel tokens

PinkiNice:
In this PR:

- Update testnet & mainnet MixPanel tokens with new ones


agileurbanite (changes requested):
Going to block this @PinkiNice let's connect

PR #2733, opened 2022-06-07
fix(translation): translation href escapes and validation

andy-haynes:
This PR adds escaping to the `href` on `v2.connectConfirm.desc` as well as a new test to catch missing quotes around `href` attributes across translation files.


Patrick1904 (approved):
LGTM

PR #2738, opened 2022-06-08
chore(flag): Remove IMPORT_ACCOUNT_WITH_LINK_V2 flag

andy-haynes:


packages/frontend/src/utils/wallet.js
```diff
@@ -859,10 +857,6 @@ class Wallet {
             accountIds = [accountId];
         }
 
-        if (!IMPORT_ACCOUNT_WITH_LINK_V2) {
-            accountIds.push(implicitAccountId);
-        }
-
         // TODO: getAccountIds returns all accounts including any implicit account.
         // Once 'IMPORT_ACCOUNT_WITH_LINK_V2' feature is shipped:
```
andy-haynes, line 861: Does this comment just refer to the removed check above? Not sure if there's anything else that needs to be removed here (besides this comment 🙂)

Patrick1904: Sweet! Yep the comment is only for the above check so can remove it now :)

Patrick1904 (approved):
LGTM 🙏

Patrick1904 (approved):

PR #2743, opened 2022-06-10
KWLT-19 Disable convenient recovery methods

lapkoshka:
**NOTE**: it is not full refactoring, at this pull request we can solve problem faster and remove JSX only.
Users can't use sms/phone recovery method more. At next pull request all related code will be removed competly.

Create Wallet form:
image

Profile page:
image


andy-haynes (changes requested):
Looks great! Just a few more unreferenced things that would be good to remove while you're in here.

Despite this not being a "full" refactor it would be best to remove the remaining code in these modules that is now deprecated, in particular the remaining state and methods which are not called or referenced. Translations can wait for now but let's make sure they are accounted for in the follow up PR please.

Patrick1904 (changes requested):
@loosefer it seems that this PR removes email/phone completely for all users, including users who currently have an active email/phone recovery method. We'll need to allow these users to manage/disable their email/phone recovery method.

loosefer:
image

Patrick1904 (changes requested):
Thanks @loosefer! Let's make sure remove the title for this section ("Most Convenient") when there are no options and avoid abnormal spacing

Screen Shot 2022-06-20 at 11 14 17 AM

packages/frontend/src/components/profile/Recovery/ActiveMethod.js
```diff
@@ -183,6 +188,24 @@ class ActiveMethod extends Component {
                                 <div>
                                     <Translate id={`recoveryMgmt.${data.kind !== 'phrase' ? 'disableTextLink' : 'disableTextPhrase'}`}/>
                                 </div>
+                                {
+                                    data.kind === 'email' && (
+                                        <UnsupportedMethod>
+                                            <Translate>
+                                                {({ translate }) => translate('recoveryMgmt.unsupportedMethodEmail')}
```
andy-haynes, line 195: Is there any reason we can't use the props syntax here?

i.e.
```js
<Translate id='recoveryMgmt.unsupportedMethodEmail' />
```

Patrick1904 (dismissed):
LGTM

gutsyphilip (dismissed):
This looks good! Just needs to be rebased.

loosefer:
@gutsyphilip @Patrick1904 
Done

gutsyphilip (approved):
Awesome! Nice work @loosefer

packages/frontend/src/components/profile/Profile.js
```diff
@@ -231,6 +232,11 @@ export function Profile({ match }) {
 
     const MINIMUM_AVAILABLE_TO_TRANSFER = new BN('10000000000000000000000');
 
+    // show most convenient methods if it was enabled
+    const activeMethodsMap = getActiveMethodsMap(getActiveMethods(userRecoveryMethods));
+    let shouldShowEmail = Boolean(activeMethodsMap.email.publicKey);
+    let shouldShowPhone = DISABLE_PHONE_RECOVERY && Boolean(activeMethodsMap.phone.publicKey);
```
andy-haynes, line 238: All we need to know here is whether the user has recovery methods of either type `email` or `phone`, instead of using helper methods we can just check `userRecoveryMethods` directly:
```js
const shouldShowEmail = userRecoveryMethods.some(({ kind }) => kind === 'email');
const shouldShowPhone = userRecoveryMethods.some(({ kind }) => kind === 'phone');
```

Then the `getActiveMethods*` functions are no longer needed and can be removed. The `publicKey` checks should also not be necessary once we're checking `userRecoveryMethods` directly, since it won't exist if there's not a value for `publicKey`.

andy-haynes (changes requested):
Just one comment on checking recovery methods in the profile page, I think we can simplify the check and avoid duplicating the wonky logic in `RecoveryContainer.js`.

packages/frontend/src/components/profile/Profile.js
loosefer: But then user can't switching off enabled method.
PublicKey check give a possibility to detect if method enabled, then we can show selector and give a chance to disable it.

andy-haynes: I don't think that's the case, if the method exists in `userRecoveryMethods` then they have it enabled. If they have it enabled then we can show the UI to disable it.

The call to `getActiveMethods` is confusing and unnecessary here because it returns an array with an entry for _all_ recovery methods, regardless of whether the user actually has them. That's why it is required to check for `publicKey` currently, because if it didn't exist in `userRecoveryMethods` then `getActiveMethods` adds it to the array without a value for `publicKey`. So why not check `userRecoveryMethods` directly since it is the actual source of truth? Then the `getActiveMethods*` helper methods copied from `RecoveryContainer` can be removed from this PR.

I've applied the changes above to my local copy to demonstrate the flow in testnet. Note that I start without an email recovery method, can only add one using wallet.testnet.near.org, and once added I can remove it using the code running with my changes:

PR #2746, opened 2022-06-10
chore: no longer render release notes modal

andy-haynes:
This PR removes the rendering of `ReleaseNotesModal` from the main wallet page component. The information on this modal is currently woefully out of date and does not serve any productive purpose.

Until there's a process for keeping this updated with useful information it is better to not display it at all.


packages/frontend/src/components/wallet/Wallet.js
```diff
@@ -321,7 +320,6 @@ export function Wallet({
         <StyledContainer
             className={SHOW_NETWORK_BANNER ? 'showing-banner' : ''}
         >
-            <ReleaseNotesModal />
```
andy-haynes, line 324: Not sure we need to do a more in-depth removal given that the component may have some future potential.

Patrick1904 (approved):
Great!

Issue #2722, opened 2022-06-06
UX Fix: Importing implicit accounts with zero balance ux


agileurbanite:
https://www.loom.com/share/0685ac66114640b1a9df27920ac56168?replyToComment=8707792

Ryan recommends:

> recommend clarifying on the center modal that only accounts who have already experienced transactions on-chain will be found, but this secure passphrase is still valid for starting a new account. The "no activity yet" phrase isn't enough, in my opinion. Maybe offer a link from the modal to a page with further explanation.

agileurbanite:
@nall-near can you take a look at the above request? FYI @Patrick1904

nall-near:
Hi @agileurbanite, I took a look and reworked the copy a bit. Would appreciate yours + @heycorwin feedback. Next step is removing the console error window (top right align).
![image](https://user-images.githubusercontent.com/103140483/172236346-b5b25bf3-1678-499e-a209-15443a5a7c85.png)

heycorwin:
Just a thought/question: If a user has a valid passphrase, that means they posses the private key for an account. The account either contains funds or it does not, but regardless, they own the account by way of owning its private key. Why not just import the account in any case and not worry about displaying this confusing messaging? The reality is that an account will _always_ be "found" as long as the passphrase is valid. Whether or not that account has been funded or not is a bit of a different matter.

One problem with the language here is primarily with the term "active account". The only other time we use this language in the product is when we are talking about the account that is currently selected by the user in the interface. We don't use the term "active" to describe accounts that have been funded.

Thoughts @nall-near @agileurbanite

nall-near:
Great points @heycorwin. Would also prefer to just avoid the message entirely. It seems like it might be there specifically to address the "hey man, you need to fund/use this thing" agenda.

The main thing I would want to address if required would be the difference between an inaccurate passphrase and account not found on the chain. Curious if the "we can't find your account" comes after some sort of check on chain, and how that differs from going through with the import. I think we need this detail to make the right choice for the messaging.

Need to understand the point of the initial check first, and if we could bypass the "account not found" and go straight to import (would bounce back if the phrase was actually incorrect)

heycorwin:
I believe seed phrase validation was something we also had an issue for and were planning on implementing after the implicit account stuff had been completed. @Patrick1904 can confirm.

Patrick1904:
[...]

PR #2747, opened 2022-06-10
feat: Import zero balance account automatically (no prompt)

Patrick1904:
Referencing issue: https://github.com/near/near-wallet/issues/2722#issuecomment-1151111137

This PR skips the 'import anyway' prompt while importing a zero balance account. The user is notified with a modal post import instead.

To test this PR, simply try importing an implicit account (seed, link, ledger) that does not currently exist on chain.

Screen Shot 2022-06-10 at 4 10 23 PM

Additional notes:
1. Cleaned up `SignInLedgerWrapper` and removed the ability to manually specify account id's to import in case no accounts id's were found with the corresponding ledger key since we now opt to simply import the default implicit version of the ledger key.


andy-haynes (approved):
:shipit:

PR #2752, opened 2022-06-14
Migration Modal using batch import route

gutsyphilip:
In this PR:

- Wallet migration modal prompt
- Wallet migration encode account data to batch import url
- Allow select different wallet types

Deployment Preview URL: https://my-near-wallet-pr-2752.onrender.com/.

### How to test:
1. Click the link above to open up the deploy preview.
2. Import your existing account or create a new one.
3. You should see a banner at the top of the page prompting you to migrate your account. Follow the instructions therein to migrate your account to mynearwallet.


packages/frontend/src/components/wallet-migration/GenerateMigrationPin.jsx
```diff
@@ -0,0 +1,110 @@
...
+    }
+`;
+
+
+const GenerateMigrationPin = ({ handleSetActiveView, migrationPin }) => {
+    const [successSnackbar, setSuccessSnackbar] = React.useState(false);
+
+    const handleCopyPin = () => {
+        copyText(migrationPin);
+        setSuccessSnackbar(true, () => {
```
andy-haynes, line 70: There is no second argument to the setter returned from `setState`, is this doing what you expect?
```
State updates from the useState() and useReducer() Hooks don't support the second callback argument. To execute a side effect after rendering, declare it in the component body with useEffect().
```

I think you want a `useEffect` hook to facilitate this with a dependency on `successSnackBar`, e.g.
```js
useEffect(() => {
  if (successSnackBar) {
    ...
  }
}, [successSnackBar]);
```

packages/frontend/src/components/wallet-migration/GenerateMigrationPin.jsx
```diff
@@ -0,0 +1,110 @@
...
+                <p><Translate id='walletMigration.generateMigrationPin.desc'/></p>
+                <PinViewer>
+                    {migrationPin}
+                    <IconCopy onClick={handleCopyPin} />
+                </PinViewer>
+                {successSnackbar &&<p><Translate id='walletMigration.generateMigrationPin.pinCopySuccess'/></p> }
+
+              
+                <ButtonsContainer>
+                    <StyledButton className="gray-blue" onClick={()=>{handleSetActiveView(WALLET_MIGRATION_VIEWS.MIGRATION_PROMPT);}}>
```
andy-haynes, line 98: 💄 no need for the `{}` around the lambda body, you can return directly

packages/frontend/src/components/wallet-migration/MigrationPromptModal.jsx
```diff
@@ -0,0 +1,96 @@
...
+                <MyNearWalletLogo className="logo" mode={1} />
+                <h3 className='title'><Translate id='walletMigration.migrationPrompt.title'/></h3>
+                <p><Translate id='walletMigration.migrationPrompt.desc'/></p>
+            </ContentContainer>
+            <StyledDivider />
+           <ButtonsContainer>
+                <StyledButton onClick={handleRedirectToBatchImport}>
+                    <Translate id='walletMigration.migrationPrompt.transferAccountsBtn' />
+                 </StyledButton>
+                 <StyledButton className="link" onClick={()=>{handleSetActiveView(WALLET_MIGRATION_VIEWS.SELECT_DESTINATION_WALLET);}}>
```
andy-haynes, line 88: This would be a lot simpler if it didn't need to specify a parameter. By requiring a parameter the implication is that there may be some other value in `WALLET_MIGRATION_VIEWS` that we want to set here, but since it's hardcoded that's not necessary.

Instead of `handleSetActiveView` this should be `selectDestinationWallet` and should be invoked without any parameters. Then this component doesn't have to know anything about the callback other than it's name - just like how you implemented `handleRedirectToBatchImport`.

packages/frontend/src/components/wallet-migration/SelectDestinationWallet.jsx
```diff
@@ -0,0 +1,161 @@
...
+                            onClick={() => handleSetWalletType(id)}
+                            key={name}
+                            >
+                            <h4 className='name'>{name}</h4>
+                            {icon}
+                        </WalletOptionsListingItem>;
+                    })}
+                </WalletOptionsListing>
+                <ButtonsContainer>
+                    <StyledButton className="gray-blue" onClick={()=>{handleSetActiveView(WALLET_MIGRATION_VIEWS.MIGRATION_PROMPT);}}>
```
andy-haynes, line 149: Same comment as above with `handleSetActiveView` - specify the parameter when passing props to this component, not within this component. It doesn't need to know about `WALLET_MIGRATION_TYPES`, it just needs a function to call when this button is pressed that makes it clear it's going to display a migration prompt (e.g. `showMigrationPrompt`)

packages/frontend/src/components/wallet-migration/SelectDestinationWallet.jsx
```diff
@@ -0,0 +1,161 @@
...
+            modalSize='md'
+            style={{ maxWidth: '435px' }}
+        >
+            <Container>
+                <IconWallet/>
+                <h4 className='title'><Translate id='walletMigration.selectWallet.title'/></h4>
+                <WalletOptionsListing>
+                    {WALLET_OPTIONS.map(({ id, name, icon }) => {
+                        const isSelected = id === walletType;
+                        return <WalletOptionsListingItem
```
andy-haynes, line 138: Please wrap your JSX in `()` so that it's clear exactly what's being returned. I have an eslint rule pending for this but it's a good habit to get into 🙂

packages/frontend/src/components/wallet-migration/SelectDestinationWallet.jsx
```diff
@@ -0,0 +1,161 @@
...
+            disableClose={true}
+            modalSize='md'
+            style={{ maxWidth: '435px' }}
+        >
+            <Container>
+                <IconWallet/>
+                <h4 className='title'><Translate id='walletMigration.selectWallet.title'/></h4>
+                <WalletOptionsListing>
+                    {WALLET_OPTIONS.map(({ id, name, icon }) => {
+                        const isSelected = id === walletType;
```
andy-haynes, line 137: It's fine to leave this inline rather than assigning it to a variable, it will still be clear as
```js
{ active: id === walletType }
```

andy-haynes (changes requested):
Looking good so far, I have some comments around component structure and clarity that will be important as we iterate on this. I will add more feedback as I continue to test the functionality.

esaminu:
Can we only redirect / display the banner if there is at least one account to be migrated? User lands on 404 if they have no accounts yet

judith-Near:
`Can we only redirect / display the banner if there is at least one account to be migrated? User lands on 404 if they have no accounts yet` -esaminu
or maybe just conditionally render the transfer accounts button? the banner is helpful even to those without accounts, because then they know mnw is where they should create an account

judith-Near:
i imported an account, then clicked transfer my accounts, and got a 404 with the following link:
`https://testnet.mynearwallet.com/batch-import#keys=&ledgerHdPaths=&accounts=`

i think it's because the testnet rpc call to view account is erroring out, we might need to account for this? like only load the link if there are accounts present?

judith-Near:
after talking with @esaminu and @nall-near, i think the copy of the modal's buttons should be changed to "Transfer Accounts to MyNearWallet" followed by "Choose a Different Wallet". the reason is that "use a different wallet" was not intuitive/left me stumped. use a different wallet for what?

esaminu:
@gutsyphilip Please close this and resolve feedback by sending PRs to #2780 including any relevant feedback on that PR

Issue #2755, opened 2022-06-16
Bumping Linkdrop Attached GAS


BenKurrek:
I'm currently working on the [linkdrop proxy](https://github.com/near-examples/linkdrop-proxy) contract which allows for NFTs, FTs, and arbitrary function calls to be executed upon claiming a linkdrop. Currently, 100 TGas is being attached to the `claim` and `create_account_and_claim` functions which has made it extremely tight when writing the logic and creating interesting onboarding experiences for users.

Can we bump this number to 150 or 200 TGas? What are people's thoughts on this?

judith-Near:
I think this should be fine, if a user attaches more gas than the gas used then the system sends a refund -- you can see this in the refund receipt in the transaction execution plan. We still shouldn't go too high on the gas, because it might not be refunded the exact amount back, but 150 or 200 TGas should be fine.

BenKurrek:
Awesome should I throw up a PR or is that something y'all handle?

BenKurrek:
bumping gas is not needed and should not be done. Everything right now works with 100 TGas. If we bump the gas, the access key allowance for all keys will go up significantly which means higher upfront costs for users.

In the far far future, however, if we want to add much more complicated scenarios like transferring many different FTs and NFTs in the same key use, we could allow the Gas attached to be embedded into the linkdrop itself so it’s like /linkdrop/contract/secreteKey#gas=GAS_AMOUNT and this gas param would be attached. That way for specific drops that need more Gas, they can specify it and it doesn’t harm everybody else’s upfront costs

I was unaware of the access key pessimistic allowance calculation before this PR was created

PR #2756, opened 2022-06-16
Bump linkdrop GAS from 100 TGas to 200 TGs

BenKurrek:
Fixes https://github.com/near/near-wallet/issues/2755

### DIFF
--- a/packages/frontend/src/config/environmentDefaults/development.js
+++ b/packages/frontend/src/config/environmentDefaults/development.js
@@ -16,7 +16,7 @@ export default {
     EXPLORER_URL:'https://explorer.testnet.near.org',
     HIDE_SIGN_IN_WITH_LEDGER_ENTER_ACCOUNT_ID_MODAL: false,
     INDEXER_SERVICE_URL: 'https://testnet-api.kitwallet.app',
-    LINKDROP_GAS: '100000000000000',
+    LINKDROP_GAS: '200000000000000',
     LOCKUP_ACCOUNT_ID_SUFFIX: 'lockup.devnet',
     MIN_BALANCE_FOR_GAS: nearApiJs.utils.format.parseNearAmount('0.05'),
     MIN_BALANCE_TO_CREATE: nearApiJs.utils.format.parseNearAmount('0.1'),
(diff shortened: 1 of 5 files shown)


esaminu:
Thanks for the PR @BenKurrek! Do you think it's necessary to check the allowance of the funding key and do a `Math.min` or will most keys have a larger allowance than 200 TGas in this use case? If there are cases where keys don't have the allowance we may benefit from either checking before hand or retrying with less gas if the transaction fails.

BenKurrek:
@esaminu yea maybe it would make sense to attach whatever allowance the key has? If the keys have 200 TGas, attach 200 TGas but if they have 100, attch 100. Similarly, if they have 50 TGas, we attach that amount up to a max of 300 TGas?

mattlockyer:
@MaximusHaximus you're right I was aiming to reduce but we're now talking about a contract with arbitrary callbacks. If links were distributed with some complicated callback logic it would be nice to use the max possible gas.

I think a URL parameter is a nice compromise, it would allow any devs to direct users to the NEAR Wallet and specify the max gas they think they need for the linkdrop and resulting callback functionality.

Any idea on a timeline for this?

BenKurrek:
@MaximusHaximus I think we could go either way but maybe attaching it in the URL might be a better solution both in terms of implementation and reduces any chance for breaking changes.

Thoughts Matt?

gutsyphilip:
Just jumping in to refresh this conversation to be clear on what the agreed direction is here. Most liklely adding an optional url argument? 

Do we close this for now and have that done in another PR?

BenKurrek:
bumping gas is not needed and should not be done. Everything right now works with 100 TGas. If we bump the gas, the access key allowance for all keys will go up significantly which means higher upfront costs for users.

In the far far future, however, if we want to add much more complicated scenarios like transferring many different FTs and NFTs in the same key use, we could allow the Gas attached to be embedded into the linkdrop itself so it’s like /linkdrop/contract/secreteKey#gas=GAS_AMOUNT and this gas param would be attached. That way for specific drops that need more Gas, they can specify it and it doesn’t harm everybody else’s upfront costs

I was unaware of the access key pessimistic allowance calculation before this PR was created

PR #2760, opened 2022-06-17
MYNW-64: add MixPanel events to top-up options

PinkiNice:
https://mnw.atlassian.net/browse/MYNW-64

**Problem:**
MixPanel events were missing for UTORG, NearPay, RainbowBridge and all exchanges

**Fix:**
Added MixPanel events for this actions


andy-haynes (approved):
Looks good! I noticed that the `binance` and `okex` keys didn't match their values so I went ahead and swapped them even though it wasn't introduced in this PR.

Issue #2646, opened 2022-05-08
HAPI integration 


Solniechniy:
## Problem
There is a problem with managing addresses that are under regulations during sanctions, scams, and so on.

#### What are we planning to do about it?
Integrate the HAPI protocol into the near wallet, which allows checking the address before sending the funds and displaying the information about it. 

#### Job Story(s)
- User wants to send funds to another address. While entering the address of the receiver wallet sends get requests to HAPI protocol in order to check any information about this address. If there is any - show it with the red alert or input border.

## Scope
- [x]  Create hapi service which will give the information about address in module/services
- [x]  Create the design proposal which will display the information about the checked address from HAPI protocol smartcontract 
- [x]  Expanding the wallet functionality according to created design proposal

PinkiNice:
Implemented in #2840

PinkiNice:
Implemented in #2840

agileurbanite:
Wait did security review this service? @PinkiNice

PR #2763, opened 2022-06-20
[DRAFT]: Initial setup for HapiService, which used hapi-protocol for checking …

askulikov:
Initial setup for HapiService, which used hapi-protocol for checking near addresses.
For now, the result of this view call is just logging in the console. 
The expected result for any valid address is `['None', 0]`, where:
* `None` - risk category
* `0` - risk score.

You can insert the address `ref-test.rkonoval.testnet` for testing purposes. The expected result: `['WalletService', 6]`;
All possible categories you may find in this [enum](https://github.com/HAPIprotocol/near-proxy-contract/blob/2cb7699dcbd47f472b21dff1010b09991418c77a/contract/src/lib.rs#L9);
The risk score is a number from `0` to `10`, where `10` is the highest risk.

The Issue - https://github.com/near/near-wallet/issues/2646


packages/frontend/src/config/configFromEnvironment.js
```diff
@@ -85,5 +85,6 @@ module.exports = {
     FARMING_CLAIM_GAS: process.env.FARMING_CLAIM_GAS || parseNearAmount('0.00000000015'),
     FARMING_CLAIM_YOCTO: process.env.FARMING_CLAIM_YOCTO || '1',
     REF_FINANCE_API_ENDPOINT: process.env.REF_FINANCE_API_ENDPOINT,
-    REF_FINANCE_CONTRACT: process.env.REF_FINANCE_CONTRACT
+    REF_FINANCE_CONTRACT: process.env.REF_FINANCE_CONTRACT,
+    HAPI_PROTOCOL_ADDRESS: process.env.HAPI_REPORTER_ADDRESS
```
judith-Near, line 89: `HAPI_PROTOCOL_ADDRESS: process.env.HAPI_PROTOCOL_ADDRESS` so that they match

packages/frontend/src/config/environmentDefaults/development.js
```diff
@@ -50,5 +50,6 @@ export default {
     FARMING_CLAIM_YOCTO: '1',
     REF_FINANCE_API_ENDPOINT: 'https://dev-indexer.ref-finance.com/',
     REF_FINANCE_CONTRACT: 'ref-finance-101.testnet',
-    USN_CONTRACT: 'usdn.testnet'
+    USN_CONTRACT: 'usdn.testnet',
+    HAPI_PROTOCOL_ADDRESS: 'proxy.contracts.sergei24.testnet'
```
judith-Near, line 54: can you elaborate on what proxy.contracts.sergei24.testnet is?

judith-Near: ah i see it's the constant provided by the [near proxy contract](https://github.com/HAPIprotocol/near-proxy-contract)

packages/frontend/src/components/send/components/InputAccountId.js
```diff
@@ -179,6 +180,10 @@ class InputAccountId extends Component {
             }
             setAccountIdIsValid(false);
         }
+
+
+        const hapiStatus =  await HapiService.checkAddress({accountId});
+        if (hapiStatus) console.log('hapiStatus', hapiStatus);
```
andy-haynes, line 186: This is effectively just some debugging code, let's please remove it from this PR since it doesn't need to be run in production environments yet. It should get a real implementation once the integration is ready, ideally one that uses our feature flag implementation so that it can be enabled in testnet first.

judith-Near (approved):

packages/frontend/src/components/send/components/InputAccountId.js
askulikov: I wrote it mostly for demonstration purposes. Thought that this PR will be a draft for some time, till the full requirements won't be known. But for now, this part just deleted.

agileurbanite:
... wait but why? why are we doing this? why should this be an option?

agileurbanite (changes requested):
need more context on this

PR #2771, opened 2022-06-21
MYNW-62 Transfer Accounts

loosefer:
[Task](https://mnw.atlassian.net/jira/software/projects/MYNW/boards/1?selectedIssue=MYNW-62)
Users should be able to transfer keys into MyNearWallet with a single click.

- update a migration banner, change "Set a reminder" button to "Transfer my accounts"
- updating a existed transfer flow from [this](https://github.com/near/near-wallet/pull/2732) and [this](https://github.com/near/near-wallet/pull/2751) PRs
- change PIN encoding scheme to encoding with secret/public keys

image
image
image
image

image
image
image
image

PR #2780, opened 2022-06-24
Mynw 62

esaminu:
#2771 rebased on #2752 

The MNW migration can be tested via the [NEARORG preview](https://near-wallet-pr-2799.onrender.com/) (Create accounts then click "Transfer my Accounts" on the banner) and then once redirected updating the MNW url to the MNW preview: `https://my-near-wallet-pr-2799.onrender.com/`

Deploy preview branch for this PR (flag on): #2799 

This PR:

- Updates the banner from pre / post migration to a single banner that either links to MNW when no accounts present or triggers account migration modals if current wallet has accounts
- Adds migration, use a different wallet, copy secret and continue to MNW modals
- Adds encryption to the payload for migrating accounts outlined in [KWLT-8](https://nearinc.atlassian.net/browse/KWLT-8)
- Adds a decryption step to the `/batch-import` route


Patrick1904 (commented):
@esaminu can we please fix the conflicts on this PR

gutsyphilip:
This mostly looks good to me. I'll push in a few commits to remove some redundant code from combining both implementations.

packages/frontend/src/components/accounts/batch_import_accounts/index.js
```diff
@@ -75,6 +179,7 @@ const BatchImportAccounts = ({ accountIdToKeyMap, onCancel }) => {
         return <BatchImportAccountsSuccessScreen accounts={accountsApproved} />;
     }
 
+    const accounts = state.accounts.filter(({ accountId }) => Boolean(accountId));
```
esaminu, line 182: @loosefer Why is this necessary? If it is can we use a memoized version?

features/flags.json
```diff
@@ -160,9 +160,9 @@
       "lastEditedAt": "2022-05-25T01:11:55.121Z"
     },
     "testnet_NEARORG": {
-      "enabled": false,
+      "enabled": true,
```
esaminu, line 163: This is flagged on temporarily for testing on the preview. Should be flagged off before merge

packages/frontend/src/components/wallet-migration/WalletMigration.jsx
```diff
@@ -0,0 +1,131 @@
...
+
+
+const encodeAccountsToURL = async (accounts, publicKey) => {
+    const accountsData = [];
+    for (let i = 0; i < accounts.length; i++) {
+        const accountId = accounts[i];
+        const keyPair = await wallet.getLocalKeyPair(accountId);
+        accountsData.push([
+            accountId,
+            keyPair.secretKey,
```
gutsyphilip, line 39: I get an error when I try to migrate accounts:
image

packages/frontend/src/components/wallet-migration/WalletMigration.jsx
```diff
@@ -0,0 +1,131 @@
...
+            getLedgerHDPath(accountId),
+        ]);
+    }
+
+    const hash = encodeAccountsTo(accountsData, publicKey);
```
gutsyphilip, line 44: `encodeAccountsTo` not sure if you planned to update this? The function name seems incomplete.

packages/frontend/src/components/Routing.js
```diff
@@ -292,6 +297,16 @@ class Routing extends Component {
         this.pollTokenFiatValue = null;
     };
 
+    // todo refactor: move to state manager
```
gutsyphilip, line 300: This feels like something we'd never really get back to doing? Maybe we do it now or just remove it since this flow is only valid for the migration and will be removed afterwards.

packages/frontend/src/components/Routing.js
```diff
@@ -564,23 +588,10 @@ class Routing extends Component {
...
-
-                                const accountIdToKeyMap = accounts.reduce((acc, curr) => {
-                                    const [ accountId, keyIndex, ledgerHdPathIndex ] = curr.split('*');
-                                    return { ...acc, [accountId]: {key: keys[keyIndex], ledgerHdPath: ledgerHdPaths?.[ledgerHdPathIndex]} };
-                                }, {});
-                                return <BatchImportAccounts accountIdToKeyMap={accountIdToKeyMap} onCancel={() => this.props.history.replace('/')}/>;
-                            })} />
+                            <Route exact path="/batch-import" render={() =>
+                                (<BatchImportAccounts
+                                    onCancel={() => this.props.history.replace('/')} />)}
```
gutsyphilip, line 593: Not sure I see why `onCancel` needs to be passed here rather than be a function within the component?

gutsyphilip (changes requested):
Mostly looks good. Just raised a couple of things and noticed a bug.

packages/frontend/src/components/accounts/batch_import_accounts/index.js
loosefer: @esaminu Legacy from my firsts commits, deleted

packages/frontend/src/components/Routing.js
loosefer: I think no doesn't matter where declarate `history.replace`, here, on top-level component, or inside FC.

features/flags.json
andy-haynes: This should really be a draft PR then to minimize the chance of forgetting to reset this. Alternatively create a branch off of this and create a draft PR for that, or better yet test it locally against testnet 🙂

gutsyphilip (approved):

features/flags.json
esaminu: Tested both flag states in local but when choosing which one to be more public about going with flag on state to have all folks incl non technical be able to test the feature. Created #2799 for testing and reset this 👍

PinkiNice (approved):

PR #2783, opened 2022-06-28
Move Migration Landing Page to New Project

gutsyphilip:
Implements new post migration landing page for NEAR Wallet.
https://near-wallet-pr-2783.onrender.com/

<hr/>

image

<hr/>

image


gutsyphilip:
Export accounts flow will need to be updated to reflect updates in #2780.

cc: @loosefer @esaminu

packages/frontend-near-wallet/src/assets/favicon.svg
```diff
@@ -0,0 +1,26 @@
+<svg width="80" height="92" viewBox="0 0 80 92" fill="none" xmlns="http://www.w3.org/2000/svg">
```
esaminu, line 1: This should be wallet.near.org's logo not MNW

packages/frontend-near-wallet/src/utils/constants.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const WALLET_MIGRATION_VIEWS = {
+    MIGRATION_PROMPT: 'MIGRATION_PROMPT',
+    SELECT_DESTINATION_WALLET: 'SELECT_DESTINATION_WALLET',
+    MIGRATION_SECRET: 'MIGRATION_SECRET',
+    MIGRATE_ACCOUNTS: 'MIGRATE_ACCOUNTS'
+
+};
+
+export const getMyNearWalletUrl = (https = true) => {
```
esaminu, line 12: We need to make sure `NEAR_WALLET_ENV` is set to non `*_NEARORG` when we deploy the new landing page or use `getMyNearWalletUrlFromNEARORG`. Pretty sure there are conflicts here since `encodeAccountsToURL` uses `getMyNearWalletUrlFromNEARORG` and `RedirectPage` uses `getMyNearWalletUrl`. Please test the whole thing with a set `NEAR_WALLET_ENV` and flag a deploy with that `NEAR_WALLET_ENV` when it's time to deploy

packages/frontend-near-wallet/src/modules/wallet-migration/components/SelectDestinationWallet.tsx
```diff
@@ -0,0 +1,161 @@
...
+                        <h4 className='name'>{name}</h4>
+                        {icon}
+                    </WalletOptionsListingItem>;
+                })}
+            </WalletOptionsListing>
+            <ButtonsContainer>
+                <StyledButton onClick={handleCancel} variant="gray">
+                    Cancel
+                </StyledButton>
+                <StyledButton onClick={handleTransferMyAccounts}>
```
esaminu, line 46: This is broken for ledger. It proceeds with MNW secret page and migration even if you select ledger

packages/frontend-near-wallet/src/modules/wallet-migration/components/MigrateAccounts.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+const MigrateAccounts = ({ onContinue, onClose }) => {
+    const accounts = getAvailableAccounts()
+
+    return (
+        <StyledContainer>
+            <IconMigrateAccount />
+            <h3 className='ttl'>We found {accounts.length} active account(s)</h3>
+            <p className='desc'>
+                The following accounts will be transferred to MyNearWallet:
```
esaminu, line 19: i18n here? Speaking of which we should probably include a language select option on the new page. Maybe in a follow up? cc @agileurbanite @heycorwin @nall-near

packages/frontend-near-wallet/src/utils/constants.ts
```diff
@@ -0,0 +1,21 @@
+import Environments from '../../../../features/environments.json';
+import { NEAR_WALLET_ENV } from './config';
+
+export const WALLET_MIGRATION_VIEWS = {
+    MIGRATION_PROMPT: 'MIGRATION_PROMPT',
```
esaminu, line 5: I believe this screen will be removed to maintain neutrality of `near.org` cc @agileurbanite

packages/frontend-near-wallet/src/modules/redirects/pages/RedirectPage.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import { useLocation } from 'react-router-dom'
+import { getMyNearWalletUrl } from '../../../utils/constants';
+
+const RedirectPage = () => {
```
esaminu, line 5: Nice but I believe there is a screen here to prompt the user to select a wallet to redirect to cc @agileurbanite

esaminu (changes requested):
Nice structure and redirect handling. I found some issues that I pointed out in comments 👍

packages/frontend-near-wallet/src/modules/wallet-migration/components/MigrateAccounts.tsx
nall-near: Agreed @esaminu, would be good to make that a standard practice across this flow eventually.

packages/frontend-near-wallet/src/modules/redirects/pages/RedirectPage.tsx
gutsyphilip: Not at the moment, except there's some design updates in this regard. We had a redirect prompt for when the transfer is initiated but @nall-near said we could remove it earlier.

packages/frontend-near-wallet/src/modules/wallet-migration/components/SelectDestinationWallet.tsx
gutsyphilip: Yeah. There's no support for ledger here yet

packages/frontend-near-wallet/src/modules/wallet-migration/components/MigrateAccounts.tsx
gutsyphilip: To be handled in another task.

packages/frontend-near-wallet/src/utils/constants.ts
gutsyphilip: This is scheduled to be handled in a follow up task. The flow for redirect will change based on the new design. We want to allow users select a destination wallet of their choice and they'd get redirected accordingly.

packages/frontend-near-wallet/src/modules/wallet-migration/components/MigrationSecret.tsx
```diff
@@ -0,0 +1,104 @@
...
+    }, []);
+
+    return (
+        <StyledContainer>
+            <MyNearWalletLogo className="logo" />
+            <h3 className='ttl'>Password required to import accounts</h3>
+            <p className='desc'>
+                You'll need this <strong>password</strong> to securely import your accounts to MyNearWallet. Copy or write it down until you've completed transferring your accounts.
+            </p>
+            <ClickToCopy
```
roshaans, line 24: I think there should be a clear call to action button to copy the the password. If the user is on a touch screen device, they won't have any context as to why they can not hit the `continue` button.

Either, hitting `continue` button should prompt them with a popup that says something like 'I have copied down the password.' and have a button for `cancel` and `continue`

or the continue button should be in a disabled state, until the user copies the password and there should definitely be a copyToClipboard icon on the bottom right corner so the user does not have to guess that clicking the field copies to clipboard. 

wallet-test

roshaans (changes requested):
Overall looks good. Just some changes required for better UX. Also, could you please specify what conditions need to be true in order for the wallet-migration flow to show up? Where is the entry point for the flow and can the user go back to this flow manually?

packages/frontend-near-wallet/src/utils/constants.ts
esaminu: Still not clear what you intend `NEAR_WALLET_ENV` to be set to when you deploy this. It is being used to compute the export url too

packages/frontend-near-wallet/src/modules/wallet-migration/components/MigrateAccounts.tsx
roshaans: Created a new issue to discuss and implement this #2887 @esaminu @gutsyphilip

lostpebble:
image

Hi guys,

Just looking at this PR today- is this the UI that is currently planned for the migration page for Near Wallet ?

I was under the impression that the migration was going to be completely neutral as far as helping the user move their accounts to a new wallet goes. This UI as shown in this PR does not appear neutral to me at all- first of all "We've moved" is already telling users that MyNearWallet is the "new" official wallet, only that's its moved- this is not the case, right? MyNearWallet is a wholly new commercial entity? (that just borrows the code of the current Near Wallet implementation) I thought that was the point of creating MyNearWallet- to open to ecosystem up and move away for "centralization" of Near tooling.

Secondly, the buttons layout seems to imply that the main blue "action" button is going to use MyNearWallet for the migration, and the other "white-out" secondary button is for using _other_ wallets. This is also another clear bias and seems to go against the information we've been told.

We'd appreciate some clarification on this.

roshaans:
I believe we are not going to go with this approach anymore. Feel free to reopen if I'm wrong @gutsyphilip

roshaans:
We will be needing this for the new landing page project deployed on *.near.org after migration it seems.

PR #2785, opened 2022-06-28
feat: made Sentry lazy

followdarko:
This reverted commit 7b24611caa04513a30e9f1f7a5d688dbecc461f4 highlighted the problem that when main js bundle has an errror than sentry will not initialize and we don't see the problem. This PR moves Sentry to separate script that loads before the main bundle. This is the lightweight sentry lazy skeleton around main sentry SDK. For now this PR isn't completed because I don't have access to Senty web interface to get Setry_client_key. Someone who have access should pass the key into bundle. Then me or someone else could continue work and test this PR.  

Also you could see more details on https://docs.sentry.io/platforms/javascript/install/lazy-load-sentry/

### DIFF
--- a/packages/frontend/src/redux/actions/staking.js
+++ b/packages/frontend/src/redux/actions/staking.js
@@ -8,9 +8,10 @@ import {
     STAKING_GAS_BASE,
     FARMING_CLAIM_GAS,
     FARMING_CLAIM_YOCTO,
-    LOCKUP_ACCOUNT_ID_SUFFIX
+    LOCKUP_ACCOUNT_ID_SUFFIX,
+    FT_MINIMUM_STORAGE_BALANCE_LARGE
 } from '../../config';
-import { fungibleTokensService, FT_MINIMUM_STORAGE_BALANCE_LARGE } from '../../services/FungibleTokens';
+import { fungibleTokensService } from '../../services/FungibleTokens';
 import { listStakingPools } from '../../services/indexer';
 import StakingFarmContracts from '../../services/StakingFarmContracts';
 import { getLockupAccountId, getLockupMinBalanceForStorage } from '../../utils/account-with-lockup';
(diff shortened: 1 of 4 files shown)


packages/frontend/src/redux/actions/staking.js
```diff
@@ -8,9 +8,10 @@ import {
     STAKING_GAS_BASE,
     FARMING_CLAIM_GAS,
     FARMING_CLAIM_YOCTO,
-    LOCKUP_ACCOUNT_ID_SUFFIX
+    LOCKUP_ACCOUNT_ID_SUFFIX,
+    FT_MINIMUM_STORAGE_BALANCE_LARGE
 } from '../../config';
-import { fungibleTokensService, FT_MINIMUM_STORAGE_BALANCE_LARGE } from '../../services/FungibleTokens';
+import { fungibleTokensService } from '../../services/FungibleTokens';
```
followdarko, line 14: Random error in current master that prevents `yarn run commit` =/

MaximusHaximus:
1. It's worth noting that this now loads a version of Sentry that we may have never run any tests against (or even loaded in our UI previously) because it is loading from a CDN which will be swapped out without announcement; by using `@sentry/browser` directly, we ensure that the version of Sentry we are running does not change without us explicitly bumping its version (which is also desirable from a security standpoint, as it reduces our risk of pulling in a supply chain attack in a just-published release of the dependency).
2. How does the app behave if the script tag fails to load? e.g. I block the CDN that the lazy-loader tries to load Sentry from?

packages/frontend/src/utils/sentryLazy.js
```diff
@@ -0,0 +1,245 @@
...
+    });
+  }
+})(
+  window,
+  document,
+  'script',
+  'onerror',
+  'onunhandledrejection',
+  'Sentry',
+  window.__SENTRY_CLIENT_KEY__ || '', // TODO: need to pass Sentry client key
```
roshaans, line 224: I believe you should be able to get the Sentry_DSN similarly to the [following](https://github.com/near/near-wallet/blob/4582f61fd49d2337cc33d267b53594b733185098/packages/frontend/src/utils/sentry.js)

roshaans (changes requested):

followdarko:
I will return to this after Nearcon

PR #2788, opened 2022-06-28
feat: linting enhancements

andy-haynes:
This PR updates the CI configuration to run `yarn lint` before building, ensuring broken builds on branches with eslint rule violations. It also moves the local lint run to `prepush` from `precommit`, avoiding a 10+ second commit time and dropping the `--fix` to make it more difficult to push changes with eslint breaks.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -46,6 +46,7 @@ pipeline {
                     }
                     steps {
                         dir("$WORKSPACE/packages/frontend") {
+                            sh 'yarn lint'
                             sh 'yarn install --frozen-lockfile'
                         }
                     }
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -46,6 +46,7 @@ pipeline {
                     }
                     steps {
                         dir("$WORKSPACE/packages/frontend") {
+                            sh 'yarn lint'
                             sh 'yarn install --frozen-lockfile'
```
morgsmccauley, line 50:
```suggestion
                            sh 'yarn install --frozen-lockfile'
                            sh 'yarn lint'
```
Might need to swap these round? The Jenkins build is currently failing because it can't find `eslint`

PR #2790, opened 2022-06-28
feat: pull fiat value metadata according to owned tokens

judith-Near:
Screen Shot 2022-06-28 at 1 31 49 PM

**in slices/tokenFiatValues**
from getLikelyTokenContracts, we get contract names. from that, we get symbols using getCachedContractMetadataOrFetch. pass in symbols to dataloader.

**In the dataloader**
from symbols, we get the coingecko specific ids. fetch the coingecko coins. present result using symbols.

**in hooks/fungibleTokensIncludingNear**
fungibleTokenPrices is an existing hashtable of contractNames to fiatData. due to coingecko, it now also has the key value pair of symbols to fiatData. if the hashtable has the symbol, fill in the fiat metadata according to symbol. if not, fill it in according to the contract name.


packages/frontend/src/components/Routing.js
```diff
@@ -245,13 +242,15 @@ class Routing extends Component {
...
     componentDidUpdate(prevProps) {
-        const { activeLanguage, account } = this.props;
+        const { activeLanguage, fetchTokenFiatValues, account } = this.props;
 
         if (
             prevProps.account.accountId !== account.accountId &&
             account.accountId !== undefined
         ) {
             this.props.getTokenWhiteList(account.accountId);
+            fetchTokenFiatValues(account.accountId);
```
marcinbodnar, line 252: nit: can we pass it as an object?

marcinbodnar (commented):
I’m getting this error while fetching tokens fiat value
`Querying [object Object] failed: wasm execution failed with error: FunctionCallError(MethodResolveError(MethodNotFound)). { "block_hash": "BUX1ytqg9pA1jBpZgiNocoL2XH3YttwXPsGQq5QFGVoN", "block_height": 68850478, "error": "wasm execution failed with error: FunctionCallError(MethodResolveError(MethodNotFound))", "logs": [] }`
it's shown on `tokenFiatValue.status.error` redux state

Patrick1904 (approved):
LGTM

Issue #2047, opened 2021-08-18
Support Generic Signing of Data from User Chosen Accounts Without Login / Adding Key


mattlockyer:
## Problem
Currently, if you want a user to sign some data with any keys on their account they will need to first sign into your app. This flow can be a UX interruption and cost GAS, if all the app wants to know is which NEAR account you want to prove that you own.

#### How did we discover this problem?
Several partners have asked for this since... I want to say 6mo - 1yr.

#### Job Story(s)
- User is in app, has **not** signed in, app does not know if they even have a NEAR wallet
- App redirects to route in wallet e.g. `wallet.com/sign-message/[URI_ENCODED_MESSAGE]`
- Wallet shows the same route as signing for a TX e.g. **select account, click accept**
- User is redirected to app with **signature in query params**

#### What are we planning to do about it?
- Implement it

#### What are we _not_ planning to do about it?
- NA

## Measuring Success
- Apps can identify users without adding keys to their account (gas cost)

## Resources
- NA

stefanopepe:
Followup to this request:
- we already have similar functionality buried in the wallet helper codebase, but it's undocumented
- V0 is pinned for the next sprint (Monday cc @MaximusHaximus )
- it needs minimum UI review, essentially to specify the difference from an on-chain transaction (cc @corwinharrell and @Patrick1904 )

Some corrections to the card above:

## Scope

#### What are we planning to do about it?
- Release a V0 with an endpoint/route within the wallet and a confirmation panel
- In the future, this should become part of `near-api-js` and not the wallet only

#### What are we not planning to do about it (for V0)?
- Build an end-to-end developer-friendly feature with code examples, documentation and e2e tests (at least for V0)
- Submit a PR to `near-api-js`

starpause:
@stefanopepe is there any chance of getting the most fundamental documentation available before sprint end? Specifically Collab.Land is looking for this to allow for existing NEAR wallets to associate with their bot before the near.org/metabuidl (aug27).

stefanopepe:
@starpause what's the volume and the impact of not having it? Is it blocking the August 27 event? Is it more valuable than having `coin-op.near` back and working as before?

Unfortunately, we have some long-awaited OOO time for one of the team members (planned way before knowing the strict urgency of this issue) and the remaining resources can be exclusively allocated to coin-op or this one, at least this week.

Realistically, given also the interruptions suffered during the weekend, this feature may anyway go live in the sprint after this one, and the spec made available later next week.

[...]

PR #2801, opened 2022-07-04
feat: implement sign arbitrary message

xilosada:


chefsale:
cc: @MaximusHaximus @Patrick1904, this is a feature we need for Calimero pretty urgently. I know this week is not the best for you, but appreciate if somebody could take a look to land this asap, as it is blocking us on shipping authentication with Private Shards.

agileurbanite:
I agree with @PinkiNice here but let's get some more folks to review.  Adding @MaximusHaximus and @Patrick1904

agileurbanite:
Also is there an NEP i can follow to understand what the intent is better?

PinkiNice:
@agileurbanite https://github.com/near/near-wallet/issues/2047 this issue describes what we are trying to achieve.

My proposal is to instead of signing provided data, wrap this data into something like
{ 
  message: “Original Message”,
  blockId: “Most recent block hash”,
  accountId: “User account used for signature”,
  publicKey: “Public key used for signature”,
}
And give a signature to this object: this way it cannot be interpreted as a transaction which makes it safe to sign arbitrary data. This approach is cheap to implement and won't require extra transactions for creating FunctionalAK

packages/frontend/src/routes/SignMessageWrapper.js
```diff
@@ -0,0 +1,99 @@
...
+    selectSignedRequest,
+    SIGN_MESSAGE_STATUS,
+} from '../redux/slices/signMessage';
+import { isUrlNotJavascriptProtocol } from '../utils/helper-api';
+
+export function SignMessageWrapper() {
+    const dispatch = useDispatch();
+
+    const signCallbackUrl = useSelector(selectAccountUrlCallbackUrl);
+    const meta = useSelector(selectAccountUrlMeta);
```
Patrick1904, line 25: nit: While not enforced yet, we try to define selectors by simply removing the `select` part. E.g.:

`const accountUrlMeta = useSelector(selectAccountUrlMeta);`

packages/frontend/src/redux/slices/signMessage/index.js
```diff
@@ -0,0 +1,81 @@
...
+    [selectSignMessageSlice],
+    (signMessage) => signMessage.status
+);
+
+export const selectSignedRequest = createSelector(
+    [selectSignMessageSlice],
+    (signMessage) => signMessage.signedRequest || {}
+);
+
+export function addQueryParams(baseUrl, queryParams) {
```
Patrick1904, line 28: Can we please re-use/re-org this to so that we don't have do duplicate this code from the TX sign slice?

https://github.com/near/near-wallet/blob/master/packages/frontend/src/redux/slices/sign/index.js#L82

packages/frontend/src/components/sign-message/SignMessage.js
```diff
@@ -0,0 +1,70 @@
...
+                margin: 15px 0;
+                word-break: break-all;
+            }
+        }
+    }
+`;
+
+export default ({ sender, message }) => {
+    return (
+        <StyledContainer className="brs-8 bsw-l">
```
Patrick1904, line 54: nit: can we please use single quotes in jsx

packages/frontend/src/components/sign-message/SignMessageInvalid.js
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    && .text {
+        color: #72727a;
+        margin-top: 24px;
+    }
+`;
+
+const SignMessageInvalid = () => (
+    <CustomContainer className="small-centered">
```
Patrick1904, line 31: nit: can we please use single quotes in jsx

xilosada:
## Summary

Changes since the last PR:
* Renamed url, components, actions, reducers and slices to `verify-owner`. This was the name was proposed here and is more accurate as this code doesn't sign arbitrary messages.
* Modify signature method to include accountId, blockId, public key and key type to the signature.
* Apply provided design
* Apply feedback
* Wrap `isUrlNotJavascriptProtocol` in a try-catch block. It threw an exception getting the protocol field from an invalid URL.

This PR adds the verify-owner page, which allows sites to request a proof of ownership of the private key of a Near account. For that the site sends a `message` that will be signed along other parameters like. The initial PR allowed to sign arbitrary messages and it could be used by a malicious actor to sign transactions. This new approach uses @agileurbanite suggestions and combines the message with the following fileds:
`
  accountId: user selected near account id,
  message: message generated by the requesting site,
  blockId: hash of the latest block,
  publicKey,
  keyType
`

Page status:
NEEDS_CONFIRMATION(verify-owner-needs-confirmation): Initial status. The user can chose to `confirm` or `reject` the authorization requested. If there is any issue during the signature process a dialog with details of the error is shown and the page will go back to this status.

IN_PROGRESS(verify-owner-in-progress): the verification process is in progress. `Confirm` and `Cancel` buttons are disabled.

COMPLETED(verify-owner-completed): the authentication request has been rejected or completed. The user will be redirected to the `callbackUrl` if it is valid. Otherwise to `/` 


## Test plan
Clone the test page from https://github.com/calimero-is-near/calimero-examples/tree/master/near-sign-message and serve it locally  `python -m SimpleHTTPServer`.

Build and run Near Wallet `yarn && NEAR_WALLET_ENV=testnet yarn start`

Scenarios:
#### Authentication request confirmed:
* Navigate to localhost:8000 and click on the first link.
* Click on `approve`
* Verify that the user is redirected to callback url with the following query arguments:
`
accountId=calitest.testnet
&message=dassdasdasdasadasdas+34+87
&blockId=4TZUNbCGKREqeBAYumtGPBrfyoy36Z98QmSHtYHjuriV
&publicKey=ZHB5f2WZB3Hc08w%2FwfhWsqiSrRL%2BMmQTyx9E%2B9za8r0%3D
&signature=%2FW5drEvUCriBz2XUx6xLkhOmeXazkOckJqArylNdnck6KLMfcotj6fhuoYULx38pJgkGpppPhxCHFLhjvM87Dg%3D%3D
&meta=test
`

https://user-images.githubusercontent.com/2929173/178935822-067f945c-d9dd-4527-b3d0-32ca302e6a0d.mov


#### Authentication request rejected:
* Navigate to localhost:8000 and click on the first link.
* Click on `cancel`
* Verify that the user is redirected to callback url with the following query arguments:
`
meta=test
&error=User+rejected
`

https://user-images.githubusercontent.com/2929173/178936209-35f82c57-7637-4b5b-b0c4-e747508ba583.mov


#### Invalid request:
* Navigate to localhost:8000 and click on the first link.
* Click on approve
* Verify that the invalid request page is shown.

https://user-images.githubusercontent.com/2929173/178936022-ead2db18-35ec-405a-91f8-c97c8ceae57c.mov


#### Auth error:
* Throw an error in `handleAuthorizationRequestConfirmed` packages/frontend/src/redux/slices/verifyOwner/index.js line 40
* Navigate to localhost:8000 and click on the first link.
* Click on approve
* Verify that the error dialog is shown and the wallet returns to the confirmation state.

Screen Shot 2022-07-14 at 9 06 57 AM

packages/frontend/src/components/verify-owner/VerifyOwner.js
```diff
@@ -0,0 +1,109 @@
...
+        }
+    }
+`;
+
+export default ({ accountId, appName }) => {
+    return (
+        <StyledContainer>
+            <h3><Translate id='verifyOwner.summary.title' /></h3>
+            <div className='logos'>
+                <CalimeroRoundedLogo />
```
Patrick1904, line 92: I suspect it won't always be Calimero requesting signature here? Might want to opt for something more generic like what's shown on the `/login` confirm screen (`SwapGraphic.js`):

Screen Shot 2022-07-14 at 10 29 01 AM

packages/frontend/src/components/verify-owner/VerifyOwner.js
```diff
@@ -0,0 +1,109 @@
...
+                <div className='swap'><SwapIcon color='#D5D4D8' stroke='1px' /></div>
+                <NearRoundedLogo/>
+            </div>
+            <h2><Translate id='verifyOwner.summary.connetingWith' data={{accountId}}/></h2>
+            <div className={'account signer'}>
```
Patrick1904, line 97: nit: can we use regular strings instead of curly's around the strings?

packages/frontend/src/components/verify-owner/VerifyOwner.js
chefsale: No, we actually want to be clear it is Calimero in this case.

PinkiNice: @xilosada @chefsale We should stick with a placeholder for now, so using SwapGraphic is the way to go until we come up with a solution to provide customizable icons there.

PR #2807, opened 2022-07-12
KWLT-19 Remove convenient 2FA methods from view

loosefer:


andy-haynes (changes requested):
Looking great, just a few minor things otherwise ready to go 🚀

andy-haynes (approved):
🚀

PR #2814, opened 2022-07-14
feature - web3 app auth

PinkiNice:
See https://github.com/near/near-wallet/pull/2801

TL;DR
PR enables web3-auth via signature verification for invalid payloads only (does not allow to make a replay attack);

This functionality is blocked under feature-flag wall to allow signing only in testnet environments.
It will stay in testnet environments until thoroughly tested and probably will be improved and remade based on received feedback.


followdarko:
@PinkiNice take a look https://github.com/near/near-wallet/pull/2821

chefsale:
@PinkiNice bump, can we land this on testnet?

packages/frontend/src/routes/VerifyOwnerWrapper.js
```diff
@@ -0,0 +1,75 @@
...
+        if (verifyOwnerStatus === VERIFY_OWNER_STATUS.COMPLETED) {
+            if (accountUrlCallbackUrl && isValidCallbackUrl) {
+                window.location.href = buildRedirectUrl(accountUrlCallbackUrl, accountUrlMeta, signedRequest, verifyOwnerError);
+            } else {
+                dispatch(redirectTo('/'));
+            }
+        }
+    }, [verifyOwnerStatus]);
+
+    const handleApproveRequest = async () =>
```
andy-haynes, line 54: No need for `async` on these

packages/frontend/src/routes/VerifyOwnerWrapper.js
```diff
@@ -0,0 +1,75 @@
...
+
+    useEffect(() => {
+        if (verifyOwnerStatus === VERIFY_OWNER_STATUS.COMPLETED) {
+            if (accountUrlCallbackUrl && isValidCallbackUrl) {
+                window.location.href = buildRedirectUrl(accountUrlCallbackUrl, accountUrlMeta, signedRequest, verifyOwnerError);
+            } else {
+                dispatch(redirectTo('/'));
+            }
+        }
+    }, [verifyOwnerStatus]);
```
andy-haynes, line 52: Should the dependencies array include values from the other selectors?

packages/frontend/src/redux/slices/verifyOwner/index.js
```diff
@@ -0,0 +1,82 @@
...
+                keyType: publicKey.keyType
+            };
+
+            const encoded = JSON.stringify(data);
+            const signed = await wallet.signMessage(encoded, accountId);
+
+            return {
+                ...data,
+                signature: Buffer.from(signed.signed.signature).toString('base64'),
+                keyType: signed.signed.publicKey.keyType
```
andy-haynes, line 57: Will this always be the same as `data.keyType`? If so it can be removed here since it will always overwrite the value spread from `data`

andy-haynes (approved):
Minor comments inline but otherwise LGTM - thanks for feature flagging

PR #2815, opened 2022-07-15
Even if FT contract does not implement recommended method, do not error out

judith-Near:


packages/frontend/src/redux/slices/tokenFiatValues/index.js
```diff
@@ -32,6 +35,10 @@ const fetchTokenFiatValues = createAsyncThunk(
                 try {
                     const metadata = await getCachedContractMetadataOrFetch(contractName, getState());
                     symbol = metadata.symbol;
+                } catch (error) {
+                    if (!error.message.includes(METHOD_NOT_FOUND_ERROR)) {
+                        console.log(`Unknown error fetching fiat value for ${contractName}`, error.message);
+                    }
```
andy-haynes, line 41: We should continue to reject the promise here so that we're not suppressing errors when the known error case occurs. Just a `throw error;` after the message check is good

andy-haynes (changes requested):
Per @esaminu's suggestion let's also please use `Promise.allSettled` in favor of `Promise.all` when iterating over the contract fetches

andy-haynes (approved):

PR #2828, opened 2022-07-20
feat: use ref fallback for near

esaminu:


esaminu:
suggestion integrated into #2830

PR #2842, opened 2022-07-22
fix(ci): don't run PR actions on PRs targeting stable

andy-haynes:
This PR updates the PR actions to not deploy preview bundles for release PRs targeting `stable`. The PR should still test and bundle the code but not deploy to any preview environments.

### DIFF
--- a/.github/workflows/bundle-and-deploy.yml
+++ b/.github/workflows/bundle-and-deploy.yml
@@ -3,6 +3,10 @@ name: Build and Deploy Frontend
 on:
   workflow_call:
     inputs:
+      deploy_preview:
+        default: true
+        required: false
+        type: boolean
       environment:
         required: true
         type: string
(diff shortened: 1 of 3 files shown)


roshaans:
@andy-haynes Is this ready to be reviewed? Also, @Patrick1904 and I were discussing about having PRs from forks being able to also have a preview environment deployment.  Manual activation of deployment previews on a PR by a Near-wallet org member would be great if that is possible. 

For clarity, see this [PR](https://github.com/near/near-wallet/pull/2844) that fails to build previews for test environment.

andy-haynes:
> @andy-haynes Is this ready to be reviewed? Also, @Patrick1904 and I were discussing about having PRs from forks being able to also have a preview environment deployment. Manual activation of deployment previews on a PR by a Near-wallet org member would be great if that is possible.
> 
> For clarity, see this [PR](https://github.com/near/near-wallet/pull/2844) that fails to build previews for test environment.

IIRC we were looking at a simpler solution as this ended up being too much of a kludge for use in CI. The goal behind this was only to eliminate an unnecessary step in deployment PRs, in practice it's just awkward for these specific PRs (that only we create) until we're doing trunk-based development.

The S3 preview environment was/is a proof of concept that still needs lots of work to be production ready, including DNS configuration (likely requires CloudFront) and securing S3 uploads for non-contributors. This is why we were going to retain Netlify/Render for the time being, so that previews could continue to be supported until this was worked out.

PR #2844, opened 2022-07-24
feat: MYNW-88 Cache for Indexer requests

loosefer:

- cache indexer response every 30 sec
- accumulate indexer response to IndexedDb cache and concatenate it with prevously


packages/frontend/src/services/indexer.js
```diff
@@ -1,17 +1,33 @@
 import { INDEXER_SERVICE_URL } from '../config';
 import sendJson from '../tmp_fetch_send_json';
+import { buildUrlOptParam } from '../utils/url';
+import IndexerCache from './lib/indexerCache';
+
+const cache = new IndexerCache();
+cache.open();
```
PinkiNice, line 7: Is there any need to check if cache is opened before making requests?
Is it possible that some request will throw if called immediately after cache.open ?

packages/frontend/src/services/indexer.js
```diff
@@ -1,17 +1,33 @@
...
+
+function cachedRequest (accountId, kind) {
+    const url = `${INDEXER_SERVICE_URL}/account/${accountId}/${kind}`;
```
PinkiNice, line 10: I think its better to distinct logic of building access URL to access API and Cache ids (kind)

packages/frontend/src/services/indexer.js
```diff
@@ -1,17 +1,33 @@
...
+}
 
 export function listAccountsByPublicKey(publicKey) {
     return fetch(`${INDEXER_SERVICE_URL}/publicKey/${publicKey}/accounts`)
         .then((res) => res.json());
 }
 
 export function listLikelyNfts(accountId) {
-    return sendJson('GET', `${INDEXER_SERVICE_URL}/account/${accountId}/likelyNFTs`);
+    return cachedRequest(accountId, IndexerCache.LIKELY_NFT_KEY);
```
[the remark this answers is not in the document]
PinkiNice: And then original functions are incapsulated inside services/indexer.js
And all exports are coming from another file called indexerWithCache.js

packages/frontend/src/services/IndexerCache.js
```diff
@@ -0,0 +1,144 @@
...
+                if (Boolean(record)) {
+                    await this._updateRecord(accountId, kind, updated);
+                } else {
+                    await this._addRecord(accountId, kind, updated);
+                }
+
+                return updated.list;
+            }
+        } catch (e) {
+            console.error(e);
```
PinkiNice, line 134: Is it that in case of a failure there is no error handling?
What are possible outcomes that can lead there?
Lets add comments to describe scenarios which could throw an error

packages/frontend/src/services/IndexerCache.js
```diff
@@ -0,0 +1,144 @@
+import Cache from '../utils/cache';
+
+export class IndexerCache extends Cache {
```
PinkiNice, line 3: Lets rename this file to indexer-cache to avoid mixing CamelCase and kebab-case.

packages/frontend/src/utils/cache.js
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    async getIndexStore(access = 'readwrite') {
+        const store = await this.getObjectStore();
+
+        return store.index(this.indexName);
+    }
+
+    open() {
+        const open = indexedDB.open(Cache.CACHE_DB_NAME, this.dbVersion);
```
followdarko, line 47: `indexedDB` is propherty of `window`. Please, don't skip it and add -> `window.indexedDB.open(...`

Since it's `window` prop, anybody can overwrite it! So it's better to add check like:
```
if ('indexedDB' in window) {
  try {
     const request = window.indexedDB.open(DB_NAME, VERSION)
     etc...
  } catch (e) {
  }
}
```

followdarko: btw, i saw real situations from sentry logs on production, when indexeddb.open doesnt open an falls with error

packages/frontend/src/services/IndexerCache.js
```diff
@@ -0,0 +1,144 @@
...
+        try {
+            const lastTimestamp = record?.data?.timestamp;
+
+            if (this._shouldUpdate(lastTimestamp)) {
+                const response = await updater(lastTimestamp);
+                const prev = record?.data?.list || [];
+
+                const onlyUniqValues = new Set(response.concat(prev));
+                const updated = {
+                    timestamp: new Date().getTime(),
```
esaminu, line 121: This should be returned by the endpoint to indicate which is the last fetched block imo. I'll add a key to the response payload called `lastBlockTimestamp` wdyt

packages/frontend/src/services/IndexerCache.js
```diff
@@ -0,0 +1,144 @@
...
+     * The main idea is save the contract-helper from searching through the entire history of the blockchain.
+     * Each next request, we send the last timestamp, while accumulating data on the client.
+     */
+    async accumulate(accountId, kind, updater) {
+        const record = await this._getRecord(accountId, kind);
+        try {
+            const lastTimestamp = record?.data?.timestamp;
+
+            if (this._shouldUpdate(lastTimestamp)) {
+                const response = await updater(lastTimestamp);
```
esaminu, line 116: Another thing we should consider here is a version param from the endpoint in the unlikely case that some conditions in the query for an FT are added or removed. The endpoint should return a `version` key and if the version is updated on the helper we should rescan from the beginning of the blockchain

packages/frontend/src/utils/url.js
```diff
@@ -0,0 +1,16 @@
+export const buildUrlOptParam = (url, params = {}) => {
```
esaminu, line 1: We usually use [`stringify` from the query-string library](https://github.com/sindresorhus/query-string#stringifyobject-options) to build these kinds of request payloads. This may be useful here?

esaminu (commented):
Great solution! left some comments regarding unlikely but possible re-scan conditions due to the queries changing and also a comment about where we get the last fetched block 😃

packages/frontend/src/services/indexer/cache.js
```diff
@@ -0,0 +1,146 @@
...
+                        record.timestamp = 0;
+                    }
+
+                    await this._updateRecord(accountId, kind, updated);
+                } else {
+                    await this._addRecord(accountId, kind, updated);
+                }
+
+                return updated.list;
+            }
```
esaminu, line 133: We should return the latest record if we did not update

esaminu (commented):
Looking good. Left some minor comments 👍

packages/frontend/src/services/indexer/cache.js
loosefer: Yes, last record returns on line 139 in this case

packages/frontend/src/services/indexer/cache.js
```diff
@@ -0,0 +1,146 @@
...
+                    list: Array.from(onlyUniqValues),
+                    version,
+                };
+
+                if (Boolean(record)) {
+                    // If the version is updated on the helper
+                    // we should rescan from the beginning of the blockchain
+                    const isVersionChanged = version !== record.version;
+                    if (isVersionChanged) {
+                        record.timestamp = 0;
```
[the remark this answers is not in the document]
loosefer: fixed by shouldRestart flag and recall function recursively

packages/frontend/src/services/indexer/api.js
```diff
@@ -0,0 +1,94 @@
...
+            .then((r) => r.json());
+    },
+    listStakingPools: () => {
+        return fetch(`${INDEXER_SERVICE_URL}/stakingPools`, {
+            headers: {
+                ...CUSTOM_REQUEST_HEADERS,
+            }})
+            .then((r) => r.json());
+    },
+};
```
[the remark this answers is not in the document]
loosefer: This was a mistake after master merging =( Fixed

packages/frontend/src/services/indexer/cached-api.js
```diff
@@ -0,0 +1,26 @@
...
+            updater: (timestamp) => api.listLikelyNfts(accountId, timestamp),
+            timeout: UPDATE_REQUEST_INTERVAL,
+        });
+    },
+    listLikelyTokens(accountId, timestamp) {
+        return Cache.accumulate({
+            accountId,
+            kind: LIKELY_TOKENS_KEY,
+            updater: (timestamp) => api.listLikelyTokens(accountId, timestamp),
+            timeout: UPDATE_REQUEST_INTERVAL,
```
roshaans, line 23: Is the intended behavior to call `listLikelyTokens` endpoint every 30 seconds?

I was looking over network requests while running the app locally and I did not really see that happen. Messed around with `UPDATE_REQUEST_INTERVAL` and it still did not work. Am I missing something?

roshaans (commented):
Really neat addition 💯 

Just a bit confused about the refresh timer. 

 would be nice in the future to also cache calls in `/frontend/src/services/NonFungibleTokens.js` and `/frontend/src/services/FungibleTokens.js`  because every time the user changes tabs and lands back into `collectibles`, it makes consecutive RPC calls to load metadata, supply, image, e.t.c. If the user has tons of nft from different collections, like I do on my test account every time I switch tabs I see upwards of 60-70 RPC calls fire away.

packages/frontend/src/services/indexer/cached-api.js
loosefer: It is a bug, thank you.

loosefer: Current time should be converted to nanoseconds before comparison, fixed.
lastTime from backend should be parsed as integer.

Fixed.

roshaans (changes requested):
It seems to be making way too many calls now. Is this also happening for you?cache_bug

PR #2847, opened 2022-07-26
feat: remove initial MNW migration prompt

esaminu:
This PR removes initial MNW migration modal and immediately prompts the user with the select a wallet to export accounts modal


agileurbanite (approved):
LGTM! 🚢 !

PR #2840, opened 2022-07-22
[WIP]: added HAPI risc scoring system

followdarko:
Discussion starts here https://github.com/near/near-wallet/pull/2763

You can insert the address `ref-test.rkonoval.testnet` for testing purposes.

How it looks like:
Screenshot 2022-07-22 at 20 21 11
Screenshot 2022-07-22 at 20 21 19

PR #2849, opened 2022-07-26
Mynw 108 fix risk scoring bug

### LINKED ISSUE #2848: Cannot transfer NEAR on testnet. Continue button remains disabled after entering valid recipient
**Problem**
`Continue` button is not enabled after entering valid recipient address when trying to transfer NEAR on testnet wallet

Reproduced on Brave and Firefox Developer Edition

**Expected Behavior**
Button should be enabled when `Send to` field is validated

**Steps to reproduce**

1. Open testnet wallet
2. Click `Send` button
3. Enter 1 NEAR as amount
4. Click `Continue`
5. Enter valid testnet address in `Send to` field
6. `Continue` button remains disabled

PinkiNice:
This PR:
* Fixes bug with disabled button on `Send Screen` on validated account
* Makes `Send Screen` waits on risk-score result before allowing to proceed (reduces flickering)
* Adds state reset for risk-score hook between checks for different accounts
* Fixes bug with checkbox losing state between checks

Closes #2848

### DIFF
--- a/packages/frontend/src/components/send/components/InputAccountId.js
+++ b/packages/frontend/src/components/send/components/InputAccountId.js
@@ -128,7 +128,7 @@ class InputAccountId extends Component {
     }
 
     handleChangeAccountId = ({ userValue, el }) => {
-        const { handleChange, localAlert, clearLocalAlert } = this.props;
+        const { handleChange, localAlert, clearLocalAlert, setIsImplicitAccount } = this.props;
         const { wrongChar } = this.state;
         const pattern = /[^a-zA-Z0-9._-]/;
 
@@ -147,6 +147,7 @@ class InputAccountId extends Component {
             this.setState({ wrongChar: false });
         }
 
+        setIsImplicitAccount(false);
         handleChange(accountId);
 
         localAlert && clearLocalAlert();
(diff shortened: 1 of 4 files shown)


Patrick1904 (approved):
LGTM. Thanks for the quick fix @PinkiNice!

Patrick1904 (changes requested):
For some reason, it seems implicit accounts are still stuck at this step? Can we fix that?

followdarko:
@PinkiNice take a look at this draft https://github.com/near/near-wallet/pull/2852, if everything ok please merge it here

followdarko:
@Patrick1904 this UI/UX with implicit accounts  is not intuitive clear. Probably it is better to still mark it green or at least show some notification to user?

followdarko:
We could also https://github.com/near/near-wallet/pull/2853 merge it here

agileurbanite:
Where are we with this PR @PinkiNice ? it looks like there's a few conflicts and not sure if #2853 is still necessary given #2852 was merged.  Am I right to assume this @followdarko ? @Patrick1904 as FYI

Patrick1904:
@followdarko can we please add these changes into a PR for https://github.com/near/near-wallet/pull/2840 so that we can test it all together before we merge?

PR #2851, opened 2022-07-26
feat: Enable `RENDER_MIGRATION` in `mainnet` environments

morgsmccauley:
Enables `RENDER_MIGRATION` in both `mainnet` and `mainnet_NEARORG` making wallet.near.org use the contract-helper service in AWS rather than Render.


andy-haynes (dismissed):

hcho112 (approved):
LGTM

PR #2861, opened 2022-08-01
chore: Use react-gtm-module vs manual script tag

Patrick1904:
Changes:
1. Use `react-gtm-module` for Google Tag Manager to enable environment specific IDs
2. Add some initial tracking IDs for recovery buttons for testing in GTM

### DIFF
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -47,6 +47,7 @@
     "react-dom": "^16.13.1",
     "react-google-recaptcha": "^2.1.0",
     "react-google-recaptcha-v3-near": "^2.0.0",
+    "react-gtm-module": "^2.0.11",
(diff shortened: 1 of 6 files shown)


followdarko:
@Patrick1904 why we need to add additional dependency that solves nothing and already forgotten (last update 2 years ago, lot of issues and prs) ?

Patrick1904:
> @Patrick1904 why we need to add additional dependency that already forgotten (last update 2 years ago, lot of issues and prs) and does the same as current code?

The main reason for using the module is so that we can use different IDs between branding/environments, etc. This will be beneficial if MNW wants to start tracking with GTM too. `react-gtm-module` has 300k+ downloads per week.

andy-haynes (approved):
LGTM :shipit:

PR #2862, opened 2022-08-01
chore: Disable Mixpanel tracking

Patrick1904:


packages/frontend/src/mixpanel/index.js
```diff
@@ -39,7 +39,9 @@ let Mixpanel = {
     register: () => {}
 };
 
-if (BROWSER_MIXPANEL_TOKEN) {
+const shouldEnableTracking = false; // BROWSER_MIXPANEL_TOKEN
+
+if (shouldEnableTracking) {
     mixpanel.init(BROWSER_MIXPANEL_TOKEN);
```
agileurbanite, line 45: What happens here? Do we send data to MixPanel?

Patrick1904: Yep we send data to MixPanel if there's a token coming from the config

followdarko:
@Patrick1904 just one question — why?

packages/frontend/src/mixpanel/index.js
PinkiNice: @Patrick1904 Are your trying to remove mixpanel tracking for wallet.near.org domains?

andy-haynes (approved):

PR #2863, opened 2022-08-02
fix: MYNW-114 Temporary hide bridge section from topup-page

followdarko:
MYNW team decided to hide bridge section from Topup page because the only one option here is Rainbow Bridge. Right now MYNW users can't use Rainbow Bridge because of lack of integration. Work in progress on this issue, but for now best option is not to confuse users and hide section.

### DIFF
--- a/packages/frontend/src/components/buy/BuyNear.js
+++ b/packages/frontend/src/components/buy/BuyNear.js
@@ -4,6 +4,7 @@ import { useSelector } from 'react-redux';
 import styled from 'styled-components';
 
 import { getPayMethods } from '../../config/buyNearConfig';
+import { isWhitelabel } from '../../config/whitelabel';
 import { Mixpanel } from '../../mixpanel';
 import { selectAccountId } from '../../redux/slices/account';
 import { isMoonpayAvailable, getSignedUrl } from '../../utils/moonpay';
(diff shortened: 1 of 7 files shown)


packages/frontend/src/components/buy/BuyNear.js
```diff
@@ -225,11 +226,13 @@ export function BuyNear({ match, location, history }) {
                     subTitle='buyNear.nearPurchaseSubTitle'
                     actions={[PayMethods.nearPay, PayMethods.moonPay, PayMethods.utorg, PayMethods.ftx]}
                 />
-                <FundingCard
-                    title='buyNear.bridgeTokens'
-                    subTitle='buyNear.bridgeSubTitle'
-                    actions={[PayMethods.rainbow]}
-                />
+                {!isWhitelabel && (
```
Patrick1904, line 229:
```suggestion
                {!isWhitelabel() && (
```

Patrick1904 (changes requested):
LGTM - just make sure to run the function

packages/frontend/src/components/buy/BuyNear.js
Patrick1904: Not a blocker: may be worth to make `isWhiteLabel` a selector vs running this manually every time

followdarko: Oh! Thank you! Always forgot that this is a function. Actually naming like boolean variable really confusing. If we go deeper then why does it's even a function while it behaves like a real boolean flag!? No need to make any selector I refactor code to use it like a variable.

followdarko:
@Patrick1904 Sorry, I'm refactored a little bit. Hope you don't mind

Patrick1904:
> @Patrick1904 Sorry, I'm refactored a little bit. Hope you don't mind

Thanks @followdarko. Can you please just rebase this PR and make sure all usage of `isWhiteLabel` is correct after this change (some usage in latest [release](https://github.com/near/near-wallet/pull/2869)), and we'll this merged ASAP.

followdarko:
Outdated since Rainbow Bridge integrate support for MyNearWallet.

PR #2866, opened 2022-08-04
Display private key with qrcode

AmazingBeerBelly:
Display private key with qrcode to convenient the mobile wallet app to scan to import this near account.

image

### DIFF
--- a/packages/frontend/src/components/profile/export_private_key/ExportKeyModal.js
+++ b/packages/frontend/src/components/profile/export_private_key/ExportKeyModal.js
@@ -2,6 +2,7 @@ import React, { useState } from 'react';
 import { Translate } from 'react-localize-redux';
 import { useSelector } from 'react-redux';
 import styled from 'styled-components';
+import QRCode from 'qrcode.react';
 
 import CheckMarkNoBorderIcon from '../../../images/icon-check-no-border.svg';
 import { selectAccountId } from '../../../redux/slices/account';
@@ -133,6 +134,7 @@ export default ({
 };
 
 const ExportKeyModalBody = ({ currentView, setCurrentView, onClose, secretKey }) => {
+
     switch (currentView) {
         case VIEWS.ACCOUNT_ID_CONFIRMATION: 
             return (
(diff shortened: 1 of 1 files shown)


roshaans:
Thank you for the contribution @AmazingBeerBelly. I can definitely see this being helpful for mobile wallet apps import flows. It seems that on we have the functionality to export the account's full access key via a QR code, but it is **only on the testnet env.**

image

Before merging this in, i'd like to get some clarity on why this is not on mainnet yet. Would you happen to know @Patrick1904

Patrick1904:
> Before merging this in, i'd like to get some clarity on why this is not on mainnet yet. Would you happen to know @Patrick1904

@roshaans the 'Scan QR code to import' feature is only available on testnet for security reasons since the QR code contains a FAK. Currently there are no plans to introduce this feature on mainnet.

@AmazingBeerBelly is there a use case where displaying the private key on the UI is better than scanning the QR code? The QR code uses the 'auto import' functionality in the wallet, so there is no additional understanding of how to import an account needed. However, just displaying the private key requires a deeper understanding of how the wallet uses keys in order to work properly, and I'm afraid that users may find this confusing.

roshaans:
> > Before merging this in, i'd like to get some clarity on why this is not on mainnet yet. Would you happen to know @Patrick1904
> 
> @roshaans the 'Scan QR code to import' feature is only available on testnet for security reasons since the QR code contains a FAK. Currently there are no plans to introduce this feature on mainnet.
> 
> @AmazingBeerBelly is there a use case where displaying the private key on the UI is better than scanning the QR code? The QR code uses the 'auto import' functionality in the wallet, so there is no additional understanding of how to import an account needed. However, just displaying the private key requires a deeper understanding of how the wallet uses keys in order to work properly, and I'm afraid that users may find this confusing.

Agreed. I do not think think this feature will benefit users as it currently stands. The QR code flow will need to be worked out very carefully with security and usability in mind. We can revisit this at a later point and if this is a feature that you think should exist, we can bring it up to Product and see how it can be integrated.

PR #2875, opened 2022-08-09
fix: update migration banner copy

agileurbanite:
fix: update migration banner copy


esaminu (approved):
LGTM, the url needs to updated in [MigrationBanner](https://github.com/near/near-wallet/blob/master/packages/frontend/src/components/common/MigrationBanner.js#L132) when passed in or just [hardcoded](https://github.com/near/near-wallet/pull/2875/files#diff-5a21fd9991c7631bf26ef8b958f160e7b68989f82c1235fc8bebad14ad196407R773)

PR #2879, opened 2022-08-11
Fix ledger issue

marcinbodnar:


packages/frontend/src/components/accounts/ledger/SetupLedger.js
```diff
@@ -51,6 +50,10 @@ const SetupLedger = (props) => {
     const fundingOptions = parseFundingOptions(props.location.search);
     const shouldRenderRecaptcha = !fundingOptions && RECAPTCHA_CHALLENGE_API_KEY && isNewAccount && !ENABLE_IDENTITY_VERIFIED_ACCOUNT;
 
+    const has2fa = useSelector(selectAccountHas2fa);
+    // disable the Continue button if a user has 2fa enabled, or we don't know yet if it's disabled/enabled
+    const disable2fa = has2fa || typeof has2fa === 'undefined';
```
Patrick1904, line 55: nit: can we please name the selector `accountHas2fa` and change `disable2fa` --->  `accountMightHave2fa`

marcinbodnar: done

Patrick1904 (approved):

andy-haynes (changes requested):
I'm not able to create a new account using ledger, the `Continue` button stays disabled even when ledger is connected

image

packages/frontend/src/components/accounts/ledger/SetupLedger.js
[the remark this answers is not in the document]
marcinbodnar: > The only reason to use typeof is if the variable might not be defined, but here that's not a concern.

It is, we cannot allow enabling ledger until we are sure that 2fa is not enabled - 2fa info comes with a delay.

Extended the condition so it will work properly on testnet as well.

Patrick1904 (changes requested):
I'm not able to create a new account with Ledger if I have 2FA enabled on the currently active account:

Screen Shot 2022-08-12 at 10 49 59 AM

marcinbodnar:
@Patrick1904 fixed.

packages/frontend/src/components/profile/Profile.js
```diff
@@ -164,6 +165,13 @@ export function Profile({ match }) {
     const userRecoveryMethods = useSelector((state) => selectRecoveryMethodsByAccountId(state, { accountId: account.accountId }));
     const twoFactor = has2fa && userRecoveryMethods && userRecoveryMethods.filter((m) => m.kind.includes('2fa'))[0];
 
+    const accountState = useSelector(selectAccountSlice);
+    const keys = accountState.fullAccessKeys || [];
+    const publicKeys = keys.map((key) => key.public_key);
+    const hasLedger = userRecoveryMethods.filter((method) => method.kind === 'ledger').map((key) => key.publicKey).some((key) => publicKeys.includes(key));
```
Patrick1904, line 171: Haven't we already filtered out recovery methods that do not have a key on chain in [`getRecoveryMethods`](https://github.com/near/near-wallet/blob/master/packages/frontend/src/utils/wallet.js#L912)? Seems any selector should only have recovery methods with keys confirmed on chain.

packages/frontend/src/utils/twoFactor.js
```diff
@@ -42,6 +43,27 @@ export class TwoFactor extends Account2FA {
...
+            throw new WalletError('Ledger Hardware Wallet is enabled', 'initTwoFactor.ledgerEnabled');
+        }
+
+        // additional check if the ledger is not connected but exists as a recovery method
+        const userRecoveryMethods = await this.wallet.getRecoveryMethods();
+        console.log('userRecoveryMethods', userRecoveryMethods);
+        const accountState = await this.wallet.loadAccount();
+        const keys = accountState.fullAccessKeys || [];
+        const publicKeys = keys.map((key) => key.public_key);
+        const hasLedger = userRecoveryMethods.filter((method) => method.kind === 'ledger').map((key) => key.publicKey).some((key) => publicKeys.includes(key));
```
Patrick1904, line 60: Same here: https://github.com/near/near-wallet/pull/2879/files#r944876270

packages/frontend/src/utils/twoFactor.js
```diff
@@ -42,6 +43,27 @@ export class TwoFactor extends Account2FA {
...
+        const accessKeys = await this.wallet.getAccessKeys();
+        const ledgerKey = accessKeys.find((key) => key.meta.type === 'ledger');
+        // throw error if ledger is enabled
+        if (ledgerKey) {
+            throw new WalletError('Ledger Hardware Wallet is enabled', 'initTwoFactor.ledgerEnabled');
+        }
+
+        // additional check if the ledger is not connected but exists as a recovery method
+        const userRecoveryMethods = await this.wallet.getRecoveryMethods();
+        console.log('userRecoveryMethods', userRecoveryMethods);
```
Patrick1904, line 56: nit: debug?

packages/frontend/src/components/profile/hardware_devices/HardwareDevices.js
```diff
@@ -61,21 +60,23 @@ const Container = styled(Card)`
...
+    hasLedger,
+    ledgerIsConnected,
+    hasLedgerButNotConnected
+}) => {
 
     const [disabling, setDisabling] = useState(false);
     const [confirmDisable, setConfirmDisable] = useState(false);
     const dispatch = useDispatch();
-    const account = useSelector(selectAccountSlice);
+    // const account = useSelector(selectAccountSlice);
```
Patrick1904, line 75: Leftover?

packages/frontend/src/components/accounts/ledger/SetupLedger.js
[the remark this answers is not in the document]
andy-haynes: No, the variable _is_ defined - a variable being defined (declared) and having a value of `undefined` (uninitialized) are two different things.

What I'm saying is, the _only_ reason to use `typeof` rather than check for `undefined` directly is if the variable may not have be in scope. `typeof` ensures that referencing the variable doesn't break if it hasn't been declared. Here `has2fa` is defined with a `const` and so the `typeof` check is entirely unnecessary here.

I'm not aware of any scenario in NodeJS in which you would need to use `typeof x === 'undefined'` at all, especially when using `let` and `const` which throw `ReferenceError` when referenced before being declared.

packages/frontend/src/utils/twoFactor.js
```diff
@@ -42,6 +43,27 @@ export class TwoFactor extends Account2FA {
     }
 
     async initTwoFactor(accountId, method) {
+        // additional check if the ledger is enabled, in case the user was able to omit disabled buttons
```
andy-haynes, line 46: As much as I hate to say it, the new logic in `initTwoFactor` might be better as a new method on the wallet class. In this method all we need to know is `is there a ledger associated with this account?`, which relies on a multiple wallet methods and lots of wallet state but does not rely on the parameters to `initTwoFactor`.

From a clarity (and, optimistically, a future refactor) perspective, it flows a lot better if all the `initTwoFactor` method needs to be aware of is whether there's a ledger enabled - not how that determination is made. If `initTwoFactor` can simply call a wallet method then it doesn't need to take on this extra baggage, e.g.
```js
const isLedgerEnabled = await wallet.isLedgerEnabled();
if (!isLedgerEnabled) {
  throw new WalletError('Ledger Hardware Wallet is enabled', 'initTwoFactor.ledgerEnabled');
}
```

As an added bonus this new `isLedgerEnabled` method can be reused in other contexts in which we need to query whether a ledger is kinda-sorta-maybe enabled 😄

andy-haynes (changes requested):
Looking good, I think this covers our bases. I just have some comments on making the new behavior a bit more obvious, there are some subtleties here that we want to make as explicit as possible for future devs.

marcinbodnar:
@Patrick1904 @andy-haynes All changes are done.

Can we please merge it? It's crucial that this bug is fixed, and users will not fall into this trap. Any non-blockers can be addressed in a separate PR.

packages/frontend/src/components/profile/Profile.js
```diff
@@ -292,20 +300,31 @@ export function Profile({ match }) {
...
+                                ledgerIsConnected={ledgerIsConnected}
+                                hasLedgerButNotConnected={hasLedgerButNotConnected}
+                            />
+                        )}
                         <RecoveryContainer type='phrase' recoveryMethods={userRecoveryMethods}/>
                         { (shouldShowEmail || shouldShowPhone) && <h4><Translate id='profile.security.lessSecure'/><Tooltip translate='profile.security.lessSecureDesc' icon='icon-lg'/></h4>}
                         { shouldShowEmail && <RecoveryContainer type='email' recoveryMethods={userRecoveryMethods}/> }
                         { shouldShowPhone && <RecoveryContainer type='phone' recoveryMethods={userRecoveryMethods}/> }
-                        {!account.ledgerKey && (
+                        {!account.ledgerKey && !(hasLedgerButNotConnected && !twoFactor) && (
```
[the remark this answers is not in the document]
marcinbodnar: Seems like it can be simplified to this:
```js
{(twoFactor || !hasLedger) && (
```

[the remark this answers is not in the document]
andy-haynes: Could be, but we unfortunately are sourcing a value for `ledgerKey` from two different places - `useAccount(accountId).ledgerKey` and `useSelector(selectAccountSlice).ledgerKey`. The current check uses the former but `hasLedger` relies on the latter.

Unless we are 100% certain they are identical then we should incorporate the original `account.ledgerKey` reference into the new check to ensure it continues to work as expected.

Patrick1904 (approved):
LGTM.

Not a blocker: Can we please disable the 'continue' button on `/enable-two-factor` if the user has Ledger enabled on the account but used a different method to import the account. No reason to allow the user to continue in this case.

Screen Shot 2022-08-15 at 2 47 03 PM

packages/frontend/src/utils/wallet.js
```diff
@@ -235,6 +235,24 @@ export default class Wallet {
...
+
+        // additional check if the ledger is not connected but exists as a recovery method
+        const userRecoveryMethods = await this.getRecoveryMethods();
+        const accountState = await this.loadAccount();
+        const hasLedger = userRecoveryMethods.some((method) => method.kind === 'ledger');
+        const ledgerIsConnected = accountState.ledgerKey;
+        const hasLedgerButNotConnected = hasLedger && !ledgerIsConnected;
+        if (hasLedgerButNotConnected) {
+            return true;
+        }
```
andy-haynes, line 253: 💄  Can just be
```js
return hasLedger && !ledgerIsConnected;`
```

andy-haynes (approved):

Issue #2865, opened 2022-08-03
Unable to remove 2FA from near wallet


hcho112:
**Problem**

Unable to remove 2FA from my account `hcho112.near` and getting this:
<img width="529" alt="image" src="https://user-images.githubusercontent.com/6027014/182729231-5e37efc9-af00-4c44-8963-85d7ffae7f3c.png">

https://explorer.mainnet.near.org/transactions/FC5QGmsiR8SY2dGdpsAZPqGG1pwYKbVeShXUKEm9Z6nv

**Expected Behavior**
Should be able to remove 2FA

**Steps to reproduce**

1.  authenticate on near wallet
2. click `Disable` button on Two-Factor-Authentication 
3. insert account id and click `Disable 2FA` button

esaminu:
Thanks for reporting @hcho112 . So the reason you're facing this is because we need the current contract state keys to cleanup your account state when you disable 2fa by deploying and calling [`clean` on the cleanup contract](https://github.com/near/core-contracts/blob/master/state-cleanup/src/lib.rs#L17) with the state keys. We need to do this so that you can deploy other code to your account safely without bricking your account after disabling 2fa. 

There's a 50kb limit on `view_state` calls on non archival rpc nodes. This is why we attempt to delete all your requests before calling `view_state` so that your state falls under 50kb when the requests are deleted. The issue is that the multisig contract has a hardcoded [15 minute cooldown period](https://github.com/near/core-contracts/blob/master/multisig/src/lib.rs#L14) after requests are created for them to be deletable. So what's happening is that we're unable to delete the request and your state is still > 50kb and the `view_state` call fails so we can't get the state keys to call `clean` with and thus can't cleanup your state when you disable so we throw. In practice this means that in the worst case you need to disable 2fa 15 minutes after your last transactions if they added requests in the state amounting to > 50kb.

Some ideas to improve this would be if the cleanup contract did not need to know the keys when cleaning up, eliminating our dependency on `view_state` (not sure if this is possible) or if the multisig contract did not have the 15 minute cooldown period. Both improvements would require changes to the core contracts the wallet uses. 

We could pursue some improvement here but the multisig functionality is being completely eliminated from the wallet anyway as part of the migration so I think what we have now is fine until then and is a lot better than what we had before since its a 15 minute inconvenience instead of bricking users' accounts.

marcinbodnar:
[...]

PR #2880, opened 2022-08-11
feat: add blocking batch 2fa disable modal

esaminu:
This PR adds a blocking modal forcing users to disable 2fa before they can proceed with the wallet migration

Design: https://www.figma.com/file/w4SQVO6kBvAP5D65TGlm7y/Wallet-Domain-Migration?node-id=1050%3A2322

Testing instructions:

1. Ensure `SHOW_MIGRATION_BANNER` flag is enabled
2. Create an account and enable 2fa
3. Initiate migration flow by clicking "Transfer my accounts" on the banner
4. You should not be able to migrate your account without disabling 2fa on that account

### DIFF
--- a/packages/frontend/src/components/wallet-migration/Disable2FA.jsx
+++ b/packages/frontend/src/components/wallet-migration/Disable2FA.jsx
@@ -1,0 +1,166 @@
+import React, { useState, useEffect, useMemo, useRef } from 'react';
+import { Translate } from 'react-localize-redux';
+import { useDispatch, useSelector } from 'react-redux';
+import styled from 'styled-components';
+import { useImmerReducer } from 'use-immer';
+
+import { NETWORK_ID } from '../../config';
+import IconSecurityLock from '../../images/wallet-migration/IconSecurityLock';
+import { switchAccount } from '../../redux/actions/account';
+import { selectAccountId } from '../../redux/slices/account';
+import WalletClass, { wallet } from '../../utils/wallet';
+import AccountListImport from '../accounts/AccountListImport';
+import { IMPORT_STATUS } from '../accounts/batch_import_accounts';
+import sequentialAccountImportReducer, { ACTIONS } from '../accounts/batch_import_accounts/sequentialAccountImportReducer';
+import FormButton from '../common/FormButton';
+import LoadingDots from '../common/loader/LoadingDots';
+import Modal from '../common/modal/Modal';
+import { WALLET_MIGRATION_VIEWS } from './WalletMigration';
+
+
+const Container = styled.div`
+    padding: 15px 0;
+    text-align: center;
+    margin: 0 auto;
+
+    @media (max-width: 360px) {
+        padding: 0;
+    }
+
+    @media (min-width: 500px) {
+        padding: 48px 28px 12px;
+    }
+
+    .accountsTitle {
+        text-align: left;
+        font-size: 12px;
+        padding-top: 72px;
+        padding-bottom: 6px;
+    }
+
+    .title{
+        font-weight: 800;
+        font-size: 20px;
+        margin-top: 40px;
+    }
+`;
+
+const ButtonsContainer = styled.div`
+    text-align: center;
+    width: 100% !important;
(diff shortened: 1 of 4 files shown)


roshaans:
Hey @esaminu, could you please add brief instructions on how to test this PR?

agileurbanite:
Just tested locally looks like this is working as expected.

agileurbanite (dismissed):
LGTM

marcinbodnar:
It is possible to encounter the same issue as described here: https://github.com/near/near-wallet/issues/2865

I think we should inform users what type of issue it is, and inform users that there is a need to wait 15 minutes. It's should be addressed in a separate PR. Added an issue https://github.com/near/near-wallet/issues/2885

packages/frontend/src/components/wallet-migration/Disable2FA.jsx
```diff
@@ -0,0 +1,166 @@
...
+    useEffect(() => {
+        if (batchDisableNotStarted) {
+            initialAccountId.current = initialAccountIdOnStart;
+        }
+    },[initialAccountIdOnStart, batchDisableNotStarted]);
+
+    useEffect(() => {
+        const disable2faForCurrentAccount = async () => {
+            try {
+                await dispatch(switchAccount({accountId: currentAccount.accountId}));
```
marcinbodnar, line 107: Seems like `await` has no effect here.

packages/frontend/src/components/wallet-migration/Disable2FA.jsx
```diff
@@ -0,0 +1,166 @@
...
+                const account = await wallet.getAccount(currentAccount.accountId);
+                await account.disableMultisig();
+                localDispatch({ type: ACTIONS.SET_CURRENT_DONE });
+            } catch (e) {
+                localDispatch({ type: ACTIONS.SET_CURRENT_FAILED_AND_END_PROCESS });
+            } finally {
+                await dispatch(switchAccount({accountId: initialAccountId.current}));
```
marcinbodnar, line 114: Seems like `await` has no effect here.

packages/frontend/src/components/wallet-migration/Disable2FA.jsx
```diff
@@ -0,0 +1,166 @@
...
+    }, [completedWithSuccess]);
+
+    return (
+        <Modal
+            modalClass="slim"
+            id='migration-modal'
+            isOpen={true}
+            disableClose={true}
+            modalSize='md'
+            style={{ maxWidth: '435px' }}
```
marcinbodnar, line 135: I think we can use a regular size here (`modalSize`) to match other models.

packages/frontend/src/components/wallet-migration/Disable2FA.jsx
```diff
@@ -0,0 +1,166 @@
...
+            initialAccountId.current = initialAccountIdOnStart;
+        }
+    },[initialAccountIdOnStart, batchDisableNotStarted]);
+
+    useEffect(() => {
+        const disable2faForCurrentAccount = async () => {
+            try {
+                await dispatch(switchAccount({accountId: currentAccount.accountId}));
+                const account = await wallet.getAccount(currentAccount.accountId);
+                await account.disableMultisig();
```
marcinbodnar, line 109: We have `disableMultisig` redux action, probably we should use it here and additionally dispatch `fetchRecoveryMethods` as well to refresh the redux state.

marcinbodnar (approved):
Looks good.

Added a few comments, but these are not blockers so they can be addressed in a separate PR.

packages/frontend/src/components/wallet-migration/Disable2FA.jsx
esaminu: Agreed 👍 . Initially I was trying to disable 2fa without switching accounts in the wallet but was not able to. Since we are now switching before disable we should use the redux action

esaminu:
Thanks @marcinbodnar will address cleanups in the next iteration of this

PR #2883, opened 2022-08-15
Batch export to Sender extension

AmazingBeerBelly:
Presentaion: 
https://youtu.be/GX8Wr3SjlnI

Sender batch import build:
https://github.com/SenderWallet/sender-build/blob/main/sender-batch-import.zip


packages/frontend/src/components/wallet-migration/WalletMigration.jsx
```diff
@@ -57,25 +69,33 @@ const WalletMigration = ({ open, onClose }) => {
...
+    }, [handleSetActiveView]);
 
     const onContinue = useCallback(async () => {
-        const url = await encodeAccountsToURL(
-            availableAccounts,
-            state.migrationKey
-        );
+        let url = '';
+        if (state.walletType === 'sender') {
+            url = await encodeAccountsToSender(
```
agileurbanite, line 87: why duplicate the same function?

packages/frontend/src/components/wallet-migration/WalletMigration.jsx
```diff
@@ -39,6 +37,20 @@ const encodeAccountsToURL = async (accounts, publicKey) => {
         ]);
     }
 
+    return accountsData;
+};
+
+const encodeAccountsToSender = async (accounts, publicKey) => {
```
agileurbanite, line 43: encodeAccountsToURL seems like it needs to add an additional argument such as wallet or walletID this way we can do a lookup for url and keep the codebase DRY

agileurbanite: Shouldn't block this PR based on this. Will address in a separate PR.

packages/frontend/src/components/wallet-migration/WalletMigration.jsx
agileurbanite: Shouldn't block this PR based on this. Will address in a separate PR.

agileurbanite (approved):
LGTM

packages/frontend/src/components/wallet-migration/WalletMigration.jsx
esaminu: Agree with @agileurbanite here. Would appreciate if you can add a `getUrl` key to [`WALLET_OPTIONS`](https://github.com/near/near-wallet/pull/2883/files#diff-133f50f707eb791d6963507c2cdb6f305d087105bb4484ffc3a44a1af9928d55R37) and then call that from a single `encodeAccountsToURL` please

packages/frontend/src/components/wallet-migration/WalletMigration.jsx
esaminu: None of this will have to change if we do the minor refactor described in https://github.com/near/near-wallet/pull/2883/files#r946033135

esaminu:
@AmazingBeerBelly this is looking good! I added a comment regarding a minor refactor and wanted to confirm the following:

1. We can address this in a followup but is it not possible to detect the extension and call it directly from the wallet app? Seems like this would be better UX
2. When importing the accounts on your side are you adding a new key and deleting the one that's passed to you? Is it possible to link to this code?

agileurbanite:
> @AmazingBeerBelly this is looking good! I added a comment regarding a minor refactor and wanted to confirm the following:
> 
> 1. We can address this in a followup but is it not possible to detect the extension and call it directly from the wallet app? Seems like this would be better UX
> 2. When importing the accounts on your side are you adding a new key and deleting the one that's passed to you? Is it possible to link to this code?

I recommended against it as that would put too much logic in the wallet codebase specific to sender I didn't want to introduce.  Logic like:

* Is extension installed?
* Did the user attempt to export?
* Is user attempting to export?

I'd rather this logic live on their specific SPA versus ours, this will be for all other wallets that wallet.near.org supports exports to.

packages/frontend/src/components/wallet-migration/WalletMigration.jsx
agileurbanite: @esaminu i thought about this some more and we shouldn't block this PR, the original implementation made the function very MNW specific.  I'd recommend we handle the refactor.

esaminu:
@AmazingBeerBelly  I added a refactor in AmazingBeerBelly/near-wallet#1 to resolve above comments. Aside from this please make sure we are adding new keys and deleting the passed in keys and we should be good to go 👍

esaminu:
@AmazingBeerBelly In my testing I was only able to import one account unfortunately. Please make sure all accounts in the payload are added to Sender and also make sure ledger accounts are supported via the ledger hd path passed in the payload

AmazingBeerBelly:
> 

Thanks for your refactor @esaminu and I have merged this PR into my branch.

In my testing I can import multiple accounts. Could you provide some sample test case?

At present, the function of ledger accounts are still under development, so the import of ledger accounts are not yet supported.

esaminu:
@AmazingBeerBelly It's happening when some of my accounts are ledger accounts - maybe it's a parsing issue. I put some parsing code in this [repl](https://replit.com/@osmanabdelnasir/LongAgedScriptinglanguages#index.js) that you could reuse to make sure you are parsing the payload and disregarding hd paths correctly. 

We should also add a `isLedgerSupported` to `WALLET_OPTIONS` in this case and not list ledger accounts in the "We found X active accounts" modal in this case

Finally, can we submit a transaction to delete the passed in key and add a new NEAR full access key to the account and use that in the Sender app?

agileurbanite (approved):
LGTM

AmazingBeerBelly:
> 

@esaminu I can import these three accounts.

I have uploaded the latest Sender version to github. Maybe you need reinstall the Sender extension to make sure the code is latest.

https://github.com/SenderWallet/sender-build/blob/main/sender-batch-import.zip

Then open this link:

https://sender.org/transfer?keystore=NDQsMjksMzUsMiwxMDksMTExLDE0MiwxODUsMTc1LDIyMywyMjMsMTk5LDE0Nyw0NCwyMjUsNzMsMjMwLDEwNywxMjksOTcsMTY2LDc3LDExOCwyLDE4OCwxMjYsMzcsMjU0LDE5MiwxNTgsMjE0LDIyNiw1MiwyNTEsMjMyLDE3NiwyMzIsMjMyLDExLDI0NywxNjksMTEzLDE4NSwyNDAsMTg4LDEzNywxNTIsNTEsNzQsMTM0LDU0LDEwMSwxMDgsMTEzLDg3LDE1MCwxMjYsMTgwLDcxLDQyLDE3MCwyOCwxMjUsNTEsMTEzLDkwLDEzOCwxNDIsNTMsMTU1LDE1NCw2Myw5OCwyMTQsMTg0LDIzOSwyMjAsMjAxLDEyMSw0NywzMywxMDMsMTkwLDk2LDEyNSwyNDgsNjUsNTksMTE3LDE1NCwyMTgsNjcsNDUsMTU2LDE5NywxODAsMTY3LDM1LDc3LDIwMiwxOTEsMTE2LDE1NCwxMDcsMTIsMjIzLDE4MCwxNTMsMTY3LDE3NCwxNjIsMzYsMjMwLDc5LDQwLDIwNSw2NCwyNyw2MCw2MSw3MCwyOCwyMjEsMTAxLDMxLDI1MSwyOSw5OSwxMTcsMTQsOTUsMjksMTQ2LDIxMCwxMTEsMTgzLDc0LDgwLDExNyw0NiwxNzEsMTc2LDE3NiwyMzMsMjUwLDIwMCw0LDE4NSwyNywxNTYsNjAsMjIxLDIwOSwxMjUsMjQ0LDEwMSwyMTgsMTE2LDgwLDE0NSwxMiwxMiw0MCwyNDYsMjE4LDIxOCw1OCw2OSwxODgsMTUzLDM4LDIxLDcxLDIyMiwxMzUsMTI4LDQ0LDM2LDg4LDg2LDExOSw5LDE5MCwxODAsMTQ2LDU3LDE0MywxOTAsMjM0LDE4OCwxNjEsMCwzMSwwLDExNiw0MiwyNTEsMTU2LDQ2LDIzLDE5NSwyMzMsMTg4LDIyLDE2Nyw4OCwyMDIsMzMsMTIwLDEyLDk3LDE3MSwyMzIsNDAsMTQ3LDEzNyw3Myw3LDE1NCwxMjYsMTA0LDI0OSw5NSwxOTYsNjAsMjMxLDY2LDIwMywxNTUsMjIxLDE1MCw0LDExNSw1MiwxOSw1MywxMTEsMjI4LDk4LDIyNywxMjIsMjUyLDg5LDQ3LDExMywyMjksMTEsMTI3LDczLDE4MCwxMjksMTg3LDQ1LDcsMjE4LDEwMCwyMDMsNzYsMjI2LDExMCw2NCw3Nyw3NywxMjIsMTg4LDE5Miw4OSwxOTMsMTEzLDQ4LDc4LDE1NywxMzIsOTUsMjI1LDI0NiwyMDcsMTMzLDI1MywxMTQsMjQ0LDE3OSw0OCwxNzgsMjYsNjYsMTE5LDIwOSwxNjYsMjExLDEwLDIwLDE2NSwxODMsMTQ2LDE0MCwyNTAsMjUyLDE0Myw4LDM4LDIzLDIxMCwyNTEsMTgsNCwxMDksNjMsMTE2LDI1MiwxNTQsMTE4LDE2NCwyMjUsMTQ0LDU5LDU1LDg3LDE4MiwyMDgsMTE3LDExNCwxODAsOTYsMTIwLDIyOSwyMTAsMTc5LDYwLDEwNiwxMjAsMjMzLDIzNywxMzgsMTQyLDY5LDIyOCwyMiwxNzEsOTMsMTk2LDEzNSwxMDQsMTgyLDE2NiwxNDk=&network=testnet

Import the key, and import success.

Here is my test screen shot:

CleanShot 2022-08-17 at 00 30 01@2x

And for last question, Sender is not support to send transaction to delete the passed in key and create new full access key to the account. If you think it's a important feature, we can add this feature as soon as possible.

PR #2890, opened 2022-08-28
Feature (MYNW-152): Add Transak top-up integration

PinkiNice:
# Transak Integration
https://mnw.atlassian.net/browse/MYNW-152

### 1. Add Transak to the list of top-up providers
This PR adds Transak as another option for topping-up wallet with credit-card payments.
Integration is made similar to MoonPay but without server-signer signatures.
Transak integration is limited only to MNW domains, to now confuse users from wallet.near.org with MNW logo in emails and during on-ramp process. 

**New .env variables has been introduced:**
```
TRANSAK_API_KEY: '449da987-df12-4239-9e4e-9d5495062902' # Transak's partner api key
TRANSAK_BUY_URL: 'https://staging-global.transak.com/' # URL to redirect once user selects Transak
```

---

### 2. Avoid bumpy layout with disabled providers
Instead of removing disabled options from layout they are grayed out and made unclickable (see screenshots below).

## Screenshots
tg_image_3473578045
tg_image_4158939792
tg_image_2505728325


packages/frontend/src/components/buy/BuyNear.js
```diff
@@ -209,6 +218,10 @@ export function BuyNear({ match, location, history }) {
         );
     };
 
+    // To avoid user confusion with MNW logo being featured in emails
+    // Only allow Transak on MNW domains
+    const onrampMethods = [PayMethods.moonPay, isWhitelabel() ? PayMethods.transak : null, PayMethods.utorg, PayMethods.nearPay, PayMethods.ftx].filter((v) => !!v);
```
loosefer, line 223: Too long, please split to lines

packages/frontend/src/components/buy/BuyNear.js
```diff
@@ -175,26 +177,33 @@ const StyledContainer = styled.div`
 
 export function BuyNear({ match, location, history }) {
     const accountId = useSelector(selectAccountId);
-    const [moonPayAvailable, setMoonPayAvailable] = useState(null);
+    const [moonPayAvailable, setMoonPayAvailable] = useState(true);
```
roshaans, line 180: It seems like Moonpay won't be available everywhere, so by default, we should assume that it is not supported and let `checkMoonPay()` figure out whether it is available or not.

packages/frontend/src/components/buy/BuyNear.js
```diff
@@ -175,26 +177,33 @@ const StyledContainer = styled.div`
...
         setUtorgPayUrl(buildUtorgPayLink(accountId));
         setFtxPayUrl(buildFtxPayLink(accountId));
+        setTransakPayUrl(buildTransakPayLink(accountId));
         checkMoonPay();
     }, [accountId]);
 
     const PayMethods = useMemo(
-        () => getPayMethods({ accountId, moonPayAvailable, signedMoonPayUrl, utorgPayUrl, ftxPayUrl }),
+        () => {
+            const paymentMethods = getPayMethods({ accountId, transakPayUrl, moonPayAvailable, signedMoonPayUrl, utorgPayUrl, ftxPayUrl });
```
loosefer, line 200: Wanna make more short?

`() => getPayMethods({ accountId, transakPayUrl, moonPayAvailable, signedMoonPayUrl, utorgPayUrl, ftxPayUrl });`

packages/frontend/src/components/buy/FundingCard.jsx
```diff
@@ -99,23 +99,17 @@ export const FundingCard = ({ title, subTitle, actions, link }) => {
...
+                            <a href={link.url} target="_blank" rel="noreferrer">
                                 <Translate id={link.title} />
                             </a>
                         </Link>
                     )}
                 </SubTitle>
             </TextWrap>
             <FundingTypeWrap>
-                {actions
-                    .filter(({ disabled }) => !disabled)
```
roshaans, line 111: Without the disabled filter, moonPay will be shown as a TopUp method regardless if it is available in a region. 

See https://github.com/mynearwallet/near-wallet/blob/ea5dcd954eaae5634ab67203cf4f0b44ad4bbebb/packages/frontend/src/config/buyNearConfig.js#L23

and [`isMoonPayAvalaible` ](https://github.com/mynearwallet/near-wallet/blob/ea5dcd954eaae5634ab67203cf4f0b44ad4bbebb/packages/frontend/src/utils/moonpay.js#L12)

packages/frontend/src/components/buy/FundingType.jsx
```diff
@@ -45,14 +49,27 @@ const WrapperImg = styled.div`
...
+            onClick={() => track && track()}
+            disabled={disabled}
+        >
             <Wrapper>
                 <WrapperImg title={name}>
-                    <img src={icon} alt="" />
+                    <img src={icon} alt={name} />
                 </WrapperImg>
-                <Button />
+                {disabled ? null : <Button />}
```
roshaans, line 72: nit: {disabled ? null : <Button />} -> {!disabled && <Button />}

packages/frontend/src/components/buy/providers/transak.js
```diff
@@ -0,0 +1,23 @@
...
+import { TRANSAK_API_KEY, TRANSAK_BUY_URL } from '../../../config';
+
+export function buildTransakPayLink(accountId) {
+    const url = {
+        url: TRANSAK_BUY_URL,
+        query: {
+            apiKey: TRANSAK_API_KEY,
+            cryptoCurrencyList: 'NEAR',
+            walletAddress: accountId,
+            // Disabled, because ay confuse users
```
roshaans, line 12: Nit: Spelling error in comment 

Also, can you provide a bit more clarity as to why the user will be confused?

packages/frontend/src/config/buyNearConfig.js
```diff
@@ -6,13 +6,21 @@ import monoPay from '../components/buy/assets/monoPay.svg';
...
+    moonPayAvailable,
+    signedMoonPayUrl,
+    utorgPayUrl,
+    transakPayUrl,
+    ftxPayUrl,
+}) => {
+    const paymentMethods = {
         moonPay: {
-            disabled: accountId && !moonPayAvailable,
+            disabled: accountId && moonPayAvailable,
```
roshaans, line 23: We want to disable this TopUp method if moonPayAvalaible returns false. Can we revert this back to read `disabled: accountId && !moonPayAvailable`?

roshaans (commented):
Saw some changes that may affect current behaviors with MoonPay, other integration with Transak looks good. Will give this a test soon.

packages/frontend/src/components/buy/BuyNear.js
PinkiNice: I made this change from UX standpoint, let me clarify:
It looks to me that it is better to assume by default that MoonPay **is available**. 
1) There is no checks for other providers - so they are active no matter supported they are or not.
2) For users with slow internet the delay for detecting MoonPay availability may be long enough for them to decide MoonPay is not available (which would be wrong and we want to avoid misleading users).

So, in terms of progressive enhancement I see it as favorable to only block MoonPay when we know **for sure** that user will not be able to use it.

packages/frontend/src/components/buy/FundingCard.jsx
PinkiNice: It is made by design - instead of removing MoonPay block from layout we make it look "disabled" - gray and low contrast.
It helps to avoid bumpy layout behaviour during loading time.

roshaans: Good point :)

packages/frontend/src/components/buy/BuyNear.js
roshaans: Really good point :) 

As long as the MoonPay Option is grayed out when isMoonPayAvalaible() returns False, then this should all be fine.

packages/frontend/src/config/buyNearConfig.js
PinkiNice: Indeed! Forgot to revert change after testing

packages/frontend/src/components/buy/providers/transak.js
PinkiNice: Nevermind, I enabled this param back to ease user-flow. 
It works fine now

PinkiNice:
@roshaans I've fixed the issues. Thanks for spotting them!

packages/frontend/src/components/buy/FundingType.jsx
PinkiNice: I'd like to keep as it is now.
This way it is more verbose but simpler to understand (no need to make invertion in the mind) and easier to read.

packages/frontend/src/components/buy/FundingCard.jsx
```diff
@@ -99,23 +99,17 @@ export const FundingCard = ({ title, subTitle, actions, link }) => {
                     <Translate id={subTitle} />
                     {link && (
                         <Link>
-                            <a href={link.url} target='_blank' rel='noreferrer' >
+                            <a href={link.url} target="_blank" rel="noreferrer">
```
Patrick1904, line 102: Please use single quotes

packages/frontend/src/components/buy/FundingType.jsx
```diff
@@ -45,14 +49,27 @@ const WrapperImg = styled.div`
...
+    name,
+    track,
+    disabled,
+    provideReferrer,
+}) => {
     return (
-        <LinkWrap href={link} target='_blank' rel={!provideReferrer ? 'noreferrer' : 'noopener'} onClick={() => track && track()}>
+        <LinkWrap
+            href={link}
+            target="_blank"
```
Patrick1904, line 63: Single quote

Patrick1904 (approved):
LGTM

roshaans (approved):
LGTM!

PR #2897, opened 2022-09-01
Mynw 88 cache requests to indexer

loosefer:

- cache indexer response every 30 sec
- accumulate indexer response to IndexedDb cache and concatenate it with prevously

Prev PR: https://github.com/near/near-wallet/pull/2844


roshaans:
Further developments will be made on #2899.

PR #2899, opened 2022-09-01
FEAT: Loosefer/mnw cache requests to indexer

roshaans:
PR Originally by @loosefer #2897

This PR will allow us to see development previews of the changes so they can be better tested for release. 

-----

- `likelyNfts` & `likelyTokens` responses are cached upon reloading of the page if the reloading of the page happens 30 seconds after the previous request. 
- accumulate indexer response to IndexedDb cache and concatenate it with previously


Patrick1904 (approved):
Great job everyone!

PR #2892, opened 2022-08-31
Feat/MYNW-154 Added Explore page

followdarko:
Figma design https://www.figma.com/file/qVgRY71P3XjttByfO6prmG/Explore-A?node-id=1%3A175

Added:

* Explore page /explore
* NavBar link to explore
* Light sidebar banner on the wallet page

PR #2900, opened 2022-09-01
merge explore page to master

roshaans:
This PR merges #2892 into master which was previously reverted because of issues on the `/explore` page. This PR creates a flag to hide the `/explore` page while giving a way for the MNW team to iterate in the background and fix the changes. This PR also gives MNW access to see this in a dev environment.

### DIFF
--- a/packages/frontend/src/components/wallet/ExploreApps.js
+++ b/packages/frontend/src/components/wallet/ExploreApps.js
@@ -2,6 +2,7 @@ import React from 'react';
 import { Translate } from 'react-localize-redux';
 import styled from 'styled-components';
 
+import { SHOW_EXPLORE_PAGE } from '../../../../../features';
 import { EXPLORE_APPS_URL, EXPLORE_DEFI_URL } from '../../config';
 import BerryClubLogo from '../../images/berryclub-logo.svg';
 import FluxLogo from '../../images/flux-logo.svg';
@@ -10,6 +11,8 @@ import MintbaseLogo from '../../images/mintbase-logo.svg';
 import ParasLogo from '../../images/paras-logo.svg';
 import RefLogo from '../../images/ref-logo.svg';
 import FormButton from '../common/FormButton';
+
+const AWESOME_NEAR_URL = 'https://awesomenear.com/projects';
 
 const Container = styled.div`
     && {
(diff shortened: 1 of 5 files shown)


packages/frontend/src/components/wallet/ExploreNativeBanner.js
```diff
@@ -45,12 +64,18 @@ const StyledBannerLink = styled(Link)`
...
-            <img src={exploreAppsImg} alt="Explore Apps Banner" width="211" height="180"/>
-            <h2><Translate id='exploreApps.exploreNear'/></h2>
-            <div><Translate id='exploreApps.text'/></div>
-            <StyledBannerLink to="/explore">
-                <Translate id='exploreApps.exploreApps'/>
-            </StyledBannerLink>
+            <img src={exploreAppsImg} alt="Explore Apps Banner" width="211" height="180" />
+            <h2><Translate id='exploreApps.exploreNear' /></h2>
+            <div><Translate id='exploreApps.text' /></div>
+            {!SHOW_EXPLORE_PAGE ? (
```
Patrick1904, line 70: I think this should be `SHOW_EXPLORE_PAGE ?` without the `!`

packages/frontend/src/components/wallet/ExploreNativeBanner.js
```diff
@@ -26,6 +29,22 @@ const StyledContainer = styled.div`
     }
 `;
 
+const StyledBannerLinkTemporary = styled.a`
```
Patrick1904, line 32: Can we skip creating a new styled component so that there's less cleanup later. You can use the `StyledBannerLink` for the external link by doing something like:

`<StyledBannerLink to={{ pathname: 'https://awesomenear.com' }} target='_blank' />`

Let's pass the URL directly without the variable so that it's super obvious what needs to be removed once the flag is removed

packages/frontend/src/components/wallet/ExploreNativeBanner.js
roshaans: yes, was testing to make sure the condition worked but forgot to remove it :)

packages/frontend/src/components/wallet/ExploreApps.js
```diff
@@ -111,34 +114,34 @@ const ExploreApps = () => {
...
+                <Translate id='exploreApps.title' />
             </h2>
             <div className='desc'>
-                <Translate id='exploreApps.desc'/>
+                <Translate id='exploreApps.desc' />
             </div>
             <div className='buttons'>
                 <FormButton
-                    linkTo={EXPLORE_APPS_URL}
+                    linkTo={SHOW_EXPLORE_PAGE ? EXPLORE_APPS_URL : AWESOME_NEAR_URL}
```
Patrick1904, line 133: Can we please skip any changes to the awesomenear.com url since it's already pointing to the main page 🙏

packages/frontend/src/components/wallet/ExploreNativeBanner.js
roshaans: Yes, definitely much cleaner this way

Issue #2885, opened 2022-08-16
User should be aware of the problem and a solution with disabling 2fa


marcinbodnar:
If the user encounters an issue with disabling 2fa described here:
https://github.com/near/near-wallet/issues/2865#issuecomment-1205549360
https://github.com/near/near-wallet/pull/2880#issuecomment-1216996281

The user should be informed that it's not a permanent error and he should try in 15 minutes.

PR #2901, opened 2022-09-02
fix: disable2fa error messages

roshaans:
This PR addresses Issue #2885

Adds better visibility into errors while disabling 2fa
- If the user tries to disable the 2fa within the 15 minute cooldown period, they are told to wait 15 minutes. 

| User Cancels disable 2fa request      | User attempts to disable 2fa during cooldown |
| ----------- | ----------- |
| image      | image       |


Patrick1904 (approved):
LGTM

PR #2909, opened 2022-09-07
docs: added guide for devs to add their wallets as options for migration

roshaans:
Created a guide for devs to create a PR to add their wallet as an option for migration on the Near Wallet


Patrick1904 (approved):
Thank you @roshaans - this is great! 💯

PR #2910, opened 2022-09-08
fix: fixes for mobile light sidebar

followdarko:
* Made Explore Near banner always be the first in light Sidebar
* Fixed paddings for Custom Address banner on mobile light Sidebar

Screenshot 2022-09-09 at 00 45 34
Screenshot 2022-09-09 at 00 45 15


roshaans (approved):
LGTM

PR #2914, opened 2022-09-15
MYNW-196: fix(recovery-methods): skeletons for error

PinkiNice:
Keep showing loading skeletons in case when loading failed with an error
### How it works now?
A new status `isInitialized` is now available on every thunk.
`isInitialized` is set to `true` once thunk succeeds for the first time (and stays `true` even in case of errors).
Initial state is set to be `false`.

Recovery methods skeleton's are shown until `isInitialized` will be set to `true`.

image


Patrick1904 (approved):
LGTM 💯

PR #2925, opened 2022-09-27
[WEP-160] fix: update warning text for disable 2fa

roshaans:
simplification of the two-factor authentication verification popup.


roshaans:
Danny made this change in latest PR

PR #2928, opened 2022-09-30
[WEP-161 ]fix: error handling

roshaans:
The batch disable 2fa flow fails without any errors to the users. There are a few errors and hints we want to give the user:

- If the user tries to disable 2fa within 15 mins of each other for the same account
- The user hits cancel on the input 2fa code screen(let user know to wait 15 mins if they want to try again)
- When the user has a bricked account and enters the wrong key phrase into the modal. 

image

image

### DIFF
--- a/packages/frontend/src/components/wallet-migration/Disable2FA.jsx
+++ b/packages/frontend/src/components/wallet-migration/Disable2FA.jsx
@@ -7,6 +7,7 @@ import { useImmerReducer } from 'use-immer';
 import { NETWORK_ID } from '../../config';
 import IconSecurityLock from '../../images/wallet-migration/IconSecurityLock';
 import { switchAccount } from '../../redux/actions/account';
+import { showCustomAlert } from '../../redux/actions/status';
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/wallet-migration/Disable2FA.jsx
```diff
@@ -132,6 +133,12 @@ const Disable2FAModal = ({ handleSetActiveView, onClose }) => {
                     localDispatch({ type: ACTIONS.SET_CURRENT_DONE });
                 }
             } catch (e) {
+                dispatch(showCustomAlert({
+                    errorMessage: e.message,
+                    success: false,
+                    messageCodeHeader: 'error'
+                }));
+                await new Promise((r) => setTimeout(r, 3000));
```
roshaans, line 141: error shows for 3 seconds. Without this delay, error message goes away in a second.

gutsyphilip (approved):
Looks good to me!

PR #2926, opened 2022-09-28
WEP-99: Unbrick 2FA process to UI

hcho112:
This PR contains implementation for showing Disable 2FA banner for near wallet.
(Please check [ticket](https://nearinc.atlassian.net/browse/WEP-99) for further details)

Banner
image

Modal
image

Account lock Modal (only show if target account is bricked)
image

Reset via passphrase
image

If anyone need help with manually creating bricked account, please let me know.

PR #2929, opened 2022-09-30
2022-09-30 Release 

roshaans:
- #2926
- #2928
- #2916
- #2917
- #2934


packages/frontend/src/components/common/TwoFactorDisableBanner.js
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    return (
+        <Container className='banner-container'>
+            <div className='alert-container'>
+                <AlertTriangleIcon color={'#DC1F25'} />
+            </div>
+            <div className='content'>
+                <h4 className='title'>
+                    <Translate id='twoFactorDisbleBanner.title' data={{ accountsCount, context: accountsCount === 1 ? 'Account Needs' : 'Accounts Need'}} />
```
Patrick1904, line 147: `context: accountsCount === 1 ? 'Account Needs' : 'Accounts Need'}}`

The above won't work with other languages. You may have to split the title string into several strings to handle singular vs plural. You can render the number of accounts directly in the code since it won't change across languages.

hcho112: Thanks for the feedback, follow up PR has been created [here](https://github.com/near/near-wallet/pull/2942)

PR #2934, opened 2022-09-30
feat: unbrick accounts from wallet account settings + refactoring

roshaans:
Integrated flow for disabling 2fa for bricked accounts from user settings page.

### DIFF
--- a/packages/frontend/src/components/profile/two_factor/TwoFactorAuth.js
+++ b/packages/frontend/src/components/profile/two_factor/TwoFactorAuth.js
@@ -64,7 +64,7 @@ const Container = styled(Card)`
     }
 `;
 
-const TwoFactorAuth = ({ twoFactor, history, isBrickedAccount }) => {
+const TwoFactorAuth = ({ twoFactor, history, isBrickedAccount, onDisableBrickedAccountComplete }) => {
     const [confirmDisable, setConfirmDisable] = useState(false);
     const [showBrickedAccountModal, setShowBrickedAccountModal] = useState(false);
     const account = useSelector(selectAccountSlice);
@@ -89,6 +89,7 @@ const TwoFactorAuth = ({ twoFactor, history, isBrickedAccount }) => {
 
     const onAccountLockComplete = () => {
         setShowBrickedAccountModal(false);
+        onDisableBrickedAccountComplete();
     };
 
     const onAccountLockCancel = () => {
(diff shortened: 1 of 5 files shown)


packages/frontend/src/components/wallet-migration/utils.ts
```diff
@@ -0,0 +1,13 @@
+export const isAccountBricked = async (account) => {
+    // If account is bricked and unable to run checkMultisigCodeAndStateStatus
+    if (!account.checkMultisigCodeAndStateStatus) {
+        console.log('cant find func');
```
ggoshanov, line 4: Remove console.log?

hcho112, line 4: 🔪

packages/frontend/src/components/profile/two_factor/TwoFactorAuth.js
```diff
@@ -108,16 +139,19 @@ const TwoFactorAuth = ({ twoFactor, history }) => {
                     twoFactorKind={twoFactor.kind}
                 />
             )}
-            {!twoFactor && (
+            {!twoFactor && isBrickedAccount && showBrickedAccountModal && (
```
hcho112, line 142: shouldn't it be `twoFactor` not `!twoFactor`? 
since it is 2fa enabled and we want to show modal upon disabling it?

packages/frontend/src/components/profile/two_factor/TwoFactorAuth.js
```diff
@@ -64,41 +65,71 @@ const Container = styled(Card)`
...
         setConfirmDisable(false);
     };
 
+    const onAccountLockClose = () => {
+        setShowBrickedAccountModal(false);
+    };
+
+    const onAccountLockComplete = () => {
+        setShowBrickedAccountModal(false);
+        this.forceUpdate();
```
hcho112, line 93: code breaks at this point. I don't see any method of `forceUpdate`

hcho112 (changes requested):
There are few critical changes required in order to successfully disable 2fa from profile page. All of them are simple change :)
Good work tho!

packages/frontend/src/components/profile/two_factor/TwoFactorAuth.js
roshaans: No, because for brickedaccounts, there is no value for twoFactor because we are unable to parse the account to get details such as the 2fa type and date of creation.

packages/frontend/src/components/profile/two_factor/TwoFactorAuth.js
roshaans: This is an internal method for classComponents which we do not need anymore. There is a slight issue of the screen not refreshing once the 2fa has been disabled for a bricked account. I'll remove this, but that issue still stands. We can tackle it in next release.

packages/frontend/src/components/profile/two_factor/TwoFactorAuth.js
hcho112: Are you sure about this? I just created bricked account manually and I got this:

image

hcho112: When I click `Disable 2FA` button from banner it does show things correctly:
image

hcho112:
@roshaans I have tested the above issue again with fresh account and issue of not showing account lock modal persisted. As we discussed I have pushed the fix directly on top of your PR. (Also made minor style change on buttons) 

image

hcho112 (approved):
LGTM !
Please review on my latest change and if you agree with the change, let merge this work and make the release ready.

Good work!!

Patrick1904 (approved):
Great job! 👍

PR #2943, opened 2022-10-03
feat: disable 2fa enabling

roshaans:
disable users from enabling 2fa as part of the off boarding initiative

### DIFF
--- a/packages/frontend/src/components/profile/two_factor/TwoFactorAuth.js
+++ b/packages/frontend/src/components/profile/two_factor/TwoFactorAuth.js
@@ -5,7 +5,7 @@ import { useSelector, useDispatch } from 'react-redux';
 import { withRouter } from 'react-router-dom';
 import styled from 'styled-components';
 
-import { MULTISIG_MIN_AMOUNT, ALLOW_2FA_ENABLE_HASHES } from '../../../config';
+import { MULTISIG_MIN_AMOUNT } from '../../../config';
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/profile/two_factor/TwoFactorAuth.js
```diff
@@ -136,7 +135,7 @@ const TwoFactorAuth = ({ twoFactor, history, isBrickedAccount, onDisableBrickedA
                         <FormButton
                             onClick={() => history.push('/enable-two-factor')}
                             trackingId="2FA Click enable button"
-                            disabled={!account.canEnableTwoFactor || existingContract}
+                            disabled={true}
```
hcho112, line 138: I think we should only disable it if it is NEAR WALLET not MNY. 
`isWhitelabel ? (!account.canEnableTwoFactor || existingContract) : false`
Leave existing logic as it is for MNY and only apply disable on NEAR WALLET?

Also instead of just disable button, would it be nicer to add simple tooltip, showing some context around why this is disabled? something like `We are stopping users from enabling 2FA on Near Wallet as part of wallet migration requirement` (I hope we get some help on rephrase this to better wordings)

roshaans: We just got confirmation that we should disable across MNW and Near-wallet

hcho112 (approved):
LGTM

PR #2944, opened 2022-10-04
feat: remove modal all together unless the user has 2fa enabled

roshaans:
Instead of graying out the 2fa Enable button in the 2fa component, we will remove the 2fa component all together to not confuse the user. 

The change will also effect MNW. 

If the user has 2fa enabled, they will still be shown the 2fa disable component.  

**Switching between an account with and without 2fa enabled**
2fa_account_vs_non_2fa_account

**Disabling 2fa on an account**
remove_2fa


Patrick1904 (approved):
LGTM

PR #2945, opened 2022-10-05
fix/MNW correct implicit account issues

NotEternal:
- fix overflow in an accounts dropdown on the LinkDrop page: [issue MNW-184](https://mnw.atlassian.net/browse/MNW-184)

current-id-short

drop-down-short
---

- fix overflow in an NFT transfer result: [issue MNW-186](https://mnw.atlassian.net/browse/MNW-186)

nft-fix

- allows NFT sending to an implicit account: [issue MNW-173](https://mnw.atlassian.net/browse/MNW-173)

In the screen above we can see a successful tx to this not registered on-chain account:
**77a594302ed8841680e1c7a5344a59bc7ebc75de88994f84026d589f851dffe9**
[Explorer link](https://explorer.testnet.near.org/accounts?query=77a594302ed8841680e1c7a5344a59bc7ebc75de88994f84026d589f851dffe9)

can-send

### DIFF
--- a/packages/frontend/src/components/common/AccountDropdown.js
+++ b/packages/frontend/src/components/common/AccountDropdown.js
@@ -6,6 +6,7 @@ import styled from 'styled-components';
 import { switchAccount } from '../../redux/actions/account';
 import { selectAccountId } from '../../redux/slices/account';
 import { selectAvailableAccounts } from '../../redux/slices/availableAccounts';
+import { shortenAccountId } from '../../utils/account';
 import classNames from '../../utils/classNames';
 import DropDown from '../common/DropDown';
 
@@ -37,8 +38,6 @@ const Container = styled.div`
         cursor: pointer;
         color: #72727A;
         font-weight: 500;
-        text-overflow: ellipsis;
-        overflow: hidden;
 
         :last-of-type {
             border: 0;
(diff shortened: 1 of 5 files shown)


packages/frontend/src/components/common/DropDown.js
```diff
@@ -41,6 +41,8 @@ const Container = styled.div`
         }
 
         .dropdown-title {
+            overflow: hidden;
```
Patrick1904, line 44: Since we have the `shortenAccountId` utility function for implicit accounts, can we modify the `DropDown` component and utilize the `shortenAccountId` function so that we get the ellipsis (`xxx...xxx`) in the middle of implicit accounts?

NotEternal: Only for the selected ID or also in the drop-down list?

Patrick1904: It'd be great to aim for consistency by having the correct ellipsis treatment (`xxx...xxx`) everywhere we show an implicit address, unless cases where showing the full address is needed or otherwise useful.

So yes, both selected ID and each account in the dropdown 👍

NotEternal: I updated screens in the description. Desktop view is the same.
By the way, it wasn't a wise decision to change a title directly in the `DropDown` component. Because this is a common component and we don't know what type of title we can receive from outside.

packages/frontend/src/components/common/AccountDropdown.js
```diff
@@ -75,18 +76,17 @@ export default function AccountDropdown({ disabled, 'data-test-id': testId }) {
...
-                }
+                title={shortAccountId}
+                content={accountsWithoutCurrent.map((account, i) => (
+                    <div
+                        key={i}
+                        title={account}
+                        onClick={() => dispatch(switchAccount({ accountId: account }))}
+                        className='account-dropdown-toggle'
+                    >
+                        {shortenAccountId(account)}
```
hcho112, line 87: use `shortAccountId` instead of `{shortenAccountId(account)}`

hcho112 (approved):
LGTM

packages/frontend/src/components/common/DropDown.js
Patrick1904: Great point on `DropDown` being a generic component 👍

Patrick1904 (approved):

PR #2949, opened 2022-10-09
MNW-215: add simple FT block-list implementation 

PinkiNice:
Adds simplest implementation for hiding scam tokens - just a hard-coded list for now.
More flexible implementations to-be-done in next iteration

### DIFF
--- a/packages/frontend/src/hooks/useTokenBlacklist.js
+++ b/packages/frontend/src/hooks/useTokenBlacklist.js
@@ -1,0 +1,16 @@
+import { useMemo } from 'react';
+
+export function useTokenBlacklist({ tokens }) {
+    // TODO: make list dynamic, fetch from db
+    const blacklistedTokens = ['kusama-airdrop.near'];
+
+    const allowedTokens = useMemo(() => {
+        if (!tokens) {
+            return tokens;
+        }
+
+        return tokens.filter((token) => !blacklistedTokens.includes(token.contractName));
+    }, [tokens]);
+
+    return { blacklistedTokens, allowedTokens };
+}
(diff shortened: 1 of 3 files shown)


packages/frontend/src/hooks/useTokenBlacklist.js
```diff
@@ -0,0 +1,16 @@
+import { useMemo } from 'react';
+
+export function useTokenBlacklist({ tokens }) {
+    // TODO: make list dynamic, fetch from db
+    const blacklistedTokens = ['kusama-airdrop.near'];
+
+    const allowedTokens = useMemo(() => {
```
PinkiNice, line 7: Memo probably won't be working great here - as array will be recreated at every render.
But it should not pose huge performance issues anyway, as there is only 1 element for now.

packages/frontend/src/routes/WalletWrapper.js
```diff
@@ -64,7 +67,7 @@ export function WalletWrapper({
             createFromImplicitSuccess={createFromImplicitSuccess}
             createCustomName={createCustomName}
             zeroBalanceAccountImportMethod={zeroBalanceAccountImportMethod}
-            fungibleTokensList={fungibleTokensList}
+            fungibleTokensList={blacklist.allowedTokens}
```
PinkiNice, line 70: I tried to remove black-listed tokens as high in hierachy as possible, to avoid multiple filtrations. 
If there is other places where token list will be used directly from the store - tokens will not be filtered there.
Please point if there is such places.

roshaans: This would also be good to add in the filter step of `components/send/components/views/SelectToken.js`

roshaans (approved):
Changes work great. This is good to approve. If you can add filtering blacklist assets in the Select Asset Modal when the user is sending funds, that would be great too. 

Just make sure that the search feature in that modal is still functional, as my limited testing to try to implement that feature caused page reloads.

roshaans (approved):
Working good on my end!
LGTM!

PR #2954, opened 2022-10-17
feat: migration flow - rotate keys for users

roshaans:
Rotate keys for a user in a seamless way. 
Refactored the wallet migration project. 

Things to be worked on
- [x] show user the keyphrase screen to be sure they wrote down the new key. 
- [X] Key rotation functionality. Create a new key. Save it in local storage. 
- [X] Allow users to retry failed key rotation attempts. 
- [X]  Allow users to continue migrate their keys even if they did not rotate all of their keys - **due to failed attempts only.** 

Some UI bugs I noticed along the way that need to be fixed and addressed **in the next PR**

- [x] can't click banner once again after clicking it, exiting it and trying to return back to it. 
- [x] when the seed phrase recovery screen shows up. Show the account ID of the account being rotated on that. 
- [ ] error messages have a weird vertical offset.


packages/frontend/src/components/wallet-migration/modals/MigrateAccountsModal/SelectDestinationWallet.jsx
```diff
@@ -128,7 +127,7 @@ const WalletOptionsListingItem = styled.div`
             width: 11px;
             background-color: white;
             border-radius: 50%;
-            box-shadow: 1px 0px 2px 0px #0000005;
+            box-shadow: 1px 0px 2px 0px "#0000005";
```
hcho112, line 130: It should be `#000005` ?

packages/frontend/src/components/wallet-migration/modals/RotateKeysModal/RotateKeysModal.jsx
```diff
@@ -0,0 +1,300 @@
...
+import { WALLET_MIGRATION_VIEWS } from '../../WalletMigration';
+
+const ButtonsContainer = styled.div`
+    text-align: center;
+    width: 100% !important;
+    display: flex;
+`;
+
+const StyledButton = styled(FormButton)`
+    width: calc((100% - 16px) / 2);
```
Patrick1904, line 31: Can we avoid setting a width on buttons like this and instead utilize flexbox or similar for button layout?

packages/frontend/src/components/wallet-migration/modals/RotateKeysModal/RotateKeysModal.jsx
```diff
@@ -0,0 +1,300 @@
...
+            localDispatch({ type: ACTIONS.SET_CURRENT_DONE });
+            setShowConfirmSeedphraseModal(() => false);            
+        } catch (e) {
+            localDispatch({ type: ACTIONS.SET_CURRENT_FAILED_AND_END_PROCESS });
+            dispatch(showCustomAlert({
+                errorMessage: e.message,
+                success: false,
+                messageCodeHeader: 'error'
+            }));
+            await new Promise((r) => setTimeout(r, 3000));
```
Patrick1904, line 151: Can we please make it more clear why this `setTimeout` is needed?

packages/frontend/src/components/wallet-migration/modals/RotateKeysModal/RotateKeysModal.jsx
```diff
@@ -0,0 +1,300 @@
...
+                   />
+               </Modal>
+           )
+               : (
+                   <Modal
+                       modalClass="slim"
+                       id='migration-modal'
+                       onClose={() => {}}
+                       modalSize='md'
+                       style={{ maxWidth: '431px' }}
```
Patrick1904, line 251: It would be great if we can set the modal size once instead of three times here

Patrick1904:
Great job @roshaans! Is there a way for me to test this flow in the UI?

packages/frontend/src/components/wallet-migration/modals/RotateKeysModal/RotateKeysModal.jsx
roshaans: Yes, I believe all modals use this Styled Button and Container styled element, so would be great to have a followup PR to clean all of this up in one go.

packages/frontend/src/components/wallet-migration/modals/RotateKeysModal/RotateKeysModal.jsx
roshaans: Because of rerenders, the error message flashes without the user having enough time to see the error itself. This lets the error message persist for at least 3 seconds.

packages/frontend/src/components/wallet-migration/modals/RotateKeysModal/RotateKeysModal.jsx
roshaans: Yes good idea, I just copied it from a previous modal so probably worth doing a full refactor on this point for all the modals in the flow. I created an issue here -> https://github.com/near/near-wallet/issues/2959

roshaans:
> Great job @roshaans! Is there a way for me to test this flow in the UI?

Yes absolutely! Simply run this in `packages/frontend/` -> `NEAR_WALLET_ENV='development' && yarn run start`

hcho112 (approved):
LGTM, I'm happy to take over the remaining [issues](https://github.com/near/near-wallet/issues/2959)

packages/frontend/src/components/accounts/batch_import_accounts/sequentialAccountImportReducer.js
```diff
@@ -22,7 +22,10 @@ export const ACTIONS = {
     REMOVE_ACCOUNTS: 'REMOVE_ACCOUNTS',
     ADD_ACCOUNTS: 'ADD_ACCOUNTS',
     SET_CURRENT_FAILED_AND_END_PROCESS: 'SET_CURRENT_FAILED_AND_END_PROCESS',
-    RESTART_PROCESS: 'RESTART_PROCESS'
+    RESTART_PROCESS_FROM_ACCOUNT: 'RESTART_PROCESS_FROM_ACCOUNT',
+    RESTART_PROCESS: 'RESTART_PROCESS',
```
roshaans, line 26: comment to self, these don't get used

PR #2955, opened 2022-10-25
fix: only show Finer Wallet if device is iOS

hcho112:
During wallet migration, only show `Finer Wallet` from migration list if the device is iOS.
[They](https://finerwallet.io/) only support iOS apps at the moment.

This is to avoid any android (and window 😄 ) phone users accidentally selecting  Finer Wallet.


Patrick1904 (approved):
Clean fix ✌️

PR #2958, opened 2022-10-27
feat: cleanup keys migration modal

roshaans:
This PR adds the key cleanup functionality which (at the moment) occurs after the migration. The user will be prompted to select which keys to remove (defaults to all). Next they confirm they still have access to the rotated key (or other key not marked for deletion) with validation to ensure an approved key is provided.

Note that upon deleting the access key used by the wallet to sign, that key will be removed from localstorage as it is no longer on chain.

https://pagodaplatform.atlassian.net/browse/WEP-185


hcho112:
so I actually executed the migration and looks like there are still keys:
`near keys dannytest29.testnet` returns list of FAKs. (https://explorer.testnet.near.org/accounts/dannytest29.testnet)
is it mean to delete keys?

packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/AccessKeyList.jsx
```diff
@@ -0,0 +1,127 @@
...
+            flex: 1;
+            padding: 16px 0 0 0;
+        }
+    }
+`;
+
+const AccessKeyList = ({ account, onClose, onNext, selectKey, selectedKeys }) => {
+    const [expanded, setExpanded] = useState(true);
+    return (
+        <AccessKeyListContainer>
```
hcho112, line 67: we should use `Container` component from `CommonComponents` instead
(it has common style props, eg `title` class name is applied on h3 but it is not applied properly because it's expecting `Container` component)

packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/AccountKeyCleanup.jsx
```diff
@@ -0,0 +1,97 @@
...
+const AccountKeyCleanup = ({
+    accounts,
+    batchStarted,
+    offerRetry,
+    onClose,
+    onNext,
+    onRetry,
+}) => {
+    return (
+        <AccountKeyCleanupContainer>
```
hcho112, line 48: Same as above

packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/CleanKeysModal.jsx
```diff
@@ -1,53 +1,255 @@
...
+
+            removeKeysForCurrentAccount();
+        }
+    }, [currentAccount]);
 
     return (
         <MigrationModal>
             <Container>
                 <IconBackground>
                     <IconLogout />
```
hcho112, line 181: Need to use different icon. (currently it's logout but it should be key looking icon

packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/AccessKeyList.jsx
```diff
@@ -0,0 +1,127 @@
...
+            <div
+                className={classNames(['account', expanded ? 'open' : ''])}
+                id='full-access-keys'
+                onClick={() => setExpanded(!expanded)}
+            >
+                <div className='account-detail'>
+                    <p className='account-id'>{account.accountId}</p>
+                    <p className='keys-to-remove'>
+                        {account.accessKeys.length}/{account.totalAccessKeys}
+                        <Translate id='walletMigration.cleanKeys.fullAccessKeys' />
```
hcho112, line 80: Something like this?

```
{`${account.accessKeys.length}/${account.totalAccessKeys}`}
{` `}
<Translate id='walletMigration.cleanKeys.fullAccessKeys' />
```

packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/AccessKeyList.jsx
```diff
@@ -0,0 +1,127 @@
...
+                </StyledButton>
+                <StyledButton
+                    className='gray-blue'
+                    onClick={onClose}
+                    fullWidth
+                >
+                    <Translate id='button.cancel' />
+                </StyledButton>
+            </ButtonsContainer>
+        </AccessKeyListContainer>
```
hcho112, line 123: image
Some weird css issue here

packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/AccessKeyList.jsx
```diff
@@ -0,0 +1,127 @@
...
+            </h3>
+            <div
+                className={classNames(['account', expanded ? 'open' : ''])}
+                id='full-access-keys'
+                onClick={() => setExpanded(!expanded)}
+            >
+                <div className='account-detail'>
+                    <p className='account-id'>{account.accountId}</p>
+                    <p className='keys-to-remove'>
+                        {account.accessKeys.length}/{account.totalAccessKeys}
```
hcho112, line 79: when I untick some accounts, this number doesn't get updated:

image

packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/AccessKeyList.jsx
```diff
@@ -0,0 +1,127 @@
...
+                    <div className='access-key' key={publicKey}>
+                        <div className='public-key'>
+                            <span className='key-prefix'>
+                                {publicKey.replace('ed25519:', '').slice(0, 16)}
+                            </span>
+                            &nbsp;
+                            (<Translate id={`walletMigration.cleanKeys.keyTypes.${kind}`} />)
+                        </div>
+                        <div className='remove-checkbox' onClick={() => selectKey(publicKey)}>
+                            <Checkbox checked={selectedKeys[publicKey]} />
```
hcho112, line 101: recovery phrase type key should be unchecked and disabled:

packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/CleanKeysModal.jsx
```diff
@@ -1,53 +1,255 @@
...
+    });
+
+    const [loadingAccounts, setLoadingAccounts] = useState(true);
+    const [keysToRemove, setKeysToRemove] = useState({});
+    const dispatch = useDispatch();
+    const initialAccountIdOnStart = useSelector(selectAccountId);
+    const initialAccountId = useRef(initialAccountIdOnStart);
+    const [showConfirmCleanupModal, setShowConfirmCleanupModal] = useState(false);
+    const [showConfirmSeedphraseModal, setShowConfirmSeedphraseModal] = useState(false);
+    const [/*finishingSetupForCurrentAccount*/, setFinishingSetupForCurrentAccount] = useState(false);
```
hcho112, line 89: what is this for? why is it commented?

packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/CleanKeysModal.jsx
```diff
@@ -1,53 +1,255 @@
...
+    }, [rotatedKeys]);
+
+    const currentAccount = useMemo(
+        () =>  state.accounts.find(({ status }) => status === IMPORT_STATUS.PENDING),
+        [ state.accounts]
+    );
+    const currentFailedAccount = useMemo(
+        () => state.accounts.find(({ status }) => status === IMPORT_STATUS.FAILED)
+            && state.accounts.every(({ status }) => status !== IMPORT_STATUS.PENDING),
+        [ state.accounts]
```
hcho112, line 134: remove empty space

packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/CleanKeysModal.jsx
```diff
@@ -1,53 +1,255 @@
...
-                </ButtonsContainer>
+
+                {showConfirmSeedphraseModal && (
+                    <ConfirmKeyDeletion
+                        accountId={currentAccount.accountId}
+                        fakPublicKeys={currentAccount.accessKeys}
+                        onClose = {async () => {
+                            localDispatch({ type: ACTIONS.SET_CURRENT_FAILED_AND_END_PROCESS });
+                            setShowConfirmSeedphraseModal(false);
+                        }}
```
hcho112, line 191: let pull out this into a seperate const above

packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/CleanKeysModal.jsx
```diff
@@ -1,53 +1,255 @@
...
+                                await deleteKeys({
+                                    accountId: currentAccount.accountId,
+                                    publicKeysToDelete: keysToRemove,
+                                    wallet,
+                                });
+                                dispatch(switchAccount({accountId: initialAccountId.current}));
+                            } finally {
+                                setFinishingSetupForCurrentAccount(false);
+                            }
+                        }}
```
hcho112, line 206: let pull out this into a seperate const above

packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/ConfirmKeyDeletion.jsx
```diff
@@ -0,0 +1,158 @@
...
+        ].includes(keyStatus));
+
+        setKeyMessage({
+            [KEY_STATUS.DOES_NOT_EXIST]: 'doesNotExist',
+            [KEY_STATUS.INSECURE_RECOVERY_METHOD]: 'insecureRecoveryMethod',
+            [KEY_STATUS.TO_BE_DELETED]: 'toBeDeleted',
+        }[keyStatus] || null);
+    }, [keyStatus]);
+
+    console.log({ keyStatus });
```
hcho112, line 110: 🔪

packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/AccessKeyList.jsx
andy-haynes: @hcho112 where is this requirement documented?

packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/CleanKeysModal.jsx
andy-haynes: It was preferable to keeping it in a `console.log` just to skirt the `eslint` rule about unreferenced variables. But since it never ended up being needed I can remove it.

packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/AccessKeyList.jsx
hcho112: @andy-haynes my apologies, it should have been updated on the ticket. (It was briefly mentioned in the comment section of the ticket as `- Newly created FAK from rotating keys step are now stored only in memory. (on WalletMigration.jsx, variable called rotatedKeys) So you may need to reference it on clean keys step to NOT delete this one. (Perhaps hide them from UI?)` but even this isn't clear enough.

I just came up with this requirement as that if we created a new FAK from rotating and aim to use this to pass down to other wallets, it make sense for us to make sure this is not deleted.

andy-haynes: Ah the `Recovery Phrase` is not the rotated key - it's the key the wallet has designated as its seed phrase. It's there if the user created the account with a seed phrase via wallet UI or if they added a seed phrase later on. We actually only know it's the `Recovery Phrase` because it's listed in DynamoDB as a recover method. A new FAK added from a BIP39 phrase via CLI or anywhere else outside the wallet UI would not qualify.

The rotated key I've opted not to show at all. I agreed with your comment around hiding it from the UI and implemented it accordingly, as its presence feels a bit confusing; i.e. I'm considering at the subset of _removable_ keys, not the set of keys on my account. I'm good either way though, just the decision I made in the absence of a definitive approach.

packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/AccessKeyList.jsx
hcho112: hmm I tried to solve this UI issue by myself but couldn't get it around with using `Accordion` component. (I feel that if we want to solve this, we may better to write UI newly)

Either way, I will sort this out later

hcho112 (approved):
LGTM! I have tested the whole flow and now it successfully deletes all keys except rotated key.

PR #2962, opened 2022-11-01
feat: batch key conversion when disabling 2FA

andy-haynes:
Users attempting to disable 2FA with > 48 function call access keys for multisig require multiple transactions to convert these keys to FAKs. This PR adds batch conversion logic to convert keys prior to attempting to disable 2FA to facilitate a seamless disable workflow for the handful (95 at time of writing) of accounts impacted by this. Users going through the batch conversion flow will need to approve one 2FA transaction for each batch of 50 keys to be converted, e.g. 49-99 keys requires one prompt for key conversion, 100-149 requires two, etc. The 2FA prompt will still be required for the final disable as usual.

Once the design is finalized, additional code will be added to the `Disable2FA` component to indicate to affected users that multiple prompts will be required to disable.

Note that this implementation was originally written for `near-api-js`, which has better support for the tests included in this PR. To port them over it was necessary to separate out the `TwoFactor` class to decouple the dependencies (yes, plural) with circular references. If this is too much for this PR I am open to leaving it for a different piece of work.

### DIFF
--- a/packages/frontend/src/utils/twoFactor/twoFactorBase.js
+++ b/packages/frontend/src/utils/twoFactor/twoFactorBase.js
@@ -2,10 +2,10 @@ import { BN } from 'bn.js';
 import * as nearApiJs from 'near-api-js';
 import { parseSeedPhrase } from 'near-seed-phrase';
 
-import { store } from '..';
-import { ACCOUNT_HELPER_URL, MULTISIG_CONTRACT_HASHES, MULTISIG_MIN_AMOUNT, NETWORK_ID, NODE_URL } from '../config';
-import { promptTwoFactor, refreshAccount } from '../redux/actions/account';
-import { WalletError } from './walletError';
+import { store } from '../..';
+import { ACCOUNT_HELPER_URL, MULTISIG_CONTRACT_HASHES, MULTISIG_MIN_AMOUNT, NETWORK_ID, NODE_URL } from '../../config';
+import { promptTwoFactor, refreshAccount } from '../../redux/actions/account';
+import { WalletError } from '../walletError';
 
 const {
     multisig: { Account2FA },
--- a/packages/frontend/src/utils/twoFactor/twoFactorBase.js
+++ b/packages/frontend/src/utils/twoFactor/twoFactorBase.js
@@ -2,9 +2,13 @@ import { BN } from 'bn.js';
 import * as nearApiJs from 'near-api-js';
 import { parseSeedPhrase } from 'near-seed-phrase';
 
-import { store } from '../..';
-import { ACCOUNT_HELPER_URL, MULTISIG_CONTRACT_HASHES, MULTISIG_MIN_AMOUNT, NETWORK_ID, NODE_URL } from '../../config';
-import { promptTwoFactor, refreshAccount } from '../../redux/actions/account';
+import {
+    ACCOUNT_HELPER_URL,
+    MULTISIG_CONTRACT_HASHES,
+    MULTISIG_MIN_AMOUNT,
+    NETWORK_ID,
+    NODE_URL,
+} from '../../config';
 import { WalletError } from '../walletError';
 
 const {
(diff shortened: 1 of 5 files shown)


packages/frontend/src/utils/twoFactor/twoFactorBase.js
```diff
@@ -0,0 +1,214 @@
+import { BN } from 'bn.js';
```
andy-haynes, line 1: This file was moved and renamed, the history is still here despite the PR showing it as new.

packages/frontend/src/utils/twoFactor/twoFactorBase.js
```diff
@@ -0,0 +1,214 @@
...
+    }) {
+        super(connection, accountId, {
+            storage: localStorage,
+            helperUrl: ACCOUNT_HELPER_URL,
+            getCode,
+        });
+        this.has2fa = has2fa;
+        this.init2fa = init2fa;
+        this.isLedgerEnabled = isLedgerEnabled;
+        this.refreshAccount = refreshAccount;
```
andy-haynes, line 43: Aside from the new methods (the last three declared in the class), the only changes to this file are where these new methods (well, function parameters) are called. This decouples the class from the redux side effects and DOM global, making the class actually testable.

packages/frontend/ci/configFromEnvironment.js
```diff
@@ -30,7 +30,7 @@ module.exports = {
     SENTRY_RELEASE:
       process.env.SENTRY_RELEASE ||
       (parseBooleanFromShell(process.env.RENDER) &&
-          `render:${process.env.RENDER_SERVICE_NAME}:${process.env.RENDER_GIT_BRANCH}:${process.env.RENDER_GIT_COMMIT}`),
+          `render:${process.env.RENDER_SERVICE_NAME}:${process.env.RENDER_GIT_BRANCH.replace(/\//g, '-')}:${process.env.RENDER_GIT_COMMIT}`),
```
andy-haynes, line 33: The `/` character is invalid in Sentry identifiers, causing Render `near-wallet` deployments to fail. Only affects `testnet` PR previews.

packages/frontend/src/utils/twoFactor/twoFactorBase.js
```diff
@@ -0,0 +1,214 @@
...
+            helperUrl: ACCOUNT_HELPER_URL,
+        });
+
+        return await account.disableWithFAK({ contractBytes: emptyContractBytes, cleanupContractBytes });
+    }
+
+    /**
+     * Returns the list of multisig function call access keys
+     */
+    async get2faLimitedAccessKeys() {
```
andy-haynes, line 142: This logic is from the `Account2FA` class in `near-api-js`, duplicated here to facilitate this short-term functionality.

packages/frontend/src/utils/twoFactor/twoFactorBase.js
```diff
@@ -0,0 +1,214 @@
...
+                    conversionActions.push(deleteKey(nearApiJs.utils.PublicKey.from(public_key)));
+                    conversionActions.push(addKey(nearApiJs.utils.PublicKey.from(public_key), fullAccessKey()));
+
+                    return conversionActions;
+                }, []);
+
+            await this.signAndSendTransaction({
+                receiverId: this.accountId,
+                actions: conversionActions
+            });
```
hcho112, line 207: I have tried to ran the code that has many FAKs with function call. It worked smoothly but it also took some time to get things cleared. So for UX improvement, wonder if we should show some form of a notification on right top corner showing:
50/200 Access keys are converted..
100/200 Access keys are converted..
150/200 Access keys are converted..
on when above `await this.sighANdSendTransaction` function call completes.

hcho112: Also in case RPC is down or due to some limitation, if `await this.signAndSendTransaction` call fails, is there a way to display error message and stop the while loop?

hcho112 (approved):
This PR is purely awesome! And will definitely save many users who wants to disable 2FA. Great work @andy-haynes 🙏

packages/frontend/src/utils/twoFactor/twoFactorBase.js
andy-haynes: > So for UX improvement, wonder if we should show some form of a notification on right top corner showing:

It's certainly an option for the design. We can track it on the instance easily enough without needing event emitters. However it's a one-time use feature for < 40 users who require more than one prompt so it'll come down to a matter of effort.

> Also in case RPC is down or due to some limitation, if `await this.signAndSendTransaction` call fails, is there a way to display error message and stop the while loop?

Once the retry/backoff logic reaches its limit, an exception would be thrown and surface to the user as usual. Next time they resume the disable workflow it would continue the remaining work.

hcho112: Thanks for clarification!

PR #2965, opened 2022-11-02
feat: batch key conversion in disable 2FA modal

andy-haynes:
This PR changes the components currently responsible for disabling 2FA so that they have awareness around whether a 2FA-enabled account requires that its keys are converted prior to attempting to disable 2FA. I'm not aware of a design for this so the exact wording and specifics on how the user is informed of this conversion (i.e. that they'll need to confirm multiple 2FA modals) is subject to change.

Disabling from `Disable 2FA` button:

### DIFF
--- a/packages/frontend/src/translations/en.global.json
+++ b/packages/frontend/src/translations/en.global.json
@@ -1678,6 +1678,7 @@
             "desc": "Keep in mind that transactions won't have to be confirmed with 2FA once it's disabled.",
             "disable": "Disable 2FA",
             "keep": "No, keep 2FA",
+            "keyConversionRequired": "Due to the number of 2FA access keys on this account, some access keys will need to be converted to full access keys before 2FA can be disabled. The converted keys can sign transactions without a 2FA prompt, so it is important to complete the 2FA disable process once begun. You will receive one or more additional 2FA prompts to perform this key conversion when disabling 2FA.",
             "phoneDesc": "The SMS option for two-factor authentication is being deprecated, and cannot be re-enbaled.",
             "title": "Are you sure you want to disable 2FA?"
         },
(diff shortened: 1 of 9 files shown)


hcho112 (approved):
LGTM, good work!

packages/frontend/src/translations/en.global.json
```diff
@@ -1678,6 +1678,7 @@
             "desc": "Keep in mind that transactions won't have to be confirmed with 2FA once it's disabled.",
             "disable": "Disable 2FA",
             "keep": "No, keep 2FA",
+            "keyConversionRequired": "<br />Due to the number of 2FA access keys on this account, some access keys will need to be converted to full access keys before 2FA can be disabled. The converted keys can sign transactions without a 2FA prompt, so it is important to complete the 2FA disable process once begun. You will receive one or more additional 2FA prompts to perform this key conversion when disabling 2FA.",
```
hcho112, line 1681: Just small changes here, adding `<br><br>` in front of `You will receive one or more ...` and warp the last sentence with `<b></b>` maybe to elaborate a bit more?

Some users may not understand the details of this, but it is important to emphasize last sentence a bit more because they are likely to get multiple emails during the process.

hcho112 (approved):
LGTM!!

packages/frontend/src/translations/en.global.json
andy-haynes: That's much better, thanks @hcho112 !

This text could probably use some touching up, this was just some placeholder text I put down. Basically we want to convey:
- You have a lot of keys.
- You need to approve more than one 2FA transaction in order to disable 2FA.
- Once you've approved your first transaction, there are now 50 FAKs that can sign transactions without a 2FA prompt (this only applies to transactions outside of `wallet.near.org` as wallet will continue to use the multisig LAK until 2FA is disabled. If you abandon the process after you start, then you've circumvented the security guarantees made by 2FA/multisig. For users exclusively on `wallet.near.org` this incomplete access key state doesn't really weaken security on the account, it just breaks some assumptions.

hcho112 (approved):
LGTM! confirmed the change is on the branch!

image

PR #2968, opened 2022-11-08
Wallet Migration UI minor improvements [WEP-175]

hcho112:
This PR contains improvements/bug fixes for wallet migration. 
1. Bug fix on situation where there is only 1 ledger account to be migrated (previously errored out)
2. Wallet list randomizing
3. Refactor/Improvements on going 2FA disable -> Rotate keys step where previously it was fetching same account info twice, causing un-necessary loading UI showing twice.


Patrick1904 (approved):
LGTM

PR #2970, opened 2022-11-09
feat: migration to welldone-wallet

Yoon-Suji:
* Add WELLDONE Wallet to near migration option
* You can batch-import your accounts to our WELLDONE Wallet in our docs page.


hcho112:
Thank you for submitting PR @Yoon-Suji /WELLDONESTUDIO

I have ran the code locally with chrome incognito mode. My first impression is that landing page is amazing/new user friendly in a sense that it captures if user doesn't have welldone studio wallet installed on chrome extension, it guides users nicely. 👍 
After I have installed them (also make it available on incognito), I have faced these issues:

1. Ambiguity after clicking `Connect Wallet` button. 
After clicking the button, it opens a popup window as below:
image
At this point it's not so clear that if I need to click `create wallet` button or `import wallet` button. Because I was migrating from near-wallet, user may think that they are `importing wallet`  so they could click import button instead. (But I believe it is not what it is intended?)

image
So after I have clicked import wallet button and realise it is not the right place, I was easily able to go back and clicked create new wallet flow. After I have insert everything, it froze and nothing was happening after. I was trying to investigate what's going on but all I can see from inspector of the popup was:

image

So overall, these issues need to be fixed.

Yoon-Suji:
Hi @hcho112 , thanks for your review!

We found the cause of the problem in the wallet. We will deploy a new version of the wallet today after fixing that issue. You can test tomorrow with new version of wallet(It takes some time for Chrome Webstore to be updated). 

Also, thank you for your kind UI feedback. We will add a more detailed guide about ambiguous naming(`Import Wallet`) to the page!

packages/frontend/src/components/wallet-migration/modals/MigrateAccountsModal/SelectDestinationWallet.jsx
```diff
@@ -79,6 +80,13 @@ export const WALLET_OPTIONS = shuffle([
         getUrl: ({ hash }) => `finer://wallet.near.org/batch-import#${hash}`,
         checkAvailability: () => isMobile('iOS'),
     },
+    {
+        id: 'welldone-wallet',
+        name: 'WELLDONE Wallet',
+        icon: <img src={ImgWelldoneWallet} alt="WELLDONE Wallet Logo" />,
+        getUrl: ({ hash }) => `https://docs.welldonestudio.io/contribution/202211-batch-import#${hash}`,
+        checkAvailability: () => true,
```
hcho112, line 88: one last request, is it possible to add `ledgerSupport: false,` after line 88?
This is new boolean we introduced, to indicate ledger compatibility

Yoon-Suji: ok, I added it! please check it.

hcho112 (approved):
LGTM!

PR #2971, opened 2022-11-10
Wallet Migration UI/flow enhancement [WEB-175]

hcho112:
This PR contains improvements on wallet migration:

1. Ability to auto save on important steps during migration.
Save points are:
- when user reach to transfer step (moving to other wallet)
- when user completes key clean up step (at this point there is no turning back)
- when user click start over button on verify step (This will remove save point and start from beginning)

2. Show tailored UI/view on third party wallet selection upon transferring ledger account(s)
A. When there is no ledger account, show everything
B. When there is only ledger account, show categorized view, ledger support, unsupported group
image

C. When there is mixture of ledger account(s) and FAK account(s), for the wallets that do not support ledger accounts, it will show info icon indicating that selecting unsupported wallet will exclude ledger accounts being transferred. (only FAK accounts will)
image

### DIFF
--- a/features/flags.json
+++ b/features/flags.json
@@ -267,9 +267,9 @@
     "createdBy": "roshaans",
     "createdAt": "2022-09-29T12:47:40.121Z",
     "development": {
-      "enabled": false,
-      "lastEditedBy": "roshaans",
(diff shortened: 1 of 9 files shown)


Patrick1904 (approved):
Great job @hcho112 👍

PR #2979, opened 2022-11-21
feat: Disable new account creation in NW

Patrick1904:
This disables new account creation in NEAR wallet when feature flag `WEP_PHASE_ONE` is enabled. Whenever a user tries to create a new account, the wallet shows a modal informing the user that new accounts should be created in an alternative wallet going forward. The user then clicks 'More NEAR Wallets' which triggers the `wallet-selector` and displays available wallets.

More info here: https://pagodaplatform.atlassian.net/browse/WEP-204

https://user-images.githubusercontent.com/24921205/203175391-c28c3b9d-efb4-4f54-9ec2-9896d5c8e3e4.mov


packages/frontend/src/components/common/modal/Style.css.js
```diff
@@ -63,7 +64,7 @@ const Style = styled.div`
         z-index: 1050;
         width: 100%;
         background-color: white;
-        border-radius: 6px;
+        border-radius: 16px;
```
hcho112, line 67: Not a dramatic change but this change would affect other modal component's border radius. Have you checked on other modal UIs?

packages/frontend/src/components/common/wallet_selector/WalletSelectorGetAWallet.js
```diff
@@ -0,0 +1,75 @@
...
+}) {
+
+    const walletWrapperDiv = document.getElementById('near-wallet-selector-modal');
+    walletWrapperDiv?.classList.add('wallet-selector-get-a-wallet-modal');
+
+    useEffect(() => {
+
+        const initSelector = async () => {
+            const selector = await setupWalletSelector({
+                network: 'testnet',
```
hcho112, line 36: maybe replace this with referencing to real `networkId`?

packages/frontend/src/components/common/wallet_selector/WalletSelectorGetAWallet.js
```diff
@@ -0,0 +1,75 @@
...
+                    setupWelldoneWallet(),
+                    setupLedger(),
+                    setupNearFi(),
+                    setupCoin98Wallet(),
+                    setupOptoWallet()
+                ],
+            });
+
+            setWalletSelectorModal(setupModal(selector, {
+                contractId: 'test.testnet'
```
hcho112, line 54: So if we go production with this code, is it ok to use `test.testnet`? or do we need to change into something else?

packages/frontend/src/components/common/wallet_selector/WalletSelectorGetAWallet.js
Patrick1904: Good catch, this should definitely be using the correct network since the modal will link to mainnet/testnet versions depending on what network is provided. Will update 👍

packages/frontend/src/components/common/wallet_selector/WalletSelectorGetAWallet.js
Patrick1904: Since this version of the selector will never be used to sign in with a wallet, I think it's fine for it to be testnet, or what do you think?

packages/frontend/src/components/common/modal/Style.css.js
Patrick1904: Yep I changed it globally since the Figma design is using 16px. In general we should make these changes globally to keep things consistent. UI design is constantly evolving so whenever we notice an inconstancy between a new design and the current UI let's confirm the change and then make it global, especially for common elements like modal, buttons, etc.

packages/frontend/src/components/common/wallet_selector/WalletSelectorGetAWallet.js
hcho112: hmm if it is mandatory to have `contractId` filled but does not care about validity of the account, maybe we can use text like `dontcare` or have a comment above saying it's for placeholder? (Just to avoid anyone in the future)

hcho112 (approved):
LGTM

hcho112 (approved):
LGTM

PR #2985, opened 2022-12-07
feat: FT registration

andy-haynes:
This PR modifies the existing `FungibleTokenService` logic to accommodate FTs implementing registration requirements outlined in NEP-145. These changes allow the wallet to transfer USDC by registering the recipient with the contract when registration is required (determined by calling the `check_registration` method).

FTs requiring registration are currently assumed to not require storage deposits, though only in terms of calculating estimated fees. We may want to change this in the future if token contracts would implement both.


Patrick1904 (approved):
Thanks @andy-haynes. Works great! 💯

Patrick1904:
@andy-haynes should we enable the flag on `testnet` so that we can easily test it until we enable it on `mainnet`?

andy-haynes:
> @andy-haynes should we enable the flag on `testnet` so that we can easily test it until we enable it on `mainnet`?

Definitely, just wanted to land this and quickly verify there are no issues with flag-off FT transfers. I'll have the PR to enable this in `testnet` momentarily 🙂

PR #2991, opened 2023-01-06
fix: remove outer FunctionCallError in NFT metadata error matching

andy-haynes:
This PR removes the outer `FunctionCallError(...)` string used to match error "missing method" responses from RPC. Currently mainnet errors do include this value but testnet RPC error responses do not, preventing NFTs from rendering.


Patrick1904 (approved):
LGTM! Thanks for the swift fix 👍

PR #2996, opened 2023-01-23
feat: change the steps order on wallet migration

hcho112:
This PR contains changes for wallet migration. 

A. Change the order of wallet migration steps to following:

1. Disable 2FA (if any)
2. Rotate Key
3. Delete FAKs
4. Export Account(s) to third party wallet (To be replaced with wallet-selector)
5. Logout

Intention of this change is to make sure we clean up FAKs before moving to third party wallet. (Also this is only way to make sure we don't accidentally delete any FAK(s) created by third party wallet)

B. Ability to come back and continue where user left

This is achieved by leaving current user's FAK that is used to log in on browser wallet from delete FAK(s) step. This key will be deleted on the final step `Logout`.

### DIFF
--- a/packages/frontend/src/components/wallet-migration/WalletMigration.jsx
+++ b/packages/frontend/src/components/wallet-migration/WalletMigration.jsx
@@ -1,3 +1,4 @@
+import * as nearApi from 'near-api-js';
 import React, { useCallback, useEffect, useState } from 'react';
 import { useSelector } from 'react-redux';
 
@@ -37,7 +38,7 @@ const WalletMigration = ({ open, onClose }) => {
     const availableAccounts = useSelector(selectAvailableAccounts);
     const [loadingMultisigAccounts, setLoadingMultisigAccounts] = useState(true);
     const [accountWithDetails, setAccountWithDetails] = useState([]);
-
+    const [isLoggingOut, setIsLoggingOut] = useState(false);
     useEffect(() => {
         const importRotatableAccounts = async () => {
             const accounts = await wallet.keyStore.getAccounts(NETWORK_ID);
@@ -101,12 +102,20 @@ const WalletMigration = ({ open, onClose }) => {
     };
 
     const onLogout = () => {
-        // TODO: Add logic to remove FAK(s) here so everything gets cleared together
-        return Promise.all(availableAccounts.map((accountId) => wallet.removeWalletAccount(accountId)))
+        setIsLoggingOut(true);
+        return Promise.all(availableAccounts.map(async (accountId) => {
+            const account = await wallet.getAccount(accountId);
+            const keyPair = await wallet.keyStore.getKey(wallet.connection.networkId, accountId);
+            await account.signAndSendTransaction({
+                actions: [nearApi.transactions.deleteKey(keyPair.publicKey)],
+                receiverId: accountId,
+            });
+            return wallet.removeWalletAccount(accountId);
+        }))
             .then(() => {
-                location.reload();
+                setIsLoggingOut(false);
                 deleteMigrationStep();
-                onClose();
+                location.reload();
             });
     };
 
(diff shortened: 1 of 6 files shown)


Patrick1904:
@hcho112 This feels like an improved UX. A part from recovery keys and the key being exported, are there any additional FAKs left on the users account after making this change?

hcho112:
@Patrick1904 On current implementation no.

But there was a discussion that we should also exclude FAK key that user logged in with. So durning key clean up, we exclude logged in FAK and rotated key. Then, on the last step where user logs out, logged in FAK is also deleted. 

In this approach, it will allow user to come back (say if it user disconnected during the migration or away from keyboard) safely and able to continue the process all the way until the end. 

I will implement this feature in this PR as well.

packages/frontend/src/components/wallet-migration/WalletMigration.jsx
```diff
@@ -91,22 +92,30 @@ const WalletMigration = ({ open, onClose }) => {
...
-        // TODO: Add logic to remove FAK(s) here so everything gets cleared together
-        return Promise.all(availableAccounts.map((accountId) => wallet.removeWalletAccount(accountId)))
+        setIsLoggingOut(true);
+        return Promise.all(availableAccounts.map(async (accountId) => {
+            const account = await wallet.getAccount(accountId);
+            const keyPair = await wallet.keyStore.getKey(wallet.connection.networkId, accountId);
+            await account.signAndSendTransaction({
+                actions: [nearApi.transactions.deleteKey(keyPair.publicKey)],
+                receiverId: accountId,
+            });
```
andy-haynes, line 112: We have some shorthand around this already that I think would be better to use here:
```js
            const publicKey = await wallet.getPublicKey(accountId);
            await account.deleteKey(publicKey);
```

packages/frontend/src/components/wallet-migration/WalletMigration.jsx
```diff
@@ -91,22 +92,30 @@ const WalletMigration = ({ open, onClose }) => {
...
 
+    const navigateToMigrateAccounts = () => {
+        handleSetActiveView(WALLET_MIGRATION_VIEWS.MIGRATE_ACCOUNTS);
+    };
+
     const onLogout = () => {
-        // TODO: Add logic to remove FAK(s) here so everything gets cleared together
-        return Promise.all(availableAccounts.map((accountId) => wallet.removeWalletAccount(accountId)))
+        setIsLoggingOut(true);
+        return Promise.all(availableAccounts.map(async (accountId) => {
```
andy-haynes, line 106: If an account fails in here, does the UI break (i.e. does `isLoggingOut` remain `true`)? I'm wondering if a `.reflectAll` might be better here to aggregate failed removals - at least displaying an error for the accounts that failed.

Realistically I think this only fails during RPC connection issues so it's not super pressing for this PR IMO.

Patrick1904 (approved):
Great job!

packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/CleanKeysModal.jsx
```diff
@@ -33,9 +33,11 @@ async function getAccountDetails({ accountId, publicKeyBlacklist, wallet }) {
         }, {});
 
     const allAccessKeys = await wallet.getAccessKeys(accountId);
+    const currentAccount = await wallet.getLocalAccessKey(accountId, allAccessKeys);
```
andy-haynes, line 36: Can we use `wallet.getPublicKey(accountId)` here instead? It's more consistent with how we're getting the key elsewhere and is equivalent in this case.

The usage here is a little circuitous since `getLocalAccessKey` only returns a value if it's in `allAccessKeys`, but we only want to get the key to filter it out of `allAccessKeys`.

andy-haynes (changes requested):

PR #2997, opened 2023-01-24
fix: verify actions exist

andy-haynes:
This PR updates the code guard in the 2FA modal to match the expression being evaluated. This prevents the UI from breaking when the call to `get_request` fails.


Patrick1904 (approved):
Great fix!

PR #2999, opened 2023-01-26
Minor Syntax improvements on wallet migration

hcho112:
This PR contains minor syntax improvements that were commented on:
[…logout step](https://github.com/near/near-wallet/pull/2996)

### DIFF
--- a/packages/frontend/src/components/wallet-migration/WalletMigration.jsx
+++ b/packages/frontend/src/components/wallet-migration/WalletMigration.jsx
@@ -102,34 +102,31 @@ const WalletMigration = ({ open, onClose }) => {
         handleSetActiveView(WALLET_MIGRATION_VIEWS.MIGRATE_ACCOUNTS);
     };
 
-    const onLogout = () => {
+    const onLogout = async () => {
         setIsLoggingOut(true);
-       
-        return Promise.allSettled(availableAccounts.map(async (accountId) => {
-            const account = await wallet.getAccount(accountId);
+        const failedAccounts = [];
+        for (const accountId of availableAccounts) {
             const publicKey = await wallet.getPublicKey(accountId);
-            await account.deleteKey(publicKey);
-            return wallet.removeWalletAccount(accountId);
-        }))
-            .then(async (results) => {
-                setIsLoggingOut(false);
-                deleteMigrationStep();
-                if (results.some((result) => result.status === 'rejected')) {
-                    const failedKeys = await results.reduce(async (prev, current, index) => {
-                        if (current.status === 'rejected') {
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/wallet-migration/WalletMigration.jsx
```diff
@@ -103,18 +104,30 @@ const WalletMigration = ({ open, onClose }) => {
 
     const onLogout = () => {
         setIsLoggingOut(true);
-        return Promise.all(availableAccounts.map(async (accountId) => {
+       
+        return Promise.allSettled(availableAccounts.map(async (accountId) => {
```
andy-haynes, line 108: Seeing this laid out I think this is probably clearer to make `onLogout` an `async` method and implement the key deletes in a `for` loop, i.e.

```js
const failedAccounts = [];
for (const accountId of availableAccounts) {
  const publicKey = await wallet.getPublicKey(accountId);
  try {
    const account = await wallet.getAccount(accountId);
    await account.deleteKey(publicKey);
  } catch {
    failedAccounts.push(accountId);
  }
  // do we still want to delete the signer if there was an error?
  // if they don't delete it here and now then when would they?
  await wallet.removeWalletAccount(accountId);
}
...
if (failedAccounts.length) {
  dispatch(showCustomAlert({
    success: false,
    messageCodeHeader: 'error',
    errorMessage: `fail to delete keys for account(s) ${failedAccounts.join(', ')}`,
  }));
}
```

From here I'm not sure what we actually want to _do_ with these, if it really is RPC timeout then you'd probably want to retry before deleting your keystore. I don't think it needs to be addressed for the first cut, we can see if it comes up in the wild.

andy-haynes (changes requested):
Looks good, I think we can simplify it a bit and only show the accounts that failed then we're good to go.

packages/frontend/src/components/wallet-migration/WalletMigration.jsx
hcho112: Totally make sense! much better. 

In terms of deleting the key from the localStorage, I have shift the `await wallet.removeWalletAccount(accountId);` to only delete if it doesn't fail.

andy-haynes (approved):
Great work!

packages/frontend/src/components/wallet-migration/WalletMigration.jsx
andy-haynes: Good call, keeping their account active will let them know there's still some cleanup to do. Probably not something we'll see come up often but if it does at least it won't interfere with cleanup of other accounts.

PR #3005, opened 2023-02-09
feat: enable WEP_PHASE_ONE in testnet

andy-haynes:
Enables recovery method/key cleanup and account export on wallet.testnet.near.org.


hcho112 (approved):
LGTM

PR #3007, opened 2023-02-13
feat: migration banner URL

andy-haynes:
This PR updates the migration banner to include the current URL and keep account creation guarded behind a new flag (to be kept enabled on testnet for ease of testing). It also re-enables `WEP_PHASE_ONE` on testnet, allowing users to export accounts to available wallets.

### DIFF
--- a/packages/frontend/src/translations/en.global.json
+++ b/packages/frontend/src/translations/en.global.json
@@ -791,6 +791,7 @@
     },
     "migration": {
         "message": "Soon NEAR wallet (${walletUrl}) will no longer be supported. Transfer your accounts to a different wallet or export your private keys from <a href='/profile'>account settings.</a>",
+        "readMore": "Please see our <a href='https://near.org/blog/near-opens-the-door-to-more-wallets/' target='_blank' rel='noopener noreferrer'>blog post</a> for more details on this upcoming transition.",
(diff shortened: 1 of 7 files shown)


andy-haynes:
Given the initial reaction to seeing the banner in testnet, I'd like to include a link to extra resources offering more context on why we're asking users to migrate. This will reduce the number of support requests around the nature of the change.

This post seems appropriate [NEAR Wallet’s Evolution in 2023](https://near.org/blog/near-wallets-evolution-in-2023/), but I am open to suggestions.

packages/frontend/src/translations/en.global.json
```diff
@@ -791,6 +791,7 @@
     },
     "migration": {
         "message": "Soon NEAR wallet (${walletUrl}) will no longer be supported. Transfer your accounts to a different wallet or export your private keys from <a href='/profile'>account settings.</a>",
+        "readMore": "Please see our <a href='https://near.org/blog/near-opens-the-door-to-more-wallets/' target='_blank' rel='noopener noreferrer'>blog post</a> for more details on this upcoming transition.",
```
Patrick1904, line 794: ❤️

Patrick1904 (approved):
LGTM!

PR #3012, opened 2023-02-24
fix: linkdrop connection no call to getAccessKeys

mattlockyer:
ON FIRE

Linkdrops take excess of 30-40s sometimes over 1 minute to claim.

Old connection in wallet (checking for ledger on ALL accessKeys) was making 3 (yes 3) requests to getAccessKeys every time a linkdrop functionCall was being signed.

See fix for details, solution was to create basic connection that bypassed the custom signer in `wallet.ts` and use that ONLY for linkdrops.

### DIFF
--- a/packages/frontend/src/utils/wallet.js
+++ b/packages/frontend/src/utils/wallet.js
@@ -21,6 +21,8 @@ import { setAccountConfirmed, setWalletAccounts, removeActiveAccount, removeAcco
 import { TwoFactor } from './twoFactor';
 import { WalletError } from './walletError';
 
+const { Account } = nearApiJs;
+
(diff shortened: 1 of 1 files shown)


Patrick1904 (approved):
@mattlockyer thanks for the fix. All linkdrops coming to testnet (wallet.testnet.near.org) currently redirect to MNW because of the enabled wallet migration flow (flag `WEP_PHASE_ONE`). The same fix has to be made in the MNW repo, or one could disable the wallet migration flow for linkdrops in this repo for this to work.

andy-haynes:
> The same fix has to be made in the MNW repo, or one could disable the wallet migration flow for linkdrops in this repo for this to work.

@Patrick1904 @mattlockyer I'll get this cherry-picked into MNW 👍

PR #3014, opened 2023-02-27
Added Herewallet as part of account export list

hcho112:
This PR contains minor changes to support Herewallet to be listed on mainnet ready list. 

I have tested both testnet and mainnet environment locally and they work just fine.


Patrick1904 (approved):
LGTM

PR #3016, opened 2023-03-02
Fix: filter out corrupted accounts and continue wallet-migration

hcho112:
This PR contains fix/improvements for wallet-migration from having corrupted accounts included in the list. 

As part of wallet-migration, it attempts to fetch extra detail for each available accounts within the browser prior to start of the migration. Previously if any one of account is corrupted (invalid key pair), the process stuck on initial step and unable to proceed to next step.

To fix this bug, it uses `Promise.allSettled` to gracefully handle the failed accounts and filter them out so migration can continue proceed regardless of presence of corrupted account key pairs. 

`console.log` is left there in case to be used on debugging/helping purposes if end user encounters them.

image

### DIFF
--- a/packages/frontend/src/components/wallet-migration/WalletMigration.jsx
+++ b/packages/frontend/src/components/wallet-migration/WalletMigration.jsx
@@ -43,9 +43,8 @@ const WalletMigration = ({ open, onClose }) => {
     useEffect(() => {
         const importRotatableAccounts = async () => {
             const accounts = await wallet.keyStore.getAccounts(NETWORK_ID);
-            const details = await Promise.all(
-                accounts.map((accountId) => getAccountDetails({ accountId, wallet }))
-            );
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/wallet-migration/WalletMigration.jsx
```diff
@@ -43,9 +43,8 @@ const WalletMigration = ({ open, onClose }) => {
     useEffect(() => {
         const importRotatableAccounts = async () => {
             const accounts = await wallet.keyStore.getAccounts(NETWORK_ID);
-            const details = await Promise.all(
-                accounts.map((accountId) => getAccountDetails({ accountId, wallet }))
-            );
+            const details = await Promise.allSettled(accounts.map((accountId) => getAccountDetails({ accountId, wallet })))
+                .then((results) => results.filter(({ status }) => status === 'fulfilled').map(({ value }) => value));
```
Patrick1904, line 47: ❤️

Patrick1904 (approved):
LGTM!

PR #3017, opened 2023-03-06
fix: use accountWithDetails in CleanKeysModal

andy-haynes:
Update `CleanKeysModal` to use the sanitized output of `accountWithDetails`. Currently, deleted accounts in your set of available accounts will break during key rotation.


hcho112 (approved):
LGTM! good catch

hcho112 (approved):
LGTM Again, thank you for picking this up.

PR #3027, opened 2023-03-28
NODE_URL for development to be consistent with near-cli #3028

soonnear:
Configured NODE_URL on near-cli
- https://github.com/near/near-cli/blob/aff742dae27c7afce259142263c63a30a5275276/config.js#L66


soonnear:
Closed as re-opened a new one in https://github.com/near/near-wallet/pull/3029 in order to link the github issue as this is only part of the full fix. https://github.com/near/near-wallet/issues/3028

PR #3031, opened 2023-03-29
Bug fix on account export sending wrong private key

hcho112:
This PR contains a bug fix for account export where it was sending a private key from localStorage instead of newly created private key from key rotation step. 

Bug is fixed by sending a correct private key from `rotatedKeys` object. 

Conducted full end to end test to make sure that:
1. Correct private key is passed to other wallet
2. Upon logout, private key from localStorage gets deleted and rotated private key is still exist
3. After logout, attempt to transfer token from third party wallet that uses rotated private key to make sure it is available


andy-haynes (approved):
Nice catch!

hcho112:
shout out to @sunny0529 from WELLDONE Studio team for reporting this bug and assist on testing!

PR #3034, opened 2023-04-03
feat: preemptively delete 2FA keys, ignoring failures from too recent keys

andy-haynes:
This PR attempts to preemptively delete 2FA requests in the wallet prior to disabling 2FA. While this currently happens in `near-api-js` as part of the current disable 2FA workflow, the deletion there can be prematurely halted by attempts to delete multisig requests created too recently.


Patrick1904 (approved):
LGTM!

PR #3036, opened 2023-04-05
fix: reset to default

andy-haynes:
Partial revert of #3017 to remove debugging configuration that breaks only on wallet.near.org (staging unaffected).


trechriron (approved):
Simple change.

jberrytech (approved):

PR #3038, opened 2023-04-11
Added token cache reset interval of one week

amirsaran3:
A large amount of users are raising an issue that they can't see fungible tokens in their near wallet. Indexer and RPC endpoints are returning the tokens correctly, but the issue is still there. There is a big chance that it the problem with the near wallet caching not updating properly.

I added a cache reset for every week so that the cache is synced with the blockchain.

### DIFF
--- a/packages/frontend/src/services/indexer/cached-api.js
+++ b/packages/frontend/src/services/indexer/cached-api.js
@@ -14,6 +14,7 @@ export default {
             kind: LIKELY_NFT_KEY,
             updater: (timestamp) => api.listLikelyNfts(accountId, timestamp),
             timeoutNs: UPDATE_REQUEST_INTERVAL_NS,
+            resetMs: RESET_REQUEST_INTERVAL_MS,
         });
     },
     listLikelyTokens(accountId, timestamp) {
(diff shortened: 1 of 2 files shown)


MaximusHaximus (changes requested):
If I'm reading this correctly, this change will only reset the cache if it hasn't been updated for a week -- in the case of a poisoned cache, it may be updated regularly but still invalid -- we want to intermittently clear the cache _even if we've updated it regularly and recently_.  To accomplish this, we will need to set a separate timestamp to track the age of the cache since originally created, then check against _that_ value instead of the last _updated_ timestamp.

packages/frontend/src/services/indexer/cached-api.js
```diff
@@ -21,6 +22,7 @@ export default {
             kind: LIKELY_TOKENS_KEY,
             updater: (timestamp) => api.listLikelyTokens(accountId, timestamp),
             timeoutNs: UPDATE_REQUEST_INTERVAL_NS,
+            resetMs: RESET_REQUEST_INTERVAL_MS,
```
MaximusHaximus, line 25: Please add the same argument to `listLikelyNfts()` above; we can assume that since both caches use the same infrastructure/logic, either one could be inaccurate and need resetting.

MaximusHaximus (changes requested):
Just 2 changes, then this is ready to go.

AmmarHumackicSQA:
Hey @andy-haynes , would you mind checking this PR from before? Thanks!

PR #3040, opened 2023-04-17
Added ledger disclaimer banner with translations

amirsaran3:
Added alert banner when connecting ledger.
image
image


MaximusHaximus (approved):
:+1:

PR #3047, opened 2023-04-19
Updated ledger connection error text

amirsaran3:
Updated disclaimer on ledger connection error text.

image


MaximusHaximus (approved):
:+1:

PR #3052, opened 2023-05-02
fix: avoid relying on NETWORK_ID to reference correct network

hcho112:
This PR contains implementation for fixing the wallet migration wizard error. 

After reverting [this PR](https://github.com/near/near-wallet/pull/3036) , wallet migration wizard fail to operate properly as it was replying on `NETWORK_ID` to reflect the correct network. (eg. `testnet` or `mainnet`)

But revert needed to happen to fix another bug on wallet so wallet wizard also needed to make corresponding change in order to work properly.


andy-haynes (approved):
Looks good, thanks Danny!

PR #3054, opened 2023-05-21
WEP-227 Implementing segment on wallet migration

hcho112:
This PR contains `segment` implementation for wallet migration. Intention is to retrieve key metrics during the wallet migration:
1. When the user starts migration
2. What are the browsers/devices that migration started with
3. Each step of migration completion
4. How many keys has been deleted
5. Which wallet the user choose to migrate to
6. When the user completes the migration

I have confirmed that metrics are being logged on segment.


packages/frontend/src/components/wallet-migration/metrics.js
```diff
@@ -0,0 +1,63 @@
...
+                ...traits,
+                userAgent: navigator?.userAgent || 'Unknown',
+            }
+        });
+    } catch (e) {
+        console.error(e);
+    }
+};
+
+export const flushEvents = () => {
```
charleslavon, line 58: If any part of the migration process redirects the user to another site or involves a callback, you may want to call this function prior to initiating that action. That will ensure that the current stack of events are sent and not lost while segment does its default of waiting for a certain amount of time before sending all events in a batch.

packages/frontend/src/components/wallet-migration/WalletMigration.jsx
```diff
@@ -82,7 +89,11 @@ const WalletMigration = ({ open, onClose }) => {
         });
     }, [Object.keys(rotatedKeys).length]);
 
-    const navigateToRedirect = () => {
+    const navigateToRedirect = ({ accounts, walletName }) => {
+        recordWalletMigrationEvent(`${WALLET_MIGRATION_VIEWS.MIGRATE_ACCOUNTS} COMPLETED`, {
+            listOfAccounts: accounts.join(', '),
+            selectedWallet: walletName,
+        });
         setMigrationStep(WALLET_MIGRATION_VIEWS.VERIFYING);
```
charleslavon, line 97: ah yes, you likely need to flush here.

PR #3059, opened 2023-06-06
Bugfix: Account Transfer Wizard clean key bug fix on multiple accounts

hcho112:
This PR contains bug fix on `Clean Keys` step of wallet migration. 

Previous implementation had a bug where if there are two or more accounts being transferred, it was deleting FAK that is stored in localStorage. (Which it shouldn't be deleted) After investigation, the issue was using public key of first account. So the fix was made by selecting correct public key on each `accountId` given.

### DIFF
--- a/packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/CleanKeysModal.jsx
+++ b/packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/CleanKeysModal.jsx
@@ -34,8 +34,7 @@ async function getAccountDetails({ accountId, publicKeyBlacklist, wallet }) {
         }, {});
 
     const allAccessKeys = await wallet.getAccessKeys(accountId);
-    const signingPublicKey = await wallet.getLocalKeyPair(accountId)
-        .then(({ publicKey }) => publicKey.toString());
+    const signingPublicKey = (await wallet.getLocalKeyPair(accountId))?.publicKey.toString();
 
     const accessKeys = allAccessKeys
         .filter(({ public_key }) =>
--- a/packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/CleanKeysModal.jsx
+++ b/packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/CleanKeysModal.jsx
@@ -34,12 +34,12 @@ async function getAccountDetails({ accountId, publicKeyBlacklist, wallet }) {
         }, {});
 
     const allAccessKeys = await wallet.getAccessKeys(accountId);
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/CleanKeysModal.jsx
```diff
@@ -34,11 +34,12 @@ async function getAccountDetails({ accountId, publicKeyBlacklist, wallet }) {
         }, {});
 
     const allAccessKeys = await wallet.getAccessKeys(accountId);
-    const signingPublicKey = await wallet.getPublicKey();
+    const signingPublicKey = (await wallet.getLocalKeyPair(accountId))?.publicKey.toString();
```
andy-haynes, line 37: After looking at these methods more closely, I think we should actually be using `wallet.getPublicKey`:

- `getLocalKeyPair` returns more data than needed and includes the secret key
- `getPublicKey` is more commonly used across the wallet migration components

The problem is that the `accountId` parameter is omitted from the `getPublicKey` call. It also needs to be added here: https://github.com/near/near-wallet/blob/0211b43341aa46b413e18f53b8e57d70dab95145/packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/CleanKeysModal.jsx#LL61C8-L61C8

So if we fix those two instances of omitted parameters it should resolve the behavior you're seeing. Also I would not include optional chaining here - if the public key from the signer were somehow `null` then we want it to fail here rather than get included in the set of keys to remove.

andy-haynes (changes requested):
Nice catch ❤️ 

I found another instance in this module where the `accountId` parameter is being omitted. Let's get it included in those two places and it should address any other weirdness stemming from key cleanup with multiple accounts.

packages/frontend/src/components/wallet-migration/modals/CleanKeysModal/CleanKeysModal.jsx
hcho112: @andy-haynes very good point!, I should have also check the `getPublicKey` func more closely. It does reference correct PK after I passed `accountId`. Ran the test multiple times and confirm that everything works accordingly

andy-haynes: > I should have also check the `getPublicKey` func more closely

Me too or I would have gotten it right the first time 😂 

That `Wallet` class will get you every time 😅

andy-haynes (approved):
🎉

PR #3062, opened 2023-06-07
Upgrade wallet-selector to 8.1.2

hcho112:
This PR contains a package update, `wallet-selector` related packages to `8.1.2`.

This Change will allow transfer wizard to introduce a new feature, on hover to show passphrase:
image
(FYI cursor is actually hovering on top of the passphrase element)


andy-haynes (approved):
🚢

PR #3063, opened 2023-06-07
feat: handle excess keys when disabling 2FA w/FAK

andy-haynes:
This PR updates the `disableMultisigWithFAK` to use the same logic for handling an excess number (> 48) of multisig LAKs as the `batchConvertKeysAndDisable` uses. For consistency I factored the usage out of the latter into a static method that could be used by `disableMultisigWithFAK`.

I have tested both the new and refactored functionality as well as updating tests to accommodate new method signatures.


hcho112 (approved):
LGTM!, thank you for awesome work! (Nice test updates too)
I have tested them locally and disable 2FA work nicely :)

PR #3067, opened 2023-06-19
Replacing Segment to RudderStack on transfer wizard metrics

hcho112:
This PR contains changes from replacing Segment to Rudderstack. 

Currently these metrics are only active within wallet transfer functionality.

### DIFF
--- a/packages/frontend/src/components/wallet-migration/WalletMigration.jsx
+++ b/packages/frontend/src/components/wallet-migration/WalletMigration.jsx
@@ -9,7 +9,7 @@ import { MAINNET, TESTNET } from '../../utils/constants';
 import { wallet } from '../../utils/wallet';
 import LoadingDots from '../common/loader/LoadingDots';
 import { MigrationModal, ButtonsContainer, StyledButton, Container } from './CommonComponents';
-import { flushEvents, initSegment, recordWalletMigrationEvent, recordWalletMigrationState } from './metrics';
+import { flushEvents, initSegment, recordWalletMigrationEvent, recordWalletMigrationState, rudderAnalyticsReady } from './metrics';
 import CleanKeysCompleteModal from './modals/CleanKeysCompleteModal/CleanKeyCompleteModal';
 import CleanKeysModal from './modals/CleanKeysModal/CleanKeysModal';
 import Disable2FAModal from './modals/Disable2faModal/Disable2FA';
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/wallet-migration/metrics.js
```diff
@@ -1,63 +1,65 @@
...
-        console.error(e);
-    }
+    return new Promise((resolve) => {
+        if (rudderAnalyticsReady) {
+            return;
+        }
+      
+        const rudderAnalyticsKey = IS_MAINNET
+            ? '2RPBJoVn6SRht1E7FnPEZLcpNVh'
+            : '2RPBBhH1Dka2bYG6HKQmvkzNUdz';
```
roshaans, line 18: @pkudinov Will we have a separate key for mainnet/testnet?

packages/frontend/src/components/wallet-migration/metrics.js
```diff
@@ -1,63 +1,65 @@
...
     }
 };
 
 export const flushEvents = () => {
-    if (!segment) {
+    if (!rudderAnalyticsReady) {
         return;
     }
-    return segment.flush();
+    return rudderanalytics.reset();
```
roshaans, line 64: `rudder-stack-js` does not expose a publicly available method similar to the functionality of `segment.flush()` which is intended for forcing all queued events to be sent immediately to the server. `.reset()` does not do that here though.

 It simply resets the anonymousId as well as the userAccountId that we are tracking as per the implementation [here](https://github.com/rudderlabs/rudder-sdk-js/blob/66c998ba2785cd33445119f9468d0d3488b34f38/src/core/analytics.js#L1071).
 
I think we can rename this method to `resetUserState()` instead.

roshaans (approved):
Nice job! Overall looks good. Just some minor changes in naming...

I tested the code in the browser and it seems to forward the events as expected!

packages/frontend/src/components/wallet-migration/WalletMigration.jsx
```diff
@@ -94,11 +95,9 @@ const WalletMigration = ({ open, onClose }) => {
             listOfAccounts: accounts.join(', '),
             selectedWallet: walletName,
         });
-        flushEvents()
-            .finally(() => {
-                setMigrationStep(WALLET_MIGRATION_VIEWS.VERIFYING);
-                handleSetActiveView(WALLET_MIGRATION_VIEWS.REDIRECTING);
-            });
+        flushEvents();
```
roshaans, line 98: At the moment, when `flushEvents()` gets called we will have descrepencies in our data as `flushEvents()` calls `rudderStack.reset()` which will reset the current logged in user data. We should remove flushEvents() here.

roshaans (changes requested):

hcho112:
@roshaans thank you for thoughtful review!

hcho112:
I have rebased master. This PR needs to be reviewed & merged prior to releasing transfer wizard, so would be appreciated if I can get more reviews

andy-haynes (approved):
LGTM 👍

PR #3068, opened 2023-06-20
feat: add FT to blacklist

andy-haynes:
This PR adds an FT blacklist entry for the spam token `youwon400neartoclaimyourgainwwwlotte.laboratory.jumpfinance.near` on wallet.near.org.


hcho112 (approved):
LGTM,

PR #3073, opened 2023-06-22
Fix: Use correct environment variable to match staging

hcho112:
Minor bug fix on rudderstack config to use correct environment variable to match if target is staging or not.

### DIFF
--- a/packages/frontend/src/components/wallet-migration/metrics.js
+++ b/packages/frontend/src/components/wallet-migration/metrics.js
@@ -1,15 +1,15 @@
 import * as rudderanalytics from 'rudder-sdk-js';
 
-import Environment from '../../../../../features/environments.json';
-import { IS_MAINNET, NEAR_WALLET_ENV } from '../../config';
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/wallet-migration/metrics.js
```diff
@@ -24,7 +24,7 @@ export const initAnalytics = () => {
             return;
         }
       
-        const rudderAnalyticsKey = IS_MAINNET
```
morgsmccauley, line 27: should we fix `IS_MAINNET` rather than inlining the correct logic here?

hcho112: I'm not sure. `IS_MAINNET` is being used in many other places with existing logic. If we make change here just to make it work for `RudderStack` it could break on other part. 

Also I think it was just wrong on using `IS_MAINNET` here. (`IS_MAINNET` is also considered as `true` for `staging`) Which is why I decided to change the condition

morgsmccauley (approved):

PR #3075, opened 2023-06-23
Show Error Modal if user has adblocker during transfer wizard

hcho112:
This PR contains implementation to show Error Modal if user is using adblocker during transfer wizard process.

It is essential for us to obtain metrics of users during transfer wizard and having adblocker stops `RudderStack` to work and therefore it is import to turn off adblocker prior to start of the transfer wizard.

Greenshot 2023-06-23 15 51 25

Do not merge this PR if we decided to merge https://github.com/near/near-wallet/pull/3074 instead


hcho112:
Closing this PR as we decided to go with https://github.com/near/near-wallet/pull/3074 instead of this PR

PR #3079, opened 2023-06-29
Lint fixes

hcho112:
This PR contains updates on lint fixes. Some of them ran through `yarn run fix` and some are manually fixed. 

There are parts where I put lint ignore because the implementation breaks when the order of import changes.

This PR should not change any logic.

### DIFF
--- a/packages/frontend/ci/ParcelBundler.js
+++ b/packages/frontend/ci/ParcelBundler.js
@@ -3,8 +3,8 @@ const path = require('path');
 
 const Bundler = require('parcel-bundler');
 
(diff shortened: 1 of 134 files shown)


andy-haynes:
I'm not seeing any lint issues on the head of `master` and `yarn lint --fix` produces no changes. Are you maybe picking up a lint configuration external to the repository?

hcho112:
@andy-haynes hmm strange, as you said, it is actually working just fine.. Might be an issue with my local machine. Will close this PR first

PR #3084, opened 2023-07-15
fix: ru translation of migration message

frolvanya:
Fixed ru translation of migration message (origin from English version)

PR #3085, opened 2023-07-17
Minor translation fix related to transfer wizard in RU

hcho112:
This PR contains minor translation fix written in RU. We no longer support `startDate` and this part of translation only exist in RU. Based on commit date, it is very old and out dated. 

Before:
image

Now:
image

Thanks to @esaminu and other RU members on reporting this.

### DIFF
--- a/packages/frontend/src/translations/ru.global.json
+++ b/packages/frontend/src/translations/ru.global.json
@@ -551,7 +551,7 @@
         }
     },
     "migration": {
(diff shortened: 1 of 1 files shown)


esaminu (approved):
LGTM

andy-haynes:
@esaminu @hcho112 the translation is still problematic since it references MNW. Maybe whoever raised this issue can independently verify this [PR](https://github.com/near/near-wallet/pull/3084)? If it's a good translation we should just fix the whole thing by taking this contributor PR 🙂

hcho112:
@andy-haynes yes you are right! good catch. 
and the PR you referenced looks like a proper fix as well. (However would be nicer to get a proper confirmation from Russian speaking person.

PR #3089, opened 2023-07-31
Replace account id on rudderstack with hash id

hcho112:
This PR contains implementation to replace existing `account id` details on RudderStack metrics with hash id.

The team decided to not track account Id from RudderStack on Transfer Wizard. So similar to how it is implemented on `near-discovery` we hash the accountId and use it instead.

This change need to be published to prod as soon as it passes the review.

### DIFF
--- a/packages/frontend/src/components/wallet-migration/metrics.js
+++ b/packages/frontend/src/components/wallet-migration/metrics.js
@@ -1,3 +1,4 @@
+import sha256 from 'js-sha256';
 import * as rudderanalytics from 'rudder-sdk-js';
 
 import Environments from '../../../../../features/environments.json';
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/wallet-migration/metrics.js
```diff
@@ -88,3 +91,23 @@ export const resetUserState = () => {
     }
     return rudderanalytics.reset();
 };
+
+export function accountIdToHash(accountId) {
+    const hash = sha256.create();
+    hash.update(accountId);
+    return hash.hex();
+};
```
andy-haynes, line 99: Per our discussion: unless there's a good reason to persist each hash, I think you should drop the functions that get/set localstorage and only use this method to compute the hash each time.

packages/frontend/src/components/wallet-migration/metrics.js
```diff
@@ -88,3 +93,16 @@ export const resetUserState = () => {
     }
     return rudderanalytics.reset();
 };
+
+export function accountIdToHash(accountId) {
+    if (!accountIdHash)  {
```
andy-haynes, line 98: Introducing state here now has introduced the potential for an event to be reported under the wrong accountId. Can we please compute the hash every time so as not to add any further complexity to event reporting?

andy-haynes (changes requested):
Looking good overall, but I remain strongly in favor of computing the hash every time. We don't get any noticeable benefit from caching, we only introduce the possibility of reporting the wrong data.

andy-haynes (approved):

PR #3115, opened 2024-01-09
Add Call To Action Button on Transfer Wizard Doc

hcho112:
This PR contains minor change on transfer wizard doc to include call to action button at the bottom of the doc, allowing user to initiate the transfer wizard from doc page:

image

This button will only be visible if there is at least one account already signed in.

https://github.com/near/wallet-experience/issues/15

### DIFF
--- a/packages/frontend/src/components/Routing.js
+++ b/packages/frontend/src/components/Routing.js
@@ -520,7 +520,13 @@ class Routing extends Component {
                             <Route
                                 exact
                                 path="/transfer-wizard"
-                                component={TransferWizardWrapper}
+                                render={() => (
+                                    <TransferWizardWrapper
+                                        account={account}
+                                        onTransfer={this.handleTransferClick} 
+                                    />
+                                )}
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/Routing.js
```diff
@@ -520,7 +520,12 @@ class Routing extends Component {
                             <Route
                                 exact
                                 path="/transfer-wizard"
-                                component={TransferWizardWrapper}
+                                render={() => (
+                                    <TransferWizardWrapper
+                                        account={account}
+                                        onTransfer={this.handleTransferClick}
```
andy-haynes, line 526: Let's please invoke this in an arrow function to avoid potential binding issues, e.g.:
```js
onTransfer={() => this.handleTransferClick()}
```

andy-haynes (changes requested):
Just one thing otherwise LGTM

andy-haynes (approved):

PR #3123, opened 2024-02-12
Move mobile wallet UI to last in the list

hcho112:
This PR contains small change, moving mobile wallet to last in the list:
image

Since this PR doesn't create any side-effect/functional change, please merge them as soon as it gets approved.


trechriron (approved):
Ship it!

PR #3126, opened 2024-03-12
Replace INDEXER_SERVICE_URL to use nearblock

hcho112:
This PR contains implementation to repalce `INDEXER_SERVICE_URL` to nearblock. 
kitwallet API is currently stable and this is stopping users to utilise transfer wizard


charleslavon:
@andy-haynes is there a staging env where this PR can be tested or is it worth pushing and testing live?

andy-haynes:
> is there a staging env where this PR can be tested or is it worth pushing and testing live?

Not anymore, but given the indexer endpoint replacement I don't think this PR is still relevant.

PR #3128, opened 2024-03-19
Adding Explore NEAR Button

s-n-park:
Context:
- The 3rd most popular query for getting to https://near.org/ is "near wallet" yet wallet.near.org often shows up in the first position
- From wallet.near.org, if a users is interested in exploring the NEAR ecosystem, there is no option to get to the main near.org  page

PR:
- This PR adds a button on the wallet.near.org landing page with CTA "Explore NEAR" which takes users to https://near.org/ in the event that they want to continue exploring NEAR regardless of their wallet situation.
image
- Also added are the translations for the various languages supported (disclaimer: these were translated from "Explore NEAR" using chatGPT and not verified with native speakers)


trechriron (approved):
Nicely Done! Shipit.

